PR #91, opened 2021-08-17
[DRAFT] control-service: deprecated jobsList endpoint cleanup

ivakoleva:
jobsList endpoint deprecated in favour of jobsQuery cleanup.

The deprecated endpoint jobsList deleted.


tozka:
> > My understanding is that we are not ready to delete it because it's used.
> 
> That is correct; however the cleanup is ready for review, we can prepare for merging once we have the green light

Should we merge it, I believe all customers have moved on ?

PR #321, opened 2021-10-01
vdk-plugin-control-cli: introduce job execution skipping functionality

mrMoZ1:
why: As a user, when starting a data job execution manually through
the new execution API I don't want a scheduled job to run in parallel
with my manual execution. This is why we need a plugin that is able
to check for a different running execution and, skip it if necessary.

what: The new implementation will make a call to the control-service
API and query for any executions in RUNNING or SUBMITTED state.
If we have such executions present, it will SKIP the current execution.
Added the implementation to the vdk-plugin-control-cli existing
plugin as discussed. The entry point for the new functionality is through
the run_job hookimpl as we need log_config_type, job_name, execution_id
and job_team to be initialized by the initialize_job hooks. We also have the
tryfirst param enabled.

testing: Verified skipping logic is working as intended. 
Added unit tests. CI/CD


projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -0,0 +1,164 @@
...
+            f"Error while checking for another data job excecution: {str(exc)} "
+        )
+        log.warning("Printing stack trace and proceeding with execution:")
+        log.warning(exc)
+    return None
+
+
+@hookimpl(tryfirst=True)
+def run_job(context: JobContext) -> None:
+    log.info("Plugin hook method. Obtaining arguments for skip execution check.")
```
gageorgiev, line 153: Maybe drop this log since it would be called on every job run and clutter the logs a bit too much.

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -0,0 +1,164 @@
...
+        log.info("Authenticating.")
+        auth = Authentication()
+        auth.update_api_token_authorization_url(api_token_authorization_url)
+        auth.update_api_token(api_token)
+        auth.update_auth_type(auth_type)
+        auth.acquire_and_cache_access_token()
+        log.info("Login successful.")
+
+
+class ExitWithMessage:
```
gageorgiev, line 94: What's the purpose of this class? It seems like it has one static method which gets used once; there are no objects instantiated from the ExitWithMessage class.

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -0,0 +1,164 @@
...
+        ]  # We query only for running or manual jobs
+        response = self.execution_api_client.data_job_execution_list(
+            job_name=job_name, team_name=team_name, execution_status=job_status
+        )
+        log.info(f"Api call returned job execution list: {response}")
+        return response
+
+    @staticmethod
+    def _authenticate(api_token_authorization_url, api_token, auth_type) -> None:
+        log.info("Authenticating.")
```
gageorgiev, line 85: What are you authentication for and against?

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -0,0 +1,164 @@
...
+        auth = Authentication()
+        auth.update_api_token_authorization_url(api_token_authorization_url)
+        auth.update_api_token(api_token)
+        auth.update_auth_type(auth_type)
+        auth.acquire_and_cache_access_token()
+        log.info("Login successful.")
```
gageorgiev, line 91: Same as previous comment; either make these log calls more descriptive or get rid of them.

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -0,0 +1,164 @@
...
+                 submitted or running with a different id and in that case we should skip current execution
+                 until the other finishes.
+        """
+        jobs = self._get_running_job_executions(job_name, job_team)
+        job_execution_running = False
+        for job_execution in jobs:
+            # We break only if we have a job with a different execution id
+            # Sometimes ids returned by VDK execution_state can slightly differ than
+            # ids returned by the API. VDK ids  have '-xxxxx' appended at the end,
+            # where x is a random alphameric char. Otherwise ids start the same.
```
gageorgiev, line 60: nit: alphanumeric*

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -0,0 +1,164 @@
...
+    action: WriteToFileAction,
+):
+    try:
+        log.info("Checking if job should be skipped:")
+        log.info(
+            f"Job : {job_name}, Team : {job_team}, Log config: {log_config}, execution_id: {execution_id}"
+        )
+        vdk_config = VDKConfig()
+        # vdk_config.
+        if log_config != "CLOUD":
```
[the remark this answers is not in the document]
gageorgiev: I think we can proceed as is for now and fix all of these cases in one PR next sprint.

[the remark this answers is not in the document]
mrMoZ1: I agree with you @tozka , however I think another PR will introduce scope creep and the execution API will be delayed even further. IMO if the proposed solution works we should use it. Nevertheless, if you think that this isn't a good pattern and we cannot afford to use it as is, I can open another PR for that. Otherwise, I can create a ticket and refactor this in the next sprint. Let me know how to proceed.

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
mrMoZ1: It's am alternative form for the same word, though alphanumeric is used more often.

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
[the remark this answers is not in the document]
tozka: I think next spring is OK.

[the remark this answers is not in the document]
mrMoZ1: Created issue for next spring:
https://jira.eng.vmware.com/browse/SC-996

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
mrMoZ1: Addressed feedback in latest commit.

gageorgiev (approved):

tozka (approved):

PR #324, opened 2021-10-04
vdk-heartbeat: Add option for multiple env var names

doks5:
Some of our CI/CD pipelines set configurations with env var names
starting with `VDKCLI_`, while others with names starting with `VDK_`.
We need to unify them, but until that happens, the two possibilities
need to be taken into account.

This change makes it possible to make specific configurations using two
different names for environment variables, e.g., `VDKCLI_OAUTH2_REFRESH_TOKEN`
and `VDK_OAUTH2_REFRESH_TOKEN` would both be valid.

Testing Done: Tested locally by running vdk-heartbeat -f vdk-options.ini


projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -21,8 +22,12 @@ def __init__(self, config_file=None):
             log.info(f"Configuration loaded: {self._config_ini.items()}")
 
         # VDK CLI Arguments for vdkcli login
-        self.vdkcli_api_refresh_token = self.get_value("VDKCLI_OAUTH2_REFRESH_TOKEN")
-        self.vdkcli_oauth2_uri = self.get_value("VDKCLI_OAUTH2_URI")
+        self.vdkcli_api_refresh_token = self._get_atleast_one_value(
+            "VDKCLI_OAUTH2_REFRESH_TOKEN", "VDK_OAUTH2_REFRESH_TOKEN"
```
tozka, line 26: Let's name the new variables VDK_HEARTBEAT_API_TOKEN - That's the official name. 

And let's prefix all vdk heartbeat (only) specific options with "VDK_HEARTBEAT_" 

Refresh token is an old name that was not changed in vdk-heartebat. (see vdk login --help) .

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -21,8 +22,12 @@ def __init__(self, config_file=None):
...
+        )
+        self.vdkcli_oauth2_uri = self._get_atleast_one_value(
+            "VDKCLI_OAUTH2_URI", "VDK_OAUTH2_URL"
```
tozka, line 29: VDK_API_TOKEN_AUTHORIZATION_URL.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -70,7 +75,33 @@ def __init__(self, config_file=None):
...
+                "PROPERTIES_API_TOKEN_AUTHORIZATION_URL",
+                self.properties_api_authorization_url,
+            )
+
+        # Set the VDK properties API token
+        self.properties_api_token = self.get_value(
+            "VDK_PROPERTIES_TOKEN", is_required=False
+        )
+        if self.properties_api_token:
+            self.set_value("PROPERTIES_API_TOKEN", self.properties_api_token)
```
tozka, line 104: Those are not necessary to be set by vdk heartbeat.

doks5: Removed, thanks for the review

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
doks5: Renamed, thanks

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
doks5: Modified, thanks

tozka (approved):
We should do it separately but all options should be prefixed with "VDK_HEARTBEAT_" I think.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -177,6 +184,25 @@ def get_value(self, key: str, default_value: str = None, is_required=True):
...
+        if (not value1) and (not value2):
+            raise ValueError(
+                "Error occurred:\n"
+                f"What: Cannot configure heartbeat test\n"
+                f"Why: Missing required configuration with either key {key1} or key {key2}\n"
+                f"Consequences: The heartbeat will abort\n"
+                f"Countermeasures: Set as required configuration either as environment variable or in the config file."
+            )
+
+        return value1 or value2
```
tozka, line 200: This would break if value is boolean . 

Perhaps return value1 if value1 is not None else value2
```suggestion
        if (value1 is None) and (value2 is None):
            raise ValueError(
                "Error occurred:\n"
                f"What: Cannot configure heartbeat test\n"
                f"Why: Missing required configuration with either key {key1} or key {key2}\n"
                f"Consequences: The heartbeat will abort\n"
                f"Countermeasures: Set as required configuration either as environment variable or in the config file."
            )

        return value1 if value1 is not None else value2
```

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -177,6 +184,25 @@ def get_value(self, key: str, default_value: str = None, is_required=True):
...
+
+    @staticmethod
+    def set_value(key: str, value: str):
```
tozka, line 203: Where is this used?

tozka:
Look good. My comments are not a blocker.  So address them as you see fit and merge it (when other reviewers are finished as well)

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
doks5: Nowhere. Fixed, thanks

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
doks5: Ooops, good catch. Fixed, thank you

PR #326, opened 2021-10-04
control-service: fix NullPointerException when reading job deployment status

tpalashki:
Testing done: Tested against a local control service, configured to run against
the environment where the bug initially appeared.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -1480,7 +1480,10 @@ public abstract class KubernetesService implements InitializingBean {
                 deployment.setImageName(image); // TODO do we really need to return image_name?
             }
             var labels = cronJob.getSpec().getJobTemplate().getMetadata().getLabels();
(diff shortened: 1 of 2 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -1480,7 +1480,7 @@ private V1Secret buildV1Secret(String name, Map<String, byte[]> data) {
                 deployment.setImageName(image); // TODO do we really need to return image_name?
             }
             var labels = cronJob.getSpec().getJobTemplate().getMetadata().getLabels();
-            if (labels.containsKey(JobLabel.VERSION.getValue())) {
+            if (labels != null && labels.containsKey(JobLabel.VERSION.getValue())) {
                 deployment.setGitCommitSha(labels.get(JobLabel.VERSION.getValue()));
             } else {
                 // Legacy approach to get version:
```
tozka, line 1486: can you add log for which we do not have labels expected. It could be even warning.

tpalashki: I will. This particular one was the ecr-cred-helper.

PR #329, opened 2021-10-04
vdk-test-utils, vdk-plugin-control-cli: fix version of vdk-core

tozka:
We are fixing the version of vdk-core since we are planning a breaking change in IPropertiesServiceClient in a next version. We would adopt it in subsequent PR and un-fix the version. 

Testing Done: existing automated tests


tozka:
Discarded

PR #333, opened 2021-10-05
plugins: fix entry-point check

tozka:
The entry point check was failing for vdk-test-utils and quickstart-vdk
which are not actual plugins.
quickstart-vdk is a distribution of vdk (aggregation of plugins) to get
started quickly .
vdk-test-utils is a library for testing plugins or vdk-core.

So they are added as expcetions explicitly to the check.

Also decoupled the building of plugins in separate plugin to make it
easier to develop and reuse locally

Testinng Done: This change CI and locally from quickstart-vdk,
vdk-ingest-file and vdk-test-utils directory run `../build-plugin.sh`
successfully


projects/vdk-core/plugins/build-plugin.sh
```diff
@@ -0,0 +1,32 @@
...
+
+pip install -U pip setuptools pre-commit
+pre-commit install
+
+pip install --extra-index-url $PIP_EXTRA_INDEX_URL -r requirements.txt
+pip install -e . --extra-index-url $PIP_EXTRA_INDEX_URL
+
+# List exceptions to below check here.
+# Those are not technically plugins so they would not have entry point defined.
+if [ "$PLUGIN_NAME" != 'quickstart-vdk' ] && [ "$PLUGIN_NAME" != 'vdk-test-utils' ]; then
```
gageorgiev, line 19: This is probably fine for now, but we should think about a programmatic way of separating plugins from supplementary packages like test-utils. Also, what about taurus-vdk?

projects/vdk-core/plugins/build-plugin.sh
```diff
@@ -0,0 +1,32 @@
...
+
+  echo "Check plugin entry point is configured correctly ..."
+  if ! vdk version 2>&1 | grep -q "$PLUGIN_NAME"; then
+    echo "Plugin entry point seems to be mis-configured."
+    echo "Make sure to set setup.py entry_points for the plugin or update an exception case in above if statement."
```
gageorgiev, line 24: No exit statement? Would this CI job succeed if the entry point is misconfigured?

projects/vdk-core/plugins/README.md
```diff
@@ -6,3 +6,5 @@ Generally those are
  * Update .plugin-ci.yml file with name of the plugin - make sure to follow comments.
  * src folder is which contain the source code of your plugin
  * tests folder will contain tests which will be run by CI framework automatically.
+
+In order to build and test a plugin go to plugin directory and use `../build-plugin.sh` script to build it
```
gageorgiev, line 10:
```suggestion
In order to build and test a plugin go to the plugin directory and use `../build-plugin.sh` script to build it.
```
nit

projects/vdk-core/plugins/build-plugin.sh
tozka: I plan to delete taurus-vdk. It is not necessary here

projects/vdk-core/plugins/build-plugin.sh
tozka: Oops . Thanks :)

gageorgiev (approved):

PR #339, opened 2021-10-05
vdk-plugin-control-cli: define vdk-control-cli configuration options

tozka:
Define in a plugin vdk-control-cli config optoins. This way they can be
seen with `vdk config-help` CLI command which shows the detailed
configuration of the CLI

Testing Done: see unit tests

### DIFF
--- a/projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
+++ b/projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
@@ -19,6 +19,8 @@ from vdk.internal.control.command_groups.login_group.login import login
 from vdk.internal.control.command_groups.logout_group.logout import logout
 from vdk.internal.control.configuration.default_options import DefaultOptions
 from vdk.internal.control.plugin import control_plugin_manager
+from vdk.internal.core.config import ConfigurationBuilder
+from vdk.plugin.control_cli_plugin import control_service_configuration
 
 
 @hookimpl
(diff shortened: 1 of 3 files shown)


projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
```diff
@@ -41,3 +43,9 @@ def vdk_command_line(root_command: click.Group):
     default_options = DefaultOptions(plugins)
     if default_options.get_default_map():
         root_command.context_settings["default_map"] = default_options.get_default_map()
+
+
+@hookimpl(tryfirst=True)
+def vdk_configure(config_builder: ConfigurationBuilder) -> None:
+    """ """
+    control_service_configuration.add_definitions(config_builder)
```
gageorgiev, line 51: This is totally unrelated to this change, but seeing this code pattern makes me feel weird. Usually in a OOP paradigm, if we execute `A.method(B)` we expect A to mutate, and its mutation to depend on B in some way; however, here we see the opposite, where `config_builder` mutates depending on the `control_service_configuration` object. I don't know if this is something that requires refactoring necessarily or if I'm being pedantic (or maybe I'm entirely wrong and this isn't that unusual, who knows).

tozka: I am open to better suggestions. You may have a point. We need to be consistent though. 

Another thing I don't like I did was using get_xxx() methods in the DomainOrientedConfiguration class. Python properties would have been a more pythonic way to do it. get_xxx is very java like. But I am sticking to it everywhere until we decide to change it everywhere.

PR #340, opened 2021-10-05
vdk-control-cli: impelment properties plugin reusing control service client

tozka:
We had custom implementation with custom configuration of
PropertiesServiceClient plugin which caused unnecessary issues .

Testing Done: existing  tests passed


projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/control_service_properties_client.py
```diff
@@ -0,0 +1,42 @@
...
+    Implementation of PropertiesServiceClient which connects to Pipelines Control Service Properties API
+    """
+
+    def __init__(self, rest_api_url: str):
+        self.__properties_api = ApiClientFactory(rest_api_url).get_properties_api()
+        log.debug(f"Initialized Properties against {rest_api_url}.")
+
+    def read_properties(self, job_name: str, team_name: str):
+        data = self.__properties_api.data_job_properties_read(
+            team_name=team_name, job_name=job_name, deployment_id="todo"
```
doks5, line 32: Why is deployment_id set to "todo"?

tozka: Because we currently do not support multiple deployments per job in the Control Service so it is ignored.

gageorgiev (approved):

PR #341, opened 2021-10-05
vdk-logging-json: Escape double quotes in message

gageorgiev:
Double quotes in the log message would mess up the parsing, as
JSON separates fields and values based on double quote pairs.
This change escapes the double quotes in the message string.

Testing done: verified that the formatted log with double quotes
in the message is valid JSON
Logs before:
```
{"@timestamp":"2021-10-05T14:15:00.959Z","created":"1633432500.959478","jobname":"gg-job2","level":"INFO","modulename":"vdk.internal.builtin_plugins.run.cli_run","filename":"cli_run.py","lineno":"66","funcname":"run_job","attemptid":"1633432500-7b4ee0-db01f0","message":"Data Job execution summary: {\n  "data_job_name": "gg-job2",\n  "execution_id": "1633432500-7b4ee0",\n  "start_time": "2021-10-05T11:15:00.958685",\n  "end_time": "2021-10-05T11:15:00.959378",\n  "status": "success",\n  "steps_list": [\n    {\n      "name": "10.py",\n      "type": "python",\n      "start_time": "2021-10-05T11:15:00.958709",\n      "end_time": "2021-10-05T11:15:00.959355",\n      "status": "success",\n      "details": null,\n      "exception": null\n    }\n  ],\n  "exception": null\n}"}
```
This causes a JSON parsing error.

Logs after:
```
{"@timestamp":"2021-10-05T14:18:15.617Z","created":"1633432695.617323","jobname":"gg-job2","level":"INFO","modulename":"vdk.internal.builtin_plugins.run.cli_run","filename":"cli_run.py","lineno":"66","funcname":"run_job","attemptid":"1633432695-a89cbb-d5a5c7","message":"Data Job execution summary: {\n  \"data_job_name\": \"gg-job2\",\n  \"execution_id\": \"1633432695-a89cbb\",\n  \"start_time\": \"2021-10-05T11:18:15.616559\",\n  \"end_time\": \"2021-10-05T11:18:15.617217\",\n  \"status\": \"success\",\n  \"steps_list\": [\n    {\n      \"name\": \"10.py\",\n      \"type\": \"python\",\n      \"start_time\": \"2021-10-05T11:18:15.616584\",\n      \"end_time\": \"2021-10-05T11:18:15.617192\",\n      \"status\": \"success\",\n      \"details\": null,\n      \"exception\": null\n    }\n  ],\n  \"exception\": null\n}"}
```

### DIFF
--- a/projects/vdk-core/plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
+++ b/projects/vdk-core/plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
@@ -19,14 +19,14 @@ format_template = (
 # this class serves to escape newline characters in the log message
 # as according to https://stackoverflow.com/questions/42068/how-do-i-handle-newlines-in-json
 # it is currently experimental and might be removed
-class RemoveNewlinesFormatter(logging.Formatter):
+class EscapeNewlinesAndQuotesFormatter(logging.Formatter):
     def __init__(self, fmt):
         super().__init__(fmt=fmt)
         self.default_time_format = "%Y-%m-%dT%H:%M:%S"
         self.default_msec_format = "%s.%03dZ"
 
     def format(self, record):
-        record.msg = record.msg.replace("\n", "\\n")
+        record.msg = record.msg.replace("\n", "\\n").replace('"', '\\"')
         return super().format(record)
 
 
@@ -36,7 +36,7 @@ def vdk_start(plugin_registry: IPluginRegistry, command_line_args: List):
         "{" + format_template.format(job_name="", attempt_id="") + "}"
     )  # appending the braces
     for handler in logging.getLogger().handlers:
-        formatter = RemoveNewlinesFormatter(fmt=format_copy)
+        formatter = EscapeNewlinesAndQuotesFormatter(fmt=format_copy)
 
         handler.setFormatter(formatter)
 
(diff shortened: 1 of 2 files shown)


tozka:
Can you explain a bit more how does it mess it up ? Maybe provide an example of the logs ( attach file)?

projects/vdk-core/plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
```diff
@@ -19,14 +19,14 @@
 # this class serves to escape newline characters in the log message
 # as according to https://stackoverflow.com/questions/42068/how-do-i-handle-newlines-in-json
 # it is currently experimental and might be removed
-class RemoveNewlinesFormatter(logging.Formatter):
+class EscapeNewlinesAndQuotesFormatter(logging.Formatter):
```
tozka, line 22: I do not see change in tests. 

Most changes but bug fix especially should be covered in tests - 
a) to make sure the fix is truly fixed (the test should fail before the fix
b) to make sure we do not regress again as bug tend to resurface 
c) make it easier to understand the problem and review

gageorgiev: Included example logs. As for the issue, a JSON parser will stop at the first double quote, as it sees it as the end of the message field, but the extra text afterwards causes a parsing error.

Will include a more extensive test as the formatter subclass was not being tested until now.

tozka (approved):

PR #347, opened 2021-10-05
vdk-core: extend pod termination massage

mivanov1988:
In order to be able to use the K8S Pod termination message
to transfer various data (such as vdk version) we need to switch
from plain text to json format of termination message.

Testing: unit tests


mrMoZ1:
vdk-core: extend pod termination **massage**?

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
```diff
@@ -25,13 +30,25 @@ def vdk_configure(self, config_builder: ConfigurationBuilder):
...
-    def _write_termination_message(self, error_overall, user_error, configuration):
+    # Returns the version of the package named dist_name
+    @staticmethod
+    def version(
+        dist_name="vdk-core",
+    ):  # Change default value if project is renamed and does not equal the setuptools metadata.name
+        try:
+            return get_distribution(dist_name).version
+        except DistributionNotFound:  # pragma: no cover
+            return "unknown"
```
gageorgiev, line 47: Why not just use the `_version` function from the version module? Seems preferable to duplicating a chunk of code. If you agree with my comment, you might also want to rename that function to just `version`, as it is no longer a protected function.

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
```diff
@@ -56,20 +73,26 @@ def _write_termination_message(self, error_overall, user_error, configuration):
                         f'Writing termination message to file "{termination_message_output_file}"'
                     )
                 self._execute_termination_action(
-                    termination_action, error_overall, user_error
+                    self, file_util, error_overall, user_error, execution_skipped
                 )
-        except:
-            log.exception("Failed to write termination message.")
+        except Exception as e:
+            log.exception(e)
```
gageorgiev, line 79: Provide some context as to where and why this exception occured.
```suggestion
            log.exception("Failed to write termination message. See exception: ", e)
```

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
```diff
@@ -25,13 +30,25 @@ def vdk_configure(self, config_builder: ConfigurationBuilder):
...
             errors.get_blamee_overall(),  # TODO: get this from context
             errors.get_blamee_overall_user_error(),  # TODO: get this from context
             context.configuration,
         )
 
-    def _write_termination_message(self, error_overall, user_error, configuration):
+    # Returns the version of the package named dist_name
+    @staticmethod
+    def version(
+        dist_name="vdk-core",
```
tozka, line 42: This would record the version of vdk-core. 

I think we need the version of the main distribution of vdk that aggregates different plugins and vdk-core (e.g quickstart-vdk, company-x-vdk). 

From user point of view, this is the main version they'd be using and they know. Also from the vdk distribution version we can see the version of the plugins potentially.

mivanov1988:
> vdk-core: extend pod termination **massage**?

before I made that change I went for a massage :D

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
tozka: You pray to the gods of ancient Rome and Greece and then fast for 30 days and the path would show for you. 

Alternatively, we can re-use PACKAGE_NAME - https://github.com/vmware/versatile-data-kit/blob/4cdbe3b0cfedb8ddc675bec27aee4a88389ce16d/projects/vdk-core/src/vdk/internal/builtin_plugins/version/new_version_check_plugin.py#L30 - whose idea is to set the correct main python package (or distribution) name to use to check for the new version. 

But we'd need to. fix the description and possibly refactor a bit (so it's not used only for new version check)

mivanov1988: I agree, but let's do it iteratively in a separate PR.

PR #357, opened 2021-10-07
control-service: change CI/CD to release a new version on each PR merge

tpalashki:
Up until now Control Service releases had to be done manually
by making changes to several files and then making a new PR
and merge it.

With this commit a new version of the Control Service is
released automatically every time a new image of the Control Service
is built, which happens every time a PR with changes to the code
base of the Control Service is merged. The MAJOR and MINOR components
of the version are taken from the version.txt file. The PATCH
component is automatically generated.

Updated the CONTRIBUTING.md file to reflect the new release
process. Cleaned up the CHANGELOG.md file.

Testing done: Tested the new functionality by temporarily making
it trigger on pull requests in addition to merges and verified that
the service is released with a correctly generated version and that
the referenced image is correct as well.


projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -226,17 +223,20 @@ control_service_release:
...
     - curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
-    - helm plugin install https://github.com/chartmuseum/helm-push
+    - helm plugin install https://github.com/chartmuseum/helm-push --version 0.10.0
+    - helm plugin install https://github.com/thynquest/helm-pack --version 0.2.2
     - cd projects/control-service/projects/helm_charts
     - bash -ex ../../cicd/release-pipelines-service.sh
   retry: !reference [.control_service_retry, retry_options]
   only:
     refs:
+      - external_pull_requests
```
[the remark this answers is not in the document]
tpalashki: Correct.

mivanov1988:
Since we are moving to automatic releasing we should change the `control_service_vdk_heartbeat_release` step as well.

mivanov1988:
In order to test the pipeline you should do the following:

1. Add the `external_pull_requests` ref to the required steps (build, publish, release, etc.). This will trigger the step on pull request:

```
only:
    refs:
      - external_pull_requests
``` 
2. Change the version.txt to something like `0.1`.
3. Verify that the artifacts are published.

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -249,7 +250,7 @@ control_service_vdk_heartbeat_release:
   stage: release
   script:
     - export CHART_NAME=pipelines-control-service
-    - export CHART_VERSION=$(cat projects/control-service/projects/helm_charts/$CHART_NAME/version.txt)
+    - export CHART_VERSION=$(cat projects/control-service/projects/helm_charts/$CHART_NAME/version.txt).$CI_PIPELINE_ID
```
tozka, line 253: How are we sure someone will not set patch version in the version.txt file e.g 1.2.0  next time they edit it. 

I'd parse out the major and minor version (as it is in other projects) to be on the safe side.

mivanov1988 (approved):

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -249,7 +250,7 @@ control_service_vdk_heartbeat_release:
   stage: release
   script:
     - export CHART_NAME=pipelines-control-service
-    - export CHART_VERSION=$(cat projects/control-service/projects/helm_charts/$CHART_NAME/version.txt)
+    - export CHART_VERSION=$(cat projects/control-service/projects/helm_charts/$CHART_NAME/version.txt | grep -o '[0-9]\.[0-9]').$CI_PIPELINE_ID
```
tozka, line 253: That would not work if verison if 1.22 or 22.1 (double digits) or 1.22.2.3 . 

Maybe this will:

```suggestion
    - export CHART_VERSION=$(cat projects/control-service/projects/helm_charts/$CHART_NAME/version.txt | grep -o '^[0-9]\+\.[0-9]\+').$CI_PIPELINE_ID
```

tpalashki: Good point. I will actually omit the `^` at the start to allow whitespaces at the beginning of the version.

tpalashki: Actually, the last example still won't work without `^`.

tozka: I think it is more likely for people to leave a space than to use 4 versions. But either way is fine.

PR #359, opened 2021-10-08
vdk-trino: Ensure tests would catch the 'LIKE view' issue

gageorgiev:
A previous PR fixed an issue with the SCD1 template attempt to
create a table LIKE a view, which isn't permitted. This change
ensures that the SCD1 template test would catch that issue.

Testing done: ensured that the new test succeeds on the fixed
template, but fails on the old template


projects/vdk-core/plugins/vdk-trino/tests/jobs/load_dimension_scd1_template_job/01_prepare_input_data.py
```diff
@@ -52,7 +52,7 @@ def run(job_input: IJobInput) -> None:
 
     job_input.execute_query(
         f"""
-            CREATE TABLE IF NOT EXISTS "{source_schema}"."{source_view}" (
+            CREATE TABLE IF NOT EXISTS "{source_schema}"."{source_view}_data" (
```
tozka, line 55: nit: extract `"{source_view}_data"` as local variable.

tozka (approved):

ivakoleva:
Looking at the scd1 template structure, there is `backup table` mentioned, 
Find comment here https://github.com/vmware/versatile-data-kit/pull/356#issuecomment-938428889

gageorgiev:
> > > Looking at the scd1 template structure, there is `backup table` mentioned, Find comment here [#356 (comment)](https://github.com/vmware/versatile-data-kit/pull/356#issuecomment-938428889)
> > 
> > 
> > Because a backup table is created as part of step 6, and step 0 attempts to recover from that backup. If that backup existed, and the recover was successful, the backup should be dropped as it is no longer useful.
> 
> Maybe I missed it, but could you link where the backup table is created?

The last step of the template executes this call https://github.com/vmware/versatile-data-kit/blob/3f2ca68739bdf7bc3306fe5ded138fb69d7707b6/projects/vdk-core/plugins/vdk-trino/src/vdk/plugin/trino/trino_utils.py#L167

PR #365, opened 2021-10-10
vdk-core: enable setting all vdk options using config file

tozka:
Currently, for most of the plugin or vdk configuration, the only way for
users to set it is using environment variables. This is a bit cumbersome
because
- environment variables are temporarily and cannot be shared (e.g in
the job source code). This is making it harder to create working example
jobs for vdk without requiring manual steps. The same goes for the
sample job(s) created during vdk create
- They also are set per shell which makes it harder/error-prone to
switch between shells or IDE and shell - you need to remember to set
them in both
- Since they are per shell. If we have multiple jobs and each requires
different value for the same configuration (e.g VDK_INGEST_METHOD is
different for each), you need to constantly unset and set them.

Here we are adding a new section in the config.ini file called `[vdk]`.
If you set key-value pair in this section it would automatically inejct
it in the configuration class.

Testing Done: unit tests, set those settings in sample job and saw it
succeeds.


gageorgiev (approved):
Great, this is something that has been bugging me for a while.

PR #368, opened 2021-10-11
vdk-core: configuration managment bugfixes

antoniivanov:
There are 2 bugfixes on configuration in this PR (I'd merge them as separate commits) 

     vdk-core: normalize (ignore case) of configuration keys
    
    Configuration keys being case sensitive is just unnecessary problematic
    . It's often pattern when using env variables to use capital letter but
    in files to use lower letters. Currently this is making it difficult
    when we mix config keys from environment variables and files due to mix
    cases.
    
    This is fixing by lower casing all keys passed in configuration.
    
    Testing Done: unit tests

vdk-core: use Enum.value instead of Enum
    
    Using the enum directly as key for configuration makes it impossible to
    set it directly as normal string since they are not equal (being of
    different types). This is using JobConfigKeys.NAME.value to  return
    string representation of the enum
    
    Testing Done: unit tests

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/properties_router.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/properties_router.py
@@ -95,9 +95,9 @@ class PropertiesRouter(IPropertiesRegistry):
                 errors.log_and_throw(
                     errors.ResolvableBy.CONFIG_ERROR,
                     log,
-                    "",  # set by handler
-                    f"Too many choices.",
-                    f"",  # set by handler
+                    "Properties API client cannot be chosen.",  # set by handler
+                    f"Too many choices for properties client implementation.",
+                    f"Properties API functionality does not work.",  # set by handler
                     f"Configure which properties client implementation "
                     f"to use with properties_default_type config option. "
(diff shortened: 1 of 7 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -158,21 +156,24 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             log.info("Detected config.ini. Will try to read config.ini.")
 
             job_config = JobConfig(self.__job_path)
-            config_builder.set_value(JobConfigKeys.TEAM, job_config.get_team())
+            config_builder.set_value(JobConfigKeys.TEAM.value, job_config.get_team())
             config_builder.set_value(
-                JobConfigKeys.NOTIFIED_ON_JOB_FAILURE_USER_ERROR,
+                JobConfigKeys.TEAM, job_config.get_team()
+            )  # kept for backward compatibility
```
gageorgiev, line 162: Why does this need to be kept exactly?

tozka: Because it's used in one plugin. I will update the plugin after releasing this and then remove it.

PR #375, opened 2021-10-12
control-service: change latest_job_termination_status converter

mivanov1988:
Since the database column `latest_job_termination_status` type
is VARCHAR we need to change the hibernate column converter to work with String
DB value rather than Integer.

Testing: unit tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.0.0
+    rev: v4.0.1
     hooks:
     -   id: check-yaml
         args: ['--unsafe']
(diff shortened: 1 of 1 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/model/ExecutionTerminationStatusConverter.java
```diff
@@ -9,18 +9,22 @@
 import javax.persistence.Converter;
 import java.util.Arrays;
 
+import com.google.common.primitives.Ints;
+
 @Converter
-public class ExecutionTerminationStatusConverter implements AttributeConverter<ExecutionTerminationStatus, Integer> {
+public class ExecutionTerminationStatusConverter implements AttributeConverter<ExecutionTerminationStatus, String> {
```
tozka, line 15: It is a good idea to have a test that verifies that a bug fix is truly fixed. This way we increase our confidence in the fix, make sure we won't regress again.

mivanov1988: Done

PR #378, opened 2021-10-12
vdk-plugin-control-cli: add enable skip logic flag and set execution id correctly

mrMoZ1:
why:A change was needed to make the skipping logic work as intended.
Moreover, the need to easily switch off the skipping logic was discussed 
and a turn off plug-able mechanism is introduced with this change.

what: Added hookable key which when overridden disables
the skipping logic entirely. Changed execution_id used from 
configuration to state.

testing: Started local execution with vdk run and made sure
we are using execution_id from state, made sure the turn off
logic works through a local execution.


projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
```diff
@@ -128,14 +131,31 @@ def _skip_job_if_necessary(
     return None
 
 
+@hookimpl(tryfirst=True)
+def vdk_configure(config_builder: ConfigurationBuilder):
+    config_builder.add(
+        key=VDK_EXECUTION_SKIP_CHECKER_ENABLED,
```
tozka, line 137: Remove the `VDK_` prefix. For configuration key XXX it is picked up from environment variable VDK_XXX or from config.ini file option "xxx"

tozka (commented):
Split the change in 2 commits or 2 PRs. If you split it in two commits remember to use Rebase option when merging. 

We do not want to mix changes in different components (plugin and vdk-core in this case). They are managed, developed and released independently.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+            f"Setting: OP_ID: {op_id}, ATTEMPT_ID: {attempt_id}, EXECUTION_ID: {execution_id}"
+        )
         context.state.set(CommonStoreKeys.OP_ID, op_id)
         context.state.set(CommonStoreKeys.EXECUTION_ID, execution_id)
         context.state.set(CommonStoreKeys.ATTEMPT_ID, attempt_id)
 
         context.state.set(CommonStoreKeys.VDK_VERSION, vdk_build_info.RELEASE_VERSION)
         context.state.set(CommonStoreKeys.START_TIME, datetime.utcnow())
 
+    def _set_ids_from_env(self, context: CoreContext):
```
tozka, line 97: you can make it private with double underscore: "__"
```suggestion
    def __set_ids_from_env(self, context: CoreContext):
```

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+        """
+        Private method used to set the attempt id and
+        execution id from values present in the environment.
+        We are not setting  attempt id or execution id anywhere.
+        """
+        attempt_id = getenv(VDK_ATTEMPT_ID, None)
```
tozka, line 103: Why need to use env variables? 

use plugin configuration. The way plugin configuration works is explained if you type `vdk config-help` on top of it . There are different configuration providers. We have two: 
 * One that reads from environment variables 
 * One that reads from config.ini (with lower priority)
 
 So you do not need to use getenv.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+            attempt_id = f"{str(uuid.uuid4())[:6]}"
 
         if not execution_id:
-            execution_id = f"{op_id}-{str(uuid.uuid4())[:6]}"
+            execution_id = attempt_id
 
-        if not attempt_id:
-            attempt_id = f"{execution_id}-{str(uuid.uuid4())[:6]}"
+        if not op_id:
+            op_id = str(int(time.time()))
```
tozka, line 85: If there is not an op_id. I think it should be same as execution_id, right??

What if there is op_id but not execution_id and attempt_id ?

tozka (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
 
         Configuration can override an attempt/op/execution id.
 
-        Otherwise each is auto-generated in format:
-        * Attempt id format is "execution_id-random_suffix"
-        * Execution id format is "op_id-random_suffix"
+        Attempt id and Execution id are extracted from env variables.
+        IF not present they are auto-generated.
```
gageorgiev, line 66: nit
```suggestion
        If not present they are auto-generated.
```

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+
+        If auto generated:
```
gageorgiev, line 68: nit: 'auto-generated' in one place but 'auto generated' in another.
```suggestion
        If auto-generated:
```

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+        * Attempt id format is random string.
```
gageorgiev, line 69:
```suggestion
        * Attempt id format is a random string.
```

projects/vdk-core/plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/execution_skip.py
mrMoZ1: I removed the prefix and split the PR in two.

tozka:
> Would not have been more efficient to leave the other change here. THere are far more comments about the other change.
> 
> Fix the title . `vdk-core: fix execution_id not set properly` is no longer accurate.

I think this is perhaps less informative :) vdk-core: add enable skip logic flag and set execution id correctly

tozka (approved):

PR #380, opened 2021-10-12
control-service: fix vdk-heartbeat

tozka:
The post-deployment test(vdk-heartbeat) seems to be failing with
authorization errors. I guess it is because of misconfiguration.

I set the correct VDK_OPTIONS file in Gitlab CI (it can be overridden by Gitlab CI Variable).
I think it might be better to leave it and edit it directly in Gitlab CI ? 
And keep the one in source control for reference purposes. 

Testing Done: I set VDK_OPTIONS file in
https://gitlab.com/vmware-analytics/versatile-data-kit/-/settings/ci_cd
and ran manually to deploy control service with new options in pipeline https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/386027936

And then ran post-deployment test in same pipeline

### DIFF
--- a/projects/control-service/cicd/vdk-options.ini
+++ b/projects/control-service/cicd/vdk-options.ini
@@ -8,11 +8,6 @@ VDK_TRINO_CATALOG=memory
 VDK_TRINO_USER=data-job-user
 VDK_DB_DEFAULT_TYPE=TRINO
 
-# TODO: this must not be set here.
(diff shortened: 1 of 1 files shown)


gageorgiev (approved):
Let's try to avoid non-specific PR names. Something like `control-service: fix vdk-heartbeat authorization issue` is far more descriptive.

PR #381, opened 2021-10-12
vdk-core: change how we keep execution_id

mrMoZ1:
why: During testing of the new skipping functionality we noticed
that the execution_id was not being properly set and was completely
different compared to the pod name in K8S. A change was needed to
make the skipping logic work as intended.

what: Changed the way we set execution_id and attempt_id.
We now try to read them from the environment and only if not
present we populate them with random values. Previously we
didn't set them anywhere in the code, thus defaulting to setting
random strings. 

testing: Verified execution_id attempt_id are being set with/
without env variable being set through a local execution 
and tracing the logs.

Signed-off-by: Momchil Zhivkov mzhivkov@vmware.com


projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
         context.state.set(CommonStoreKeys.VDK_VERSION, vdk_build_info.RELEASE_VERSION)
         context.state.set(CommonStoreKeys.START_TIME, datetime.utcnow())
 
+    def _set_ids_from_env(self, context: CoreContext):
+        """
+        Private method used to set the attempt id and
+        execution id from values present in the environment.
+        We are not setting  attempt id or execution id anywhere.
+        """
+        attempt_id = getenv(VDK_ATTEMPT_ID, None)
```
tozka, line 103: Why need to use env variables?

use plugin configuration. The way plugin configuration works is explained if you type vdk config-help on top of it . There are different configuration providers. We have two:

One that reads from environment variables
One that reads from config.ini (with lower priority)
So we do not need to use getenv.

tozka:
```suggestion
        attempt_id = context.configuration.get_value(vdk_config.EXECUTION_ID)
```

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+        attempt_id = context.state.get(CommonStoreKeys.ATTEMPT_ID)
 
-        if not op_id:
-            op_id = str(int(time.time()))
+        if not attempt_id:
+            attempt_id = f"{str(uuid.uuid4())[:6]}"
 
         if not execution_id:
-            execution_id = f"{op_id}-{str(uuid.uuid4())[:6]}"
+            execution_id = attempt_id
```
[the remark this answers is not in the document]
mivanov1988: @mrMoZ1 please schedule a meeting for that.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -58,31 +62,50 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
-        attempt_id = context.configuration.get_value(CommonStoreKeys.ATTEMPT_ID.key)
+        self._set_ids_from_env(context)
+        op_id = context.state.get(CommonStoreKeys.OP_ID)
+        execution_id = context.state.get(CommonStoreKeys.EXECUTION_ID)
+        attempt_id = context.state.get(CommonStoreKeys.ATTEMPT_ID)
 
-        if not op_id:
-            op_id = str(int(time.time()))
+        if not attempt_id:
+            attempt_id = f"{str(uuid.uuid4())[:6]}"
```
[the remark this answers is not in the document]
mivanov1988: @mrMoZ1 please schedule a meeting for that.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -50,24 +50,36 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
         execution_id = context.configuration.get_value(vdk_config.EXECUTION_ID)
         attempt_id = context.configuration.get_value(vdk_config.ATTEMPT_ID)
 
-        if not op_id:
-            op_id = str(int(time.time()))
+        if not attempt_id:
+            # UUID is 36 chars
+            attempt_id = (
+                f"{str(uuid.uuid4())}-{str(int(time.time()))}-{str(uuid.uuid4())[:5]}"
+            )
```
tozka, line 69: I think if EXECUTION_ID is provdied we should generate attempt id out of execution id . 
And only if it is not provided generate it from empty. 

Otherwise, We cannot be sure attempt id is in format [execution-id]-{5. chars}  which is what we want.


```suggestion
        if not attempt_id:
            if not execution_id:
                attempt_id = f"{execution_id}-{str(uuid.uuid4())[:5]}"
            else:
                # UUID is 36 chars
                attempt_id = (
                    f"{str(uuid.uuid4())}-{str(int(time.time()))}-{str(uuid.uuid4())[:5]}"
                )
```

mrMoZ1: Shouldn't it be the other way around?

`        if not attempt_id:
            if not execution_id:
                                # UUID is 36 chars
                attempt_id = (
                    f"{str(uuid.uuid4())}-{str(int(time.time()))}-{str(uuid.uuid4())[:5]}"
                )
            else:
                attempt_id = f"{execution_id}-{str(uuid.uuid4())[:5]}"`

tozka: Write unit tests and it would show. 

But the idea is if execution id is NOT empty to use it, otherwise generate it with `{str(uuid.uuid4())}`

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -50,24 +50,42 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
         """
         op_id = context.configuration.get_value(vdk_config.OP_ID)
         execution_id = context.configuration.get_value(vdk_config.EXECUTION_ID)
         attempt_id = context.configuration.get_value(vdk_config.ATTEMPT_ID)
 
-        if not op_id:
-            op_id = str(int(time.time()))
+        if not attempt_id:
+            # Generate attempt id if execution id not present.
+            if not execution_id:
```
mivanov1988, line 67: I don't think we have such a case. We should focus only on what we have.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -50,24 +50,42 @@ def vdk_initialize(self, context: CoreContext) -> None:
...
+            # If env is cloud we must print a warning.
+            if context.configuration.get_value(vdk_config.LOG_CONFIG) == "CLOUD":
+                log.warning(
+                    f"Attempt ID not found in env or configuration. Using auto generated attempt_id: {attempt_id}"
+                )
 
         if not execution_id:
-            execution_id = f"{op_id}-{str(uuid.uuid4())[:6]}"
+            # Delete everything after the last '-' character in the attempt id (inclusive)
+            execution_id = "-".join(attempt_id.split("-")[:-1])
```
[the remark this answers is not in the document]
mrMoZ1: This is how it was initially, but a change was requested: https://github.com/vmware/versatile-data-kit/pull/381#discussion_r728746656

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
mrMoZ1: I did refactored/add a unit test. Could you please check again?

projects/vdk-core/tests/vdk/internal/builtin_plugins/test_builtin_hook_impl.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_op_id_len(self):
+        # same as execution_id
+        assert len(self.context.state.get(CommonStoreKeys.OP_ID)) == 47
+
+    def test_attempt_id_len(self):
+        # execution id len + - + 5 chars = 53
+        assert len(self.context.state.get(CommonStoreKeys.ATTEMPT_ID)) == 53
+
+
+class TestExternalId(unittest.TestCase):
```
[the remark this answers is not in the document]
mivanov1988: ok, but for me it remains an over engineering :)

PR #383, opened 2021-10-13
vdk-core: configure correct attempt, execution and op id.

tozka:
For a configuration variable to be used it needs to be defined with
ConfigurationBuilder#add. If it is not defined the environment variables
configuration provider will likely not detect it.

In our case we define attempt id using:

config_builder.add(CommonStoreKeys.ATTEMPT_ID.key, None)
...

This is the configuration store key which as string look like this:

    ATTEMPT_ID: StoreKey[str] = StoreKey[str]("vdk.attempt_id")

So we basically defined it as `vdk.attempt_id` and the (environment
variables) configuration provider appends VDK_ (and replaces dot with an
underscore). So it searches for
the environment variable named "VDK_VDK_ATTEMPT_ID" in this example.

execution_id and op_id already had correct definitions. In this change
we adding one for op id. And removing redundant ones.
Also try to improve their description a bit.

Testing Done: add new test. I decide to use functional type of test
since the logic spans more than one vdk core component/class instead of
a unit test.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
@@ -18,6 +18,7 @@ JOB_GITHASH = "JOB_GITHASH"
 LOG_CONFIG = "LOG_CONFIG"
 LOG_LEVEL_VDK = "LOG_LEVEL_VDK"
 WORKING_DIR = "WORKING_DIR"
+ATTEMPT_ID = "ATTEMPT_ID"
 EXECUTION_ID = "EXECUTION_ID"
 OP_ID = "OP_ID"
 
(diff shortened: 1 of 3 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -70,20 +71,33 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
-            "Data Job execution can run a Data Job one or more times. ",
+            "Data Job execution can run a Data Job one or more times. "
+            "It can be passed externally (in case of external re-tries are considered part of 1 execution).",
+        )
+        config_builder.add(
+            ATTEMPT_ID,
+            "",
+            True,
+            "An identifier to be associated with the current VDK execution attempt."
+            "If left empty it will be auto-generated. "
```
mivanov1988, line 97: If it is auto-generated the skip logic will not work properly. We should discuss that.

tozka: I am not specifying how it will be auto-generated at this point. 

 If you want I can be more specific: ` If it is left empty it will be auto-generated from op_id or attempt_id if they are provided`

mivanov1988 (approved):

PR #420, opened 2021-10-19
control-service: expose alerting metrics for execution cleanup job

mrMoZ1:
why: The new execution API introduced a periodically running job
which cleans up older executions, and makes sure the executions
stored in the database do not increase over a certain threshold.
As a user I need metrics to monitor the status of the cleanup job,
if I had cleanup failures, or how many executions were deleted. 

what: Exposed execution cleanup metrics which include information
about the number of successful/failed deletions and the number of
invocations of the cleanup job.

testing: added unit tests. Started a locally running tpcs and made sure
metrics are exposed.


tozka:
How are those metrics going to be used? Can you write example PromQL queries that show some monitoring use-case so we are sure the data model (metrics name, tags, values) is correct? Before we jump into implementation first thing we need to make sure that the data model works.

tpalashki:
We can have an offline chat to discuss the rationale about my suggestions.

mrMoZ1:
> How are those metrics going to be used? Can you write example PromQL queries that show some monitoring use-case so we are sure the data model (metrics name, tags, values) is correct? Before we jump into implementation first thing we need to make sure that the data model works.

In the latest changes we have a Counter which counts the invocations of the cleanup job. We can create a metric based on invocation rate in a limited amount of time for example over 24h. This way if there are no invocations that means the cleanup job is not being executed.

In the latest changes we have a gauge that has a value for failed or successful cleanups of data job executions. We can create a metric based on failed execution cleanups.

In the latest changes we have a gauge that has a value for the number of deleted executions per data job. This is optional but we can create an alert based on individual data jobs that we know for a fact should have deletions over a fixed number of time.

PR #458, opened 2021-10-28
control-service: Remove logging of credentials in builder

doks5:
Currently, we are leaking AWS credentials in builder pods. These
are then logged, which is a security concern.

This change removes the leaked credentials from the logs.

Testing Done: Manually changed the version of the builder job to a
dev version, in order to push an image to docker hub. Afterwards, started
a local instance of Control Service, and manually deployed a test data job
and verified that the AWS credentials are no longer being logged.


mivanov1988 (approved):

mivanov1988:
Just to remind you that you should change the helm chart builder version in a separate PR.

projects/control-service/projects/job-builder/build_image.sh
```diff
@@ -6,8 +6,12 @@
...
 # TODO: replace those as env variables
+# Turn off debug mode temporarily to prevent credentials from being logged
+if [[ $- =~ x ]]; then debug=1; set +x; fi
 aws_access_key_id=$1
 aws_secret_access_key=$2
+[[ $debug == 1 ]] && set -x
+
 aws_region=$3
 docker_registry=$4
 git_username=$5
```
tozka, line 17: What about git creds or registry creds? 

Would it not be simpler to disable debug mode of the shell and add prints to log what is needed.

doks5: The focus of the story was the removal of AWS credentials. However, yes, turning off debug mode would be much better. The question is, what things then need to be logged?

tozka: Use your own judgment. Very least the nonsensitive arguments make sense to be logged.

PR #462, opened 2021-10-29
[control-service]:fix memoryToMB conversion overflow in KubernetesService

mrMoZ1:
why: We noticed in the logs that a running control service 
would update a data job execution with a negative value for 
memory allocation. As a user I want to know for certain the 
allocated memory resources for a data job and a negative
value is an obvious bug.

what: Changed culprit function's implementation to use BigInteger instead
of int as the calculations it was performing would easily overflow the int
value at around 2gb or more.

testing: Added unit tests that verify function is working as intended
on small and large conversion numbers.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
+++ b/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
@@ -218,4 +218,47 @@ public class KubernetesServiceTest {
         }
     }
 
+    @Test
+    public void testQuantityToMbConversionMegabytes() throws Exception {
+        var hundredMb = "100M";
+        testQuantityToMbConversion(100, hundredMb);
+
+        var thousandMb = "1000M";
+        testQuantityToMbConversion(1000, thousandMb);
+        // this should be enough for overflow errors on the method to manifest.
+        var tenThousandMb = "10000M";
+        testQuantityToMbConversion(10000, tenThousandMb);
+
+        var hundredThousandMb = "100000M";
+        testQuantityToMbConversion(100000, hundredThousandMb);
+
+    }
+
+    @Test
+    public void testQuantityConversionGigabytes() throws Exception {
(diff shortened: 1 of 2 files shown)


projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
```diff
@@ -218,4 +218,47 @@ public void testCreateV1beta1CronJobFromInternalResource() {
...
+        testQuantityToMbConversion(3000, threeG);
+
+        var fourG = "4G";
+        testQuantityToMbConversion(4000, fourG);
+
+        var sixtyFourG = "64G";
+        testQuantityToMbConversion(64000, sixtyFourG);
+    }
+
+    public void testQuantityToMbConversion(int expectedMb, String providedResources) throws Exception {
```
mivanov1988, line 255: I prefer you to make the method "default". If something change an error will occur during the compile time.

mivanov1988:
Good job.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
mrMoZ1: Done.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
```diff
@@ -218,4 +218,45 @@ public void testCreateV1beta1CronJobFromInternalResource() {
...
+        KubernetesService service = new DataJobsKubernetesService("default", "someConfig");
```
mivanov1988, line 256: Since convertMemoryToMBs() is static it should be available at class level - KubernetesService.convertMemoryToMBs(q);

mivanov1988 (approved):

PR #470, opened 2021-11-02
control-service: fix NPE in JobExecutionCleanup service

mrMoZ1:
why: During monitoring of a running control-service's logs
we noticed that the cleanup task would fail for some jobs.
This was caused by a null end_time for the execution in the
database, which was in turn caused by an issue that is already
fixed, yet the null data remained. However, the null end_time
was causing the cleanup job to fail for the entire data job thus
no execution cleanup was happening at all for that particular
data job's executions.

what: Introduced a null check in the culprit code. Executions
with null end_time will now be ignored when selecting executions
to delete by the age criteria. However, executions will still be
deleted when they exceed over 100 even if the end_time is null.

testing: Added unit tests.


doks5 (approved):
LGTM

Issue #377, opened 2021-10-12
Override vdk version of deployed job


tozka:
**What is the feature request? What problem does it solve?**
 Currently, all jobs use the same vdk version in which to run. This makes it easier to maintain thousands of jobs as they are uniform. 
 But if we want to do a canary release of vdk by rolling out the initial version of vdk to a subset of users as an initial test. 
 Or we want to do A/B testing with different flavors of vdk. We need to be able to override the vdk version for subset of jobs.

Another use-case is when vdk introduces a major release (aka breaking change). We'd need to enable users to use both previous and new major releases concurrently so they can gradually switch to the new release.

**Suggested solution**
 During deploy we can set not only job version (as we do not) but also [vdk version]([https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/model/apidefs/datajob-api/api.yaml#L1123]) (by default it's left empty so it uses latest) 
 This would all us to override the vdk version per deployment basis

**Acceptance Criteria**
 - I can deploy a single job with a specific (possibly non-officially released) version of vdk in cloud runtime. While all the other jobs use the common version.

tozka:
Control Service API has an option by design to set vdk_version which will override default vdk version. 

Use cases 

1. To enable users to upgrade in a self-service manner when there is major (breaking) upgrade of vdk.  
   When there is a breaking change (in the VDK library for example) it may take time for user to migrate data jobs to new "syntax"/python interfaces. So we need to enable them to do it in their own time. 
   
 2. Another one is to enable to run canary release of vdk. All jobs run in the Versatile Data Kit managed runtime use the same version and are upgraded automatically (and at the same time). This make sure that the new release indeed work , it is configured correctly and job would succeed. 
 
 3. Similarly to 2 A/B testing - introduce different version of vdk with different semantics - e.g different query optimization logic and compare which one works better. 
 
4. Local development. vdk-server plugin seem useful for local development. If you are making changes on VDK project only , there's no need to use the Control Service code base at all. Start vdk server locally and test how the job un in managed runtime. But vdk-server need to allow to set a local version of vdk. One option is to be addressed by https://github.com/vmware/versatile-data-kit/issues/392 But if user can override vdk image per job during deployment - that's another option (vdk deploy --update --vdk-version=image:v.1.0.1

PR #476, opened 2021-11-04
control-service: return and udpate vdkVersion in Deployment

tozka:
VDK version is the version of the VDK library with which the data job
run. We need it to return it when querying for deployment informatin in
order to know with which version of vdk the data job is configured to
run in cloud(kubernetes) deployment. 
See details on the analysis of how we update the vdk version in https://github.com/vmware/versatile-data-kit/issues/377

GET deployment (or list of deployments) returns the configured vdk
version

Testing Done: integration tests, unit tests.
Issue: https://github.com/vmware/versatile-data-kit/issues/377


YanaZhivkova (approved):
Look good to me, maybe we need some documentation for the new class.

mivanov1988:
Most of the time VDK version will be _latest_ right?

tozka:
> Most of the time VDK version will be _latest_ right?

Yes, Or "release" since we use release tag - https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/cicd/deploy-base-vdk-image.sh#L13

mivanov1988 (approved):

PR #478, opened 2021-11-04
control-service: GraphQL executions query

mivanov1988:
Extended the GraphQL schema (resources/schema.graphqls) by adding
an additional query called _executions_. It will allow us to filter
and sort executions by the following fields: status, startTime, endTime.

Sample request:
```
{{ baseUrl }}/data-jobs/for-team/no-team-specified/jobs?query={
  executions(filter: {
		endTime_gte: "2021-10-19T10:14:47Z",
		status_in: ["failed", "cancelled"]
             },
	     order: {property: "endTime", sort: DESC}) {
    content {
          id
          startTime
          endTime
          status
          jobName
    }
  }
}
```

Еvaluated an advanced filtering approach, but it seems to be out of scope
of this issue. It will take more implementation time since we should
introduce a new query engine.

```
{{ baseUrl }}/data-jobs/for-team/no-team-specified/jobs?query={
  executions(filter: [
		{op: gte, property: "endTime", value: "2021-10-19T10:14:47Z"},
		{op: in, property: "status", values: ["failed", "cancelled"]}
	     ],
             order: {property: "status", sort: DESC}) {
    content {
          id
          startTime
          endTime
          status
          jobName
    }
  }
}
```

Testing: unit tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
+++ b/projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
@@ -1,9 +1,12 @@
+scalar DateTime
+
 schema {
     query: Query
 }
 
 type Query {
     jobs(pageNumber: Int = 0, pageSize: Int = 20, filter: [Predicate], search: String): DataJobPage
+    executions(pageNumber: Int, pageSize: Int, filter: DataJobExecutionFilter, order: DataJobExecutionOrder): DataJobExecutionResponse
 }
 
 input Predicate {
@@ -17,9 +20,26 @@ enum Direction {
     DESC
 }
 
+input DataJobExecutionFilter {
+    startTimeGte: DateTime,
+    endTimeGte: DateTime,
+    statusIn: [DataJobExecutionStatus]
+}
+
+input DataJobExecutionOrder {
+    property: String,
+    direction: Direction
+}
+
 # TODO use auto-generation [TAUR-1376]
 type DataJobPage {
     content: [DataJob]
+    totalPages: Int
+    totalItems: Int
+}
+
+type DataJobExecutionResponse {
+    content: [DataJobExecution]
     totalPages: Int
     totalItems: Int
 }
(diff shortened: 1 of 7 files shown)


projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
```diff
@@ -17,6 +18,17 @@ enum Direction {
     DESC
 }
 
+input DataJobExecutionFilter {
+    startTime_gte: String,
```
tozka, line 22: If you need `gte`, you would then `lte` (less than). Searching for between two periods is pretty normal. 
 
Actually for pagination might be the only way

tozka:
> Еvaluated an advanced filtering approach, but it seems to be out of scope

I was just looking at how [dgraph](https://dgraph.io/docs/graphql/schema/search/) and [hasura](https://hasura.io/docs/latest/graphql/core/databases/postgres/queries/query-filters.html#greater-than-or-less-than-operators-gt-lt-gte-lte) handle that problem. Even if it's out of scope for now. Do you think it's worth going in that direction?

mivanov1988:
> > Еvaluated an advanced filtering approach, but it seems to be out of scope
> 
> I was just looking at how [dgraph](https://dgraph.io/docs/graphql/schema/search/) and [hasura](https://hasura.io/docs/latest/graphql/core/databases/postgres/queries/query-filters.html#greater-than-or-less-than-operators-gt-lt-gte-lte) handle that problem. Even if it's out of scope for now (with which I agree). Do you think it's worth going in that direction?

Yes, I think that's the way to go. I really like the _where_ clause approach. I have played with the [ontotext tutorial](https://platform.ontotext.com/tutorials/graphql-query.html) it is pretty good.

projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
mivanov1988: Pagination has not been added. Do you think we need one for this particular use case?

tozka: Pagination needs depend on the number of items the API could return. If it is potentially a high number (maybe over 100 in our case though it depends on the size of each record) then pagination would be necessary. 

Using page_size and page_number might be the normal way to paginate.

 I just think if you need GTE , user would need less (LTE). So if the cost is minimal (e.g 10% to add both) , it might be better to add both now. But I am ok either way.

tozka (approved):

projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
mivanov1988: Ok, thanks.

PR #479, opened 2021-11-04
control-service: Classify OOM as User Errors

doks5:
Currently, if a data job execution exceeds the allowed memory
quota, the pod of the execution is immediately killed by
Kubernetes and restarted. Which in turn usually results in a
similar failure. This causes a Platform Error to be raised, whereas
it should be classified as a User Error.

This change re-classifies such errors as User Errors.

Testing Done: Added unit test.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -140,7 +140,7 @@ public abstract class KubernetesService implements InitializingBean {
         String executionType;
         String jobName;
         String terminationMessage;
-        String terminationReason;
+        String jobTerminationReason;
         Boolean succeeded;
         String opId;
         OffsetDateTime startTime;
@@ -153,6 +153,7 @@ public abstract class KubernetesService implements InitializingBean {
         Integer resourcesMemoryLimit;
         OffsetDateTime deployedDate;
         String deployedBy;
+        String containerTerminationReason;
     }
 
     @AllArgsConstructor
@@ -799,8 +800,14 @@ public abstract class KubernetesService implements InitializingBean {
             lastTerminatedPodState
                   .map(v1ContainerStateTerminated -> StringUtils.trim(v1ContainerStateTerminated.getMessage()))
                   .ifPresent(s -> jobExecutionStatusBuilder.terminationMessage(s));
-            jobExecutionStatusBuilder.terminationReason(jobStatusCondition.getReason());
+            jobExecutionStatusBuilder.jobTerminationReason(jobStatusCondition.getReason());
+
+            // Termination Reason of the data job pod container
+            lastTerminatedPodState
+                    .map(v1ContainerStateTerminated -> StringUtils.trim(v1ContainerStateTerminated.getReason()))
+                    .ifPresent(s -> jobExecutionStatusBuilder.containerTerminationReason(s));
         }
+
         // Job resources
         Optional<V1Container> containerOptional = Optional.ofNullable(job.getSpec())
               .map(V1JobSpec::getTemplate)
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
@@ -30,6 +30,7 @@ public class JobExecutionResultManager {
    static final String TERMINATION_MESSAGE_ATTRIBUTE_STATUS = "status";
    static final String TERMINATION_MESSAGE_ATTRIBUTE_VDK_VERSION = "vdk_version";
    static final String TERMINATION_REASON_DEADLINE_EXCEEDED = "DeadlineExceeded";
+   static final String TERMINATION_REASON_OUT_OF_MEMORY = "OOMKilled";
 
    @Data
    @Builder
@@ -51,7 +52,7 @@ public class JobExecutionResultManager {
       ExecutionTerminationStatus terminationStatus = getTerminationStatus(terminationMessage.getTerminationStatus());
 
       terminationStatus = updateTerminationStatusBasedOnExecutionStatus(
-              terminationStatus, executionStatus, terminationMessage.getTerminationStatus(), jobExecution.getTerminationReason());
+              terminationStatus, executionStatus, terminationMessage.getTerminationStatus(), jobExecution.getJobTerminationReason(), jobExecution.getContainerTerminationReason());
       executionStatus = updateExecutionStatusBasedOnTerminationStatus(executionStatus, terminationStatus);
 
       return ExecutionResult.builder()
@@ -109,7 +110,8 @@ public class JobExecutionResultManager {
     * @param terminationStatus termination status based on the K8S Pod termination status
     * @param executionStatus execution status based on K8S Job status
     * @param terminationStatusString termination status returned from K8S Pod (e.g. "Success", "User error", etc.)
-    * @param terminationReason condition reason as reported by K8s Job (e.g. "DeadlineExceeded", "BackoffLimitExceeded", etc.)
+    * @param jobTerminationReason condition reason as reported by K8s Job (e.g. "DeadlineExceeded", "BackoffLimitExceeded", etc.)
+    * @param containerTerminationReason termination reason for pod container as reported by K8s Job (e.g., "OOMKilled", etc.)
     * @return if there is no termination message due to the missing K8S Pod
     * returns termination status based on execution status otherwise returns
     * termination status based on the K8S Pod termination status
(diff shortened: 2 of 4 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
```diff
@@ -123,7 +124,8 @@ private static ExecutionTerminationStatus updateTerminationStatusBasedOnExecutio
       if (StringUtils.isEmpty(terminationStatusString) && ExecutionStatus.FINISHED.equals(executionStatus)) {
          terminationStatus = ExecutionTerminationStatus.SUCCESS;
       } else if (StringUtils.isEmpty(terminationStatusString) && ExecutionStatus.FAILED.equals(executionStatus)) {
-         if (StringUtils.equalsIgnoreCase(terminationReason, TERMINATION_REASON_DEADLINE_EXCEEDED)) {
+         if (StringUtils.equalsIgnoreCase(terminationReason, TERMINATION_REASON_DEADLINE_EXCEEDED) ||
+                 StringUtils.equalsIgnoreCase(terminationReason, TERMINATION_REASON_OUT_OF_MEMORY)) {
```
tozka, line 128: Now it is user error. In what way would we indicate to the end-user , if it was deadline exceeded or memory. 

In other words we need to follow https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/CONTRIBUTING.md#error-handling

tozka (approved):
LGTM. Let's merge this. But we need to make sure we can report to the user what is going on correctly (see my other comment). We can handle this in separate PR.

tpalashki:
Are you sure that the `reason` field of the Kubernetes job, in this case, will be "OOMKilled"? I think the job will have a `status.condition[0].reason: BackoffLimitExceeded` and the `OOMKilled` will be set as a status of the pod itself:
```
status:
  containerStatuses:
  - ...
    lastState:
      terminated:
        containerID: ...
        exitCode: 137
        finishedAt: "2021-10-31T09:55:22Z"
        reason: OOMKilled
        startedAt: "2021-10-31T09:47:17Z"
```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
mivanov1988: In my opinion we should split the FAILED status into USER_ERROR and PLATFORM_ERROR. Then we will be able to fill in the _message_ field correctly - e.g. "OOM".

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
```diff
@@ -118,12 +119,14 @@ private static ExecutionTerminationStatus updateTerminationStatusBasedOnExecutio
          ExecutionTerminationStatus terminationStatus,
          ExecutionStatus executionStatus,
          String terminationStatusString,
-         String terminationReason) {
+         String terminationReason,
```
tpalashki, line 122: Maybe rename this to `jobTerminationReason` now that we have two different reasons.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -801,6 +802,14 @@ private static String getContainerName(V1Job job) {
                   .ifPresent(s -> jobExecutionStatusBuilder.terminationMessage(s));
             jobExecutionStatusBuilder.terminationReason(jobStatusCondition.getReason());
         }
+
+        Optional<V1ContainerStateTerminated> lastContainerState = getTerminationStatus(job);
```
tpalashki, line 806: Is there any reason for not putting this into the `if` above, where the `lastTerminatedPodState` is already acquired? This will save an extra call to the K8s API to list all of the job pods.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -801,6 +802,14 @@ private static String getContainerName(V1Job job) {
...
+        if (lastContainerState.isPresent()) {
```
mivanov1988, line 807: This check is redundant since the lastContainerState is _Optional_.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
doks5: Created issue for this here: https://github.com/vmware/versatile-data-kit/issues/484

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
doks5: Done, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: From what I understand in terms of the way kubernetes works, it is possible that at the time the `if` block above is executed, the pod of the data job might no longer exist, which would set the container termination reason as `null`.
If my understanding is incorrect, I am happy to move the logic inside the `if`, as that would indeed save extra calls to the K8s API.

tpalashki: The code inside the `if` processes the termination message of the pod, so I believe the pod should still exist at this point.

doks5: Moved the logic inside the `if` block. Thanks for the review

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -800,7 +801,14 @@ private static String getContainerName(V1Job job) {
                   .map(v1ContainerStateTerminated -> StringUtils.trim(v1ContainerStateTerminated.getMessage()))
                   .ifPresent(s -> jobExecutionStatusBuilder.terminationMessage(s));
             jobExecutionStatusBuilder.terminationReason(jobStatusCondition.getReason());
+
+            // Termination Reason of the data job pod container
+            Optional<V1ContainerStateTerminated> lastContainerState = getTerminationStatus(job);
```
tpalashki, line 806: You already have the status on line 797 - that was the idea behind moving it in the `if`.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
```diff
@@ -109,7 +110,7 @@ private static ExecutionTerminationStatus getTerminationStatus(String podTermina
     * @param terminationStatus termination status based on the K8S Pod termination status
     * @param executionStatus execution status based on K8S Job status
     * @param terminationStatusString termination status returned from K8S Pod (e.g. "Success", "User error", etc.)
-    * @param terminationReason condition reason as reported by K8s Job (e.g. "DeadlineExceeded", "BackoffLimitExceeded", etc.)
+    * @param jobTerminationReason condition reason as reported by K8s Job (e.g. "DeadlineExceeded", "BackoffLimitExceeded", etc.)
     * @return if there is no termination message due to the missing K8S Pod
```
tpalashki, line 114: You can add a comment for the new parameter.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: Good catch, removed. Thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionResultManager.java
doks5: Comment added, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -153,6 +153,7 @@ private static long fromDateTime(DateTime value) {
         Integer resourcesMemoryLimit;
         OffsetDateTime deployedDate;
         String deployedBy;
+        String containerTerminationReason;
```
tpalashki, line 156: nit: maybe change `terminationReason` to `jobTerminationReason` here as well.

tpalashki (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: I was thinking of doing it, but the attribute is directly invoked in a large number of places, and the cost of the change seemed unreasonable.

PR #480, opened 2021-11-05
control-service: add last execution info to the data_job table

tpalashki:
As part of preparation for the implementation of the ability to filter
and sort data jobs by their last execution status, time, and duration,
add the respective columns to the data_job table.

Testing done: Start the service and verify that the columns are created
in the data_job table.


projects/control-service/projects/pipelines_control_service/src/main/resources/db/migration/V20213110120000__add_last_execution_info_data_job.sql
```diff
@@ -0,0 +1,4 @@
+alter table data_job
```
mivanov1988, line 1: The script should be idempotent (https://www.cockroachlabs.com/docs/stable/add-column.html).

`alter table if exists ... add column if not exists ...`

tpalashki: Will do, thanks!

mivanov1988 (approved):

PR #482, opened 2021-11-05
vdk-control-cli: add command to set vdk version

tozka:
Setting vdk version is someitmes necessary to enable canary releases and
A/B Testing. See https://github.com/vmware/versatile-data-kit/issues/377

Adding command to set vdk version. It is hidden for now as it's
experimental

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
@@ -4,6 +4,7 @@ import glob
 import json
 import logging
 import os
+from typing import Optional
 
 import click
 import click_spinner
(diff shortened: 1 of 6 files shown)


projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
```diff
@@ -14,7 +15,6 @@
 from taurus_datajob_api import DataJobContacts
 from taurus_datajob_api import DataJobDeployment
 from taurus_datajob_api import DataJobSchedule
```
gageorgiev, line 17: Unrelated but isn't taurus_datajob_api an internal vmware resource? Should we have it as a vdk-control-cli dependency?

gageorgiev (approved):

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
tozka: No, this is part of vdk-control-service-api  which is autogenerate from api.yaml open API spec. We have set the package name to be this: https://github.com/vmware/versatile-data-kit/blob/4cdbe3b0cfedb8ddc675bec27aee4a88389ce16d/projects/control-service/projects/model/apidefs/datajob-api/config-python.json#L2

We can change it to vdk_datajob_api probably as it makes more sense.

gageorgiev: Gotcha.

Issue #461, opened 2021-10-29
Impala Connection Plugin Query Support


ivakoleva:
**What is the feature request? What problem does it solve?**
Generic plugin for Impala connection and query support.
Use latest driver.

**Suggested solution**
Create vdk-impala plugin, providing the Impala connection (similar to vdk-greenplum).

Acceptance criteria:
Support for query execution.

PR #485, opened 2021-11-08
vdk-impala: Create plugin

gageorgiev:
VDK should provide connectors for all popular databases.
This change provides a connection module for Impala.
Addresses #461

Testing done: included unit test which is executed against
a local Impala instance which is raised as part of the test
cycle

### DIFF
--- a/projects/vdk-core/plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
+++ b/projects/vdk-core/plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
@@ -1,0 +1,143 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import click
+from tabulate import tabulate
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins.run.job_context import JobContext
+from vdk.internal.core.config import Configuration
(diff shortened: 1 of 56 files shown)


projects/vdk-core/plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -0,0 +1,91 @@
...
+        key="IMPALA_HOST",
+        default_value="localhost",
+        description="The Impala host we want to connect to.",
+    )
+    config_builder.add(
+        key="IMPALA_PORT",
+        default_value=21050,
+        description="The Impala port we want to connect to.",
+    )
+    config_builder.add(key="IMPALA_DATABASE", default_value=None, description="")
```
ivakoleva, line 47: For descriptions, you can find details in impala.dbapi connect() pydocs

ivakoleva (approved):

ivakoleva:
Maybe link PR to https://github.com/vmware/versatile-data-kit/issues/461

projects/vdk-core/plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
gageorgiev: Thank you for the suggestion.

PR #486, opened 2021-11-09
control-service: Update the GraphQL model with the last execution

tpalashki:
Extended the GraphQL and service models to reflect the properties that were added
as part of the implementation of the ability to sort and filter data jobs by their
last execution status, time, and duration.

Wire the two models so that the newly added properties can be queried via GraphQL.

Testing done: Run the service locally and verified that the new properties are
returned when requested.


mivanov1988:
LGTM: do we have any tests for this?

tpalashki:
I haven't added tests yet as it is mostly plumbing code. I will cover when I have the actual implementation of the filtering/sorting in place.

mivanov1988:
> I haven't added tests yet as it is mostly plumbing code. I will cover when I have the actual implementation of the filtering/sorting in place.

Yes, but since it is possible to adopt it in production it may be a good idea to add some.

mivanov1988 (approved):

PR #490, opened 2021-11-09
control-serivice: make smtp server configurable

tozka:
The smtp server used for deployment notifications was hard-coded which
means it would not have worked .

Testing Done: set datajobs.notifications.smtp.host in
application-properties and saw it is setting the correct value.


YanaZhivkova (approved):
What do you mean "it would not have worked"? Where?

mivanov1988 (approved):

PR #491, opened 2021-11-09
control-service: fix NPE when non-datajob cron job exists

tozka:
When there's cronjob in teh kubernetes namespace that is not a data job
deployment, we fail with NPE

We should likely filter them better so data job specific logic is not
applied to non-datajob cronjobs

Testing Done: integration test (created cron job in the namespace where
integration test runs)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -1355,7 +1355,7 @@ private V1Secret buildV1Secret(String name, Map<String, byte[]> data) {
                 deployment.setImageName(image); // TODO do we really need to return image_name?
             }
             var initContainers = cronJob.getSpec().getJobTemplate().getSpec().getTemplate().getSpec().getInitContainers();
-            if (!initContainers.isEmpty()) {
+            if (initContainers != null && !initContainers.isEmpty()) {
```
mivanov1988, line 1358: How about CollectionUtils.isNotEmpty()?

tpalashki (approved):

PR #492, opened 2021-11-09
[vdk-core]: Post-ingestion specification

doks5:
With the increase in the number of use-cases for ingestion,
there arises the need for some users to do post-processing
of ingestion data (e.g, to collect ingestion telemetry, etc.).

This change aims at introducing a new interface for plugin developers
to use when they want to do further processing of payloads after the
ingestion operation is complete, or to send telemetry data based on the
ingestion operation.

Testing Done: TODO after the spec is finalised.

### DIFF
--- a/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
+++ b/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
@@ -8,6 +8,7 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import List
+from typing import NewType
 from typing import Optional
 from typing import Union
 
@@ -164,13 +165,16 @@ class IManagedConnectionRegistry:
 
 
 class IIngesterPlugin:
+    IngestionResult = NewType("IngestionResult", Dict)
+
+    @abstractmethod
     def ingest_payload(
         self,
         payload: List[dict],
         destination_table: Optional[str],
         target: Optional[str] = None,
         collection_id: Optional[str] = None,
-    ):
+    ) -> Optional[IngestionResult]:
         """
         Do the actual ingestion of the payload
 
@@ -198,6 +202,61 @@ class IIngesterPlugin:
             invocations belong to same collection. Defaults to "data_job_name|OpID",
             meaning all method invocations from a data job run will belong to the
             same collection.
+
+        :return: [Optional] IngestionResult, containing plugin-defined
+        information about the result from the ingestion operation. This result
+        could be used for post-ingestion processing operations like sending
+        telemetry data about the ingested payload.
+
+        :exceptions: Exceptions raised while ingesting data can either be
+        handled at plugin level, or they will be automatically caught at the
+        vdk-core level.
+        """
+        pass
+
+    def post_ingest_process(
+        self,
+        payload: Optional[List[dict]] = None,
+        ingestion_result: Optional[IngestionResult] = None,
+        exceptions: Optional[List] = None,
+    ) -> None:
+        """
+        Do post-ingestion processing of the ingestion payload
+        or result data from the ingestion operation.
+
+        Example Implementation for telemetry:
+
+        .. code-block:: python
+
+            def post_ingest_process(
+                self,
+                payload: Optional[List[dict]],
+                ingestion_result: Optional[IngestionResult]
+                exceptions: Optional[List]
+            ) -> None:
+                # Prepare telemetry
+                telemetry = {}
+                payload_sizes = [sys.getsizeof(i) for i in payload]
+                telemetry['payload_size'] = sum(payload_sizes)
+                telemetry['caught_ingest_exceptions'] = exceptions
+                telemetry |= ingestion_result
+
+                # Send telemetry to wherever is needed.
+                requests.post('example.com', data=telemetry)
+
+
+        :param payload: Optional[List[dict]]
+            The payloads that have been ingested. Depending on the number
+            of payloads to be processed, there might 0 or many dict objects.
+            Each dict object is a separate payload.
+        :param ingestion_result: Optional[IngestionResult]
+            The result from the ingestion operation, or some information about
+            the ingestion operation and payload that plugin developers have
+            decided to process further. This result can be either an
+            IngestionResult object (which is effectively a dictionary), or
+            None.
+        :param exceptions: Optional[List]
+            A list of all exceptions (if any) encountered while ingesting data.
         """
         pass
 
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
@@ -7,7 +7,9 @@ import sys
 import threading
 from collections import defaultdict
 from typing import Iterable
+from typing import List
 from typing import Optional
+from typing import Tuple
 
 from vdk.api.job_input import IIngester
 from vdk.api.plugin.plugin_input import IIngesterPlugin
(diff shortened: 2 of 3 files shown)


projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -198,6 +202,33 @@ def ingest_payload(
...
+        pass
+
+    def post_ingest_process(
+        self,
+        payload: Optional[List[dict]] = None,
+        ingestion_result: Optional[IngestionResult] = None,
+    ) -> None:
+        """
+        Do post-ingestion processing of the ingestion payload
+        or result data from the ingestion operation.
```
ivakoleva, line 220: What if ingestion raised an exception, is the error a specific outcome of the ingestion operation? If so, there would be an use case that telemetry would use this capability to monitor ingestion errors also? 

Or perhaps another capability in addition to this one, trying to understand should we have one or more hooks for both use cases

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -198,6 +202,33 @@ def ingest_payload(
             invocations belong to same collection. Defaults to "data_job_name|OpID",
             meaning all method invocations from a data job run will belong to the
             same collection.
+
+        :return: [Optional] IngestionResult, containing plugin-defined
```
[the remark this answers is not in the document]
doks5: A possible use-case could be if a plugin developer decides that their ingestion mechanism returns some data, e.g., after a request is made to an ingestion endpoint, the said endpoint may return information like status code, an ingestion id, etc., and this data may, then be subject of further processing as telemetry.
Does the example make sense?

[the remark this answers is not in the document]
ivakoleva: I am not aware of such an use case, for ingestion to return metadata on data ingestion itself.
Telemetry-wise, I believe we are interested if data was ingested or not, if not eventually what the issue was.

There is no explicit pydoc on ingest_payload() in case of ingestion error, yet I would presume an exception is raised (evaluated in try block).

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Did some refactoring. Now, it should be able to catch all exceptions and pass them as a list to the post-ingestion hook.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
doks5: Well, different ingestion plugins may have different ways of handling ingestion. For example, a plugin may calculate payload size (compressed, uncompressed) before ingesting the data, or do some other calculation. Additionally, an ingestion plugin may do its own exception handling, and passing the information about the exceptions as an IngestionResult would allow for more granular data gathering.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -510,9 +517,23 @@ def _payload_poster_thread(self):
                         "One or more rows were not ingested.\n"
                         "Exception was: {}".format(str(e))
                     )
+                    exceptions_list.append(e)
             finally:
                 self._payloads_queue.task_done()
 
+            # Complete Post-Ingestion operations
+            try:
+                self._ingester.post_ingest_process(
```
ivakoleva, line 526: In case of ingestion exception, should the exception be a part of the telemetry?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
ivakoleva: Got it, thank you for elaborating

ivakoleva (approved):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -198,6 +202,40 @@ def ingest_payload(
...
+        pass
+
+    def post_ingest_process(
+        self,
+        payload: Optional[List[dict]] = None,
+        ingestion_result: Optional[IngestionResult] = None,
+        exceptions_list: Optional[List] = None,
+    ) -> None:
+        """
+        Do post-ingestion processing of the ingestion payload
```
tozka, line 224: Provide an example in the pydoc of how it can be used - simple implementation of the method.

tozka (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
doks5: Well, all exceptions caught in the ingestion process would be added to the exceptions list, and passed to the post-ingestion process. It would, however, be up to the plugin developer if they want to add the exceptions to the telemetry, or not.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -198,6 +202,40 @@ def ingest_payload(
             invocations belong to same collection. Defaults to "data_job_name|OpID",
             meaning all method invocations from a data job run will belong to the
             same collection.
+
+        :return: [Optional] IngestionResult, containing plugin-defined
+        information about the result from the ingestion operation. This result
+        could be used for post-ingestion processing operations like sending
+        telemetry data about the ingested payload.
+
+        :exceptions: Exceptions raised while ingesting data can either be
```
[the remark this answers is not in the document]
doks5: Good point. Fixed

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Done, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -198,6 +202,63 @@ def ingest_payload(
...
+            The payloads that have been ingested. Depending on the number
+            of payloads to be processed, there might 0 or many dict objects.
+            Each dict object is a separate payload.
+        :param ingestion_result: Optional[IngestionResult]
+            The result from the ingestion operation, or some information about
+            the ingestion operation and payload that plugin developers have
+            decided to process further. This result can be either an
+            IngestionResult object (which is effectively a dictionary), or
+            None.
+        :param exceptions: Optional[List]
```
ivakoleva, line 260: Is this a valid use case, having a list of exceptions?

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
```diff
@@ -184,3 +184,37 @@ def test_ingest_payload_multiple_destinations():
...
+    ingester_base._ingester.ingest_payload.assert_called_with(
+        payload=test_aggregated_payload,
+        destination_table=destination_table,
+        target=target,
+        collection_id=collection_id,
+    )
+    ingester_base._ingester.post_ingest_process.assert_called_with(
+        payload=test_aggregated_payload,
+        ingestion_result=test_ingestion_result,
+        exceptions=[],
```
ivakoleva, line 219: Maybe add a test where an exception is propagated to the post-ingest hook implementation?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: It might be possible, but it is highly improbable. Decided to use a list just in case.

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
doks5: Added, thanks

tozka (approved):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -198,6 +202,63 @@ def ingest_payload(
...
+        """
+        Do post-ingestion processing of the ingestion payload
+        or result data from the ingestion operation.
+
+        Example Implementation for telemetry:
+
+        .. code-block:: python
+            import sys
+            import requests
+
```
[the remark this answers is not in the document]
doks5: Good point. Removed, thanks

PR #494, opened 2021-11-10
control-service: fix a bug when updating the notification status

tpalashki:
There is a bug when checking whether to update the termination status of
a data job. Rather than checking if the executions are different, we
are checking that they are the same.

This commit fixes the bug and also adds more logging when an update is skipped.

Testing done: new unit tests are added

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/monitoring/DataJobMonitorTest.java
+++ b/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/monitoring/DataJobMonitorTest.java
@@ -221,7 +221,7 @@ public class DataJobMonitorTest {
         dataJobMonitor.watchJobs();
 
         var gauges = meterRegistry.find(DataJobMetrics.TAURUS_DATAJOB_TERMINATION_STATUS_METRIC_NAME).gauges();
-        Assertions.assertEquals(5, gauges.size());
+        Assertions.assertEquals(4, gauges.size());
         jobStatuses.forEach(s -> {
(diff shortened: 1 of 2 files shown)


YanaZhivkova (approved):

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/monitoring/DataJobMonitorTest.java
```diff
@@ -221,7 +221,7 @@ public void testWatchJobsWithTheSameStatus() throws IOException, ApiException {
         dataJobMonitor.watchJobs();
 
         var gauges = meterRegistry.find(DataJobMetrics.TAURUS_DATAJOB_TERMINATION_STATUS_METRIC_NAME).gauges();
-        Assertions.assertEquals(5, gauges.size());
+        Assertions.assertEquals(4, gauges.size());
```
YanaZhivkova, line 224: I don't understand this line, why is it changed from 5 to 4?

YanaZhivkova (approved):

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/monitoring/DataJobMonitorTest.java
tpalashki: The `watchJob` is invoked for this job while its termination status and execution id are the same as those persisted in the database. As a result, its termination status metrics are not updated (due to the fix) and no gauge is created.

PR #502, opened 2021-11-11
vdk-core: Separate config variables with None as a default value and config variables with no default value

gageorgiev:
Vdk-core makes no distinction between configuration variables with no
default values and ones with None as a default value - both have None
as a default_value as far as the Configuration object is concerned. This
is suboptimal, as it means that when running config-help, variables with
None as default do not have their default value shown as it is presumed
that they have no default value.
This change amends this by introducing a singleton class NoDefaultValue
which replaces None for variables without a default value. It also makes it
so when introducing config variables, one does not have to specify a default
value if the variable does not require one (previously you still had to write
None for the default_value parameter of the add method of ConfigBuilder).
Additionally, config variables which previously had None set as their default
value had that line removed.

Testing done: ran config-help and observed the shown None default value,
tested for both values of show_default_value parameter when adding
config variables


tozka:
I am not sure I understood the issue? If the default value is "None" , it means there's no default value , so it should not be shown, right?

gageorgiev:
> I am not sure I understood the issue? If the default value is "None" , it means there's no default value , so it should not be shown, right?

Not in the case where you explicitly want the default value to be None. This was brought to my attention while testing the impala connection plugin, as many of the config variables have their values set to None, but they weren't shown in the config-help printout. To me this implies that they have no default value, which isn't true.

gageorgiev:
> > > I am not sure I understood the issue? If the default value is "None" , it means there's no default value , so it should not be shown, right?
> 
> > Not in the case where you explicitly want the default value to be None. This was brought to my attention while testing the impala connection plugin, as many of the config variables have their values set to None, but they weren't shown in the config-help printout. To me this implies that they have no default value, which isn't true.
> 
> Sorry for being a but dumb here. Can you give an example

So for example, the IMPALA_TIMEOUT variable sets the time in seconds that it takes to timeout the connection. The default value is None, which means the connection never timeouts, but if a user runs config-help and doesn't see a default value, they might falsely assume that this value is required of them.

Now that I think about it, I realize that None is used as a replacement for variables which have no default value, as the add function of config_builder has no default value for the default_value parameter.

tozka:
> So for example, the IMPALA_TIMEOUT variable sets the time in seconds that it takes to timeout the connection. The default value is None, which means the connection never timeouts, but if a user runs config-help and doesn't see a default value, they might falsely assume that this value is required of them.
> 
> Now that I think about it, I realize that None is used as a replacement for variables which have no default value, as the add function of config_builder has no default value for the default_value parameter.

Sorry I missed that. You can ping me (slack) or request review in the PR. 

OK. I think I got it. The problem we are trying to solve is "How will a user understand which configuration variables are optional and which are required" ? If we do config-help. It lists default value but just because it's missing default value - it doesn't mean it's required. 

That's a good point. Options. What options do we have though for solution: 
 
1. the PR here or something similar. I would not use a string through (could be a valid value) but maybe enum or object() singleton instance or something else.
2. Do not show this in vdk config-help. But Let user of the configuration  determine - (current approach) - e.g if user tries to use (configuration.get_required_value) execution will fail 
3. Add new field is_required = True/false (by default False) and print it. 
4. Similar to PR but special value that means field is required (value=VALUE_REQUIRED)

We need to come with pros/cons on these approaches before we decide.

PR #514, opened 2021-11-15
vdk-heartbeat: trino ingestion test

mivanov1988:
The test sends object for ingestion and then verifies
whether the object has been ingested.

Testing: tested manually against ETL and Trino

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
@@ -74,6 +74,33 @@ class Config:
             self.get_value("RUN_TEST_TIMEOUT_SECONDS", "1200")
         )
 
+        # Ingestion test configuration
+        # Target identifies where the data should be ingested into.
+        self.INGEST_TARGET = self.get_value(
+            "VDK_HEARTBEAT_INGEST_TARGET",
+            "vdk-heartbeat-datasource",
+        )
+
+        """
+        Indicates the ingestion method to be used. Example:
+                    method="file" -> ingest to file
+                    method="http" -> ingest using HTTP POST requests
+                    method="kafka" -> ingest to kafka endpoint
+        """
+        self.INGEST_METHOD = self.get_value("VDK_HEARTBEAT_INGEST_METHOD", "http")
+
+        # The name of the table, where the data should be ingested into.
+        self.INGEST_DESTINATION_TABLE = self.get_value(
+            "VDK_HEARTBEAT_INGEST_DESTINATION_TABLE",
+            "vdk_heartbeat_ingestion_test",
+        )
+
+        # The time for which the data should be ingested.
+        self.INGEST_TIMEOUT = self.get_value(
+            "VDK_HEARTBEAT_INGEST_TIMEOUT",
+            "300",
+        )
+
         # The Control Service API URL (http://url/data-jobs) without data-jobs suffix
         self.control_api_url = self._get_atleast_one_value(
             "CONTROL_API_URL", "VDK_HEARTBEAT_CONTROL_SERVICE_URL"
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/database_ingestion/20_verify_data.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/database_ingestion/20_verify_data.py
@@ -1,0 +1,45 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import logging
+import time
+
+from vdk.api.job_input import IJobInput
+
+log = logging.getLogger(__name__)
+
+
+def run(job_input: IJobInput):
+    # If db property includes catalog name, get only the schema name ('default' from 'memory.default').
+    # This is done because the templates' implementation escapes db, table and column names passed as arguments, so that
+    # reserved words could be used for their values
+    db_name = job_input.get_property("db").split(".")[-1]
+
+    props = job_input.get_all_properties()
+    destination_table = props["ingest_destination_table"]
+    id1 = props["id1"]
+    id2 = props["id2"]
+    id3 = props["id3"]
+
+    timeout = int(props["ingest_timeout"])  # [seconds]
+    timeout_start = time.time()
+
+    while time.time() < timeout_start + timeout:
+        # check if data was correctly ingested
+        result = job_input.execute_query(
+            f"SELECT id1, id2, id3 FROM {db_name}.{destination_table} "
+            f"WHERE id1 = '{id1}' "
+            f"AND id2 = '{id2}' "
+            f"AND id3 = '{id3}'"
+        )
+        log.info(f"Query result: {result}")
+
+        if result:
+            if result[0][0] != id1 and result[0][1] != id2 and result[0][2] != id3:
+                raise Exception("The data is not ingested correctly")
+            else:
+                props["succeeded"] = "true"
+                break
+        else:
+            time.sleep(10)
+
+    job_input.set_all_properties(props)
--- a/projects/vdk-heartbeat/README.md
+++ b/projects/vdk-heartbeat/README.md
@@ -63,3 +63,62 @@ Versioning follows https://semver.org.
 
 * A release step in Gitlab CI is automatically triggered after merging changes if build/tests are successful.
 * Update major or minor version when necessary only.
+
+## Tests
+### Database ingestion
+The testing job ingests data into a database and reads it from that database to verify the results.
+#### Configuration
+##### Target
+Target identifies where the data should be ingested into.
+
+The value for this parameter depends on the ingest method chosen.
+* For "http" method, it would require an HTTP URL.
+    Example: http://example.com/<some>/<api>/<endpoint>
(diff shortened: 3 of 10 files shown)


projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/requirements.txt
```diff
@@ -0,0 +1 @@
+vdk-trino
```
tozka, line 1: I was wondering if we can generalize the job. The only reason to call it trino ingestion is because of this. But ingestion is entirely configurable. And for some other deployments, users of vdk-heartbeat may want to test impala ingestion or HTTP ingestion. So can we make it configurable ? 

In order to make it configurable we need 
* ingest_method_default to be set 
* the correct plugin to be installed

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/10_verify_data.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import logging
```
tozka, line 3: I think it will be better to be 20_verify_data . I know that V is after I so the order is preserved but it's very non-obvious. I had to think about it.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/10_ingest_data.py
```diff
@@ -0,0 +1,19 @@
...
+
+def run(job_input: IJobInput):
+    props = job_input.get_all_properties()
+
+    payload = {"id1": props["id1"], "id2": props["id2"], "id3": props["id3"]}
+
+    # Ingest the data
+    log.info(f"Sending the following payload for ingestion: {payload}")
+    job_input.send_object_for_ingestion(
+        payload=payload, destination_table=props["destination_table"], method="http"
```
tozka, line 18: vdk-trino provides its own ingestion method: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py 

vdk-ingest-http provides a method for ingesting via http - https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-core/plugins/vdk-ingest-http 

Those are different. 

It is possible to ingest via http into trino database indirectly as we do in our internal deployment or via segment or other similar service (you can send to segment via HTTP -> which can write to trino for example). 

But that's separate case.

tozka: So to summarize. Our testing job needs to be able to ingest data into a database and read from that database (to verify the results). 

So ingest_method_default and db_default_type need to be set to ultimately go to the same database. 

This can happen in following cases 

A) ingest_method_default is set to "TRINO"  and db_default_type is set to "TRINO" and the connection string for both is the same (same trino database instance)
B) ingest_method_default is set to "HTTP" or "segment" (which writes to trino)  and db_default_type and connection settings are set to same trino instance 

Trino is used as an example. it could be any data warehouse - redshift, impala, snowflake.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/10_verify_data.py
mivanov1988: Fixed

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/requirements.txt
mivanov1988: I don't think it is in the scope of this issue. It sounds like a future improvement. What do you think?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/10_ingest_data.py
mivanov1988: Could you please check whether it already covers this requirment.

tozka: Can you link the issue? What problem are we trying to solve?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/requirements.txt
tozka: Can you please link the issue? It's a good point I should have asked first about what is the problem we are trying to solve here?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -74,6 +74,19 @@ def __init__(self, config_file=None):
...
+        # Ingestion configuration
+        self.INGEST_TARGET = self.get_value(
+            "INGEST_TARGET",
+            "vdk-heartbeat-datasource",
+        )
+
+        self.INGEST_METHOD = self.get_value("INGEST_METHOD", "http")
+
+        self.INGEST_DESTINATION_TABLE = self.get_value(
+            "INGEST_DESTINATION_TABLE",
```
tozka, line 86: Let's prefix them with VDK_HEARTBEAT ?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/database_ingestion/20_verify_data.py
```diff
@@ -0,0 +1,45 @@
...
+    # reserved words could be used for their values
+    db_name = job_input.get_property("db").split(".")[-1]
+
+    props = job_input.get_all_properties()
+    destination_table = props["destination_table"]
+    id1 = props["id1"]
+    id2 = props["id2"]
+    id3 = props["id3"]
+
+    timeout = 300  # [seconds]
```
tozka, line 23: timeout should be configurable. It's basically the first thing people need to configure in their test environment.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -74,6 +74,19 @@ def __init__(self, config_file=None):
             self.get_value("RUN_TEST_TIMEOUT_SECONDS", "1200")
         )
 
+        # Ingestion configuration
```
tozka, line 77: Currently, our only documentation for those configuration keys is in this file. Can you come with a bit more clear description in the comments for each new field?

 Imagine a new user using vdk-heartbeat for a first time would be using this.

mivanov1988: Sure

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/database_ingestion/20_verify_data.py
mivanov1988: Agreed

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
mivanov1988: Sounds meaningful. I aligned them with the rest of the variables.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/vdk_heartbeat_data_job/trino_ingestion/requirements.txt
tozka: Ok. we synced we do not have issue for that. 

But I think we have it currently pretty generalized. 
1. Users can configure ingest_method and db default type in vdk heartbeat config. 
2. Can technically configure plugin as both local and cloud installation distribution vdk is configurable (pip install or through the helm chart for cloud runs) .

tozka (approved):
See if you can update the README to make it easier to users to use it in different scenarios. Maybe even describe the scenarios? But either way it looks good to me. Thanks for taking the change now we will have more confidence that ingestion works :)

mivanov1988:
> See if you can update the README to make it easier to users to use it in different scenarios. Maybe even describe the scenarios? But either way it looks good to me. Thanks for taking the change now we will have more confidence that ingestion works :)

Sure, thanks :)

mivanov1988:
> > See if you can update the README to make it easier to users to use it in different scenarios. Maybe even describe the scenarios? But either way it looks good to me. Thanks for taking the change now we will have more confidence that ingestion works :)
> 
> Sure, thanks :)

Done.

Issue #424, opened 2021-10-20
GraphQL API - add job success rate


mivanov1988:
**What is the feature request? What problem does it solve?**
Currently there is no easy way to retrieve job success rate. We want to expose this metric thought the GraphQL API so that users can retrieve this data.

**Suggested solution**
1. Modify the GraphQL Job Deployment schema: add separate fields (numberOfsuccess/executionsSucceeded, numberOfFailures/executionsFailed) at deployment level.
2. In order to reduce the size of the dateset on which the calculations will be made we should calculate these metrics during the Post Pagination phase (populateDataJobsPostPagination).

    Sample query for success rate calculation (data_job_execution.status 2 is Finished; data_job_execution.status 3 is Failed):
    ```
    SELECT COUNT(status), status, job_name FROM data_job_execution
    WHERE status IN (2, 3) AND job_name IN ('job_name1', 'job_name2') 
    GROUP BY status, job_name
    ```
3. Calculate the executionsSucceeded and executionsFailed only if they are present in the request (GraphQL query).
4. Provide proper unit tests
5. Provide proper Integration tests

**Additional context**
Sample GraphQL query:
```
{{ baseUrl  }}/data-jobs/for-team/no-team-specified/jobs?query={
  jobs(pageNumber: 1, pageSize: 100, filter: [{property: "jobName", sort: ASC}]) {
    content {
      jobName
      config {
        team
        description
        schedule {
          scheduleCron
        }
      }
      deployments {
        enabled
        status
        executionsSucceeded,
        executionsFailed,
        executions(pageNumber: 1, pageSize: 100, filter: []) {
          id
          startTime
        }
      }
    }
  }
}
```
Sample GraphQL response:
```json
{
  "errors": [],
  "data": {
    "content": [
      {
        "jobName": "tms-staging-vdk-1633680894",
        "config": {
          "team": "taurus",
          "description": "",
          "schedule": {
            "scheduleCron": "* * * * *"
          }
        },
        "deployments": [
          {
            "enabled": true,
            "status": null,
            "executionsSucceeded": 100,
            "executionsFailed": 10,
            "executions": [
              {
                "id": "tms-staging-vdk-1633680894-1634638440",
                "startTime": "2021-10-19T10:14:47Z"
              },
              {
                "id": "tms-staging-vdk-1633680894-1634638500",
                "startTime": "2021-10-19T10:15:47Z"
              },
              {
                "id": "tms-staging-vdk-1633680894-1634638560",
[...]

PR #518, opened 2021-11-16
control-service: add execution status fail/success to graphql schema

mrMoZ1:
why: Users requested the ability to produce reports
based on data job execution failure/success numbers.
This is a part of the graphQL epic.

what: Added new fields in the schema and required
Java class.

testing: Added unit test which tests if keys are present.
Unit test will be refactored/extended to cover the full
functionality of the feature when we implement
the last story of  #424.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/model/V2DataJobDeployment.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/model/V2DataJobDeployment.java
@@ -28,6 +28,6 @@ public class V2DataJobDeployment {
    private DataJobExecution.StatusEnum lastExecutionStatus;
    private OffsetDateTime lastExecutionTime;
    private Integer lastExecutionDuration;
-   private Integer successfulDataJobExecutions;
-   private Integer failedDataJobExecutions;
+   private Integer successfulExecutions;
+   private Integer failedExecutions;
 }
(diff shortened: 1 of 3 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/model/V2DataJobDeployment.java
```diff
@@ -28,4 +28,6 @@
    private DataJobExecution.StatusEnum lastExecutionStatus;
    private OffsetDateTime lastExecutionTime;
    private Integer lastExecutionDuration;
+   private Integer successfulDataJobExecutions;
```
mivanov1988, line 31:
```suggestion
   private Integer successfulExecutions/executionsSucceeded;
```

What do you think which one is better? (The `DataJob` in the property name is redundant since the element is part of the job itself.) I am in favor of `executionsSucceeded`. @tpalashki @mrMoZ1

mrMoZ1: First option is more descriptive, second is a bit shorter. To me it doesn't really matter which one we pick.

tpalashki: The first option is more associative with a count. However, in Kubernetes jobs, the number of successful pods is named `succeeded` and is a number, so the second option makes sense as well.

mivanov1988: :)

mivanov1988: @mrMoZ1 it's up to you.

PR #522, opened 2021-11-17
vdk-test-utils: Apply changes after fully managed cursor merge

ivakoleva:
Fully managed cursor implementation required extending vdk-test-utils.
Due to present modules testing and release lifecycles, changes of
interdependent packages need to be made rather sequentially, in a
backwards-compatible fashion - for the pipeline to complete.
A decision was made for the classes changed to be placed within
vdk-core test itself, then moved to the test suite afterwards.

Moved test_run_sql_queries.py helper classes to the dedicated
util_plugins.py. Cleaned up the todos.

Testing Done: did run locally vdk-core and vdk-test-utils tests


projects/vdk-core/plugins/vdk-test-utils/src/vdk/plugin/test_utils/util_plugins.py
```diff
@@ -77,35 +78,80 @@ def initialize_job(self, context: JobContext) -> None:
...
 
-    def _cursor_execute(self, cur: ManagedCursor, query: str) -> None:
-        self.history.append(query)
-        super()._cursor_execute(cur, query)
-
     def _connect(self) -> PEP249Connection:
         return self.db.new_connection()
 
 
+class ValidatedSqLite3MemoryDbPlugin:
```
tozka, line 89: Why move them to vdk-test-utils? Are they necessary anywhere else ?

ivakoleva: My presumption was to place classes ValidatedSqLite3MemoryDbPlugin and SyntaxErrorRecoverySqLite3MemoryDbPlugin along with DecoratedSqLite3MemoryDbPlugin - each of those is a customised opinionated implementation hooked to the corresponding managed cursor handling phase (we have 3 of those).

So DecoratedSqLite3MemoryDbPlugin is tracking statements_history. It is also too specific, however it's used by two tests, instead of one.
ValidatedSqLite3MemoryDbPlugin and SyntaxErrorRecoverySqLite3MemoryDbPlugin are not used by more than one specific test (yet). So that is why I agree they could stay within the scope of the test itself. What do you think?

tozka:
Hi @ivakoleva 

You have some conflicts that you need to resolve and there are 248 file changes which seems unlikely to be correct.

ivakoleva:
> 
There are actually 2 file changes,

Issue #390, opened 2021-10-14
[TimeBoxed]Support Python 3.10


mnozhchev:
**What is the feature request? What problem does it solve?**
pip install quickstart-vdk fails on Python 3.10 with a very uninformative message. Also README.md says that vdk requires "Python 3.7+" which implies that Python 3.10 is supported.

**Suggested solution**
Test vdk on Python 3.10 and, then, update the setup.cfg metadata to include 3.10.

**Additional context**
![image (1)](https://user-images.githubusercontent.com/6613495/137323191-541ea2fe-1e0f-47ef-ab65-fd8735d8386f.png)

tozka:
There are 3 components that need to be changed. We may split it into 3 bugs 

vdk-core: 

Test vdk on Python 3.10: 
Add [Gitlab CI Job](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/.gitlab-ci.yml) (similar to the other ones for 3.7-3.9) vdk-core-simple_func_test_with_py310 and vdk-core-build_with_py310 (to vdk-core file -https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/.gitlab-ci.yml)
and if PR passes that's enough

vdk plugins: 

The same as vdk-core goes for each plugin. This means editing each plugin CI file: For example https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/plugins/vdk-postgres/.plugin-ci.yml

vdk-control-cli: 
Similarly new CI job "vdk-control-cli-build-with-py310" : 
https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/.gitlab-ci.yml


I don't anticipate anything to break. But you never know.

PR #536, opened 2021-11-19
vdk-control-cli: Support 3.10

gageorgiev:
This change enables support for Python 3.10.
Addresses #390 


projects/vdk-control-cli/cicd/build.sh
```diff
@@ -30,6 +30,7 @@ pip install -e .
 
 echo "Run unit tests and generate coverage report"
 pip install pytest-cov
+pip install pytest==6.2.4 # TODO: there is some regression with the newest version, remove this when possible
```
tozka, line 33: What is the regression? Can you link an issue/ticket to pytest  
Do we need to open an internal github issue to address this?

projects/vdk-control-cli/.gitlab-ci.yml
```diff
@@ -10,6 +10,7 @@ image: "python:3.9"
   stage: build
   before_script:
     - cd projects/vdk-control-cli
+    - pip install --upgrade pytest pip # TODO: REMOVE THIS AT SOME POINT
```
tozka, line 13: TODO? Is this necessary? as it's installed by build.sh ?

gageorgiev: I tested and it's not. I had it included at the start due to some issue, but I guess the cause was different.

projects/vdk-control-cli/cicd/build.sh
tozka: Please add this information in the comment or at least git description. We should be clear to future developer why we decided to make something if it's not clear from the code itself . 

Write in the ticket https://github.com/pytest-dev/pytest/issues/8539 that you are still seeing the issue in python v 3.10.xyz and in pytest vx.y ; Include snippet of logs and some details to help them root-cause.

gageorgiev: I was accidentally using an older version of pytest due to it being pinned in requirements.txt. Removed the pin and the extra pytest install, hopefully it should work. Any idea why it was pinned originally?

tozka: No idea. That's why it's important to put comments. It was copied over from the previous repo, so git history is gone to see details in the commit message (if there were any). Likely for no specific reason (copy pasted from somewhere else :) )

tozka (approved):

YanaZhivkova (approved):

PR #540, opened 2021-11-19
vdk-heartbeat: Make vdkcli_oauth2_uri and control_api_url optional

YanaZhivkova:
If we want to test VDK's configuration and default options, we want
to be able to set those params values in the default vdkcli options
and run vdk-heartbeat.

Make vdkcli_oauth2_uri and control_api_url params optional and don't
pass values if they are not set in the config.ini - thus whatever is
set as default in VDK's configuration will be used.

Tested by: Run locally built vdk-heartbeat in an environment with
properly configured defaults with a config.ini looking like this:
[DEFAULT]
VDKCLI_OAUTH2_REFRESH_TOKEN=***

JOB_RUN_TEST_MODULE_NAME=vdk.internal.heartbeat.simple_run_test
JOB_RUN_TEST_CLASS_NAME=SimpleRunTest
DATAJOB_DIRECTORY_NAME=simple
VDK_COMMAND_NAME=vdk

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
@@ -25,9 +25,11 @@ class Config:
         self.vdkcli_api_refresh_token = self._get_atleast_one_value(
             "VDKCLI_OAUTH2_REFRESH_TOKEN", "VDK_HEARTBEAT_API_TOKEN"
         )
+        # If no value is found, argument will not be passed and default will be used
         self.vdkcli_oauth2_uri = self._get_atleast_one_value(
-            "VDKCLI_OAUTH2_URI", "VDK_HEARTBEAT_API_TOKEN_AUTH_URL"
+            "VDKCLI_OAUTH2_URI", "VDK_HEARTBEAT_API_TOKEN_AUTH_URL", is_required=False
         )
+
         self.op_id = self.get_value("VDK_HEARTBEAT_OP_ID", Config.DEFAULT_OP_ID)
         table_suffix = re.sub("[^a-z0-9_]", "_", self.op_id.lower())
         job_suffix = re.sub("[^a-z0-9-]", "-", self.op_id.lower())
@@ -102,8 +104,9 @@ class Config:
         )
 
         # The Control Service API URL (http://url/data-jobs) without data-jobs suffix
+        # If no value is found, argument will not be passed and default will be used
         self.control_api_url = self._get_atleast_one_value(
-            "CONTROL_API_URL", "VDK_HEARTBEAT_CONTROL_SERVICE_URL"
+            "CONTROL_API_URL", "VDK_HEARTBEAT_CONTROL_SERVICE_URL", is_required=False
         )
 
         # Job name deployed during the test
@@ -211,11 +214,11 @@ class Config:
             )
         return value
 
-    def _get_atleast_one_value(self, key1: str, key2: str) -> Optional[str]:
+    def _get_atleast_one_value(self, key1: str, key2: str, is_required=True):
         value1 = self.get_value(key1, is_required=False)
         value2 = self.get_value(key2, is_required=False)
 
-        if (value1 is None) and (value2 is None):
+        if is_required and (value1 is None) and (value2 is None):
             raise ValueError(
                 "Error occurred:\n"
                 f"What: Cannot configure heartbeat test\n"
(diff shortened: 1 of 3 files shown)


projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -102,9 +104,10 @@ def __init__(self, config_file=None):
         )
 
         # The Control Service API URL (http://url/data-jobs) without data-jobs suffix
-        self.control_api_url = self._get_atleast_one_value(
-            "CONTROL_API_URL", "VDK_HEARTBEAT_CONTROL_SERVICE_URL"
-        )
+        # If no value is found, argument will not be passed and default will be used
+        self.control_api_url = self.get_value(
+            "CONTROL_API_URL", is_required=False
+        ) or self.get_value("VDK_HEARTBEAT_CONTROL_SERVICE_URL", is_required=False)
```
tozka, line 110: minor comment: Wouldn't be simpler to add `is_required` argument to `_get_atleast_one_value` ?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -23,10 +23,18 @@ class JobController:
 
     def __init__(self, config: Config):
         self.config = config
+        control_api_url = (
```
tozka, line 26: control_api_url_message seems more explicit and clear name for the variable here I think

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -23,10 +23,18 @@ class JobController:
...
+            config.control_api_url
+            or "Not set (default from VDK's configuration will be used)"
+        )
+        auth_endpoint = (
```
tozka, line 30: `auth_endpoint_message` ?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -46,102 +54,174 @@ def _execute(self, command):
...
+        else:
+            self._execute(
+                [
+                    "login",
+                    "-a",
+                    f"{self.config.vdkcli_api_refresh_token}",
+                    "-t",
+                    "api-token",
+                ]
+            )
```
tozka, line 78:
```

    def __control_api_url_arg(self):
        if self.config.vdkcli_oauth2_uri:
            return ["-u", f"{self.config.vdkcli_oauth2_uri}"]
        else 
            return []

    @LogDecorator(log)
    def login(self):
        self._execute(["login"] + ["-t", "api-token"] + self.__control_api_url_arguments())
```

Perhaps something akin to above would be slightly simpler and clearer. Here in those arrays people would be playing a bit [Spot the difference](https://en.wikipedia.org/wiki/Spot_the_difference) :)

tozka (commented):
Looks good to me. I have only a few minor readability comments.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
YanaZhivkova: Yes, good point

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
YanaZhivkova: Done.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
YanaZhivkova: Yes, thank you! I spent some time trying to find a way to do that, but everything I could think of, was awful :D

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -224,7 +227,7 @@ def _get_atleast_one_value(self, key1: str, key2: str) -> Optional[str]:
                 f"Countermeasures: Set as required configuration either as environment variable or in the config file."
             )
 
-        return value1 if value1 is not None else value2
+        return value1 or value2 or False
```
tozka, line 230: No, this won't work. Don't you have some unit test failing? If not consider adding unit tests that return empty string for eample

If value1 and value2 are a boolean (with value False) or are a empty string or empty array   

 they all evaluate to "false" in if check which would break the logic.

See https://docs.python.org/2.4/lib/truth.html

The previous logic seemed fine. Why changing it?

YanaZhivkova: Ok, I added some tests. I did not find a way to set value1/value2 to boolean values, they are always parsed as strings. But the empty string was a valid case.

tozka (approved):

PR #544, opened 2021-11-22
control-service: fix deploy notifications and add cc mails

tozka:
Deploy notificatoins were not working properly due to auto-wiring issue

We also enable operator teams to set cc mail for all notifications
(using helm chart) in order to have visibilility into what notificatiosn
are sent to data job users

Testing Done: unit tests, locally deployed Control service and reproduce
scenario to get deploy anotifiactions


doks5 (approved):
LGTM

PR #547, opened 2021-11-23
vdk-trino, vdk-greenplum: Fix ingestion

gageorgiev:
Both plugins' CICD started failing recently. This might be
related to the changes made for the fully managed cursor,
but investigating would be too time-consuming. 
This change fixes the apparent issues.


YanaZhivkova (commented):
Have you tested this somehow?

tozka:
What was the issue before? And how does it fix it? We need to have some idea otherwise we cannot be sure our fix really works permanently.

gageorgiev:
> Have you tested this somehow?

Yes, both locally against Trino and Greenplum instances, and through the CI.

> What was the issue before? And how does it fix it? We need to have some idea otherwise we cannot be sure our fix really works permanently.

The issue with Trino was that the execution of the query which gets the table columns was iterated against directly, which doesn't work as the `execute` invocation does not return the query output, we need a `fetchall` call after it. I've no clue how this was working previously.

The issue with Greenplum was that the ingestion query execution was not being committed, which meant that querying for the ingested data returned a newline character. This issue might be related to the newest psycopg2 release, which was made 11 days ago, although that's highly unlikely as these API's tend to be fairly ossified and searching for newly reported issues gave me nothing.

tozka:
> > Have you tested this somehow?
> 
> Yes, both locally against Trino and Greenplum instances, and through the CI.
> 
> > What was the issue before? And how does it fix it? We need to have some idea otherwise we cannot be sure our fix really works permanently.
> 
> The issue with Trino was that the execution of the query which gets the table columns was iterated against directly, which doesn't work as the `execute` invocation does not return the query output, we need a `fetchall` call after it. I've no clue how this was working previously.
> 
> The issue with Greenplum was that the ingestion query execution was not being committed, which meant that querying for the ingested data returned a newline character. This issue might be related to the newest psycopg2 release, which was made 11 days ago, although that's highly unlikely as these API's tend to be fairly ossified and searching for newly reported issues gave me nothing.

Thanks. Please provide those details in a commit message as well. 

I would advise splitting into 2 separate commits the changes on vdk-trion and vdk-Greenplum (in the same PR - but use rebase and not squash when merging) so that if we need to revert one or other we can do it independently. Also, the more detailed git description would be more to the point.

tozka (approved):

PR #549, opened 2021-11-23
control-service: set build job impagePullPolicity to default to IfNot…

tozka:
…Present

We rarely change the version (tag) of the builder job image. There's not
point of trying to pull it every time , so this is setting it to
IfNotPresent. This way we can limit the chance of hitting docker limits.

It's made configurable using
`datajobs.deployment.builder.imagePullPolicy` system property in case in
some environments it is necessary.

Testing Done: integration test passed

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
@@ -31,7 +31,6 @@ import static java.util.Map.entry;
 public class JobImageBuilder {
    private static final Logger log = LoggerFactory.getLogger(JobImageBuilder.class);
 
-   static final String BUILDER_IMAGE_PULL_POLICY = "Always";
    private static final int BUILDER_TIMEOUT_SECONDS = 1800;
    private static final String REGISTRY_TYPE_ECR = "ecr";
    private static final String REGISTRY_TYPE_GENERIC = "generic";
@@ -59,7 +58,9 @@ public class JobImageBuilder {
    @Value("${datajobs.deployment.dataJobBaseImage:python:3.9-slim}")
    private String deploymentDataJobBaseImage;
    @Value("${datajobs.deployment.builder.extraArgs:}")
-   private String extraArgs;
+   private String builderJobExtraArgs;
+   @Value("${datajobs.deployment.builder.imagePullPolicy:IfNotPresent}")
+   private String builderJobImagePullPolicy;
    @Value("${datajobs.git.ssl.enabled}")
    private boolean gitDataJobsSslEnabled;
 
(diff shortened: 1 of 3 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -59,7 +58,9 @@
    @Value("${datajobs.deployment.dataJobBaseImage:python:3.9-slim}")
    private String deploymentDataJobBaseImage;
    @Value("${datajobs.deployment.builder.extraArgs:}")
-   private String extraArgs;
+   private String builderJobExtraArgs;
+   @Value("${datajobs.deployment.builder.imagePullPolicy:IfNotPresent}")
```
mivanov1988, line 62: Shouldn't the default value be in application.properties?

mivanov1988:
I think we need the same for the application container. What do you think?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
tozka: I am not sure. I have always put them here I think. But perhaps it is better for the default to be only in application.properties (alternatively in both places or only here) ?
 What is the recommended way? What are the pros/cons ?

tozka:
> I think we need the same for the application container. What do you think?

It's already configurable: https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml#L36

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mivanov1988: If someone uses it elsewhere he/she won't be able to benefit from the default value. Also it is easier to see the default values by looking application.properties.

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
tozka: > If someone uses it elsewhere he/she won't be able to benefit from the default value

What do you mean by this?

mivanov1988: If I inject (`@Value("${datajobs.deployment.builder.imagePullPolicy}")`) this property in another class it wouldn't have a default value.

tozka: Got it. That make sense.

PR #552, opened 2021-11-24
control-service: fix a bug in the GraphQL implementation

tpalashki:
The recently introduced functionality to filter and sort
by last execution of data jobs has a bug where older executions
are not properly checked and as a result, the last execution
is always updated, even when the current execution is older
than the last one.

This commit fixes that.

Testing done: new unit test for this use case


mivanov1988:
LGTM

mivanov1988 (approved):

Issue #432, opened 2021-10-21
Move plugins folder to top level project


tozka:
**What is the feature request? What problem does it solve?**

From a design point, each plugin is independent of one another and of vdk-core.  
Each plugin depends on vdk-core so they know aboutn vdk-core. 

But vdk-core does not depend on any plugin and it should not even know about plugin existence. 

Currently, plugins are put in vdk-core project folder. It was done mostly when vdk-core was moved from its previous repository. But it would be better if there are further decoupled as this may cause confusion to users and developers. 
 

**Suggested solution** 
Move plugins directory to top-level projects directory. 
From projects/vdk-core/plugins (now) to projects/plugins 

Add the include in vdk-core https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/.gitlab-ci.yml#L4 

should be moved to the top level .gitlab-ci.yaml file -> https://github.com/vmware/versatile-data-kit/blob/main/.gitlab-ci.yml

PR #553, opened 2021-11-24
vdk-plugins: Move all plugins into separate directory

gageorgiev:
Until now, all plugins were stored inside the vdk-core directory.
This was counterintuitive, as plugins are separate packages from
vdk-core.
This change addresses this by moving all the plugins to a new
directory, and making the necessary changes to CI pipelines and
documentation.
Addresses #432 


tozka:
Link https://github.com/vmware/versatile-data-kit/issues/432 to the PR

projects/vdk-core/cicd/simple-functional-test.sh
```diff
@@ -54,19 +54,19 @@ python setup.py sdist --formats=gztar && pip install dist/* || fail "VDK Install
...
-  pip install --extra-index-url $PIP_EXTRA_INDEX_URL -e ./plugins/vdk-trino
-  pip install --extra-index-url $PIP_EXTRA_INDEX_URL -e ./plugins/vdk-plugin-control-cli
+  pip install --extra-index-url $PIP_EXTRA_INDEX_URL -e ./../vdk-plugins/vdk-trino
+  pip install --extra-index-url $PIP_EXTRA_INDEX_URL -e ./../vdk-plugins/vdk-plugin-control-cli
 }
 trap restore_vdk_in_editable_mode EXIT
 
 echo "Install database plugin for presto support"
-pushd plugins/vdk-trino || exit
+pushd ../vdk-plugins/vdk-trino || exit
```
tozka, line 63: Let's not change it here in this PR as this is way too big change.

 But I want each high-level project (https://github.com/vmware/versatile-data-kit/tree/main/projects) to be independent of each other (as if they are in separate git repositories). 
 
 So we'd need to see how to fix that so we do not have cross project references (../vdk-plugins). But definitely in separate PR

tozka (commented):
Looks ok. 

You probably want to update https://github.com/vmware/versatile-data-kit/blob/main/CONTRIBUTING.md 

And also after merging update the links in the wiki

I think it would be easier to  clone the wiki  (git clone https://github.com/vmware/versatile-data-kit.wiki) and do search and replace and git push master  and not update the pages one by one.

tozka (approved):
Looks good. Thanks for doing that. Moving it this way would make it easier to for users to understand and use the plugins and would be less confusing.

PR #564, opened 2021-11-26
control-service: add IT for graphQL sort, filter by next run

mrMoZ1:
why: Users requested additional sorting functionality.
We need to add sort on next run in the graphQl API.
The change should be implemented in the back-end.

what: Change is already implemented. Added IT test
that verifies it and informs us of any bugs.

testing: Added new test.


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/graphql/it/GraphQLExecutionsNextRunIT.java
```diff
@@ -0,0 +1,163 @@
...
+@AutoConfigureMockMvc
+public class GraphQLExecutionsNextRunIT {
+
+   @Autowired
+   MockMvc mockMvc;
+
+   @Autowired
+   JobsRepository jobsRepository;
+
+   private final String uri = "/data-jobs/for-team/supercollider/jobs";
```
doks5, line 38: Is the URI hard-coded somewhere in the control service code, or can we change team name?

doks5 (approved):
I have only one comment, so, overall, LGTM.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/graphql/it/GraphQLExecutionsNextRunIT.java
mrMoZ1: The uri is defined in the open api model definition, we can certainly change the team name, however I'm not sure if we use this uri parameter in the graphQl queries since we retrieve most of the needed query params from the graphql query itself. Maybe we use it to check permissions etc. ?

mivanov1988: You should extend BaseIT. It already has this propery.

mivanov1988: You can change the name as well.

mrMoZ1: I don't think I should extend BaseIT here, since this test doesn't need the extra functionality that BaseIT provides (k8s namespace) and in this case extending BaseIT will incur a performance penalty for the test. The only thing that could be reused from BaseIT is this property, however It can't be imported since it's protected in BaseIT and the tests are in different packages.

mivanov1988: What do you mean by `extra functionality that BaseIT provides (k8s namespace) `?

mrMoZ1: The methods annotated with @BeforeEach and @AfterEach in BaseIT incur a performance penalty, and this test doesn't rely on this setup.

mrMoZ1: Changes as we discussed.

PR #566, opened 2021-11-29
vdk-core: Fix error in user code classified as Platform Error

doks5:
Currently, if an exception is raised in user code (the run method in a data
job python script), the error is classified as a Platform Error, even though
it is caused by something outside of vdk control.

This happens because in the DataJobDefaultHookImplPlugin class, the exception catching
logic uses errors.find_whom_to_blame_from_exception() method, instead of
job_input_error_classifier.whom_to_blame() method. The former just checks the exception
class, while the latter actually parses the exception traceback, and tries to properly
classify the error.

This change moves error classification from errors.find_whom_to_blame_from_exception() to
job_input_error_classifier.whom_to_blame(), and improves the robustness of the latter in case
there are changes in vdk file paths which could lead to different file paths in user code exceptions.

Testing Done: Unit tests

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
@@ -32,6 +32,7 @@ def whom_to_blame(exception, executor_module):
 
 
 def _is_exception_from_vdk_code(exception, executor_module):
+    exception_in_vdk = False
     executor_module = os.path.abspath(executor_module)
     vdk_code_directory = os.path.dirname(executor_module)
     call_list = traceback.format_tb(exception.__traceback__)
@@ -42,12 +43,13 @@ def _is_exception_from_vdk_code(exception, executor_module):
     for call in call_list:
         caller_module = call.split('"')[1]  # Extract module path from stacktrace call.
         if vdk_code_directory in caller_module and caller_module != executor_module:
-            return True
+            exception_in_vdk = True
         elif (
             caller_module == executor_module
         ):  # User code starts from this module always.
-            break
-    return False
+            return False
+
+    return exception_in_vdk
 
 
 def is_user_error(received_exception: Exception) -> bool:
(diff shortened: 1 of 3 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
```diff
@@ -42,11 +43,14 @@ def _is_exception_from_vdk_code(exception, executor_module):
...
+            exception_in_vdk = True
         elif (
             caller_module == executor_module
         ):  # User code starts from this module always.
-            break
+            return False
+
+    if exception_in_vdk:
+        return True
     return False
```
gageorgiev, line 54:
```suggestion
    return exception_in_vdk
```

gageorgiev (approved):
solid stuff, I noted 1 antipattern that you should amend

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
doks5: Good point. Will fix.

projects/vdk-core/tests/vdk/internal/builtin_plugins/run/job_input_error_classifier_test.py
```diff
@@ -87,18 +98,10 @@ def test_unknown_generic_error(self, mock_is_user_error, mock_traceback_format_t
...
 
-    # Generic error thrown by job_input that is not specifically recognised by VDK should be VAC error.
-    @patch(f"{traceback.format_tb.__module__}.{traceback.format_tb.__name__}")
-    @patch(f"{is_user_error.__module__}.{is_user_error.__name__}")
-    def test_job_input_generic_error(
-        self, mock_is_user_error, mock_traceback_format_tb
-    ):
-        exception = Exception("!")
-        mock_is_user_error.return_value = False
-        mock_traceback_format_tb.return_value = self.PLATFORM_ERROR_STACKTRACE
```
tozka, line 98: Why delete this? Where is the test covering  PLATFORM_ERROR_STACKTRACE ? It correctly test that "PLATFORM_ERROR_STACKTRACE" is a platform error.

PR #567, opened 2021-11-29
vdk-heartbeat: Make oauth2 refresh token optional

YanaZhivkova:
We want to be able to run vdk-heartbeat against a local
control service.

If no VDKCLI_OAUTH2_REFRESH_TOKEN/VDK_HEARTBEAT_API_TOKEN
is set, run login with no params - it defaults to localhost.

Testing done: run locally built vdk-heartbeat with
and without oauth2 refresh token set.

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
@@ -74,16 +74,19 @@ class JobController:
 
     @LogDecorator(log)
     def login(self):
-        self._execute(
-            [
-                "login",
-                "-a",
-                f"{self.config.vdkcli_api_refresh_token}",
-                "-t",
-                "api-token",
-            ]
-            + self.__get_api_token_authorization_url_arg()
-        )
+        if self.config.vdkcli_api_refresh_token:
+            self._execute(
+                [
+                    "login",
+                    "-a",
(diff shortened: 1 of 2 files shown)


gageorgiev (approved):

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -68,16 +68,17 @@ def _execute(self, command):
...
+            self._execute(
+                [
+                    "login",
+                    "-a",
+                    f"{self.config.vdkcli_api_refresh_token}",
+                    "-t",
+                    "api-token",
+                ]
+                + self.__get_api_token_authorization_url_arg()
+            )
```
gageorgiev, line 81:
```suggestion
        if self.config.vdkcli_api_refresh_token:
            self._execute(
                [
                    "login",
                    "-a",
                    f"{self.config.vdkcli_api_refresh_token}",
                    "-t",
                    "api-token",
                ]
                + self.__get_api_token_authorization_url_arg()
            )
        else:
            log.info("Running against local control service.")
```

Maybe include a log statement.

tozka: +1 for log statement. But not local  - running against control service without authentication

PR #570, opened 2021-11-29
vdk-core: simplify log line

tozka:
Remove redundant "created" timestamp we always print. It's never really
used for anything.
Rename OpId: to Id: since we print attempt id there (and not really
opId).
But attempt id is rather long "id" seems better to fit the space.


gageorgiev (approved):
Good change. We should probably switch all .format calls to f-strings in our entire codebase, as they are much more readable imo.

PR #572, opened 2021-11-29
control-service: change the query format of nested executions

tpalashki:
This commit unifies the parameter format of the nested
jobs.deployments.executions to match those used by the root
executions query. The existing implementation is adapted and
reused.

Testing done: unit and integration tests pass;
Manually tested the nested executions with a query that uses
the new format and verified that the returned result is correct.
Also tested the execution query to verify that this commit does
not break it. Below are the queries that were used for both
tests:

```
query jobsQuery($filter: [Predicate], $executionFilter: DataJobExecutionFilter, $executionOrder: DataJobExecutionOrder, $search: String, $pageNumber: Int, $pageSize: Int) {
  jobs(
    pageNumber: $pageNumber
    pageSize: $pageSize
    filter: $filter
    search: $search
  ) {
    content {
      jobName
      config {
        team
        description
        sourceUrl
        schedule {
          scheduleCron
          nextRunEpochSeconds
          __typename
        }
        contacts {
          notifiedOnJobFailureUserError
          notifiedOnJobFailurePlatformError
          notifiedOnJobSuccess
          notifiedOnJobDeploy
          __typename
        }
        __typename
      }
      deployments {
        id
        enabled
        lastExecutionStatus
        lastExecutionTime
        lastExecutionDuration
        executions(pageNumber: 1, pageSize: 5, filter: $executionFilter, order: $executionOrder) {
          id
          status
          startTime
        }
        __typename
      }
      __typename
    }
    totalPages
    totalItems
    __typename
  }
}
```
with variables:
```
{
  "pageNumber":1,
  "pageSize":25,
  "filter":[
    {
      "property":"deployments.lastExecutionTime",
      "pattern":null,
      "sort":"DESC"
    }
  ],
  "search":"",
  "executionFilter":{
    "statusIn": ["FINISHED","FAILED"]
  },
  "executionOrder":{
    "property": "startTime",
    "direction": "DESC"
  }
}
```

And for the root executions:
```
query($filter: DataJobExecutionFilter, $order: DataJobExecutionOrder, $pageNumber: Int, $pageSize: Int) {
  executions(pageNumber: $pageNumber, pageSize: $pageSize, filter: $filter, order: $order) {
    content {
      id
      jobName
      startTime
      endTime
      status
    }
    totalPages
    totalItems
  }
}
```
with variables:
```
{
  "filter": {
    "statusIn": ["FAILED", "FINISHED"]
  },
  "order": {
    "property": "startTime",
    "direction": "ASC"
  },
  "pageNumber": 1,
  "pageSize": 25
}
```


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobExecutionFilterSpec.java
```diff
@@ -24,11 +24,16 @@
 public class JobExecutionFilterSpec implements Specification<DataJobExecution> {
 
    private DataJobExecutionFilter filter;
+   private String jobName;
 
    @Override
    public Predicate toPredicate(Root<DataJobExecution> root, CriteriaQuery<?> query, CriteriaBuilder builder) {
       List<Predicate> predicates = new ArrayList<>();
 
+      if (jobName != null) {
```
mivanov1988, line 33: I wonder if we could reuse filter.getJobNameIn()?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/ExecutionDataFetcher.java
```diff
@@ -76,19 +75,19 @@
...
    /**
-    * Currently, it does not provide filtering by specifying fields due
-    * to its post-pagination loading - the execution data is attached after slicing of the requested page.
+    * Populates the executions of the specified data jobs, that match the specified criteria.
     */
    List<V2DataJob> populateExecutions(List<V2DataJob> allDataJob, DataFetchingEnvironment dataFetchingEnvironment) {
-      final ExecutionQueryVariables queryVariables = fetchQueryVariables(dataFetchingEnvironment);
-      final Pageable pageable = constructPageable(queryVariables);
+      final Map<String, Object> arguments = dataFetchingEnvironment.getSelectionSet().getField(JobFieldStrategyBy.DEPLOYMENT_EXECUTIONS.getPath()).getArguments();
+      final DataJobExecutionQueryVariables dataJobExecutionQueryVariables = fetchDataJobExecutionQueryVariables(arguments);
```
mivanov1988, line 82: I think we need validation of _jobNameIn_ property. This field must be disabled in context of jobs query and we must send an error to the client. In context of particular job it does not make sense to filter by another one.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobExecutionFilterSpec.java
tpalashki: That was my initial consideration, but I decided against it for simplicity's sake. Using the getJobNameIn for this purpose would require changes to the filter for each individual data job. Now the implementation is much simpler.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/ExecutionDataFetcher.java
tpalashki: I will add this validation.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobExecutionFilterSpec.java
mivanov1988: The method behavior is a bit strange to me. If you set both _jobName_ and _jobNameIn_ the query will look like this:
```
SELECT * FROM data_job_execution WHERE job_name = "A" and job_name IN ['B', 'C']
```
`Using the getJobNameIn for this purpose would require changes to the filter for each individual data job. ` - what do you mean by `require changes to the filter for each individual data job` isn't it just adding an job name to the _jobNameIn_ list?

tpalashki: I will refactor to reuse `jobNameIn`.

mivanov1988 (approved):

PR #575, opened 2021-11-29
vdk-core: Rework new version printout

gageorgiev:
If VDK or some plugin has a newer version, VDK automatically
prints out a message stating this and prompting the user to
update. However, this is printed out at the start of the log,
and can be easily missed.
This change moves the printout to the end of the log.
This change also reworks the printout to collect multiple
components into one printout.

Testing done: verified printout appears at the end

if only one package needs an update, we get the following:
```
    ******************************************************************************************

    New version for vdk-core is available.

    Please update to latest version by using:
    pip install --upgrade-strategy eager -U vdk-core

    ******************************************************************************************
```

If multiple packages need an update:
```
    ******************************************************************************************

    New versions for vdk-core, vdk-ingest-http, vdk-ingest-file, vdk-plugin-control-cli,
    vdk-plugin-control-cli, vdk-plugin-control-cli, vdk-server, vdk-sqlite are available.


    Please update to latest version by using:
    pip install --upgrade-strategy eager -U vdk-core vdk-ingest-http vdk-ingest-file vdk-plugin-control-cli vdk-plugin-control-cli vdk-plugin-control-cli vdk-server vdk-sqlite

    ******************************************************************************************
```

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/version/new_version_check_plugin.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/version/new_version_check_plugin.py
@@ -1,7 +1,9 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import logging
+import textwrap
 from enum import Enum
+from typing import List
 
 import click
 from vdk.api.plugin.hook_markers import hookimpl
@@ -19,7 +21,7 @@ class ConfigKey(str, Enum):
     VERSION_CHECK_PLUGINS = "VERSION_CHECK_PLUGINS"
 
 
-def new_package(package_index, package_name):
+def new_package(package_name: str, package_index: str) -> Package:
     return Package(package_name, package_index)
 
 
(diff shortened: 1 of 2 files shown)


tozka:
> I'm not sure if pip install can take multiple extra index URLs.

It can but it is not necessary. I don't think it's expected (or supported) to have multiple index urls

tozka (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/version/new_version_check_plugin.py
```diff
@@ -43,48 +43,68 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
-                if is_plugin
-                else f"New Version of {package_name} is Available"
-            )
-            # We are using eager strategy so that if there's new version of a plugin that satisfies the requirements
-            # of the main package we want to upgrade to it.
-            pip_command = "pip install --upgrade-strategy eager -U"
-            click.echo(
-                f"""
-                **********************************************************************************************************************
+    def check_version(package_list, package_index):
```
tozka, line 70: If it doesn't need to be public make it protected or private: https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#clarification-on-access-modifiers

projects/vdk-core/src/vdk/internal/builtin_plugins/version/new_version_check_plugin.py
```diff
@@ -43,48 +43,68 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
+    Please update to latest version by using:
+    {install_message}
+
+    ******************************************************************************************
+            """,
+            err=True,
+        )
+
+
+def chunk_message(message, line_len):
```
tozka, line 101: Make private or protected . Also write pydoc.  

But is it necessary. Wouldn't https://docs.python.org/3/library/textwrap.html work ?

gageorgiev: It would. I knew there should be a library for this but i couldn't figure out how to find it, so I settled on rewriting it.

projects/vdk-core/src/vdk/internal/builtin_plugins/version/new_version_check_plugin.py
gageorgiev: done; also included a pydoc for it

tozka (approved):

PR #579, opened 2021-11-30
vdk-core: Introduce Ingestion Plugin execution sequence interfaces

doks5:
As part of the improvements to the Ingestion API, it is necessary to make it
possible for multiple ingestion plugins to be chained and provide capability for
complex payload processing.

This change introduces new interface for the IngesterPlugin to facilitate payload
processing without ingesting, and a new environment variable in the ingestion configuration
to allow clients to specify the sequence in which ingestion plugins get executed.

Testing Done: Trivial, documentation change mainly.


tozka:
Can you link the issue ?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -214,6 +213,16 @@ def ingest_payload(
         """
         pass
 
+    def process_payload(self, payload: List[dict]) -> List[dict]:
```
ivakoleva, line 216: My understanding is, hook implementations chaining is already facilitated by pluggy. Does pluggy support passing the parameter from a returned value from a previous execution? Otherwise, I would suggest a param referencing a mutable object, encapsulated within the the scope of the chaining method.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -29,6 +29,14 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             default_value=None,
             description="Default Ingestion target to be used.",
         )
+        config_builder.add(
+            key="INGEST_PLUGIN_EXECUTION_SEQUENCE",
```
ivakoleva, line 33: Not sure this is required - we should support expressions or at least a convention to determine what happens with the rest of the hook implementations that are not explicitly stated within the execution sequence. Or what happens if duplicated somewhere in the sequence. 

My understanding is, pluggy's hook impls evaluation is sequential in favour of parallel. Is that correct?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
tozka: Note that we are not using pluggy here. But our own (called router) extensibility. 

This is because pluggy does not support conditional hooks (hooks activated only if certain configuration is set for example HTTP ingest hook used only if ingest_method is HTTP for example)

doks5: As @tozka has mentioned above, we do not rely on pluggy here. With the current plugin configuration, we can set a plugin to be placed at the beginning of the execution queue, the end, or somewhere in between, but we cannot set it to be executed at a specific point before or after another plugin.

As for your other point, could you elaborate what you mean by param referencing?

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
doks5: Well, when ingestion API is used, the data engineer uses either send_object_for_ingestion() or send_tabular_data_for_ingestion(). Both of these methods require an ingestion `method` to be set (either as an env var, or passed as an argument to the method call), so there will always be at least one ingestion plugin executed.

doks5:
> Can you link the issue ?

Done, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -214,6 +213,16 @@ def ingest_payload(
...
+        """
+        Do some processing on the ingestion payload before passing it to the actual
```
tozka, line 218: What kind of processing .Can you add example usage?

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -29,6 +29,14 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
+            default_value=None,
+            description="""A string of coma-separated ingestion plugin method names,
```
tozka, line 35: Maybe some short example to clarify what you mean here ...

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -258,7 +267,7 @@ def post_ingest_process(
         :param exceptions: Optional[List]
             A list of all exceptions (if any) encountered while ingesting data.
         """
-        pass
+        raise NotImplementedError("This method is not implemented.")
```
tozka, line 270: Why Is this necessary?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Example added, thanks

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
doks5: Added.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: I was thinking that it might be useful to have some sort of warning in case one or more methods are not implemented, but it probably not necessary in this case, so I have reverted back to the use of `pass`.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -167,7 +167,6 @@ def initialize_job(context: JobContext) -> None:
 class IIngesterPlugin:
```
tozka, line 167: Can you write pydoc here explaining the the lifecycle of the 3 methods .

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -29,6 +29,18 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
+            indicating the sequence in which the ingestion plugins would process the
+            payload. For example:
+                   INGEST_PLUGIN_EXECUTION_SEQUENCE="pre-ingest-process,ingest-http"
```
tozka, line 38: Also mention how does it work with ingest method configuration. 

Also why execution - but otherwise you are mentioning process? Would this mean we can send data to multiple locations ?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
gageorgiev: > Otherwise, I would suggest a param referencing a mutable object, encapsulated within the the scope of the chaining method.

@ivakoleva could you clarify what you mean here?

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -29,6 +29,18 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             default_value=None,
             description="Default Ingestion target to be used.",
         )
+        config_builder.add(
+            key="INGEST_PLUGIN_EXECUTION_SEQUENCE",
```
gageorgiev, line 33: What would happen if I've set both this parameter and INGEST_METHOD_DEFAULT?

gageorgiev (commented):
Personally I would really benefit from seeing an actual implementation example as part of this PR, either as actual code or just a scenario inside the PR description. Right now my understanding of what the goal here is and how it's being achieved is very hazy.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
mivanov1988: I totally agree with @gageorgiev. @doks5 could you please add examples and use cases. I am not sure what problem we are trying to solve.

mivanov1988:
I totally agree with @gageorgiev. @doks5 could you please add examples and use cases. I am not sure what problem we are trying to solve.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
tozka: >Why would a data engineer register multiple ingestion plugins?

Users can have installed: 
vdk-trino - provide ingest to trino 
vdk-impala provides ingest to impala
vdk-ingest-http provides ingest ot HTTP 
And plugin that needs to convert data to a certain format (to json ld is the concrete example we have in VMware). 


And they all could be installed at the same time. 

We give options to data engineers to specify which of those plugins they want to use with configuration.
VDK_INGEST_TARGET_DEFAULT=trino


> Does Pluggy support parallelism configuration when evaluating implementations, for example .__sample_hook_spec.decorate_operation()
For this PR we will use the router functionality. Perhaps we could have written the router logic to be based on pluggy but when I was working on vdk-core initially I did not see how. Now I have some ideas but the ship has sailed for now. But I can open a ticket to look into this again.

doks5: There seems to have been some discussions on using pluggy in the context of plugin chaining ([enhancement request](https://github.com/pytest-dev/pluggy/issues/51)), and a ([discussion](https://github.com/pytest-dev/pluggy/issues/150)), but I am not sure if anything has been actually implemented as part of pluggy.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Done, thanks

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
doks5: Added additional information.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
doks5: INGEST_METHOD_DEFAULT would be ignored.

doks5:
> I totally agree with @gageorgiev. @doks5 could you please add examples and use cases. I am not sure what problem we are trying to solve.

I've added workflow examples as pydoc to the interface.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -165,9 +165,72 @@ def initialize_job(context: JobContext) -> None:
...
+
+    Example plugin chaining workflows:
+
+    Workflow 1
+    ----------
+    _____________________      _________________________      ______________
+    |      Data Job     |      |    vdk-ingest-http    |      |            |
+    |-------------------|      |-----------------------|      | Data Lake/ |
+    |                   |      |                       |      | Warehouse/ |
+    |send_object_       | ---> |   ingest_payload()    | ---> | Third-party|
```
ivakoleva, line 200: It seems to me pydocs would rather benefit from a (pseudo)code snippet in favour of ascii art.
Maybe we could also add a page in the wiki about ingestion, pre- and post-processing.
I would suggest adding to PR test coverage that ensures the newly added methods are evaluated when implementation provided. That would be the full pre-process->ingest->post-process usage example (happy path), that also verifies chaining.

ivakoleva (changes requested):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: @ivakoleva Could you elaborate what examples you're looking for? Generally, method definitions contain code examples: https://github.com/vmware/versatile-data-kit/pull/579/files#diff-65fe93ce6e4724edf9ad045d4ac6c1bc46a189cabf33ab6cc9eb4a00304d68aeR289

tozka: I like the ASCII art. I think it clearly shows the data flow.

tozka: But I agree a bit more about code snippet. But perhaps we can make an example for this - https://github.com/vmware/versatile-data-kit/wiki/Examples

The thing that is not clear   is what would be the implementation of vdk-ingest-http. 
For example vdk-ingest-http implements only ingest_payload ? While vdk-validate would implement only process_payload?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -165,9 +165,72 @@ def initialize_job(context: JobContext) -> None:
...
+    Workflow 2
+    ----------                                                  ______________
+                                                                |            |
+    _____________________      ___________________________      | Data Lake/ |
+    |      Data Job     |      |     vdk-ingest-http     |      | Warehouse/ |
+    |-------------------|      |-------------------------|      | Third-party|
+    |                   |      | -- ingest_payload()     | ---> | endpoint   |
+    |send_object_       | ---> | |                       |      |____________|
+    |    for_ingestion()|      | |                       |      ______________
+    |                   |      | -->post_ingest_process()| ---> |            |
```
tozka, line 214: Most likely it would be a separate plugin that implements telemetry. 

For example, we can send telemetry to apache Atlas using vdk-atlas plugin or opentelemetry using vdk-opentelemetry plugin or some internal telemetry using vdk-http-plugin (which is the case internally in VMware).

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: @tozka With what we have now, **vdk-validate** would need to implement only process_payload(), while **vdk-ingest-http** would need to implement *ingest_payload()* and *post_ingest_process()* ([ingester base implementation](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py#L526))

If we want to have complete separation, i.e., different plugins to implement only one method (e.g., **vdk-validate** -> *process_payload()*; **vdk-ingest-http** -> *ingest_payload()*; **vdk-telemetry** -> *post_ingest_process()*), then we would need to do substantial redesign of **IngesterRouter** and **IngesterBase**.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: See my comment in the thread above. Thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
tozka: We need vdk-telemetry to be separated as we'd need same telemetry regardless of database used (we know of use-cases where it would be necessary for both impala and trino). 

But I see no need of any substantial redesign . 

The default implementation of all methods could be -a no-op . For example: 
```
def ingest_payload(): 
      pass 
def process_payload(payload)
      # no -op return same payload
      return payload
 ```

Then a plugin can choose which method to overwrite

Comment (doks5) [reply projects/vdk-core/src/vdk/api/plugin/plugin_input.py:200]: I agree that having separate plugins is the best option. My concern, however, is that if for example, the **vdk-ingest-http** plugin wants to calculate the size of the payload before the post request is made, and pass it to the telemetry plugin, we would need to add functionality to return that information to the IngestRouter, and from there, pass it to the telemetry plugin. This may have performance penalty which we would need to account for.

Comment (tozka) [reply projects/vdk-core/src/vdk/api/plugin/plugin_input.py:200]: This should not be vehichle for communication between plugins. Plugin need to be independant. If telemetry plugin needs to know the size of the payload it can calculate it itself.

If trino plugin wants to send trino specific telemetry it'd need to hook itself to telemetry hooks provided by telemetry plugin perhaps. But that's out of scope for this change.

File: projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py (hunk 2 above)
Comment (ivakoleva) [reply projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py]: So plugins granularity presently is such that for example, when a data job using Impala runs, then the entire `vdk-impala` is wired. This plugin bootstraps the data source connection configuration. Also, the plugin includes ingestion payload processing capabilities. All the capabilities need to be enabled by the data job -specific configurations explicitly, is that correct? 
For example, VDK_DB_DEFAULT_TYPE=IMPALA needs to be configured to actually connect to Impala data source for executing queries purposes. What needs to be configured to pre-process the ingestion payload? What needs to be configured to actually ingest the payload on pre-processing completion?

Maybe add a test implementation, covering the Processing data job workflow? The test provides a real-world code snippet on happy path example usage. Like connect to Impala, then select data to load in-memory and transform (for example, convert to uniform lower/upper case of all string values via a test-scope implementation). Next, that data payload is automatically decorated to a specific conventional format (for example append a system timestamp property that is also auto-populated by a custom implementation in the scope of the test). After processing, that data job is submitted for ingestion via vdk-ingest-http.

File: projects/vdk-core/src/vdk/api/plugin/plugin_input.py [hunk 11]
```diff
@@ -165,9 +165,72 @@ def initialize_job(context: JobContext) -> None:
...
+    together.
+
+    The lifecycles of the three methods are as follows:
+
+    ingest_payload() - plugins implementing this method are intended
+                       to do actual data ingestion. Versatile Data Kit will
+                       call this method when the payload that data engineers
+                       have passed is ready to be ingested.
+
+    process_payload() - plugins implementing this method do not aim at
```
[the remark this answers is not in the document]
doks5: Done, thanks for the suggestion

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Added tests as discussed offline.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
doks5: Tests added. Thanks for the feedback

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Discussed offline. Thanks

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -0,0 +1,57 @@
...
+        return [{k: str(v) for (k, v) in i.items()} for i in payload]
+
+    @hookimpl
+    def initialize_job(self, context: JobContext) -> None:
+        log.info("Initialize data job with ConvertPayloadValuesToString Plugin.")
+
+        context.ingester.add_ingester_factory_method("convert-to-string", lambda: self)
+
+
+def test_chained_ingest():
```
tozka, line 34: Shouldn't the method has configuration ? 
```
INGEST_PLUGIN_PROCESS_SEQUENCE=convert-to-string,memory
```

another alternative
```
INGEST_PLUGIN_PROCESS_SEQUENCE=convert-to-string
```

and so on for the other cases ....

tozka (approved):
Looks good to me.

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
doks5: Good catch. Fixed, thanks

tozka: I think we'd need test for the other cases. 
cases: 
 
1. INGEST_PLUGIN_PROCESS_SEQUENCE=A,B in job code no method set
2. INGEST_PLUGIN_PROCESS_SEQUENCE=A,B in job code method = C
3. INGEST_PLUGIN_PROCESS_SEQUENCE=A,B and VDK_INGEST_METHOD_DEFAULT=C
4. INGEST_PLUGIN_PROCESS_SEQUENCE=A,B and  VDK_INGEST_METHOD_DEFAULT=C and job code is D
5. INGEST_PLUGIN_PROCESS_SEQUENCE=A,B and in job code method = B also (which I think is the current shown case)
6. INGEST_PLUGIN_PROCESS_SEQUENCE="" 

But as long as we are on the same page on their behaviour we can do the actual tests later: 
So what is expected behavior?
Maybe?
1. A,B are used for pre-processing, B is used for ingesting and post-procesing 
2. A,B is ...

projects/vdk-core/tests/functional/ingestion/utils.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import os
+import pathlib
+
+from vdk.plugin.test_utils.util_funcs import get_test_job_path
+
+
+def job_path(job_name: str) -> str:
+    return get_test_job_path(
```
ivakoleva, line 10: If you think get_test_job_path could be simplified, have you considered contributing to vdk-test-utils to reuse everywhere?

doks5: Will refactor the method and maybe add it to vdk-test-utils in a separate PR. Thanks

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -29,6 +29,26 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
+            indicating the sequence in which the ingestion plugins would process the
+            payload. For example:
+                   INGEST_PLUGIN_PROCESS_SEQUENCE="pre-ingest-process,http"
+            would mean that the payload sent for ingestion would be first processed by
+            a `pre-ingest-process` plugin, before being ingested using the `ingest-http`
+            plugin.
+            NOTE: The value of this environment variable takes precedence
+            over the value in the `INGEST_METHOD_DEFAULT` environment
+            variable, i.e., if both variables are set, the value of
+            `INGEST_METHOD_DEFAULT` would be ignored. If, however, a method
```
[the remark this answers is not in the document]
doks5: Renamed the env var. Could you elaborate on the first sentence?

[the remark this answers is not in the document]
doks5: Refactored as discussed offline.

doks5:
> Could you link to specific issue?

Not sure there is a specific issue for this change.

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -0,0 +1,105 @@
...
+    @hookimpl
+    def initialize_job(self, context: JobContext) -> None:
+        log.info("Initialize data job with ConvertPayloadValuesToString Plugin.")
+
+        context.ingester.add_ingester_factory_method("convert-to-string", lambda: self)
+
+
+@mock.patch.dict(
+    os.environ,
+    {"INGEST_PAYLOAD_PREPROCESS_SEQUENCE": "convert-to-string,memory"},
```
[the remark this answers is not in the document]
doks5: Oopsie, forgot to remove memory from the list. In this specific test case no, but I will add such capability to the memory plugin, or will create a new plugin with both pre-processing and ingestion capabilities in a separate PR.

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -0,0 +1,108 @@
...
+    # assert ingest_plugin.payloads[0].destination_table == "object_table"
+
+    # expected_rows_object = {"first": "two", "second": "2"}
+    # assert ingest_plugin.payloads[1].payload[0] == expected_rows_object
+    # assert ingest_plugin.payloads[1].destination_table == "tabular_table"
+
+
+@mock.patch.dict(
+    os.environ,
+    {"INGEST_PAYLOAD_PREPROCESS_SEQUENCE": "convert-to-string"},
```
ivakoleva, line 81: Is the test testing a chain of preprocessing plugins?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -214,6 +277,25 @@ def ingest_payload(
...
+        string before the actual ingestion happens, we could have a plugin,
+        implementing this method, and doing some payload processing:
+
+        .. code-block:: python
+            def pre_ingest_process(self, payload: List[dict]) -> List[dict]:
+                # Ensure all values in the payload are strings
+                return [{k: str(v) for (k,v) in i.items()} for i in payload]
+
+        :param payload: List[dict] the ingestion payload to be processed.
+        :return: List[dict], the processed payload.
```
ivakoleva, line 295: Could we also recieve and return, along with the payload, a dict containing metadata? For example:
```
NewType("IngestionMetadata", Dict)

def pre_ingest_process(self, payload: List[dict], metadata: IngestionMetadata) -> List[dict], Optional?[IngestionMetadata]:
```
This is common to the Optional[IngestionResult] we return after (successful) ingestion. 


The use case for this ask is, we need to be able to send metadata about ingestion start time for example. Or length of the payload in case it has been non-successfully ingested. For example, a pre-processor would populate:
```
'send_start_time': datetime.datetime.utcnow().isoformat(),
'send_end_time': None,
'http_status': -1
```
Then, a post-processor could overwrite to populate send_end_time.

That metadata dictionary, needs to be available for each pre-processing, ingestion and and post-processing step - to be read or modified. Note: all of those steps happen sequentially.

ivakoleva (changes requested):

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
doks5: It does now.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: I've added the additional interface feature, though am not sure to what extent would it be used.

ivakoleva (approved):
Great, thanks for elaborating thoroughly, this change of interface surface that adds support for ingestion pre- and post- processing along with metadata, is fundamental

A small note, I would suggest using IngestionMetadata consistently in favour of swapping to IngestionResult. It seems simpler, and no copying needs to be implemented, and described as expected behaviour.

PR #574, opened 2021-11-29
control-service: data job execution statuses refinement

mivanov1988:
Currently the Job execution statuses are not well designed.
We have the following statuses:

    SUBMITTED
    RUNNING
    FINISHED
    FAILED
    CANCELLED
    SKIPPED

Where the FAILED represents two statuses: USER_ERROR and PLATFORM_ERROR.
One of the our internal clients wants to see the exact status
(e.g. USER ERROR rather than FAILED). Also the client wants to filter
the executions by USER_ERROR or PLATFORM_ERROR.

The FAILED status is split into USER_ERROR and PLATFORM_ERROR. Also the
FINISHED is renamed to SUCCEEDED.

Testing done: unit and integration tests pass

Signed-off-by: Miroslav Ivanov miroslavi@vmware.com

PR #581, opened 2021-11-30
vdk-control-cli: bump vdk-control-service-api version

mivanov1988:
Bump vdk-control-service-api version to 1.0.6. The new version
changes DataJobExecution statuses.

This change will be merged after https://github.com/vmware/versatile-data-kit/pull/574.

### DIFF
--- a/projects/vdk-control-cli/requirements.txt
+++ b/projects/vdk-control-cli/requirements.txt
@@ -6,7 +6,7 @@ pluggy~=0.13
 tabulate
 click-spinner
 
-vdk-control-service-api==1.0.5
+vdk-control-service-api==1.0.6
 requests_oauthlib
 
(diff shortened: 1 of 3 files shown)


projects/vdk-control-cli/requirements.txt
```diff
@@ -6,7 +6,7 @@ pluggy~=0.13
 tabulate
 click-spinner
 
-vdk-control-service-api==1.0.5
+vdk-control-service-api==1.0.6
```
doks5, line 9: Is the new API package released, as it does not seem to be in PYPI?

doks5 (approved):
Apart from the fact that the package does not seem to be in PYPI, LGTM.

projects/vdk-control-cli/requirements.txt
mivanov1988: That's why I added in the description "This change will be merged after #574.".

PR #582, opened 2021-11-30
control-service: graphQl add sorting by notifications

mrMoZ1:
why: Users requested the ability to sort DataJobs
by notifications. This is a part of the graphql epic.

what: Added the ability to sort by notifications enabled
or not. We check if one of the contacts is populated and
if it is, we threat it as a notification enabled. The DataJobs
will then get sorted by notification enabled status

testing: Added unit tests. 

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
@@ -29,7 +29,8 @@ public enum JobFieldStrategyBy {
    DESCRIPTION("config.description", "content/config/description"),
    SOURCE_URL("config.sourceUrl", "content/config/sourceUrl"),
    NEXT_RUN_EPOCH_SECS("config.schedule.nextRunEpochSeconds", "content/config/schedule/nextRunEpochSeconds"),
-   SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron");
+   SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron"),
+   DATA_JOB_CONTACTS("config.contacts", "content/config/contacts");
 
    private final String field;
    private final String path;
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
@@ -30,7 +30,7 @@ public enum JobFieldStrategyBy {
    SOURCE_URL("config.sourceUrl", "content/config/sourceUrl"),
    NEXT_RUN_EPOCH_SECS("config.schedule.nextRunEpochSeconds", "content/config/schedule/nextRunEpochSeconds"),
    SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron"),
-   DATA_JOB_CONTACTS("config.contacts", "content/config/contacts");
+   DATA_JOB_CONTACTS("config.contacts.present", "content/config/contacts/present");
 
    private final String field;
    private final String path;
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyByDataJobContacts.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyByDataJobContacts.java
@@ -11,10 +11,12 @@ import com.vmware.taurus.service.graphql.model.Filter;
 import com.vmware.taurus.service.graphql.model.V2DataJob;
 import com.vmware.taurus.service.graphql.strategy.FieldStrategy;
 import org.apache.commons.collections.CollectionUtils;
+import org.springframework.stereotype.Component;
 
 import java.util.Comparator;
 import java.util.function.Predicate;
 
+@Component
 public class JobFieldStrategyByDataJobContacts extends FieldStrategy<V2DataJob> {
    private static final Comparator<V2DataJob> COMPARATOR_DEFAULT = Comparator.comparing(e -> e.getConfig().getContacts(),
          Comparator.nullsLast((o1, o2) -> {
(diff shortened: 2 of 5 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyByDataJobContacts.java
```diff
@@ -0,0 +1,49 @@
...
+         }));
+
+   @Override
+   public JobFieldStrategyBy getStrategyName() {
+      return JobFieldStrategyBy.DATA_JOB_CONTACTS;
+   }
+
+   @Override
+   public Criteria<V2DataJob> computeFilterCriteria(Criteria<V2DataJob> criteria, Filter filter) {
+      var predicate = criteria.getPredicate();
```
mivanov1988, line 33: It would be good if validate the filter as it is not supported. You can take a look at JobFieldStrategyByNextRun for reference.

mivanov1988:
We need IT test as well.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
```diff
@@ -29,7 +29,8 @@
    DESCRIPTION("config.description", "content/config/description"),
    SOURCE_URL("config.sourceUrl", "content/config/sourceUrl"),
    NEXT_RUN_EPOCH_SECS("config.schedule.nextRunEpochSeconds", "content/config/schedule/nextRunEpochSeconds"),
-   SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron");
+   SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron"),
+   DATA_JOB_CONTACTS("config.contacts", "content/config/contacts");
```
mivanov1988, line 33: What do you think about `config.contacts.enabled`?

mrMoZ1: Maybe config.contacts.present would be a more accurate name?

mivanov1988: `config.contacts.present` will work.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyByDataJobContacts.java
mrMoZ1: Added a test

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
mrMoZ1: changed in latest commit.

mrMoZ1:
> We need IT test as well.

added IT.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
```diff
@@ -29,7 +29,8 @@
    DESCRIPTION("config.description", "content/config/description"),
    SOURCE_URL("config.sourceUrl", "content/config/sourceUrl"),
    NEXT_RUN_EPOCH_SECS("config.schedule.nextRunEpochSeconds", "content/config/schedule/nextRunEpochSeconds"),
-   SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron");
+   SCHEDULE_CRON("config.schedule.scheduleCron", "content/config/schedule/scheduleCron"),
+   DATA_JOB_CONTACTS("config.contacts.present", "content/config/contacts/present");
```
tozka, line 33: what is present?

tozka: I am askng because our GraphQL model should be similar to our API model which is resource based. 
So when you query graphql you get a list of resources (data jobs in our case) which some fields being filtered out - and also possibly sorted by some fields. 

Each data job can have configuration which can have contracts. 

https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls#L119

mrMoZ1: If we have a DataJobContact present or not. For more context see this conversation https://github.com/vmware/versatile-data-kit/pull/582#pullrequestreview-819074713 . The config.contacts.present field is intended to be used only when sorting. This type of sort was requested by users.

mrMoZ1: We don't have to since we aren't planning on returning anything associated with this field and if we update it, if this field is requested the returned value will always be null.

mivanov1988: Actually, @tozka has a point. We should consider adding this field to the schema, but definitely in separate PR.

mrMoZ1: The problem with adding it to the schema is that it will require a large (ish) refactor since right now DataJobContact is a class generated by the openApi specification that we return as is. This means that we will have to have another V2DataJobContact class to return in the graphQL API which will have the extra property and translate the openApi generated class? Or is there another solution?

mivanov1988: We just need to add the new column to the DataJobContact.

mivanov1988 (approved):

tozka (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/strategy/datajob/JobFieldStrategyBy.java
mrMoZ1: Created https://github.com/vmware/versatile-data-kit/issues/610

PR #590, opened 2021-12-02
vdk-impala: Fix issue caused by timeout parameter having a float value

gageorgiev:
If the VDK_IMPALA_TIMEOUT env variable has a value, any query fails with
```
debug: Retrying vdk.internal.builtin_plugins.connection.managed_connection_base.ManagedConnectionBase.connect in 30.0 seconds as it raised TypeError: can't multiply sequence by non-int of type 'float'.
```
This change fixes this issue by converting the parameter to int.

Testing done: observed query was successful after the implemented change


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_connection.py
```diff
@@ -52,7 +52,7 @@ def _connect(self):
             host=self._host,
             port=self._port,
             database=self._database,
-            timeout=self._timeout,
+            timeout=int(self._timeout) if self._timeout else None,
```
doks5, line 55: Out of curiosity, the timeout is set through an env var, but what would happen if a user mistypes the value and adds a symbol like a `,` by mistake? The interpreter would generally throw a ValueError, but would we have a clear way of showing that to users?

doks5 (approved):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_connection.py
gageorgiev: Doesn't seem like it, right now we get:
```
Retrying vdk.internal.builtin_plugins.connection.managed_connection_base.ManagedConnectionBase.connect in 60.0 seconds as it raised ValueError: invalid literal for int() with base 10: '4,'.
```
and this is only in the debug logs, as far as a normal user is concerned VDK is frozen.

gageorgiev: Eventually it times out with `ValueError: invalid literal for int() with base 10: '4,`.

doks5: I see. Well, the fact that these are visible only in debug logs is probably not the best UX, but I guess this could be a topic for another discussion.

tozka: The type of the variable is determined by the default value - https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py#L54 

Maybe we can set default value '-1' or '0' ?

PR #596, opened 2021-12-03
vdk: improve badges, address some minor things from CII evaluaotin

tozka:
We did self-evaluation of https://bestpractices.coreinfrastructure.org/

And this is addressing some very minor issues and recommendations noted
in above link which I fixed while I was filling the evaluation.

You can see link of all current issues that we still need to fix here:
https://bestpractices.coreinfrastructure.org/en/projects/5363

This is also adding a few badges and details on how to subscribe to
youtube channel


doks5 (approved):
LGTM

PR #600, opened 2021-12-06
vdk-control-cli: enable setting many options at once with vdk deploy

tozka:
vdk deploy --update should be able to set more than one option at once -
e.g you may want to enable and vdk version at same time in single atomic
operation. For example this is necessary in vdk version to ensure next
run uses the specified version.

We also want to keep `vdk deploy --enable` as well as `vdk deploy --update --enable --vdk-version` 

Testing Done: unit tests

Signed-off-by: Antoni Ivanov <aivanov@vmware.com>`

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
@@ -3,6 +3,7 @@
 import os
 from enum import Enum
 from enum import unique
+from typing import Optional
 
 import click
 from vdk.internal.control.command_groups.job.deploy_cli_impl import JobDeploy
@@ -19,8 +20,6 @@ class DeployOperation(Enum):
 
     CREATE = "create"
     UPDATE = "update"
-    ENABLE = "enable"
-    DISABLE = "disable"
     REMOVE = "remove"
     SHOW = "show"
 
@@ -104,14 +103,16 @@ vdk deploy --update --job-version <job-version-here> -n example-job -t job-team
 )
 @click.option(
     "--enable",
-    "operation",
-    flag_value=DeployOperation.ENABLE,
+    "enabled",
+    flag_value=True,
+    default=None,
     help="Enable a job. That will basically un-pause the job.",
 )
 @click.option(
     "--disable",
-    "operation",
-    flag_value=DeployOperation.DISABLE,
+    "enabled",
+    flag_value=False,
+    default=None,
     help="Disable a job. Will not schedule a new cloud execution. Effectively pausing the job. "
     "Currently running job will be allowed to finish.",
 )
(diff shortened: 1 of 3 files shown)


projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
```diff
@@ -154,13 +155,18 @@ def update(
...
+
         if job_version:
             self.__update_job_version(name, team, deployment, output)
-        elif vdk_version:
+        elif vdk_version or enabled is not None:
             self.__update_deployment(name, team, deployment)
-            log.info(
-                f"Deployment of Data Job {name} updated to use vdk version {vdk_version}."
-            )
+            msg = f"Deployment of Data Job {name} updated: "
```
doks5, line 164: This message seems a bit confusing. Maybe could be re-written to something like, `Data Job deployment updated; Status: `?

doks5 (changes requested):

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
```diff
@@ -163,27 +164,21 @@ def deploy(
...
     output: str,
 ):
     cmd = JobDeploy(rest_api_url, output)
-    if operation == DeployOperation.UPDATE:
+    if operation == DeployOperation.UPDATE or enabled is not None:
         name = get_or_prompt("Job Name", name)
         team = get_or_prompt("Job Team", team)
-        if not vdk_version and not job_version:
+        # default to ask for job version to update
+        if not vdk_version and not job_version and enabled is None:
```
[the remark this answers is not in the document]
tozka: We want to default to ask for job version to update if non is explicitly set. 

E.g 
```
vdk deploy --update
```
And CLI prompts for Job Version 

But if user has set something 
```
vdk deploy --update --vdk-version xx
```
or 
```
vdk deploy --update --enable
``` 
Then we do not prompt for anything and simply update based on the passed options .

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
tozka:
```
Deployment of Data Job xxx updated: vdk version: 1.3; enabled; 
```

doks5: For consistency's sake, maybe replace the `:` after 'updated' with a `;`, and add 'status: ' before 'enabled'.

Also, `nit:` on line 168 below, there is NO `;` after "enabled", while there is one after "disabled".

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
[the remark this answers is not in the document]
doks5: I see. Ok.

doks5 (approved):

PR #602, opened 2021-12-06
vdk-control-cli: remove unnnecessary options

tozka:
vdk-control-cli no longer has option to disable or enable authentication
as if auth is needed is authomatically detected instead (https://github.com/vmware/versatile-data-kit/pull/305)

So we no longer neet `vdk set-default --authentication-disable` so I am
removing it.

Testing Done: unit tests


doks5 (approved):
LGTM

Issue #587, opened 2021-12-01
Impala default resource pool configuration


ivakoleva:
**What is the feature request? What problem does it solve?**
Supports a configurable default resource pool.

**Suggested solution**
Extend vdk-impala plugin.
Use db_connection_decorate_operation hook specification to execute SET REQUEST_POOL= to resource pool configured value.

Acceptance criteria:
Configurable default resource pool.

PR #611, opened 2021-12-09
vdk-impala: Allow users to specify SYNC_DDL and query pool in config

gageorgiev:
Until now, if a user wanted to either enable SYNC_DDL or run queries
in a specific pool, they had to manually set them through a query.
After this change, they can now be set as config variables and
queries will be decorated with the necessary additions.
Additionally, this change includes an extra error classification method
in the Impala error handler, which checks for pool errors.
Addresses #587 

Testing done: ran job locally and confirmed queries are correctly
decorated depending on the config variables IMPALA_SYNC_DDL and
IMPALA_QUERY_POOL, included functional test for the decoration hook


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -134,28 +48,149 @@ def impala_query(ctx: click.Context, query):
...
-@hookimpl
-def vdk_command_line(root_command: click.Group):
-    root_command.add_command(impala_query)
+class ImpalaPlugin:
+    @hookimpl
+    def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
+        """
+        Here we define what configuration settings are needed for Impala with reasonable defaults
+        """
+        config_builder.add(
```
tozka, line 57: Let's separate configuration in separate module  and use Wrapper class - similar to https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/control_service_configuration.py

tozka (approved):
Looks good to me. I have a suggestion for small refactoring which it would be nice if you can do .

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -1,161 +1,94 @@
...
+        self._impala_cfg = ImpalaPluginConfiguration(
+            context.core_context.configuration
+        )  # necessary for the query decoration hook
+        context.connections.add_open_connection_factory_method(
+            "IMPALA",
+            lambda: _connection_by_configuration(context.core_context.configuration),
+        )
 
+    @hookimpl
+    def db_connection_recover_operation(self, recovery_cursor: RecoveryCursor) -> None:
```
tozka, line 70:
```suggestion
    @hookimpl
    @staticmethod
    def db_connection_recover_operation(recovery_cursor: RecoveryCursor) -> None:
```

nit: you may consider adding `@staticmethod` annotation to get rid of the codacy warning

tozka (approved):

ivakoleva:
I am looking for test coverage, that when configured, those set statements do happen

tozka:
> I am looking for test coverage, that when configured, those set statements do happen

That's a good point. 
I recommend functional type of test aka https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-impala/tests/test_impala_connection.py (basically data job). 

I recommend to have such test for each new feature we add (unless it's trivial maybe)


Also split the tests into functional directory ala vdk-core layout: https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-core/tests 

It is not split this way for all plugins  because most plugins have 1 or 2 test files. vdk-impala is going to be a bit more

ivakoleva (approved):

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_connection.py
```diff
@@ -42,3 +45,26 @@ def test_execute_query(self) -> None:
...
+        os.environ,
+        {
+            VDK_DB_DEFAULT_TYPE: "IMPALA",
+            VDK_IMPALA_HOST: "localhost",
+            VDK_IMPALA_PORT: "21050",
+            VDK_IMPALA_SYNC_DDL: "True",
+            VDK_IMPALA_QUERY_POOL: "testing_grounds",
+        },
+    )
+    @patch("vdk.plugin.impala.impala_plugin.DecorationCursor.execute")
```
tozka, line 59: Hmm. I'd prefer we do not use patching in functional tests. 

Functional tests often serves as a proxy/example for users using given feature. 

We have a pretty powerful plugin system that's perfect for way to "mock" in a way an end user would

You can create a a plugin - similar to https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-test-utils/src/vdk/plugin/test_utils/util_plugins.py#L102  QueriesRecorderPlugin which can record SQL statements. And now you can verify below that 'sync_ddl" was recorded by the plugin

gageorgiev: Unfortunately the approach seen in the linked place doesn't work - the SYNC_DDL and REQUEST_POOL decoration queries are executed through the decoration cursor, and as such they do not get ran through the decoration hook - I verified this. I'm open to other suggestions but I see no clear way of plugging into the execution of those queries, except through some sort amendment of the decoration implementation.

Personally, I don't see an issue with the patching that happens here, as no behaviour is altered, and only the call_args_list is checked for the necessary queries. It's possible my understanding of the mocking module is wrong, though.

tozka: In this case the problem is not that behaviour is altered. The problem is that the functional test relies on internal implementation classes instead of relying on public contracts (even if it's in verification step only)

If in the future some developer decides to rename DecorationCursor or change the method signature of DecorationCursor#execute (add new parameter) -  it would break this test despite that the actual logic being tested is not broken. 
Also I do think that functional test could be even used as an example to show users how things can be done. They need to mimic what users can do. 

But if it is not possible. I guess we can live with it. I might have been wrong it's a good idea to be a functional test this.

tozka (approved):

PR #615, opened 2021-12-10
control-service: fix lastDeployedDate and lastDeployedBy

mivanov1988:
Currently the lastDeployedBy and lastDeployedDate do not return from
GraphQL API.

Testing Done: unit tests


tozka:
Shouldn't we add some check in integration tests? In other words how we'd catch if we do not get them populated  in the future again?

mivanov1988:
> Shouldn't we add some check in integration tests? In other words how we'd catch if we do not get them populated in the future again?

I will add in a separate PR, since it is a bit urgent.

tozka (approved):

PR #619, opened 2021-12-10
control-service: improve integration tests stability

mivanov1988:
Currently the integration tests depend on the same data job.
Sometimes it may take a little longer to complete the data job execution
between the tests. As a result the API will return 409
and the particular test will fail.

This change adds synchronization between data job executions.

Testing Done: integration tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/BaseDataJobDeploymentIT.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/BaseDataJobDeploymentIT.java
@@ -5,6 +5,14 @@
 
 package com.vmware.taurus.datajobs.it.common;
 
+import static org.awaitility.Awaitility.await;
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.user;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+
+import java.util.UUID;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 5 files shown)


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/BaseDataJobDeploymentIT.java
```diff
@@ -38,4 +49,35 @@ public TaskExecutor taskExecutor() {
...
+            "release");
+
+      // Wait for the job execution to complete, polling every 15 seconds
+      // See: https://github.com/awaitility/awaitility/wiki/Usage
+      String finalOpId = opId;
+      return await()
+            .atMost(5, TimeUnit.MINUTES)
+            .with()
+            .pollInterval(15, TimeUnit.SECONDS)
+            .until(() -> mockMvc.perform(post(triggerDataJobExecutionUrl)
```
tpalashki, line 75: Shouldn't you execute the job first, then wait for the completion?

tpalashki (commented):
Can you elaborate on how the synchronization is done? Are you awaiting for the execution to complete before proceeding further?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/BaseDataJobDeploymentIT.java
mivanov1988: Maybe `synchronization` is not the right term. The data job execution may not have been completed by the previous test and the API may return 409 instead of 202. This method waits for data job execution completion and then starts a new one.

PR #620, opened 2021-12-10
vdk-impala: Move time.sleep fix to conftest

gageorgiev:
The vdk-impala functional tests would require a time.sleep statement
at their start due to an issue where the Impala would return a response
with exit code 0, but the service would not yet be available, which
would cause tests to fail.
This change moves that time.sleep call to the pytest fixture function
for two reasons: it logically makes more sense for it to be there, and
this way it will only ever be invoked once, instead of at the start of
every single functional test.

### DIFF
--- a/projects/vdk-plugins/vdk-impala/tests/conftest.py
+++ b/projects/vdk-plugins/vdk-impala/tests/conftest.py
@@ -53,3 +53,4 @@ def impala_service(docker_ip, docker_services):
     docker_services.wait_until_responsive(
         timeout=90.0, pause=0.3, check=lambda: _is_responsive(runner)
     )
+    time.sleep(10)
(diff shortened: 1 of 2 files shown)


doks5 (approved):
LGTM

projects/vdk-plugins/vdk-impala/tests/conftest.py
```diff
@@ -53,3 +53,4 @@ def impala_service(docker_ip, docker_services):
     docker_services.wait_until_responsive(
         timeout=90.0, pause=0.3, check=lambda: _is_responsive(runner)
     )
+    time.sleep(10)
```
ivakoleva, line 56: Shouldn't those wait workarounds be encapsulated within _is_responsive(), to guarantee responsiveness?

gageorgiev: The issue is that _is_responsive doesn't guarantee it, since a 0 exit code is returned before the service is up.

PR #624, opened 2021-12-13
vdk-core: Publish release candidate distribution and image

YanaZhivkova:
We want to avoid releasing a broken vdk-core distribution as
much as possible. This is why we are working on adding a
pre-release step in vdk-core's CI/CD which would run vdk-heartbeat
with the new version (release candidate) before actually releasing
it.

Create release candidate vdk-core distribution and base image.
Patch version of the release-candidate distribution:
${CI_PIPELINE_ID}dev.${CI_PIPELINE_IID} - corresponding to
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#pre-release-versioning

Testing Done:
- released a test distribution to
https://test.pypi.org/legacy/ by manually executing the steps
from vdk-core-release-candidate CI/CD job. Then installed it and ran
"vdk version" to check that it is as expected.
- test CI/CD pipeline on pull request - https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/428658230 - it successfully released the release candidate distribution to test.pypi.org (with overridden URL and password env vars) and we can see that deploy vdk image script fails near its end when reading a protected password var. The rest of the image deployment script will be tested in the pipeline on merge to main.


.gitlab-ci.yml
```diff
@@ -15,6 +15,7 @@ stages:
   - build
   - publish_artifacts
   - pre_release
+  - pre_release_image
```
tozka, line 18: I am adding the stage in this PR: https://github.com/vmware/versatile-data-kit/pull/631 

This also would change it so control service test are not triggered unnecssarily

tozka (commented):
What about running vdk-heartbeat ? Is that planned in separate change?

tozka:
Can you please link the github issue ?

YanaZhivkova:
> 

Yes - https://github.com/vmware/versatile-data-kit/issues/530

YanaZhivkova:
> 

Yes - https://github.com/vmware/versatile-data-kit/issues/530

projects/vdk-core/.gitlab-ci.yml
```diff
@@ -4,6 +4,9 @@
 
 image: "python:3.9"
 
+variables:
```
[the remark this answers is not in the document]
YanaZhivkova: Done.

projects/vdk-core/.gitlab-ci.yml
```diff
@@ -83,12 +83,60 @@ vdk-core-simple_func_test_with_py310:
...
+    - echo "Build and release candidate vdk image of vdk-core"
+    - apk add --no-cache git bash python3
+    - python3 -V
+    - python3 -m pip install -U setuptools configparser
+    - export VDK_RELEASE_OVERRIDE_PATCH_VERSION=${CI_PIPELINE_ID}.dev${CI_PIPELINE_IID}
+    - ./cicd/set-patch-version.sh "$VDK_RELEASE_OVERRIDE_PATCH_VERSION"
+    - export VDK_PACKAGE=$(python3 setup.py --name)
+    - export VDK_VERSION=$(python3 setup.py --version)
+    - export BUILD_TYPE=pre_release
+    - bash -x ./cicd/deploy-base-vdk-image.sh
```
tozka, line 124: This is base image with only vdk-core (without any plugins). That means for vdk-heartbeat I think we can only use "empty" type of test (data job that does nothing - just prints)

 https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py#L167  
 
 
 Which I think it's good enough actually as it would test vdk-core efficiently.

tozka (approved):

PR #625, opened 2021-12-14
control-service: Allow job builder run as non-root

doks5:
Currently, the data job builder requires root priviledges to
run, which is not ideal in situations, where users may have access
to kubernetes clusters which allow pods to run only under SecurityContext.

This change allows the job builder pods to run under security context.

Testing Done: TODO


projects/control-service/projects/job-builder/Dockerfile
```diff
@@ -4,19 +4,22 @@ FROM gcr.io/kaniko-project/executor
 
 FROM alpine
 
-COPY --from=0 /kaniko /kaniko
-
+COPY --chown=1000:2000 --from=0 /kaniko /kaniko
```
mivanov1988, line 7: Make user and group configurable and set the default values - respectively user=1000 and group=2000.

projects/control-service/projects/job-builder/Dockerfile
```diff
@@ -4,19 +4,22 @@ FROM gcr.io/kaniko-project/executor
...
 ENV SSL_CERT_DIR=/kaniko/ssl/certs
 ENV DOCKER_CONFIG /kaniko/.docker/
 
 WORKDIR /workspace
 
-COPY Dockerfile.python.vdk /workspace/Dockerfile
-COPY build_image.sh /build_image.sh
-RUN chmod +x /build_image.sh
+RUN cd .. \
+    && chown -R 1000:2000 workspace \
```
mivanov1988, line 16: You can use full path to folder
```suggestion
    RUN chown -R 1000:2000 /workspace
```

projects/control-service/projects/job-builder/Dockerfile
```diff
@@ -25,6 +28,7 @@ RUN apk add --no-cache --update \
     bash
 
 RUN apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.10/main python3=3.7.10-r0 py3-pip \
+    && pip3 install -U pip \
```
mivanov1988, line 31: What is the purpose of this command?

projects/control-service/projects/job-builder/Dockerfile
mivanov1988: So you don't need `cd ..` and `cd workspace`.

doks5: Good point. Refactored, thank you

mivanov1988 (approved):

mivanov1988:
LGTM

PR #626, opened 2021-12-14
vdk-core: provide needed error details for vdk run 

tozka:
There was some corner cases that error message was missing - if job
failed with an exception for example during initialize or finalize phases.

Also made sure job finishes with summary of job execution so it's easier
for users to see what happened.
Also if job step failed - it was printing the exception twice (at step
level and job level) which is just cluttering the logs

Tesitng Done: unit tests.
Error message looks like this: 
```
  "data_job_name": "sql-job-syntax-error",
  "execution_id": "63d0ccf3-258f-46e1-9d50-b246402ff926-1639474925",
  "start_time": "2021-12-14T09:42:05.656806",
  "end_time": "2021-12-14T09:42:05.690940",
  "status": "error",
  "steps_list": [
    {
      "name": "bad-query.sql",
      "type": "sql",
      "start_time": "2021-12-14T09:42:05.656809",
      "end_time": "2021-12-14T09:42:05.690905",
      "status": "error",
      "details": "An exception occurred, exception message was: Trino query failed\n  WHAT HAPPENED : Trino query failed with user error\nWHY IT HAPPENED : Error message was: line 3:10: mismatched input ';'. Expecting: '%', '*', '+', ',', '-', '.', '/', 'AND', 'AS', 'AT', 'EXCEPT', 'FETCH', 'FROM', 'GROUP', 'HAVING', 'INTERSECT', 'LIMIT', 'OFFSET', 'OR', 'ORDER', 'UNION', 'WHERE', 'WINDOW', '[', '||', <EOF>, <identifier>, <predicate>. Likely the query has an error that needs to be address. See above error for more details.\n   CONSEQUENCES : The SQL query will fail and the job step likely will fail.\nCOUNTERMEASURES : Please fix the error and try again.",
      "exception": {
        "message": "Trino query failed\n  WHAT HAPPENED : Trino query failed with user error\nWHY IT HAPPENED : Error message was: line 3:10: mismatched input ';'. Expecting: '%', '*', '+', ',', '-', '.', '/', 'AND', 'AS', 'AT', 'EXCEPT', 'FETCH', 'FROM', 'GROUP', 'HAVING', 'INTERSECT', 'LIMIT', 'OFFSET', 'OR', 'ORDER', 'UNION', 'WHERE', 'WINDOW', '[', '||', <EOF>, <identifier>, <predicate>. Likely the query has an error that needs to be address. See above error for more details.\n   CONSEQUENCES : The SQL query will fail and the job step likely will fail.\nCOUNTERMEASURES : Please fix the error and try again.",
        "is_logged": true,
        "cause_exception_name": "TrinoUserError",
        "cause_exception": {
          "_error": {
            "message": "line 3:10: mismatched input ';'. Expecting: '%', '*', '+', ',', '-', '.', '/', 'AND', 'AS', 'AT', 'EXCEPT', 'FETCH', 'FROM', 'GROUP', 'HAVING', 'INTERSECT', 'LIMIT', 'OFFSET', 'OR', 'ORDER', 'UNION', 'WHERE', 'WINDOW', '[', '||', <EOF>, <identifier>, <predicate>",
            "errorCode": 1,
            "errorName": "SYNTAX_ERROR",
            "errorType": "USER_ERROR",
            "errorLocation": {
              "lineNumber": 3,
              "columnNumber": 10
            }, 
      },
      "blamee": "User Error"
    }
  ],
  "exception": null,
  "exception_name": "UserCodeError"
}
```

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
@@ -10,8 +10,6 @@ from typing import Optional
 
 import click
 from vdk.internal.builtin_plugins.run import job_input_error_classifier
-from vdk.internal.builtin_plugins.run.data_job import DataJob
-from vdk.internal.builtin_plugins.run.data_job import DataJobDefaultHookImplPlugin
 from vdk.internal.builtin_plugins.run.data_job import DataJobFactory
 from vdk.internal.builtin_plugins.run.execution_tracking import (
     ExecutionTrackingPlugin,
@@ -47,7 +45,7 @@ class CliRunImpl:
                 wrap_in_vdk_error=True,
             )
 
-    def run_job(
+    def create_and_run_data_job(
         self,
         context: CoreContext,
         data_job_directory: pathlib.Path,
@@ -61,12 +59,11 @@ class CliRunImpl:
         )
         args = self.__validate_and_parse_args(arguments)
 
+        execution_result = None
         try:
             execution_result = job.run(args)
             log.info(f"Data Job execution summary: {execution_result}")
-            if execution_result.is_failed() and execution_result.get_exception():
-                raise execution_result.get_exception()
-        except Exception as e:
+        except BaseException as e:
             errors.log_and_rethrow(
                 job_input_error_classifier.whom_to_blame(
                     e, __file__, data_job_directory
@@ -80,6 +77,8 @@ class CliRunImpl:
                 + " logs for details and ensure the prerequisite for the failed component (details in stacktrace).",
                 exception=e,
             )
+        if execution_result.is_failed() and execution_result.get_exception_to_raise():
+            raise execution_result.get_exception_to_raise()
 
 
 @click.command(
@@ -115,4 +114,6 @@ def run(ctx: click.Context, data_job_directory: str, arguments: str) -> None:
     )
     context: CoreContext = cast(CoreContext, ctx.obj)
     run_impl = CliRunImpl()
-    run_impl.run_job(context, pathlib.Path(data_job_directory), arguments)
+    run_impl.create_and_run_data_job(
+        context, pathlib.Path(data_job_directory), arguments
+    )
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py
@@ -8,6 +8,9 @@ from typing import List
 from typing import Optional
 
 from vdk.internal.builtin_plugins.run.run_status import ExecutionStatus
+from vdk.internal.core import errors
+from vdk.internal.core.errors import ErrorMessage
+from vdk.internal.core.errors import PlatformServiceError
 
 
 @dataclass(frozen=True)
@@ -30,9 +33,8 @@ class StepResult:
     details: Optional[str]
     # Exception if thrown
     exception: Optional[BaseException] = None
-
-    def __repr__(self):
-        return self.__dict__.__repr__()
+    # who is responsible for resolving the error
+    blamee: Optional[errors.ResolvableBy] = None
 
 
 class ExecutionResult:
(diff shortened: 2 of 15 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -80,6 +77,8 @@ def run_job(
                 + " logs for details and ensure the prerequisite for the failed component (details in stacktrace).",
                 exception=e,
             )
+        if execution_result.is_failed() and execution_result.get_exception():
```
YanaZhivkova, line 80: Shouldn't there be a check if execution_result is not None?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py
```diff
@@ -26,13 +29,12 @@ class StepResult:
     end_time: datetime
     # status ?
     status: ExecutionStatus
-    # string with details.
+    # string with detxails.
```
YanaZhivkova, line 32:
```suggestion
    # string with details.
```

projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py
```diff
@@ -65,21 +67,53 @@ def is_success(self):
...
-            data["exception"] = str(self.exception.__class__.__name__)
+            data["exception_name"] = str(self.exception.__class__.__name__)
 
         step_lists_of_dicts = list()
         for step in self.steps_list:
             d = step.__dict__.copy()
             d["start_time"] = d["start_time"].isoformat()
             d["end_time"] = d["end_time"].isoformat()
+            if step.exception:
+                data["exception_name"] = str(step.exception.__class__.__name__)
```
YanaZhivkova, line 105:
```suggestion
                d["exception_name"] = str(step.exception.__class__.__name__)
```

projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_data_job.py
```diff
@@ -23,7 +23,8 @@ def test_run_when_step_fails():
 
     result = data_job.run()
     assert result.is_failed()
-    assert isinstance(result.exception, IndentationError)
+    assert result.exception is None
```
YanaZhivkova, line 26: Why should result.exception be None, as result.get_exception() should be IndentationError? Is this not contradictory?

YanaZhivkova (commented):
I think there are some issues which should be addressed

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
tozka: It cannot be really None .
a) Either it is initialized in the `try:` block
b) or in `except` block `log_and_rethrow` would raise an exception

 I am setting `execution_result = None` because python linter/compiler does not realize  `log_and_rethrow` always throws and complains.

projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_data_job.py
tozka: Hmm. I need to change `get_exception` name perhaps.

If a step fails - exception is recorded only for the step in StepResult
If the job fails for other reason (for example initialize_job fails or something else outside of the step fails) then we set ExecutionResult.exception is populated only then.

get_exception aims to figure out what is the appropriate exception to raise to the user - if there's ExectuionResult level exception then it raises it otherwise it raises the first StepResult exception . If none is found it raises its own exception.

tozka: > I need to change get_exception name perhaps.


Any ideas?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py
tozka: Thanks! Good catch

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
YanaZhivkova: Yes, I somehow skipped the part that log_and_rethrow always raises and exception.

projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_data_job.py
YanaZhivkova: It seems to me that you already renamed it to get_exception_to_raise, which I think is a good candidate.

YanaZhivkova (approved):
Look good to me.

tozka:
> Look good to me.

Thanks for the review

PR #628, opened 2021-12-14
vdk-trino: extract TrinoConfiguration in separate module

tozka:
Change trino configuration to use the established pattern for
configuration

Testing Done: unit tests

### DIFF
--- a/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/trino_config.py
+++ b/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/trino_config.py
@@ -1,4 +1,102 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+from typing import cast
+
+from vdk.internal.core.config import Configuration
+
+TRINO_HOST = "TRINO_HOST"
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/trino_config.py
```diff
@@ -1,4 +1,93 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+from typing import cast
+
+from vdk.internal.core.config import Configuration
 
 trino_templates_data_to_target_strategy: str = ""
+
+
```
gageorgiev, line 9: Extract the strings as constants.
```suggestion
TRINO_HOST = "TRINO_HOST"
TRINO_PORT = "TRINO_PORT"
...etc
```

gageorgiev (approved):
Extract all the config strings as constant strings.

projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/trino_config.py
tozka: Yes, good point. thanks.

PR #631, opened 2021-12-14
vdk: add new stage (pre_release_image) and fix trigger rules

tozka:
Control Service CICD should not be triggered if top level gitlab-ci is
changed. As top level CI contains only orchestration (stages and top
level CICDs); No other project CI is triggered if top-level is changed
so it is consistent

Testing Done: this PR


YanaZhivkova (approved):

doks5:
Judging by the Gitlab CI/CD, it looks like the Control Service pipelines are still run. Is this expected?

.gitlab-ci.yml
```diff
@@ -15,6 +15,7 @@ stages:
   - build
   - publish_artifacts
   - pre_release
+  - pre_release_image
```
doks5, line 18: Judging by the Gitlab CI/CD, the Control Service pipelines are still run. Is this expected?

tozka: Yes, because I have edited gitlab-ci yaml file in control service

PR #632, opened 2021-12-14
control-service: fix graphql executions loading

mivanov1988:
If we request jobs and their executions without filters at the execution level,
the API will return all executions from the database per each data job.

Testing Done: unit tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/ExecutionDataFetcher.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/ExecutionDataFetcher.java
@@ -114,14 +114,16 @@ public class ExecutionDataFetcher {
    }
 
    private Page<DataJobExecution> findAllExecutions(DataJobExecutionQueryVariables dataJobExecutionQueryVariables, String dataJobName) {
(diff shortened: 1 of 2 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/ExecutionDataFetcher.java
```diff
@@ -114,14 +114,16 @@
...
-            throw new GraphQLException("The jobNameIn filter is not supported for nested executions");
-         }
-         filter = filter.toBuilder()
-                 .jobNameIn(List.of(dataJobName))
-                 .build();
+      DataJobExecutionFilter filter = dataJobExecutionQueryVariables.getFilter() != null ?
+            dataJobExecutionQueryVariables.getFilter() :
+            DataJobExecutionFilter.builder().build();
+
+      if (filter.getJobNameIn() != null) {
```
tpalashki, line 121: Shouldn't we still check if `dataJobName` is not null here? Otherwise, the exception will be raised when getting all executions with a filter containing `jobNameIn`.

mivanov1988: Good catch :)

tpalashki (approved):

PR #639, opened 2021-12-16
control-service: add tracking for and automatic release of locks

tpalashki:
Currently, if the Control Service is shut down all (shed)locks remain active.
This causes subsequent instances of the service to fail when trying to
obtain said locks and as a result, the functionality behind these locks
is delayed until the previous locks expire.

This change attempts to release all active locks on service' shutdown
by keeping track of all acquired and released locks and handling the
`@PreDestroy` hook where all active locks are manually released.

Testing done: manually tested the change by running the service,
waiting for a lock to be acquired, then shutting down the service
and verifying that the lock was released (inspecting the database directly).
Also, running the service again has no issues acquiring the lock.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/locks/CustomLockProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        private final SimpleLock wrappedLock;
+        private final String name;
+        private final Map<String, SimpleLock> activeLocks;
+
+        public LockWrapper(SimpleLock wrappedLock, String name, Map<String, SimpleLock> activeLocks) {
+            this.wrappedLock = wrappedLock;
+            this.name = name;
+            this.activeLocks = activeLocks;
+            this.activeLocks.put(name, this);
+            log.debug("Acquiring lock {}", name);
```
mivanov1988, line 67: May be the log should be in the beginning of method. If something go wrong we will have a marker before that.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/locks/CustomLockProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        @Override
+        public void unlock() {
+            wrappedLock.unlock();
+            activeLocks.remove(name);
+            log.debug("Releasing lock {}", name);
```
mivanov1988, line 74: May be the log should be in the beginning of method. If something go wrong we will have a marker before that.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/locks/LockConf.java
```diff
@@ -0,0 +1,48 @@
...
+     * monitoring in our case) can only occur in one of the service instances
+     * (in case of a multi-instance service deployment). This is achieved
+     * by using a database table and the LockProvider returned by this method
+     * is used to connect to that table.
+     *
+     * @see   <a href="https://github.com/lukas-krecan/ShedLock">ShedLock</a>
+     */
+    @Bean
+    public LockProvider lockProvider(DataSource dataSource) {
+        customLockProvider = new CustomLockProvider(dataSource, "shedlock");
```
mivanov1988, line 38:
```suggestion
        return new CustomLockProvider(dataSource, "shedlock");
```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/locks/LockConf.java
```diff
@@ -0,0 +1,48 @@
...
+        return customLockProvider;
+    }
+
+    @PreDestroy
+    void releaseActiveLocks() {
+        if (customLockProvider != null) {
```
mivanov1988, line 44: Log message?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/locks/LockConf.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+import javax.annotation.PreDestroy;
+import javax.sql.DataSource;
+
+@EnableScheduling
+@EnableSchedulerLock(defaultLockAtMostFor = "10m")
+@Configuration
+public class LockConf {
```
mivanov1988, line 20: Can we use more descriptive names?

mivanov1988:
Thanks for doing this change. It will eliminate the time after deployment in which the monitoring logic does not work.

mivanov1988 (approved):

PR #641, opened 2021-12-16
control-service: fix job monitoring

mivanov1988:
At the moment, after the control service is being redeployed,
all K8s jobs and K8s PODs to them are retrieved in order to
synchronize potentially missed K8s events.
Usually, the process synchronizes the K8s Jobs for the last few minutes
but in case of redeploy it synchronizes all jobs
as the variable `lastWatchTime`=0.

This change sets the appropriate `lastWatchTime` default value, 
thus reducing the synchronization time.

Testing Done: unit tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMonitor.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMonitor.java
@@ -27,6 +27,7 @@ import org.springframework.stereotype.Component;
 
 import javax.transaction.Transactional;
 import java.io.IOException;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


tozka:
So if the control service is down for 1 hour, it is going to miss the first half an hour of events?

ivakoleva:
Sounds like a persistent dynamic value to me

tpalashki:
`` is something that we can persist

tpalashki:
Closed by mistake. Apologies.

mivanov1988:
> So if the control service is down for 1 hour, it is going to miss the first half an hour of events?

Yes, that's correct, but it is temporary solution. We can use DB column `shedlock.locked_at` as a default value.

tpalashki (approved):
I agree with what others say, but I think that this change will solve more problems than it would cause until we implement a more sophisticated solution. I don't believe we had a case with the service down for 30+ minutes. On the other hand, there is a delay during every deployment, which could cause issues.

PR #645, opened 2021-12-17
<artifact/project>: Incremental ingestion from DB example

dvalkova:
Signed-off-by: Desislava Valkova <dvalkova@vmware.com>

# Why?
The example illustrates a common use case of VDK - ingestion into a data lake in an incremental fashion.
It also showcases how to use data job properties which haven't been illustrated in any of the current examples.

# What?
A mock SQLite database is also committed with which users could start off. This was chosen as the easier approach than making them create one from scratch.
Then they are encouraged to add new records to the source in order to see in practice the incremental ingestion effect.

# How has this been tested?
Tested locally with the open source VDK by adding/removing different records from the tables.

New feature - new example to illustrate VDK features.


doks5:
There is something with this diff, that is somewhat unsettling. Are you sure you meant to push all these changes?

PR #648, opened 2021-12-20
vdk-control-cli: Fix traceback shown for broken config.ini

doks5:
Currently, if the config.ini file of a data job contains a special unicode
character, Versatile Data Kit prints the whole traceback, which may confuse
the user as to what is happening.

This change handles the exception and prints a message with more user-friendly
details as to what might be the possible cause for the error. The complete traceback
is printed only in debug mode.

Testing Done: Added a unit test, and tested locally by deploying a data job, whose config.ini
file contained a special unicode character. Verified that no traceback appears in the terminal.
The message shown was:
```
Error: ¯\_(ツ)_/¯

what: Cannot parse the Data Job configuration file.
why: Configuration file config.ini is probably corrupted. Error: File contains no section headers.
file: '/home/andy/Documents/test-jobs/my-second-job/config.ini', line: 1
'Œ; Supported format: https://docs.python.org/3/library/configparser.html#supported-ini-file-structure\n'
consequences: Cannot deploy and configure the data job without properly set config.ini file.
countermeasures: config.ini must be UTF-8 compliant. Make sure the file does not contain special Unicode characters, or that your text editor has not added such characters somewhere in the file.
```

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
@@ -7,6 +7,7 @@ from configparser import ConfigParser
 from pathlib import Path
 
 from vdk.internal.control.exception.vdk_exception import VDKException
+from vdk.internal.control.utils.control_utils import read_config_ini_file
 
 log = logging.getLogger(__name__)
 
@@ -208,7 +209,12 @@ class VDKConfigFolder:
             log.debug(f"Reading configuration file: {configuration_file_path} ...")
 
             config_parser = ConfigParser()
-            config_parser.read(configuration_file_path)
+
+            read_config_ini_file(
+                config_parser=config_parser,
+                configuration_file_path=configuration_file_path,
+            )
+
             return config_parser.get(section=section, option=option, fallback=fallback)
         except OSError as e:
             raise VDKException(
--- a/projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
@@ -3,10 +3,15 @@
 # TODO: this is basically a copy of VDK JobConfig in vdk-core
 import configparser
 import fileinput
+import logging
 import os
 import sys
 
 from vdk.internal.control.exception.vdk_exception import VDKException
+from vdk.internal.control.utils.control_utils import read_config_ini_file
+
+
+log = logging.getLogger(__name__)
 
 
 class JobConfig:
@@ -27,7 +32,9 @@ class JobConfig:
                 "Make sure the file is created "
                 "or double check the data job path is passed correctly.",
             )
-        self._config_ini.read(self._config_file)
+        read_config_ini_file(
+            config_parser=self._config_ini, configuration_file_path=self._config_file
+        )
 
     def get_team(self):
         return self._get_value("owner", "team")
--- a/projects/vdk-control-cli/src/vdk/internal/control/utils/control_utils.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/utils/control_utils.py
@@ -1,0 +1,38 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import logging
+from configparser import ConfigParser
+from configparser import MissingSectionHeaderError
+
+from vdk.internal.control.exception.vdk_exception import VDKException
+
+
+log = logging.getLogger(__name__)
+
+
+def read_config_ini_file(
+    config_parser: ConfigParser, configuration_file_path: str
+) -> None:
+    """
+    Read the Data Job config.ini file.
+
+    :param config_parser: ConfigParser instance to be used for reading the
+    configuration file.
+    :param configuration_file_path: Path of the config.ini file
+    """
+    try:
+        config_parser.read(configuration_file_path)
+    except (MissingSectionHeaderError, Exception) as e:
+        log.debug(e, exc_info=True)  # Log the traceback in DEBUG mode.
+        raise VDKException(
+            what="Cannot parse the Data Job configuration file"
+            f" {configuration_file_path}.",
+            why=f"Configuration file config.ini is probably corrupted. Error: {e}",
+            consequence="Cannot deploy and configure the data job "
+            "without "
+            " properly set config.ini file.",
+            countermeasure="config.ini must be UTF-8 compliant. "
+            "Make sure the file does not contain special "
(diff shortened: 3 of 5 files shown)


gageorgiev:
Could you include what the printed traceback looked like before?

doks5:
> Could you include what the printed traceback looked like before?

Original traceback is available in the linked issue.

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
```diff
@@ -208,7 +209,25 @@ def read_configuration(self, section, option, fallback=None):
...
+                    consequence="Cannot deploy and configure the data job "
+                    "without "
+                    " properly set config.ini file.",
+                    countermeasure="config.ini must be UTF-8 compliant. "
+                    "Make sure the file does not contain "
+                    "special "
+                    "Unicode characters, or that your text "
+                    "editor "
+                    "has not added such characters somewhere "
+                    "in the file.",
```
gageorgiev, line 229: Did the precommit hooks mess up the text formatting here?

projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
```diff
@@ -27,7 +31,22 @@ def __init__(self, data_job_path):
                 "Make sure the file is created "
                 "or double check the data job path is passed correctly.",
             )
-        self._config_ini.read(self._config_file)
+        try:
+            self._config_ini.read(self._config_file)
+        except (configparser.MissingSectionHeaderError, Exception) as e:
```
gageorgiev, line 36: configparser is only used for two objects, might be better to import them specifically but this is a nitpick on my part

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
```diff
@@ -208,7 +209,25 @@ def read_configuration(self, section, option, fallback=None):
             log.debug(f"Reading configuration file: {configuration_file_path} ...")
 
             config_parser = ConfigParser()
-            config_parser.read(configuration_file_path)
+            try:
+                config_parser.read(configuration_file_path)
+            except (MissingSectionHeaderError, Exception) as e:
+                log.debug(e, exc_info=log.getEffectiveLevel() == logging.DEBUG)
```
tozka, line 215: What is this ? log.getEffectiveLevel() == logging.DEBUG seems unnecessary since log.debug ...

gageorgiev (approved):

projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
```diff
@@ -27,7 +31,22 @@ def __init__(self, data_job_path):
...
+            log.debug(e, exc_info=log.getEffectiveLevel() == logging.DEBUG)
+            raise VDKException(
```
tozka, line 38: you can create helper method to open config.ini instead of duplicating the error handling. 
 
 We should try to follow https://en.wikipedia.org/wiki/Don%27t_repeat_yourself when possible

gageorgiev:
Also, if this is a common issue caused by this character at the start, why not include some logic stripping it to prevent this issue?

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
doks5: Good catch. This was meant for log error, not debug. Fixed, thanks

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
doks5: Either the hooks, or my IDE :D
Fixed, I think. Thanks

projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
doks5: Did some refactoring. Hopefully, it is clearer now

projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
doks5: Agreed. Fixed, thanks

doks5:
> Also, if this is a common issue caused by this character at the start, why not include some logic stripping it to prevent this issue?

It is not necessary to be at the beginning of the file. If a user copy/paste-s from one file into another, it is possible for special Unicode characters to end up anywhere in the config file.

gageorgiev (approved):

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
```diff
@@ -208,7 +209,12 @@ def read_configuration(self, section, option, fallback=None):
             log.debug(f"Reading configuration file: {configuration_file_path} ...")
 
             config_parser = ConfigParser()
-            config_parser.read(configuration_file_path)
+
+            read_job_configuration(
```
tozka, line 213: Here we are not reading job configuration . 

This file is read for any VDK CLI command - not those related to job configuration. If that was so we would have used JobConfig class. 

Maybe name the method read_ini_config_file ? or something like this

projects/vdk-control-cli/src/vdk/internal/control/utils/control_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    :param config_parser: ConfigParser instance to be used for reading the
+    configuration file.
+    :param configuration_file_path: Path of the config.ini file
+    """
+    try:
+        config_parser.read(configuration_file_path)
+    except (MissingSectionHeaderError, Exception) as e:
+        log.debug(e, exc_info=True)  # Log the traceback in DEBUG mode.
+        raise VDKException(
+            what="Cannot parse the Data Job configuration file.",
```
tozka, line 28: Which file cannot be parsed ? Add `configuration_file_path` in the message.

```suggestion
            what=f"Cannot parse the Data Job configuration file {configuration_file_path}.",
```

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
doks5: Renamed, thanks

projects/vdk-control-cli/src/vdk/internal/control/utils/control_utils.py
doks5: Added, thanks

tozka (approved):

PR #649, opened 2021-12-20
vdk-impala: Classify impala errors as User error if needed

gageorgiev:
This change makes sure that if an error occurs during a job run, the
error is analyzed to check if it was caused by the user. If this is
true, an appropriate error message is printed.

Testing done: unit test and a functional test


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_classifier.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _is_query_memory_error(exception):
+    return errors.exception_matches(
+        exception,
+        classname_with_package="impala.error.OperationalError",
+        exception_message_matcher_regex=".*Memory limit exceeded.*",
+    )
+
+
+def _is_redshift_error(exception):
```
ivakoleva, line 130: the redshift user error classification seems out of scope

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_classifier.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _is_property_unsupported_value_error(exception):
+    return errors.exception_matches(
+        exception,
+        classname_with_package="ValueError",
+        exception_message_matcher_regex=".*[pP]roperty.*is of unsupported type.*",
+    )
+
+
+def _is_impala_memory_limit_exceeded(exception):
```
ivakoleva, line 106: Looking at memory limit exceeded conditions, I am missing 
```
(errors.exception_matches(
        exception,
        classname_with_package="MemoryError",
        exception_message_matcher_regex=".*memory limit.*")
```

Also, could combine with the common `_is_query_memory_error()` one into one memory-related block

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -67,6 +70,18 @@ def initialize_job(self, context: JobContext) -> None:
             lambda: _connection_by_configuration(self._impala_cfg),
         )
 
+    @hookimpl(tryfirst=True)
+    def vdk_exception(self, exception: Exception):
+        if is_impala_error(exception):
+            log_and_throw(
```
ivakoleva, line 76: One note - looking at the current `vdk_exception()` wiring, any exception raised within such hook implementation, would not be handled by any plugin in the chain but seems to be raised directly instead. Maybe we need to extend the core hookspec docs on exception raising:
```
 @hookspec
 def vdk_exception(self, exception: Exception) -> bool:
        """
        Called in case the CLI is about to exit with error.

        :return: True if exception is handled and does not need to be logged by main
        """
        pass
```

ivakoleva (commented):
LGTM, would suggest an additional review and adding the tests

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_classifier.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.internal.core import errors
+
+
+def is_impala_error(received_exception):
```
tozka, line 6:
```suggestion
def is_impala_user_error(received_exception):
```

tozka:
I see it's a draft. But where are the unit tests. 

We also need a functional test - it doesn't need to cover all the cases (if they are covered with unit tests) but it should cover 1 or 2 . The functional test should work similar to this one: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-trino/tests/test_trino_sql.py
It is particularly important to check termination message is correctly set.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -67,6 +70,18 @@ def initialize_job(self, context: JobContext) -> None:
             lambda: _connection_by_configuration(self._impala_cfg),
         )
 
+    @hookimpl(tryfirst=True)
+    def vdk_exception(self, exception: Exception):
```
tozka, line 74: Hooking to vdk_exception may not work since classification happens earlier. This way the termination message will not be udpated correctly. 

See how I've done it for trino: https://github.com/vmware/versatile-data-kit/pull/650/files

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_classifier.py
tozka:
```
(errors.exception_matches(
        exception,
        classname_with_package="MemoryError",
        exception_message_matcher_regex=".*memory limit.*")
```
This is not an impala error. It is a generic python memory error and it is covered by the core job input error classifier: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py#L72

ivakoleva:
> I see it's a draft. But where are the unit tests.
> 
> We also need a functional test - it doesn't need to cover all the cases (if they are covered with unit tests) but it should cover 1 or 2 . The functional test should work similar to this one: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-trino/tests/test_trino_sql.py It is particularly important to check termination message is correctly set.

The functional test and unit tests are now present; the unit tests seem to cover all subcases except `_is_udf_error()`

ivakoleva (approved):

PR #654, opened 2021-12-21
control-service/vdk-cli/vdk-heartbeat: set Pipeline ID as op_id for vdk-heartbeat run

tozka:
This PR contains 3 commits 2 of them set op id to be same as $CI_PIPELINE_ID 

By setting op_id to be the Pipeline ID we can easier track a given run of vdk-heartebaet as op_id is printed in the logs, it is passed to the job logs, the job name also contains the Op ID . So one can easier see which pipeline generated which job and vice-versa

And small change in vdk-heartbeat that make sure do not accidentally create job name with too long name (as op id is included in the job name


doks5 (approved):
LGTM

PR #656, opened 2021-12-22
vdk-core: Release acceptance test job added to CI/CD

YanaZhivkova:
We want to avoid releasing broken distributions of vdk-core as much
as possible. The newly added acceptance test would allow us to
release vdk-core only after vdk-heartbeat runs successfully with
a release candidate vdk-core version containing latest changes.

Run vdk-heartbeat on stage pre_release_test of CI/CD with
release candidate distribution and image created during
previous stages - pre_release and pre_release_image.

Testing done: testing CI/CD pipeline on pull request -
https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/434951809
with hard-coded release candidate version of distribution and image (0.0.430145938.dev3240),
which were already created during a previous pipeline - https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/430145938

### DIFF
--- a/projects/vdk-core/.gitlab-ci.yml
+++ b/projects/vdk-core/.gitlab-ci.yml
@@ -129,6 +129,27 @@ vdk-core-release-candidate-base-image:
     changes: *vdk_core_locations
 
 
+vdk-core-release-acceptance-test:
+  stage: pre_release_test
+  before_script:
+    - cd projects/vdk-core
+  script:
+    - pip install -U pip
+    - pip install vdk-heartbeat --extra-index-url $PIP_EXTRA_INDEX_URL
+    - export VDKCLI_OAUTH2_REFRESH_TOKEN=$VDK_API_TOKEN
+    - export VDK_HEARTBEAT_DEPLOY_JOB_VDK_VERSION=$(cat version.txt | grep -o '[0-9]\.[0-9]').${CI_PIPELINE_ID}.dev${CI_PIPELINE_IID}
+    - export JOB_NAME=vdk-core-release-test-job$(date +%s)
+    - vdk-heartbeat -f cicd/release_acceptance_test_heartbeat_config.ini
+  artifacts:
+    when: always
+    reports:
+      junit: tests.xml
+  only:
+    refs:
+      - main
+    changes: *vdk_core_locations
+
+
 vdk-core-release:
   stage: release
   before_script:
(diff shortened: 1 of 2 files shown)


projects/vdk-core/.gitlab-ci.yml
```diff
@@ -129,6 +129,26 @@ vdk-core-release-candidate-base-image:
     changes: *vdk_core_locations
 
 
+vdk-core-release-acceptance-test:
+  stage: pre_release_test
+  before_script:
+    - cd projects/vdk-core
+  script:
+    - pip install vdk-heartbeat --extra-index-url $PIP_EXTRA_INDEX_URL
```
doks5, line 137: nit: Maybe add `pip install -U pip` on the line above, to avoid installation issues with old pip versions.

projects/vdk-core/.gitlab-ci.yml
```diff
@@ -129,6 +129,26 @@ vdk-core-release-candidate-base-image:
...
+    - export VDKCLI_OAUTH2_REFRESH_TOKEN=$VDK_API_TOKEN
+    - export VDK_HEARTBEAT_DEPLOY_JOB_VDK_VERSION=$(cat version.txt | grep -o '[0-9]\.[0-9]').${CI_PIPELINE_ID}.dev${CI_PIPELINE_IID}
```
doks5, line 139: Out of curiosity, is there a specific reason why we use `CI_PIPELINE_IID` to specify the dev suffix? After looking at the linked CI/CD pipelines, it seems like this number can get pretty high. :)

doks5 (approved):
LGTM

projects/vdk-core/.gitlab-ci.yml
YanaZhivkova: Great, thanks!

projects/vdk-core/.gitlab-ci.yml
YanaZhivkova: I took the job name directly from SC analogue CI/CD job. After digging a little on the subject, it seems to me that CI_PIPELINE_IID is supposed to be significantly smaller than CI_PIPELINE_ID and is actually suitable for versioning - https://gitlab.com/gitlab-org/gitlab-foss/-/issues/3691.
On the other hand, CI_PIPELINE_ID is completely unique and I don't see the need of appending another number after "dev". But I decided to keep the code consistent with SC. Maybe I should drop the second number, @doks5 @tozka, what do you think?

doks5: It's fine to use it, I was just curious about the choice, as the number was four-digit. Nevertheless, it is generally expected that there is a number after "dev", as this is something package managers take into account.

YanaZhivkova: How can I add another auto-incrementing value per pipeline, do have suggestions?

doks5: You can keep this one, it's not an issue.

YanaZhivkova: Thanks, Andy!

PR #659, opened 2022-01-04
control-service: Data Job Executions GraphQL API returns null propert…

mivanov1988:
GraphQL API for Data Job Executions returns null for every field
related to Deployment resources and schedule.

Testing Done: Integration tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
+++ b/projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
@@ -70,6 +70,20 @@ type DataJobResources {
     netBandwitdthLimit: Int
 }
 
+type DataJobExecutionDeployment {
+  id: String
+  enabled: Boolean
+  contacts: DataJobContacts
+  jobVersion: String
+  deployedDate: String
+  deployedBy: String
+  mode: String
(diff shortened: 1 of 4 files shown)


projects/control-service/projects/pipelines_control_service/src/main/resources/schema.graphqls
```diff
@@ -70,6 +70,20 @@ type DataJobResources {
     netBandwitdthLimit: Int
 }
 
+type DataJobExecutionDeployment {
```
tpalashki, line 73: What necessitates this to be a separate model? Is it not confusing to have two deployment models with an almost similar structure - one as a parent of the executions and one as a child?

mivanov1988: GraphQL doesn't support inheritance and the DataJobDeployment differs from the model defined in Model project.

mivanov1988: https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/model/apidefs/datajob-api/api.yaml#L1267

tpalashki (approved):

PR #665, opened 2022-01-05
quickstart-vdk: Run vdk-heartbeat before release

YanaZhivkova:
Currently, plugin versions are not fixed in quickstart-vdk
and when we upgrade, we can accidentally break it. That is why it is
necessary to run some verification before we release it.

Run vdk-heartbeat before releasing quickstart-vdk (during pre-release stage).
Create release-candidate package and image of quickstart-vdk and
run vdk-heartbeat with them during pre-release stage of
each quickstart-vdk dependency CI/CD.

Testing Done: CI/CD pipeline - https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/470250661.
Also, check that the release-candidate image of quickstart-vdk is run on the vdk-heartbeat job k8s pod.


projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
```diff
@@ -31,6 +31,32 @@ build-py310-quickstart-vdk:
...
+    - set -x
+    - pip install -U pip
+    - python setup.py sdist --formats=gztar
+    - pip install dist/* --extra-index-url $PIP_EXTRA_INDEX_URL
+    - pip install vdk-heartbeat --extra-index-url $PIP_EXTRA_INDEX_URL
+    - pip list
+    - export VDKCLI_OAUTH2_REFRESH_TOKEN=$VDK_API_TOKEN
+    - export VDK_HEARTBEAT_OP_ID="quickstart-vdk-$CI_PIPELINE_ID"
+    - export JOB_NAME=quickstart-vdk-release-test-job-$(date +%s)
+    - vdk-heartbeat -f cicd/release_acceptance_test_heartbeat_config.ini
```
ivakoleva, line 50: Should the heartbeat use the local k8s setup bootstrapped by quickstart-vdk for the purposes of testing it?

YanaZhivkova: I have configured it to run the test against the demo control service installation. I think quickstart-vdk is configured to run against a local control service by default. Do you mean to run a control service locally and then run the test against it?

ivakoleva: The only `release_acceptance_test_heartbeat_config.in` file found in the project is https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/cicd/release_acceptance_test_heartbeat_config.ini#L5 

It is btw not in cicd/ dir that is relative to `cd projects/vdk-plugins/quickstart-vdk`.

Also, it points to the demo installation the way you explained. At the same time, the purpose of quickstart-vdk is to bootstrap a local environment instead of a remote one like the demo one. 
Therefore, the purpose of heartbeat tests in the context of vdk-quickstart, will be running in that newly bootstrapped environment, to verify operability.

YanaZhivkova: Yes, it seems that I was mistakenly thinking that the purpose of this test is to check if vdk-core specifically is functional when releasing quickstart-vdk. As I see it, I should run "vdk server --install" in the CI/CD job before running vdk-heartbeat and remove all configuration values pointing to the demo installation. Does that sound right?

ivakoleva: > Yes, it seems that I was mistakenly thinking that the purpose of this test is to check if vdk-core specifically is functional when releasing quickstart-vdk
Seems about right to me, the purpose is to check vdk-core in the environment bootstrapped by quickstart-vdk. The operability in demo environment is checked on vdk-core release itself, correct?

YanaZhivkova: Yes, vdk-core tests run against the demo installation. Thank you!

projects/vdk-plugins/quickstart-vdk/cicd/release_acceptance_test_heartbeat_config.ini
```diff
@@ -0,0 +1,9 @@
+[DEFAULT]
+# Generate with https://console-stg.cloud.vmware.com/csp/gateway/portal/#/user/tokens
+# VDKCLI_OAUTH2_REFRESH_TOKEN= passed as environment variable
+VDKCLI_OAUTH2_URI=https://console-stg.cloud.vmware.com/csp/gateway/am/api/auth/api-tokens/authorize
+CONTROL_API_URL=http://cicd-control-service-svc:8092
```
doks5, line 5: Should these links be hard-coded?

YanaZhivkova: Why not? Do you think they should not be public?

doks5: I don't think it's about them being public. More like, what happens if the links change, that would require a code change, as well.
However, I don't think it's anything major, so you can ignore the comment and resolve the conversation.

YanaZhivkova: Yes, I see, you have a point. I checked and the URLs are hard-coded in .ini files frequently. I suppose they do not change often. Maybe, if we decide to use environment vars, we should do it in an effort that addressed all such places. I will check this and write a comment here if we decide to pursue such an effort. Thanks!

projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
tozka: `vdk server --install` would not work in CICD since it requires [kind](https://kind.sigs.k8s.io/) which cannot be easily installed. So for lack of better option, using the demo environment seems fine to me.

projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
```diff
@@ -8,6 +8,22 @@ variables:
   DOCKER_DRIVER: overlay2
   DOCKER_TLS_CERTDIR: ""
 
+# Add all quickstart-vdk dependencies locations which trigger the dependency release when changed.
+# In this way, quickstart-vdk will always be tested before one or more dependencies are changed, and
+# this is important since quickstart-vdk has unpinned dependencies and when they are upgraded,
+# "pip install quickstart-vdk" installs their newest versions
```
tozka, line 14: I am not super fan of adding "hard" dependencies (to the locations/paths) to other plugins / projects as it's better if they are completely independent (e.g if we can pic and move a plugin into separate repo and it would work with little changes).

 But I think we can make an exception for quickstart-vdk as it is an "aggregation" distribution for the purposes of aggregating multiple plugins in a single deliverable. 

But I wanted to note it as it should not be considered good practice. I'd add a comment

tozka (approved):

projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
YanaZhivkova: Yes, I agree with you. I initially did not want to hard-code them, but I could not come up with a solution which is not over-complicated. I added a comment.

PR #666, opened 2022-01-06
vdk-kerberos-auth: provide variable to specify krb5.conf configuration

tpalashki:
Currently, the Kerberos authentication is always using a generic
krb5.conf file built into the plugin. This commits enables specifying
the path to the krb5.conf file via configuration (`KRB5_CONF_FILENAME`).
If the configuration is not set, it defaults to the built-in file.

Testing done: manually running the `run` command with and without the
new configuration and observing that the appropriate krb5.conf file is
being used.

### DIFF
--- a/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/authenticator_factory.py
+++ b/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/authenticator_factory.py
@@ -23,6 +23,7 @@ class KerberosAuthenticatorFactory:
     ) -> Optional[BaseAuthenticator]:
         if authentication_type == "minikerberos":
             return MinikerberosGSSAPIAuthenticator(
+                self.__configuration.krb5_conf_filename(),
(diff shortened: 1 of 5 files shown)


ivakoleva (commented):
Looking at the description, and Testing Done section specifically, I think adding test coverage to verify the configurations are being set as expected, is a trivial task to do; Of course, the need for this would be obsolete, when we come up with functional tests to check entire operability, that are missing presently

ivakoleva:
Have you considered something like https://hub.docker.com/r/openmicroscopy/apacheds ?

tpalashki:
I am currently looking at https://hub.docker.com/r/gcavalcante8808/krb5-server/ and exploring if I can use it to functionally test the entire plugin.

mivanov1988:
LGTM

ivakoleva (approved):
Maybe create another PR with tests with a docker image running?

tpalashki:
> Maybe create another PR with tests with a docker image running?

Thank you, I will.

PR #669, opened 2022-01-10
versatile-data-kit: Establish release process

gageorgiev:
EDIT: This PR is closed due to its irrelevancy. It may be reopened at some point in the future.

This change establishes a release process for the entire VDK repo.
The repo now has a global version.txt file, which, when changed,
caused CICD release steps to trigger for all components, with
versions set to the version in the global file. The automatic
releases made are now set to release as dev builds, and will not
be installed by default unless `pip install` is ran with the
`--pre` flag enabled.
The exact process is described in the CONTRIBUTING.md file.

Testing done: TBD


CONTRIBUTING.md
```diff
@@ -95,3 +95,25 @@ Entrypoint of CICD is the file .gitlab-ci.yml.
...
 For more details see the [CICD wiki in our Dev Guide](https://github.com/vmware/versatile-data-kit/wiki/Gitlab-CICD)
+
+# How to make a new VDK release
+
+There are two kinds of releases made - development releases and stable releases.
+
+Development releases are made automatically and separately for each component when
+any changes to their directory are merged to the main branch. These have the `dev` tag
+in their version and are labelled `pre-release` in PyPI, meaning to install them by
+default one must include the `--pre` flag in their `pip install` call.
```
ivakoleva, line 106: I didn't see an usage of `--pre` flag in the MR, my impression from the documentation is, it would apply to all the dependencies recursively (direct and transient ones). I would suggest specifying a concrete pre-release version if needed in favour of `--pre` flag in that case

ivakoleva:
Reading the PR description, the goal is to have all packages latest snapshots tagged and released with an unified pipeline-unique version - no matter if codebase changes present or not. Is that correct?

For example, if `vdk-impala` plugin had not changed yet `vdk-core` had changed, both will be released - and the latest `vdk-impala` package would be the same file like the previous package published?

CONTRIBUTING.md
gageorgiev: Given dev releases are intended for development, it seems reasonable to pull dev builds for all packages. Unless you mean that dev builds will be pulled for external dependencies as well?

PR #670, opened 2022-01-14
vdk-impala: Cleanup of DecoratedSqLite3MemoryDbPlugin during test phase

ivakoleva:
The DecoratedSqLite3MemoryDbPlugin is being registered for the purposes
of testing Impala, that seems unnecessary.

Dropped DecoratedSqLite3MemoryDbPlugin in context of
`test_impala_connection.py` to simplify.

Testing Done: did run vdk-impala tests locally to verify


gageorgiev (approved):
If I remember correctly I had added that to try to capture a history of the queries executed by the decoration cursor, which ended up not working. Good to have it removed.

PR #672, opened 2022-01-18
vdk-core: Revise parts of the IIngesterPlugin Interface

doks5:
Currently, the IIngesterPlugin defines two new data types, IngestionResult
and IngestionMetadata, that are to be used when doing ingestion tasks. These
two types are essentially overlapping in terms of their purpose, as both are
meant to contain data about the ingestion process itself.

This change removes the IngestionResult type, and leaves only the IngestionMetadata
one, in order to make the interface logic more consistent.

Testing Done: Documentation change, no testing needed.

### DIFF
--- a/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
+++ b/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
@@ -229,7 +229,6 @@ class IIngesterPlugin:
     |___________________|   |_____________|   |___________|   |____________|
     """
 
-    IngestionResult = NewType("IngestionResult", Dict)
     IngestionMetadata = NewType("IngestionMetadata", Dict)
 
     def ingest_payload(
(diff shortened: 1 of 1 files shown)


projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -267,10 +266,10 @@ def ingest_payload(
             meaning all method invocations from a data job run will belong to the
             same collection.
 
-        :return: [Optional] IngestionResult, containing plugin-defined
-        information about the result from the ingestion operation. This result
-        could be used for post-ingestion processing operations like sending
-        telemetry data about the ingested payload.
+        :return: [Optional] IngestionMetadata, containing plugin-defined
+        information about the result from the ingestion operation. This
```
ivakoleva, line 270: information about the ingestion result, along with plugin-populated ingestion metadata?
also, maybe add an example

ivakoleva (approved):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Refactored to make it clearer. Thanks for the review

PR #677, opened 2022-01-25
pipelines-control-service: publish Swagger UI to /data-jobs path

ivakoleva:
The Swagger UI is currently available at root / by default. Since the
API is prefixed by /data-jobs, we would like for the Swagger to also be
available under same path, in a consistent fashion. This unblocks
services composition by dedicated path for accessing all features.

Did add Swagger configuration for additional controllers redirect and
resource handler. Updated an integration test to verify the path.
Updated also SecurityConfiguration swagger paths.

Testing Done: did run unit tests, and also verified swagger-ui.html is
accessible under /data-jobs locally

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobCrudIT.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobCrudIT.java
@@ -68,7 +68,7 @@ public class DataJobCrudIT extends BaseIT {
       testDataJobPostCreateWebHooks();
 
       // Execute get swagger with no user
-      mockMvc.perform(get("/swagger-ui.html")
+      mockMvc.perform(get("/data-jobs/swagger-ui.html")
               .content(dataJobRequestBody)
               .contentType(MediaType.APPLICATION_JSON))
               .andExpect(status().isOk());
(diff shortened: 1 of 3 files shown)


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobCrudIT.java
```diff
@@ -68,7 +68,7 @@ public void testDataJobCrud() throws Exception {
       testDataJobPostCreateWebHooks();
 
       // Execute get swagger with no user
-      mockMvc.perform(get("/swagger-ui.html")
+      mockMvc.perform(get("/data-jobs/swagger-ui.html")
```
doks5, line 71: Out of curiosity, should we think about a way to make this configurable in the future, in case the jobs' path changes again?

doks5 (approved):
LGTM

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobCrudIT.java
ivakoleva: A generic way would be with a configurable the spring/servlet context path, however this would require refactoring

doks5: Gotcha

tpalashki:
I've noticed another SwaggerConfig in the base module: https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/doc/SwaggerConfig.java. Do you know what's its purpose and if it requires any attention in the context of this change?

ivakoleva:
> I've noticed another SwaggerConfig in the base module: https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/doc/SwaggerConfig.java. Do you know what's its purpose and if it requires any attention in the context of this change?

That's good to know; seems like a bare implementation, probably a marker one, would consider adding a comment why added to base module

PR #682, opened 2022-01-27
[vdk-core] Introduce post-ingest-sequence env var

doks5:
As part of the effort to extend the Ingestion API to support plugin
chaining, we need to improve the interfaces that are to be used by
data engineers and plugin developers.

This change adds a new environment variable, `INGEST_PAYLOAD_POSTPROCESS_SEQUENCE`
that data engineers can use to specify a plugin/s that would deal with post-ingestion
processing tasks, such as sending telemetry, etc.

Additionally, the interfaces that plugin developers use in ingestion pre-processing
and post-processing tasks, are extended to pass payload, destination_table, target and
collection_id as read-only parameters to allow for better workflows.

Testing Done: Unit tests pass, and ran a local test ingestion data job as described in
https://github.com/vmware/versatile-data-kit/wiki/Ingesting-data-from-DB-into-Database
to verify that the existing ingestion workflows are not affected by the interface changes.

### DIFF
--- a/projects/vdk-core/cicd/build.sh
+++ b/projects/vdk-core/cicd/build.sh
@@ -23,20 +23,32 @@ export PIP_EXTRA_INDEX_URL=${PIP_EXTRA_INDEX_URL:-https://test.pypi.org/simple/}
 echo "Update pip to newest version"
 pip install -U pip
 
+# Below line uses --use-deprecated=legacy-resolver to temporary workaround a
+# dependency backtracking issue.
+# TODO: Remove the use of the legacy resolver once latest pip works as expected.
 echo "install dependencies from requirements.txt (used for development and testing)"
-pip install --extra-index-url $PIP_EXTRA_INDEX_URL -r requirements.txt
+pip install --use-deprecated="legacy-resolver" --extra-index-url $PIP_EXTRA_INDEX_URL -r requirements.txt
 
 echo "Setup git hook scripts with pre-commit install"
 pre-commit install
 
+# Below line uses --use-deprecated=legacy-resolver to temporary workaround a
+# dependency backtracking issue.
+# TODO: Remove the use of the legacy resolver once latest pip works as expected.
 echo "Install the vdk-core in editable mode (develop mode)"
-pip install -e .
+pip install --use-deprecated="legacy-resolver" -e .
 
+# Below line uses --use-deprecated=legacy-resolver to temporary workaround a
+# dependency backtracking issue.
+# TODO: Remove the use of the legacy resolver once latest pip works as expected.
 echo "Install common vdk test utils library (in editable mode)"
-pip install -e ../vdk-plugins/vdk-test-utils
+pip install --use-deprecated="legacy-resolver" -e ../vdk-plugins/vdk-test-utils
 
+# Below line uses --use-deprecated=legacy-resolver to temporary workaround a
+# dependency backtracking issue.
+# TODO: Remove the use of the legacy resolver once latest pip works as expected.
 echo "Run unit tests and generate coverage report"
-pip install pytest-cov
+pip install --use-deprecated="legacy-resolver" pytest-cov
 pytest --junitxml=tests.xml --cov vdk --cov-report term-missing --cov-report xml:coverage.xml
 
 echo "Done"
--- a/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
+++ b/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
@@ -10,6 +10,7 @@ from typing import Dict
 from typing import List
 from typing import NewType
 from typing import Optional
+from typing import Tuple
 from typing import Union
 
 from vdk.internal.builtin_plugins.connection.managed_connection_base import (
@@ -230,6 +231,7 @@ class IIngesterPlugin:
     """
 
     IngestionMetadata = NewType("IngestionMetadata", Dict)
+    ExceptionsList = NewType("ExceptionsList", List)
 
     def ingest_payload(
         self,
@@ -242,29 +244,32 @@ class IIngesterPlugin:
         Do the actual ingestion of the payload
 
         :param payload: List[dict]
-            The payloads to be ingested. Depending on the number of payloads to be
-            processed, there might 0 or many dict objects. Each dict object is a
-            separate payload.
+            The payloads to be ingested. Depending on the number of payloads
+            to be processed, there might be 0 or many dict objects. Each dict
+            object is a separate payload.
             Note: The memory size of the list is dependent on the
             payload_size_bytes_threshold attribute.
         :param destination_table: Optional[string]
-            (Optional) The name of the table, where the data should be ingested into.
-            This argument is optional, and needs to be considered only when the name
-            of the destination table is not included in the payload itself.
+            (Optional) The name of the table, where the data should be
+            ingested into. This argument is optional, and needs to be
+            considered only when the name of the destination table is not
+            included in the payload itself.
         :param target: Optional[string]
             (Optional) Used to identify where the data should be ingested into.
                 Specifies a data source and its destination database.
                 The values for this parameter can be in the format
                 `<some-data-source_and-db-table>`, or as a URL.
-                Example: http://example.com/<some-api>/<data-source_and-db-table>
+                Example:
+                    http://example.com/<some-api>/<data-source_and-db-table>
             This parameter does not need to be used, in case the
-            `INGEST_TARGET_DEFAULT` environment variable is set. This can be made by
-            plugins, which may set default value, or it can be overwritten by users.
+            `INGEST_TARGET_DEFAULT` environment variable is set. This can be
+            made by plugins, which may set default value, or it can be
+            overwritten by users.
         :param collection_id: string
-            (Optional) An identifier to indicate that data from different method
-            invocations belong to same collection. Defaults to "data_job_name|OpID",
-            meaning all method invocations from a data job run will belong to the
-            same collection.
+            (Optional) An identifier to indicate that data from different
+            method invocations belong to same collection. Defaults to
+            "data_job_name|OpID", meaning all method invocations from a data
+            job run will belong to the same collection.
 
         :return: [Optional] IngestionMetadata, containing data about the
         ingestion process (information about the result from the ingestion
@@ -294,15 +299,20 @@ class IIngesterPlugin:
 
                 return metadata
 
-        :exceptions: Exceptions raised while ingesting data can either be
-        handled at plugin level, or they will be automatically caught at the
-        vdk-core level.
+        :exception: Exceptions raised while ingesting data can either be
+            handled at plugin level, or they will be automatically caught at
+            the vdk-core level.
         """
         pass
 
     def pre_ingest_process(
-        self, payload: List[dict], metadata: Optional[IngestionMetadata] = None
-    ) -> (List[dict], Optional[IngestionMetadata]):
+        self,
+        payload: List[dict],
+        destination_table: Optional[str] = None,
+        target: Optional[str] = None,
+        collection_id: Optional[str] = None,
+        metadata: Optional[IngestionMetadata] = None,
+    ) -> Union[List[Dict], Tuple[List[Dict], IngestionMetadata]]:
         """
         Do some processing on the ingestion payload before passing it to the
         actual ingestion.
@@ -317,21 +327,63 @@ class IIngesterPlugin:
                 return [{k: str(v) for (k,v) in i.items()} for i in payload]
 
         :param payload: List[dict] the ingestion payload to be processed.
+            NOTE: A read-only parameter. Whatever modifications are done to
+            this object, once returned, it is treated as a new object.
+        :param destination_table: Optional[string]
+            (Optional) The name of the table, where the data should be
+            ingested into. This argument is optional, and needs to be
+            considered only when the name of the destination table is not
+            included in the payload itself.
+            NOTE: A read-only parameter. It is not expected to be modified and
+            returned.
+        :param target: Optional[string]
+            (Optional) Used to identify where the data should be ingested into.
+                Specifies a data source and its destination database.
+                The values for this parameter can be in the format
+                `<some-data-source_and-db-table>`, or as a URL.
+                Example:
+                      http://example.com/<some-api>/<data-source_and-db-table>
+            This parameter does not need to be used, in case the
+            `INGEST_TARGET_DEFAULT` environment variable is set. This can be
+            made by plugins, which may set default value, or it can be
+            overwritten by users.
+            NOTE: A read-only parameter. It is not expected to be modified and
+            returned.
+        :param collection_id: string
+            (Optional) An identifier to indicate that data from different
+            method invocations belong to same collection. Defaults to
+            "data_job_name|OpID", meaning all method invocations from a data
+            job run will belong to the same collection.
+            NOTE: A read-only parameter. It is not expected to be modified and
+            returned.
         :param metadata: Optional[IngestionMetadata] dict object containing
-        metadata produced by the pre-ingest plugin, and possibly used by
-        other pre-ingest, ingest, or post-ingest plugins.
-        :return: (List[dict], Optional[IngestionMetadata]), a tuple
-        containing the processed payload and possibly an IngestionMetadata
-        object.
+            metadata produced by the pre-ingest plugin, and possibly used by
+            other pre-ingest, ingest, or post-ingest plugins.
+            NOTE: A read-only parameter. Whatever modifications are done to
+            this object, once returned, it is treated as a new object.
+        :return: Union[List[Dict], Tuple[List[Dict], IngestionMetadata]],
+            either only the processed payload objects, or a tuple containing
+            the processed payload objects and an IngestionMetadata object with
+            ingestion metadata information.
+
+        :exception: If an exception occurs during this operation, all other
+            pre-processing operations and the ingestion of the payload would
+            be terminated to prevent data corruption. In case the ingestion is
+            expected to happen regardless if there are exceptions or not,
+            then the exception handling needs to be implemented within this
+            method.
         """
         pass
 
     def post_ingest_process(
         self,
         payload: Optional[List[dict]] = None,
+        destination_table: Optional[str] = None,
+        target: Optional[str] = None,
+        collection_id: Optional[str] = None,
         ingestion_metadata: Optional[IngestionMetadata] = None,
-        exceptions: Optional[List] = None,
-    ) -> None:
+        exception: Optional[Exception] = None,
+    ) -> Optional[IngestionMetadata]:
         """
         Do post-ingestion processing of the ingestion payload
         or metadata from the ingestion operation.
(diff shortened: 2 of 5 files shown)


projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -320,6 +327,33 @@ def pre_ingest_process(self, payload: List[dict]) -> List[dict]:
         :param metadata: Optional[IngestionMetadata] dict object containing
         metadata produced by the pre-ingest plugin, and possibly used by
         other pre-ingest, ingest, or post-ingest plugins.
+        :param destination_table: Optional[string]
+            (Optional) The name of the table, where the data should be
+            ingested into. This argument is optional, and needs to be
+            considered only when the name of the destination table is not
+            included in the payload itself.
+            NOTE: A read-only parameter. It is not expected to be modified and
```
ivakoleva, line 335: is any parameter mutable, for example `payload`?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -230,6 +231,7 @@ class IIngesterPlugin:
     """
 
     IngestionMetadata = NewType("IngestionMetadata", Dict)
+    ExceptionsList = NewType("ExceptionsList", List)
```
ivakoleva, line 234: why list of exceptions? even with eventual exceptions chaining, would expect one instance when having one ingestion performed

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -320,6 +327,33 @@ def pre_ingest_process(self, payload: List[dict]) -> List[dict]:
...
+            overwritten by users.
+            NOTE: A read-only parameter. It is not expected to be modified and
+            returned.
+        :param collection_id: string
+            (Optional) An identifier to indicate that data from different
+            method invocations belong to same collection. Defaults to
+            "data_job_name|OpID", meaning all method invocations from a data
+            job run will belong to the same collection.
+            NOTE: A read-only parameter. It is not expected to be modified and
+            returned.
```
ivakoleva, line 356: maybe include documentation on expectation if an eventual exception is raised by a pre-processor, regarding ingestion itself

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Yes, payload and metadata are mutable. The notes are only for read-only parameters, that are not expected to be modified.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: It is possible to have exceptions caused by pre-ingestion as well, so it seemed reasonable to have a list, as it is easier to work with.
However, it is not necessary to be a list. Can make it a single exception instance, if you think we don't need to catch multiple exceptions?

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Is it necessary? If something happens during pre-processing, ingestion should be blocked to avoid data corruption.

ivakoleva: It seems fair, if any pre-processor in the sequence configured fails, the ingestion workflow should be interrupted;
This I think should be explicitly added to pre-processor documentation - what happens if an exception is raised - and maybe a note to ingestion docs also

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
ivakoleva: **Case 1 - pre-processing exception**
Given pre-processors "pre1,pre2,pre3" configured, then
if "pre2" fails, "pp3" is not triggered, and no ingestion is expected to happen, 
therefore no post-processing will happen also. Agree?

**Case 2 - ingestion exception**
Given post-processors "post1,post2, post3" configured, then
if an exception occurs during ingestion,
"post1" receives 1 exception object only, then handles post-processing,
then "post2" receives 1 exception only (raised by ingestion),
then "post3" triggered receives the same ingestion exception only.
Agree?

**Case 3 - post-processing exception**
Given post-processors "post1,post2,post3" configured, then
common to pre-processors, any exception raised will interrupt the post-processing sequence.

**Case 3 scenario 1:**
If an user wants to implement a post-processing sequence, where 
each step has a prerequisite the success of the previous one,
raising exception on an exceptional event during "post1" guarantees post-handling atomicity. 
"post2" will execute only after "post1" successful execution.
Agree?

**Case 3 scenario 2:**
If an user wants to implement multiple independent post-processors,
this could be achieved by catching(+logging) exceptions, so the post-processing chain is not interrupted.
Agree?

**Case 3 scenario 3:**
If an user wants to implement stateful post-processing sequence, where 
"step2" will be aware of an eventual exception during "step1",
this could be achieved by passing any exception metadata via Optional[IngestionMetadata].
Agree?


If you agree on all, then no list of exceptions is needed - only one could happen during ingestion,
and all the post-processors are going to be aware of the ingestion state.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
ivakoleva: Aren't all parameters in fact immutable?
Isn't the modified payload and metadata rather returned??

ivakoleva (changes requested):
lets resolve the conversations open

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Correct, from design point of view they are immutable.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Good point. Refactored, exception is now a single object, not a list. Thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Documentation updated. Thank you

doks5:
> lets resolve the conversations open

Conversations resolved. Ready for review

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -242,29 +244,32 @@ def ingest_payload(
         Do the actual ingestion of the payload
 
         :param payload: List[dict]
-            The payloads to be ingested. Depending on the number of payloads to be
-            processed, there might 0 or many dict objects. Each dict object is a
-            separate payload.
+            The payloads to be ingested. Depending on the number of payloads
+            to be processed, there might 0 or many dict objects. Each dict
```
ivakoleva, line 248:
```suggestion
            to be processed, there might be 0 or many dict objects. Each dict
```

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -317,21 +327,62 @@ def pre_ingest_process(self, payload: List[dict]) -> List[dict]:
...
         self,
         payload: Optional[List[dict]] = None,
         ingestion_metadata: Optional[IngestionMetadata] = None,
-        exceptions: Optional[List] = None,
-    ) -> None:
+        exception: Optional[Exception] = None,
+        destination_table: Optional[str] = None,
+        target: Optional[str] = None,
+        collection_id: Optional[str] = None,
+    ) -> Optional[Tuple[Optional[IngestionMetadata], Optional[Exception]]]:
```
ivakoleva, line 385: Optional[Exception] returned is not needed;
when vdk-core evaluates post-processors, it already does have the eventual ingestion exception. It will be provisioned to each of the post-processors in the sequence.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -317,21 +327,62 @@ def pre_ingest_process(self, payload: List[dict]) -> List[dict]:
...
         """
         pass
 
     def post_ingest_process(
         self,
         payload: Optional[List[dict]] = None,
         ingestion_metadata: Optional[IngestionMetadata] = None,
-        exceptions: Optional[List] = None,
-    ) -> None:
+        exception: Optional[Exception] = None,
```
ivakoleva, line 381: Would rather move exception param to last, since it is only passed for exceptional cases therefore rare.
When a user reviews the interface surface, the common to ingest_process() parameters in the known order would be easier to spot, then the newly added parameters.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -343,9 +394,14 @@ def post_ingest_process(
             def post_ingest_process(
                 self,
                 payload: Optional[List[dict]],
-                ingestion_metadata: Optional[IngestionMetadata]
-                exceptions: Optional[List]
-            ) -> None:
+                ingestion_metadata: Optional[IngestionMetadata],
+                exception: Optional[Exception],
```
ivakoleva, line 398: same here, exception goes last

ivakoleva (changes requested):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Fair point. Removed, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Refactored for consistency. Thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: ditto

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
```diff
@@ -216,7 +216,7 @@ def test_ingest_payload_and_post_ingestion_operation():
     )
     ingester_base._ingester.post_ingest_process.assert_called_with(
         payload=test_aggregated_payload,
-        ingestion_result=test_ingestion_result,
+        ingestion_metadata=test_ingestion_result,
         exceptions=[],
```
ivakoleva, line 220: look to cleanup any leftover params from previous pre- or post- ingestion format

ivakoleva (changes requested):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -294,15 +299,20 @@ def ingest_payload(self,
...
     def pre_ingest_process(
-        self, payload: List[dict], metadata: Optional[IngestionMetadata] = None
-    ) -> (List[dict], Optional[IngestionMetadata]):
+        self,
+        payload: List[dict],
+        destination_table: Optional[str] = None,
+        target: Optional[str] = None,
+        collection_id: Optional[str] = None,
+        metadata: Optional[IngestionMetadata] = None,
+    ) -> Union[List[Dict], Tuple[List[Dict], IngestionMetadata]]:
```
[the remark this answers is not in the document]
doks5: The reason for the change is that it is possible that a user wants to do pre-processing, and return the payload, but not metadata. In such case, the new return types allow for either the payload to be returned, or a tuple of payload and metadata. With the old return signature, plugin developers would have needed to return something like `(payload, )` to avoid type errors.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
doks5: Done, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -294,15 +299,20 @@ def ingest_payload(self,
...
 
     def pre_ingest_process(
-        self, payload: List[dict], metadata: Optional[IngestionMetadata] = None
-    ) -> (List[dict], Optional[IngestionMetadata]):
+        self,
+        payload: List[dict],
+        destination_table: Optional[str] = None,
+        target: Optional[str] = None,
+        collection_id: Optional[str] = None,
+        metadata: Optional[IngestionMetadata] = None,
```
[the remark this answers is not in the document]
doks5: Added. Thanks

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
doks5: I believe I have cleaned up the leftovers.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
ivakoleva: Would you share an example code snippet for usages before/after the return types change?

Also, have you considered always returning the metadata, either modified or directly the parameter value provisioned?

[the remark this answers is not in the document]
doks5: To answer your first question, sure, here's an example,
Before:
```python
def pre_ingest_process(self,
    payload: List[dict],
    destination_table: Optional[str],
    target: Optional[str],
    collection_id: Optional[str],
    metadata: Optional[IngestionMetadata]
) -> (List[dict], Optional[IngestionMetadata]):
    for i in payload:
        i["additional_column"] = "some_val"
    return (payload, None)
```

Now:
```python
def pre_ingest_process(self,
    payload: List[dict],
    destination_table: Optional[str],
    target: Optional[str],
    collection_id: Optional[str],
    metadata: Optional[IngestionMetadata]
) -> Union[List[Dict], Tuple[List[Dict], IngestionMetadata]]:
    for i in payload:
        i["additional_column"] = "some_val"
    return payload
```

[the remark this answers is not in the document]
doks5: As for your second question, you mean to force plugin developers to always return metadata?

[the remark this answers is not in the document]
ivakoleva: on the first question, I see; 
should a plugin developer return ingestion metadata always? if it's a part of the contact, and it is always being the output of an operation that always becomes the input of the next one in the sequence, it may be more clear what exactly becomes the input of the "pre2" for example?

My point is, the API should be the simplest and the most obvious way to understand the workflow.
When having a sequence, if I miss to return the second parameter, does that mean the next pre-processor will be provisioned with None metadata, or else what other metadata?
When always returning the metadata, it becomes quite obvious what the next pre-processor is going to be provisioned with.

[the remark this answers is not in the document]
doks5: @ivakoleva 
Seems fair. Ok, I've refactored the method signature, and the docstring to be more explicit as to what the contract is.

ivakoleva (approved):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
ivakoleva:
```
metadata: Optional[IngestionMetadata] = None,
    ) -> Tuple[List[Dict], IngestionMetadata]:

```
the method param is optional, yet the return type is not;
would rather keep them consistent;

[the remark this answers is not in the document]
doks5: Yup, fix it, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -237,34 +238,43 @@ def ingest_payload(
         destination_table: Optional[str],
         target: Optional[str] = None,
         collection_id: Optional[str] = None,
+        metadata: Optional[IngestionMetadata] = None,
```
ivakoleva, line 241: Optional[IngestionMetadata] seems correct; should propagate throughout entire chain, from pre- to post- processing

doks5: Gotcha. Ok, thanks

ivakoleva (approved):

projects/vdk-core/cicd/build.sh
```diff
@@ -21,7 +21,7 @@ cd ..
 export PIP_EXTRA_INDEX_URL=${PIP_EXTRA_INDEX_URL:-https://test.pypi.org/simple/}
 
 echo "Update pip to newest version"
-pip install -U pip
+pip install -U pip==22.0
```
tozka, line 24: write a comment on why it's pinned. And create a ticket to unpinn it.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -237,34 +238,43 @@ def ingest_payload(
...
         """
         Do the actual ingestion of the payload
 
         :param payload: List[dict]
-            The payloads to be ingested. Depending on the number of payloads to be
-            processed, there might 0 or many dict objects. Each dict object is a
-            separate payload.
+            The payloads to be ingested. Depending on the number of payloads
+            to be processed, there might be 0 or many dict objects. Each dict
+            object is a separate payload.
```
tozka, line 249: >  Each dict  object is a separate payload.

That does not make sense ... the variable name is called payload. Payload is list of dictionary objects .. and we get recursive definition if object is a payload as well :)

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -237,34 +238,43 @@ def ingest_payload(
...
                 Specifies a data source and its destination database.
                 The values for this parameter can be in the format
                 `<some-data-source_and-db-table>`, or as a URL.
-                Example: http://example.com/<some-api>/<data-source_and-db-table>
+                Example:
+                    http://example.com/<some-api>/<data-source_and-db-table>
             This parameter does not need to be used, in case the
-            `INGEST_TARGET_DEFAULT` environment variable is set. This can be made by
-            plugins, which may set default value, or it can be overwritten by users.
+            `INGEST_TARGET_DEFAULT` environment variable is set. This can be
```
tozka, line 265: not "environment" variable. says configuration variable better. 

INGEST_TARGET_DEFAULT can be provided using environment variable but also using config.ini

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -237,34 +238,43 @@ def ingest_payload(
...
         :param collection_id: string
-            (Optional) An identifier to indicate that data from different method
-            invocations belong to same collection. Defaults to "data_job_name|OpID",
-            meaning all method invocations from a data job run will belong to the
-            same collection.
+            (Optional) An identifier to indicate that data from different
+            method invocations belong to same collection. Defaults to
+            "data_job_name|OpID", meaning all method invocations from a data
+            job run will belong to the same collection.
+        :param metadata: Optional[IngestionMetadata] dict object containing
```
tozka, line 273: nit: dictionary instead of dict?

tozka (approved):

projects/vdk-core/cicd/build.sh
doks5: Removed.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Indeed, it looks like this docstring had turned into some shady "inception"-style description. :)
Reworded to make it clearer what it actually is. Thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Fixed, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Noted and fixed, thanks

PR #684, opened 2022-01-28
vdk-core: adjust defined type for configuration values

tozka:
Variable can be added (has its type defined) by one plugin after its
value is set by 2nd plugin.   In that cases the value would not be cast
to the correct type nor validated. This moves that casting/validation at
the point of definition (when #add is called) if value is already set
for same config key.

Testing Done: see unit tests

### DIFF
--- a/projects/vdk-core/src/vdk/internal/core/config.py
+++ b/projects/vdk-core/src/vdk/internal/core/config.py
@@ -215,7 +215,6 @@ class ConfigurationBuilder:
             self.__config_key_to_value[key] = convert_value_to_type_of_default_type(
                 key, self.__config_key_to_value[key], default_value
             )
-        pass
 
     def build(self) -> Configuration:
         """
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/internal/core/config.py
```diff
@@ -205,6 +206,17 @@ def __add_public(
...
+    def __adjust_type_if_value_set(self, key: ConfigKey, default_value: ConfigValue):
+        """
+        As set_value can be called before add (which defines the configuration) we may need to adjust
+        the type of the value if it's already set.
+        """
+        if default_value is not None and key in self.__config_key_to_value:
+            self.__config_key_to_value[key] = convert_value_to_type_of_default_type(
+                key, self.__config_key_to_value[key], default_value
+            )
+        pass
```
gageorgiev, line 218: Was this left in accidentally?

gageorgiev (approved):

projects/vdk-core/src/vdk/internal/core/config.py
tozka, line 218:
```suggestion
```

PR #616, opened 2021-12-10
Created Life Expectancy Scenario

alod83:
examples/life-expectancy - An example of Ingestion and Processing Jobs with a Streamlit Dashboard

Why?
The objective of this example is to analyze the life expectancy in the different U.S. Regions and compare them with the Region's Gross Domestic Product (GDP). The example also builds a dashboard that shows results in streamlit.

What?
The change adds a new folder under the examples directory containing the complete example of the scenario.

How has it been tested?
The following manual tests have been run:
- vdk run life-expectancy
- streamlit run dashboard.py 

What type of change are you making?
New Use Case

Signed-off-by: Angelica Lo Duca angelica.loduca [at] gmail.com

PR #685, opened 2022-01-31
vdk-trino: trino ingest to handle type casting and missing values

tozka:
Fix trino ingestion to handle different data types  more gracefully.
It will make sure the the data type (passed in python) is correctly set
based on the data type in the destination table.

Handle missing values (e.g if a value for column X is missing , it will
set it to None / NULL insted of failing

Handle case insensitive fields. The columns names (or field names) passed by the user can be in any case but trino column names are case insensitive so we normalize them to match.

Testing Done: extended functional tests to cover the new scenarios. Also ran the data job in the https://github.com/vmware/versatile-data-kit/pull/616 (with ingestion using job_input.send_...) and saw it completed successfully and ingested the data correctly in my local trino.

### DIFF
--- a/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
+++ b/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
@@ -1,13 +1,18 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import logging
+import math
+from typing import Any
 from typing import List
 from typing import Optional
+from typing import Type
 
 from trino.dbapi import Cursor
 from vdk.internal.builtin_plugins.ingestion.ingester_base import IIngesterPlugin
 from vdk.internal.builtin_plugins.run.job_context import JobContext
 from vdk.internal.core import errors
+from vdk.internal.core.errors import ErrorMessage
+from vdk.internal.core.errors import UserCodeError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


doks5:
> Fix trino ingestion to handle different types of more gracefully.

Different types of?

tozka:
> > Fix trino ingestion to handle different types of more gracefully.
> 
> Different types of?

of data. Data Types 

See unit test example.You can have a column bool and if the value passed to that type is "string" and we need to make sure they are cast correctly so the insert query succeeds

projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
```diff
@@ -57,27 +62,107 @@ def _ingest_payload(
...
+            return False
+        raise ValueError("bool cast accept only True/true/False/false values.")
+
+    def __cast(self, key: str, new_value: Any, value_type: Type) -> Any:
+        try:
+            log.debug(f"Cast {key} to type {value_type}")
+            if value_type == bool:
+                return self.__to_bool(new_value)
+            if value_type == float:
+                value_with_float_type = float(new_value)
```
doks5, line 101: Shouldn't there be a nested try-except here, in case `new_value` cannot be safely converted to float?

doks5 (approved):
LGTM

projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
tozka: It is covered by line 108

doks5: So it is assumed that if the value cannot be converted to float, then everything up until line 108 should be skipped, correct?

tozka: yes. Do you see any issue with that?

doks5: Nope, just got a bit confused by the math.isnan check that follows, and thought that value_with_float_type is expected to have value after the conversion to float. However, I see why failing at the float conversion is better than handling the exception and checking the value after that.

PR #690, opened 2022-02-01
[vdk-core] Implementation of new ingestion interfaces

doks5:
As part of https://github.com/vmware/versatile-data-kit/pull/682
and https://github.com/vmware/versatile-data-kit/pull/672 the ingestion
interfaces were extended to support ingestion plugin chaining for pre- and
post- processing operations.

This change introduces the implementation of the extended ingestion functionality.
The IngesterRouter and IngesterBase classes are extended to support the use of pre-
and post- ingestion processing plugins.

Testing Done: unit tests; functional tests to follow in a separate PR.

Signed-off-by: Andon Andonov <andonova@vmware.com>

PR #691, opened 2022-02-01
[vdk-core] Add ingestion functional tests

doks5:
As part of https://github.com/vmware/versatile-data-kit/pull/690, the
IngesterRouter and IngesterBase classes were extended to support ingestion
plugin chaining.

Although unit tests were added as part of the aforementioned PR, no functional
tests were implemented.

This change adds functional tests, that verify the new end-to-end ingestion flow
with pre- and post- processors.

Testing Done: functional tests introduced.


projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -9,68 +9,63 @@
...
-#    job code is D
-# 4) INGEST_PAYLOAD_PREPROCESS_SEQUENCE=""
-# 5) INGEST_PAYLOAD_PREPROCESS_SEQUENCE=A,B - A,B are used for pre-processing,
-#    B is used for ingesting and post-processing
-
 
 log = logging.getLogger(__name__)
 
 
+# INGEST_PAYLOAD_PREPROCESS_SEQUENCE=A,B and VDK_INGEST_METHOD_DEFAULT=C
```
ivakoleva, line 24: I think test methods need to be sorted by complexity, starting with ingestion scenarios only (no processing), then pre-processing, then post-processing, then entire end-to-end verification

projects/vdk-core/tests/functional/ingestion/jobs/ingest-method-job/1_ingest_step.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.job_input import IJobInput
+
+
+def run(job_input: IJobInput):
```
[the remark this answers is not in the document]
doks5: Indeed, there is no point in the ingest-method-job. Removed, thanks

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
doks5: Tests sorted. Thanks

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -9,95 +9,227 @@
...
+    assert ingest_plugin.payloads[1].payload[0] == expected_rows_object
+    assert ingest_plugin.payloads[1].destination_table == "tabular_table"
+
+
+# INGEST_PAYLOAD_PREPROCESS_SEQUENCE=A,B no ingest method specified
+@mock.patch.dict(
+    os.environ,
+    {"VDK_INGEST_PAYLOAD_PREPROCESS_SEQUENCE": "convert-to-string," "add-payload-size"},
+)
+def test_chained_ingest_no_method_passed():
```
tozka, line 96: In that case would the preprocess plugins be invoked ? 

I think it's fine either way as long as it's documented and we verification in the test here whichever it is .

tozka (approved):
Looks good to me. I made a very small comment about one test. But ship it after you address it.

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -9,95 +9,227 @@
...
+
+# INGEST_PAYLOAD_PREPROCESS_SEQUENCE=A,B - A,B are used for pre-processing,
+#    B is used for ingesting and post-processing
+@mock.patch.dict(
+    os.environ,
+    {
+        "VDK_INGEST_PAYLOAD_PREPROCESS_SEQUENCE": "convert-to-string,"
+        "add-payload-size",
+        "VDK_INGEST_METHOD_DEFAULT": "add-payload-size",
+        "VDK_INGEST_PAYLOAD_POSTPROCESS_SEQUENCE": "add-payload-size",
```
ivakoleva, line 206: would removing the post-processing sequence change the test result?

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
```diff
@@ -9,95 +9,227 @@
...
+
+    # Use a sample data job, in which `method` argument is not passed to
+    # ingestion method calls.
+    result: Result = runner.invoke(
+        ["run", jobs_path_from_caller_directory("no-ingest-method-job")]
+    )
+
+    cli_assert_equal(0, result)
+
+    expected_object = dict(
```
ivakoleva, line 222: It seems a lot of duplicated 
```
expected_object = dict(
        int_key=str(42),
        str_key="example_str",
        bool_key=str(True),
        float_key=str(1.23),
        nested=str(dict(key="value")),
        payload_size=163,
    )
```
and 
```
expected_rows_object = {"first": "two", "payload_size": 80, "second": "2"}
```
used by different tests; maybe define in one place then reuse?

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
doks5: No, pre-processors will not be invoked. I've added a note about this behaviour. Thanks

projects/vdk-core/tests/functional/ingestion/test_ingest_chain.py
doks5: Not at this point in time. I've added a TODO to modify the behaviour of the test plugin used for post-processing to expose the result of post-processing, in order for us to verify operations. For the time being, the tests in test_ingester_base handle such checks.

ivakoleva (approved):

PR #694, opened 2022-02-03
control-service: vdk sdk docker repository secret

mivanov1988:
Currently, VDK helm chart allows different repositories for data job images
and VDK SDK images to be set. However, it doesn't allow different credentials to be specified. 
This means that the Data Job Pod can authenticate against only one repository
and as a result, it can pull only one image.

This change provides an ability to specify credentials for VDK SDK repository.

Testing Done: unit tests, integration tests, and helm template


projects/control-service/projects/helm_charts/pipelines-control-service/templates/secret_pull_vdk_sdk_img.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if (include "shouldCreateVdkSdkDockerRepoSecret" .) }}
```
tpalashki, line 1: Would be nice to keep the two secrets consistent - i.e. introduce a template helper similar to `shouldCreateVdkSdkDockerRepoSecret`, etc.

tpalashki (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/secret_pull_vdk_sdk_img.yaml
mivanov1988: Agreed, but I prefer not to refactor the old secret.

projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl
```diff
@@ -193,9 +197,33 @@ JDBC secret name
...
+{{- define "buildImagePullSecretJson" }}
+    {{- $dockerRepository := index . 0 }}
+    {{- $dockerRegistryUsername := index . 1 }}
+    {{- $dockerRegistryPassword := index . 2 }}
+    {{- printf "{\"auths\": {\"%s\": {\"auth\": \"%s\"}}}" $dockerRepository (printf "%s:%s" $dockerRegistryUsername $dockerRegistryPassword | b64enc) | b64enc }}
+{{- end }}
+
 {{- define "imagePullSecretJson" }}
-{{- printf "{\"auths\": {\"%s\": {\"auth\": \"%s\"}}}" .Values.deploymentDockerRepository (printf "%s:%s" .Values.deploymentDockerRegistryUsernameReadOnly .Values.deploymentDockerRegistryPasswordReadOnly | b64enc) | b64enc }}
+    {{ include "buildImagePullSecretJson" (list .Values.deploymentDockerRepository .Values.deploymentDockerRegistryUsernameReadOnly .Values.deploymentDockerRegistryPasswordReadOnly) }}
```
murphp15, line 224: @mivanov1988  
I am very confused here. 
You seem to be passing in different values for the first parameter of the function. 
For the first secret(line 224) I see that you are passing in just the registry name. 
And for the second(line 227) you are passing in the registry name followed by the repository name. 


For me only the first solution works with the jfrog docker repo. 
Which repo requires it to have the second format?

PR #695, opened 2022-02-07
vdk-impala: template .sql missing from python distro

ivakoleva:
When running templates, no data is being handled. The cause is
all template .sql steps are missing when python distribution is
bundled.

Added MANIFEST.in file including *.sql files.

Testing Done: did run python sdist, then verified the .sql files
are bundled-in.


doks5 (approved):
LGTM

PR #697, opened 2022-02-08
pipelines-control-service: redirect Swagger webjars resources

ivakoleva:
Swagger UI HTML page is not completely rendered. Cause is,
the resources included in swagger-ui.html that are /webjars/**
were not redirected to /data-jobs.

Path-prefixed the webjars for auth purposes. Redirected
webjars to the new path.

Testing Done: did add an integration test to verify availability of
a webjar expected.


doks5 (approved):
LGTM

PR #701, opened 2022-02-10
vdk-ingest-http: additional request parameters support

ivakoleva:
vdk-ingest-http: additional request parameters support

Plugin was missing connect & read timeout configurations, 
along with cert file path, and configurable TLS certificate
verification that was hardcoded to False (tagged with nosec).
It was also missing retries configuration on total retries count, 
backoff factor, and status forcelist.

Added configurations for INGEST_OVER_HTTP_CONNECT_TIMEOUT_SECONDS,
INGEST_OVER_HTTP_READ_TIMEOUT_SECONDS, INGEST_OVER_HTTP_CERT_FILE_PATH,
and INGEST_OVER_HTTP_VERIFY that is enabled by default. 
Added also INGEST_OVER_HTTP_RETRY_TOTAL, 
INGEST_OVER_HTTP_RETRY_BACKOFF_FACTOR and 
INGEST_OVER_HTTP_RETRY_STATUS_FORCELIST configurations.

Testing Done: added test_ingest_over_http_request_parameters_propagation
unit test.


projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
```diff
@@ -19,6 +19,28 @@ def vdk_configure(config_builder: ConfigurationBuilder) -> None:
...
+        key="INGEST_OVER_HTTP_CONNECT_TIMEOUT_SECONDS",
+        default_value=None,
+        description="How many seconds to wait for connecting to the server",
+    )
+    config_builder.add(
+        key="INGEST_OVER_HTTP_READ_TIMEOUT_SECONDS",
+        default_value=None,
+        description="How many seconds to wait for the server to read data",
+    )
+    config_builder.add(
```
doks5, line 32: Should we be having a configuration variable that accepts two completely different data types (boolean -- string file path)?

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
```diff
@@ -102,7 +136,9 @@ def __send_data(self, data, http_url, headers) -> IngestionResult:
                 url=http_url,
                 json=data,
                 headers=headers,
-                verify=False,  # nosec # TODO: disabled temporarily for easier testing, it must be configurable
+                timeout=(self._connect_timeout_seconds, self._read_timeout_seconds),
+                cert=self._cert_file_path,  # certifi.where(),
+                verify=self._verify,
```
doks5, line 141: We don't seem to be guarding against `_verify` being a random string. Is this an acceptable risk?

doks5 (approved):
Overall, looks ok to me.

doks5 (approved):

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
ivakoleva: take a look at requests sessions.py verify param pydoc:
```
param verify: (optional) Either a boolean, in which case it controls whether we verify
            the server's TLS certificate, or a string, in which case it must be a path
            to a CA bundle to use. Defaults to ``True``. When set to
            ``False``, requests will accept any TLS certificate presented by
            the server, and will ignore hostname mismatches and/or expired
            certificates, which will make your application vulnerable to
            man-in-the-middle (MitM) attacks. Setting verify to ``False`` 
            may be useful during local development or testing.
```
both boolean or string values are supported

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
ivakoleva: seems requests is designed that way

doks5: Ok.

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
```diff
@@ -29,6 +51,24 @@ def vdk_configure(config_builder: ConfigurationBuilder) -> None:
         default_value="utf-8",
         description="Encoding used if compressing the payload",
     )
+    config_builder.add(
+        key="INGEST_OVER_HTTP_RETRY_TOTAL",
+        default_value=10,
```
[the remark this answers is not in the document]
doks5: Ok.

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
doks5: My concern was that sessions.request does not really check the value of `verify` if it is a string, https://github.com/psf/requests/blob/main/requests/sessions.py#L707. It just accepts it as-is, which means that a random sting passed as a certificate would cause an exception to be thrown down the chain.
This is a minor thing, though, so you can ignore it.

PR #706, opened 2022-02-15
vdk-control-cli: Expand missing resource error message

### LINKED ISSUE #605: Unclear error message when job name or team are not passed correctly.
**Describe the bug**  
When creating my job I used my own team name that was different from the original job’s team name. Then. When I was deploying the job I got an error and it took me a while to figure out what was going on. Vdk was insisting that I did a spelling mistake.

**Steps To Reproduce**
1. create new job vdk create -n foo -t my-team
2. configure different team name in config.ini
3. vdk deploy -p job-dir 

We get very confusing error

**Expected behavior**
Be more explicit with user that team name is wrong

**Screenshots**

image

**Version  (please complete the following information):**
 The latest as of 06. December 2021

**Additional context**

Likely need more granular error handling on "X is NotFound" error directly in teh control service. The user get the message from the control service directly usually so we need to be a bit more explicit.

gageorgiev:
In the case when a user creates a job in the control service with
a given team name, then changes the team name in the job's config
file, the message printed is not clear enough.
The message was expanded to inform the user of the aforementioned
possible issue. This approach was chosen over checking for this
issue specifically, as that approach would have required changes
to both vdk-control-cli and the control-service and would have
required much longer development time.

Closes #605

Testing done: following reproduction steps found in #605 and
verified message is printed


tozka:
That's good. change, as it makes the generic not found message a bit more helpful. 

but in the ticket https://github.com/vmware/versatile-data-kit/issues/605 I have a more bold goal. I'd like that the control service returns exactly what is not found (not generic 404) but it would say "No job with that name", "no such team exists" . So please keep the ticket open.

tozka (approved):

PR #709, opened 2022-02-16
vdk-control-cli: Make list command print all jobs on empty team param 

gageorgiev:
The `list` command will now work when no team has been supplied as
a parameter, in which case it will query for all teams.
Some other small improvements have been made as well.

Testing done: added unittest

### DIFF
--- a/projects/vdk-plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
+++ b/projects/vdk-plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
@@ -26,7 +26,7 @@ from vdk.internal.core.context import CoreContext
 from vdk.plugin.control_cli_plugin import control_service_configuration
 
 
-@hookimpl
+@hookimpl(tryfirst=True)
 def vdk_command_line(root_command: click.Group):
     root_command.add_command(login)
     root_command.add_command(logout)
@@ -50,7 +50,6 @@ def vdk_command_line(root_command: click.Group):
 
 @hookimpl(tryfirst=True)
 def vdk_configure(config_builder: ConfigurationBuilder) -> None:
-    """"""
     control_service_configuration.add_definitions(config_builder)
 
 
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/list.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/list.py
@@ -207,8 +207,6 @@ vdk list -t taurus -o json
     type=click.STRING,
     default=load_default_team_name(),
     cls=NotRequiredIfAllOperationExists,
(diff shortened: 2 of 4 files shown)


projects/vdk-plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
```diff
@@ -26,7 +26,7 @@
 from vdk.plugin.control_cli_plugin import control_service_configuration
 
 
-@hookimpl
+@hookimpl(tryfirst=True)
```
tozka, line 29: can you add a comment as to why it's set as 'tryfirst' . 

I advise on building a habit to add comments if you think the reason for the code is non-obvious.

tozka:
> Some other small improvements have been made as well.


Be specific - what other small improvements?

projects/vdk-plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
gageorgiev: noted

tozka (approved):

projects/vdk-plugins/vdk-plugin-control-cli/src/vdk/plugin/control_cli_plugin/vdk_plugin_control_cli.py
```diff
@@ -26,7 +26,9 @@
 from vdk.plugin.control_cli_plugin import control_service_configuration
 
 
-@hookimpl
+@hookimpl(
+    tryfirst=True
+)  # hook will be tried first so any hooks which decorate these commands will be able to function
```
tozka, line 31: I don't understand the reason. If it's last , what would be the problem?

gageorgiev: The commands won't be registered under the click Group, so if anyone wants to alter those commands, they risk not having access to them. To give a more concrete example, imagine I'm trying to remove the prompt for a particular command parameter (hint hint). If this hook is executed before the one registering the command, then I won't be able to remove the prompt since I don't have access to it.

tozka: Ok. Understood. My brain froze for a moment there. Maybe said that (but shorter) e.g "we add the commands first here so that they can be decorated later"

gageorgiev: noted, amended comment

PR #710, opened 2022-02-16
control-service: run data job as non-root user

tozka:
Currently data job docker container is run as root user . This is not
considered best practice
(https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user)

And in environment where root is forbidden it won't work. For example
the job would fail if data job kubernetes pod is set to run as specific
user e.g using:

```
securityContext:
  fsGroup: 1000
  runAsGroup: 1000
  runAsUser: 1000
```

We are making sure that when building the data job we are  setting up
permissions and users and starting the docker container with new user.

The UID and GID can be passed as argument during docker build but that
is not currently exposed to operators when deploying jobs. So the
UID/GID is currently required to be 1000

Testing Done: deployed locally job with runAsUser securityContext (as
above) and the job succeeded. The integration test would verify end to
end as well.


doks5 (approved):

projects/control-service/projects/job-builder-rootless/Dockerfile.python.vdk
```diff
@@ -4,15 +4,21 @@ ARG base_image=python:3.9-slim
...
+ARG GID=1000
+
 # Set the working directory
 WORKDIR /job
 
-# Make sure base image is python based
+# Validate base image is python based
 RUN python -V
+# Create necessary users and set home directory to /job
+RUN groupadd -r -g $GID vdkgroup && useradd -u $UID -g $GID -r vdkuser && chown -R $UID:$GID /job
```
mivanov1988, line 16: I wonder if it is better to add the user and the group to the base image?

tozka: You mean just this line?

 I was also wondering. Why are you wondering? What are your arguments? I will keep mine so I do not spoil you :)

mivanov1988: I guess you want to have better control over the user and the group by the ability to overwrite them through the builder. On the other hand, I don't think we will change them at all.

PR #717, opened 2022-02-18
[plugins][vdk-impala] Improve error handling to handle view errors

doks5:
Currently, the impala error handling logic fails to properly handle
TableAlreadyExist exceptions, in cases where these are caused while processing
`CREATE VIEW` sql queries.

This change modifies the error handling logic to properly handle exceptions
caused while processing `CREATE VIEW` queries.

Testing Done: Added unit test.


tozka (approved):
I should note that I am not big fan of regexes on strings that can have so many different combinations. The regex doesn't handle comments or if "create" is in a  literal string. 

But handling this is outside of the goal of this PR.

doks5:
> I should note that I am not big fan of regexes on strings that can have so many different combinations. The regex doesn't handle comments or if "create" is in a literal string.
> 
> But handling this is outside of the goal of this PR.

Agree, I'll think about some improvements in the future.
Thank you

PR #718, opened 2022-02-18
vdk-heartbeat: Successful data job run test mode

ivakoleva:
The util was missing a simple test, that verifies the (custom or not)
data job has run once and succeeded.

Introduced `successful_run_test.py`. Elaborated on `job_controller.py`
to return the data job status on `check_job_execution_finished`.
Fixed a controller bug - in case no executions were present, the flag
was still set to `job_execution_running = False`.

Testing Done: did verify locally the filtering of latest_end_date
and retrieving the status based on execution list data.

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
@@ -8,6 +8,8 @@ import shutil
 import subprocess
 import tempfile
 import time
+from datetime import datetime
+from typing import Optional
 
 from vdk.internal.heartbeat.config import Config
 from vdk.internal.heartbeat.tracing import LogDecorator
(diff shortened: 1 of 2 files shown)


projects/vdk-heartbeat/src/vdk/internal/heartbeat/successful_run_test.py
```diff
@@ -0,0 +1,49 @@
...
+    @LogDecorator(log)
+    def clean_up(self):
+        pass
+
+    @LogDecorator(log)
+    def execute_test(self):
+        status = None
+        try:
+            status = self.__job_controller.check_job_execution_finished()
+        except Exception as e:
```
doks5, line 36: Exception object seems not used. Maybe log the exception before raising the AssertionError?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -359,20 +362,31 @@ def check_job_execution_finished(self):
...
+                        break
             time.sleep(10)
         log.info(f"Data job is running : {job_execution_running}")
         assert not job_execution_running, (
             f"VDK-heartbeat timed out. Job execution {self.config.job_name} "
             + f"did not finish in {self.config.RUN_TEST_TIMEOUT_SECONDS}."
         )
+        if not execution_list:
+            return None
+        latest_end_date = max(
```
doks5, line 382: nit: Maybe add a blank line, as now it feels a bit crowded.

doks5 (approved):

PR #719, opened 2022-02-18
vdk-control-cli: Parse contacts with both comma "," as delimiter as well

tozka:
Until now we required contacts be separated by ";" semi-colon. This was
error prone as a lot of users are using "," comma and we are not really
preventing them so mails were written as "a@abv.bg,b@dir.bg" as if it is
a single email. T This also meant that people were not getting
notifications.

The fix is to add "," comma as allowed delimiter as a first step.
In a subsequent change we can add a email validator (like
https://pypi.org/project/email-validator)

Testing Done: unit tests including new ones

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
@@ -5,11 +5,13 @@ import configparser
 import fileinput
 import logging
 import os
+import re
 import sys
+from typing import Any
+from typing import List
 
 from vdk.internal.control.exception.vdk_exception import VDKException
 from vdk.internal.control.utils.control_utils import read_config_ini_file
-
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


projects/vdk-control-cli/src/vdk/internal/control/job/job_config.py
```diff
@@ -106,8 +107,9 @@ def _set_value(self, section, key, value):
...
         return success
 
-    def _get_contacts(self, key):
+    def _get_contacts(self, key) -> List[str]:
         contacts_str = self._get_value("contacts", key).strip()
+        contacts = []
         if contacts_str:
-            return [x.strip() for x in contacts_str.split(";")]
-        return []
+            contacts = [x.strip() for x in contacts_str.split(";,")]
```
gageorgiev, line 114: I don't believe this works (at it seems your test agrees with me). `.split` will want to split on the whole string you pass as a parameter instead of the individual characters in it. Try `re.split` as described here - https://www.geeksforgeeks.org/python-split-multiple-characters-from-string/

gageorgiev (changes requested):

gageorgiev (approved):

PR #721, opened 2022-02-21
[vdk-kerberos-auth] Fix keytab file in job directory

doks5:
Currently, if a `VDK_KEYTAB_FOLDER` is not specified, the kerberos plugin
looks for the keytab file in the data job directory, which causes operation
conflict between the `vdk run` and `vdk deploy` commands.

`vdk run` looks for the keytab file in the job's directory and throws an exception
if the file is not there, while
`vdk deploy` checks that there IS NO keytab file in the job's directory, and throws
an exception if there is such a file in the directory.

This change ensures that if `VDK_KEYTAB_FOLDER` is not set, the default directory to
look for a keytab file would be the data job's parent directory.

Testing Done: moved the keytab file used in functional tests outside the test job directory.


gageorgiev (approved):
You may want to review any jobs which use keytab files and make sure that their keytab is located outside the job directory. This includes non-github repos as well (I have some memory where I placed a test job's keytab inside the job dir, this is why I'm mentioning this).

ivakoleva (approved):

PR #723, opened 2022-02-21
job-builder: add amazon-ecr-credential-helper to the job builder

tpalashki:
The rootless job builder is based on an image that does not contain
the amazon-ecr-credential-helper utility. This utility is necessary
to authenticate against ECR when pushing data job images.

This commit adds the amazon-ecr-credential-helper utility to the job-builder
image by pulling it and building it locally from source. Maybe there is
a better way of provisioning it but I could not find it out.

Testing done: manually build and push a job-builder image with the change,
then create a k8s job with this image and run it to verify that it manages
to successfully push data job images to ECR.


mivanov1988:
LGTM

mivanov1988:
Please increment the builder version.

mivanov1988 (approved):

PR #725, opened 2022-02-21
vdk-ingest-http: configurable allow JSON float NaN capability

ivakoleva:
Based on requests encoder/decoder default configuration, simplejson/json
is used to handle a specification compliant JSON by default. We need a
configurable allow NaN capability, so that when enabled, ingestion
payload NaN values are parsed correctly - instead raising ValueError
interrupting the ingestion.

Introduced `INGEST_OVER_HTTP_ALLOW_NAN` plugin configuration, that is
set to False by default.

Testing Done: did add
`test_ingest_over_http_json_dumps_parameters_propagation` that verifies
`INGEST_OVER_HTTP_ALLOW_NAN` configuration propagation to `json.dumps`

### DIFF
--- a/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
+++ b/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
@@ -72,9 +72,9 @@ def vdk_configure(config_builder: ConfigurationBuilder) -> None:
     config_builder.add(
         key="INGEST_OVER_HTTP_ALLOW_NAN",
         default_value=False,
-        description="If set to True, then NaN, Infinity, and -Infinity will be encoded as such."
-        "This behavior is not JSON specification compliant, but is consistent with most JavaScript based "
-        "encoders and decoders. Otherwise, it will be a ValueError to encode such floats.",
+        description="If set to False, then it will be a ``ValueError`` to serialize out of range ``float`` "
+        "values (``nan``, ``inf``, ``-inf``) in strict compliance of the JSON specification, "
+        "instead of using the JavaScript equivalents (``NaN``, ``Infinity``, ``-Infinity``).",
     )
 
 
--- a/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
+++ b/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
@@ -69,6 +69,13 @@ def vdk_configure(config_builder: ConfigurationBuilder) -> None:
         default_value=None,
         description="A string of comma-separated HTTP status codes that we should force a retry on.",
     )
+    config_builder.add(
+        key="INGEST_OVER_HTTP_ALLOW_NAN",
+        default_value=False,
+        description="If set to True, then NaN, Infinity, and -Infinity will be encoded as such."
+        "This behavior is not JSON specification compliant, but is consistent with most JavaScript based "
+        "encoders and decoders. Otherwise, it will be a ValueError to encode such floats.",
+    )
 
 
 @hookimpl
--- a/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
+++ b/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
@@ -102,6 +102,9 @@ class IngestOverHttp(IIngesterPlugin):
             )
             else None
         )
+        self._allow_nan = context.core_context.configuration.get_value(
+            "INGEST_OVER_HTTP_ALLOW_NAN"
+        )
 
         adapter = HTTPAdapter(
             max_retries=Retry(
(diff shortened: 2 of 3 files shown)


projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
```diff
@@ -69,6 +69,13 @@ def vdk_configure(config_builder: ConfigurationBuilder) -> None:
         default_value=None,
         description="A string of comma-separated HTTP status codes that we should force a retry on.",
     )
+    config_builder.add(
+        key="INGEST_OVER_HTTP_ALLOW_NAN",
+        default_value=False,
+        description="If set to True, then NaN, Infinity, and -Infinity will be encoded as such."
+        "This behavior is not JSON specification compliant, but is consistent with most JavaScript based "
+        "encoders and decoders. Otherwise, it will be a ValueError to encode such floats.",
```
doks5, line 77: What does the last sentence mean? Maybe it could be paraphrased?

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
```diff
@@ -171,18 +174,20 @@ def __amend_payload(payload, destination_table):
     def __send_data(self, data, http_url, headers) -> IngestionResult:
         uncompressed_size_in_bytes = sys.getsizeof(data)
         compressed_size_in_bytes = None
+        data = json.dumps(data, allow_nan=self._allow_nan)
```
doks5, line 177: Wouldn't `_allow_nan` be loaded as a string, and not as a boolean?

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
```diff
@@ -226,3 +229,31 @@ def arg(a):
...
+
+    def configuration_side_effect(*args, **kwargs):
+        def arg(a):
+            return args[0] == a
+
+        if arg("INGEST_OVER_HTTP_ALLOW_NAN"):
+            return allow_nan
+        return None
+
+    job_context = MagicMock()
```
doks5, line 246: `job_context` is already mocked in the global scope at the top of the file. Do you need to mock it again?

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
```diff
@@ -226,3 +229,31 @@ def arg(a):
...
+    job_context.core_context.configuration.get_value.side_effect = (
+        configuration_side_effect
+    )
+    try:
```
doks5, line 250: Is this try-except actually needed? If you expect an exception to be raised, wouldn't it be better to use `with pytest.raises()`?

doks5 (changes requested):

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
ivakoleva: By memory, would expect to be interpreted as boolean, since default value is boolean.
Also, did verify it locally, successfully (run a data job that enables allow nan).
And, we did elaborate on same thing with "INGEST_OVER_HTTP_VERIFY", that is boolean, and interpreted the same way currently

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
ivakoleva: yes, it seems to me, since I customise the configuration_side_effect in a test-specific way;
same stuff already merged with `test_ingest_over_http_result` for example

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
ivakoleva: well, with using pytest.raises() - I would've wanted to ensure the exception is raised (requirement);
since exception-or-not does not matter for the particular test, I decided to wrap it with try - so either way the test proceeds to assert_called_with, that is the actual test

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
doks5: Ok.

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
doks5: Then, do we need job_context mocked in the global scope?

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
doks5: Ok, I'll give it a pass, though I am a bit concerned with the use of try-except in tests, when we don't know if a method call would throw an exception, or not.

doks5 (changes requested):
Not all comments seem to have been addressed yet.

projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_http_plugin.py
ivakoleva: Oh, this is from simplejson encoder documentation on the parameter we pass, literally.
I just realised, maybe json.dumps() pydoc would be even better, updated the MR

projects/vdk-plugins/vdk-ingest-http/tests/test_ingest_over_http.py
ivakoleva: Yes, it is still used by the majority of other tests, that do not introduce test-specific patches

doks5 (approved):

PR #730, opened 2022-02-22
[draft] vdk: add VDK Notebook proposal

tozka:
The PR is proposing introducing a new plugin vdk-notebook that will
provide integration between Jupyter notebooks and VDK.
The current PR covers the summary and motivation of the proposal.


tozka:
This is will be re-created by @duyguHsnHsn later

PR #735, opened 2022-02-23
control-service: fix log link to set endTime always

tozka:
In case the job is not finished, there's no end time (logically) but
VMware Log Insight links do not work if no endTime is set. That might be
problem for other engines.

It's better to set it to current time instead of leaving empty.

I was wondering if I should set it to 1 hour in the future. But uses can
use endOffset option to change this. And there's chance that some Log
services may not work well with "future" times.

Testing Done: unit tests


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionLogsUrlBuilder.java
```diff
@@ -101,7 +104,8 @@ private String convertDate(OffsetDateTime dateTime, Long offset) {
       String format = dateFormat;
 
       if (dateTime == null) {
-         return "";
+         // if not set it to current time
+         dateTime = OffsetDateTime.now(clock);
```
mivanov1988, line 108: OffsetDateTime.now() uses the system default zone (Clock.systemDefaultZone()) by default.

tozka: Yes. But I want to be able to mock the Clock to a fixed clock in teh tests. That's why I am passing it explicitly.

mivanov1988 (approved):

PR #743, opened 2022-02-28
control-service: Adopt kubernetes-client 13.0

gageorgiev:
To take advantage of the V2CronJob API and enable automatic
job cleanup for the builder jobs and manually executed data jobs,
we need to adopt a later version of kubernetes-client than the one
currently used.
This change is mostly comprised of import path fixes, inclusion of
new optional method parameters (all have null value as we do not
use them), switching from the joda-time DateTime to the JDK
OffsetDataTime as the former is considered deprecated.

Further steps are replacing the currently used CronJob object
with the new V1CronJob, and including the TTL parameter for
builder jobs and manually executed data jobs.

Testing done: project builds successfully and unittests pass


tozka (approved):
Can you please write what are the next steps so it's clear. 

Otherwise looks good to me.

PR #745, opened 2022-03-01
vdk-core: JobInput get_name and get_job_directory implementation

ivakoleva:
The IJobInput get_name and get_job_directory methods were not
implemented. Yet, we need to utilize those capabilities.

Extended JobInput instantiation performed by job_context.py.
Since modifying the JobInput constructor parameters, did unify with
JobContext constructor params convention (since params matching).

Testing Done: added unit tests for name and job directory propagation;
added functional test for verifying the name and directory retrieved;
adjusted test_substitute_query_params.py to extended JobInput params.


tozka (approved):
Fixed the codacy warning. But otherwise looks good to me.

PR #746, opened 2022-03-01
vdk-snowflake: Enable support for Python 3.10

gageorgiev:
Support for Python 3.10 was disabled due to the fact that the
python-snowflake-connector had not been updated to support it.
It has been now so we can enable support for vdk-snowflake as well.


tozka (approved):
Cool. And thanks for keeping track and following up on this.

PR #747, opened 2022-03-01
vdk-core: refactor test to use job_path method

tozka:
job_path is shorter version to get the jobs path in test in vdk-core

### DIFF
--- a/projects/vdk-core/tests/functional/run/test_run_sql_queries.py
+++ b/projects/vdk-core/tests/functional/run/test_run_sql_queries.py
@@ -2,19 +2,17 @@
 # SPDX-License-Identifier: Apache-2.0
 import logging
 import os
-import pathlib
 import sqlite3
 from unittest import mock
 
 from click.testing import Result
-from functional.run import util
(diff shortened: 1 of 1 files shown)


ivakoleva (commented):
I can see other `def job_path` implementations, that would evaluate `get_test_job_path`, only substituting the `__file__` where the function is defined.

I would suggest, instead of defining multiple `job_name` functions per each package (for example functional.run in this particular case), to introduce a common `job_name` method, that considers the location relatively to the _actual caller file location_ - in favour of `__file__` of definition.

tozka:
> I can see other `def job_path` implementations, that would evaluate `get_test_job_path`, only substituting the `__file__` where the function is defined.
> 
> I would suggest, instead of defining multiple `job_path` functions per each package (for example functional.run in this particular case), to introduce a common `job_path` method, that considers the location relatively to the _actual caller file location_ - in favour of `__file__` of definition.

I will consider re-implementing util.job_path to be more generic in separate PR. But let's merge this first.

ivakoleva (approved):

PR #749, opened 2022-03-01
vdk-core: ensure sql args are subsituted in correct priority

tozka:
The documented order is that get_property API is lower priority over
get_arguments API. But it was not working, properties were taking
precedence instead. This was not well explained in execute_query API as
well.

This was fixed in this change and new tests are added.

Testing Done: extended functional tests.


ivakoleva (approved):
except tests run (probably double curly brackets), lgtm

PR #750, opened 2022-03-01
vdk-heartbeat: Handle execution end_time not string

doks5:
Currently, if for some reason the `end_time` field of the execution object
is not a string (for example, the value is `None`), the `datetime.fromisoformat()`
method throws an exception and fails the heartbeat test.

This change introduces an explicit conversion of the value to string and adds
logging for what the error is along with indication that the heartbeat will fail.

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
@@ -379,13 +379,14 @@ class JobController:
         )
         if not execution_list:
             return None
+
         latest_end_date = max(
-            datetime.fromisoformat(e["end_time"]) for e in execution_list
+            self._datetime_from_iso_format(str(e["end_time"])) for e in execution_list
         )
         return [
             e["status"]
             for e in execution_list
-            if datetime.fromisoformat(e["end_time"]) == latest_end_date
+            if self._datetime_from_iso_format(str(e["end_time"])) == latest_end_date
         ].pop()
 
     def _update_config_ini(self, heartbeat_job_dir):
@@ -458,3 +459,16 @@ def run(job_input):
             os.path.join(heartbeat_job_dir, "06_override_properties.py"), "w"
         ) as pyfile:
             pyfile.write(python_script)
+
+    @staticmethod
+    def _datetime_from_iso_format(datetime_string: str):
+        try:
+            return datetime.fromisoformat(datetime_string)
+        except ValueError as e:
+            log.info(
+                "An exception occurred while converting datetime string "
+                f"value of -- {datetime_string} -- to "
+                f"a datetime object. The exception was {e}"
+            )
+
+        return None
--- a/projects/vdk-heartbeat/tests/vdk/internal/test_job_controller.py
+++ b/projects/vdk-heartbeat/tests/vdk/internal/test_job_controller.py
@@ -1,0 +1,43 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from unittest.mock import patch
+
+from vdk.internal.heartbeat.config import Config
+from vdk.internal.heartbeat.job_controller import JobController
+
+
+@patch.object(JobController, "_execute")
+def test_check_job_execution_finished(patched_method):
+    test_config = Config()
+    test_config.control_api_url = None
+    test_config.vdkcli_oauth2_uri = None
+
+    patched_method.return_value = (
+        '[{"status": "finished", "end_time": ' '"2022-03-01"}]'
+    )
+
+    test_controller = JobController(test_config)
+    test_controller.config.RUN_TEST_TIMEOUT_SECONDS = 10
+
+    res = test_controller.check_job_execution_finished()
+    assert res == "finished"
+
+
+@patch.object(JobController, "_execute")
+def test_check_job_execution_finished_no_exceptions_for_value_errors(patched_method):
+    test_config = Config()
+    test_config.control_api_url = None
+    test_config.vdkcli_oauth2_uri = None
+
+    patched_method.return_value = '[{"status": "finished", "end_time": "None"}]'
+
+    test_controller = JobController(test_config)
+    test_controller.config.RUN_TEST_TIMEOUT_SECONDS = 10
+
+    res = test_controller.check_job_execution_finished()
+    assert res == "finished"
+
+    patched_method.return_value = '[{"status": "finished", "end_time": ' "12345}]"
+
+    res = test_controller.check_job_execution_finished()
+    assert res == "finished"


projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, config: Config):
             f"with job {config.job_name} and team {config.job_team}. "
             f"Authorization endpoint: {auth_endpoint_message}"
         )
+        self.test_dict = {}
```
tozka, line 41: What is the idea behind this variable ? Or is it forgotten debug var ...

tozka:
> Currently, if for some reason the end_time field of the execution object is not a string
 
Can you describe in a bit more details with example? Why would not be a string (I guess it can be None).

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
doks5: Good catch. It was indeed a test var. Removed, thanks

doks5:
> > Currently, if for some reason the end_time field of the execution object is not a string
> 
> Can you describe in a bit more details with example? Why would not be a string (I guess it can be None).

I've added a clarification.

projects/vdk-heartbeat/tests/vdk/internal/test_job_controller.py
```diff
@@ -0,0 +1,44 @@
...
+    assert res == "finished"
+
+
+@patch.object(JobController, "_execute")
+def test_check_job_execution_finished_exceptions(patched_method):
+    test_config = Config()
+    test_config.control_api_url = None
+    test_config.vdkcli_oauth2_uri = None
+
+    patched_method.return_value = '[{"status": "finished", "end_time": "None"}]'
```
tozka, line 33: Shouldn't it return None without "" as https://www.pythontutorial.net/advanced-python/python-none 

Can the API return anything else but valid string timestamp or None (e.i. no such value)

doks5: I believe the result of the request is always in the form of a string, even with 'None', as otherwise, it would raise a JsonDecoderError while converting the string to dictionary on line 363 https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py#L363

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -379,14 +379,22 @@ def check_job_execution_finished(self) -> Optional[str]:
...
+        try:
+            latest_end_date = max(
+                datetime.fromisoformat(str(e["end_time"])) for e in execution_list
+            )
+            return [
+                e["status"]
+                for e in execution_list
+                if datetime.fromisoformat(str(e["end_time"])) == latest_end_date
+            ].pop()
+        except Exception as e:
```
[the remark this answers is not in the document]
doks5: What makes you believe the exception is not propagated?
The only errors that are swallowed, are ValueErrors, as they tend to cause false negatives (the heartbeat test fails, even if all the steps are successful, just because the value of `end_time` is not yet available at the time of the check)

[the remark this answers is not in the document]
tozka: Ok. I guess what I mean is do we need the `except Exception as e:` and the log.info ..? The exception would be propagated to the caller and logs eventually at appropriate place ?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -379,14 +379,27 @@ def check_job_execution_finished(self) -> Optional[str]:
...
+                datetime.fromisoformat(str(e["end_time"])) for e in execution_list
+            )
+            return [
+                e["status"]
+                for e in execution_list
+                if datetime.fromisoformat(str(e["end_time"])) == latest_end_date
+            ].pop()
+        except ValueError as e:
+            log.info(
+                "Something happened while checking the end_time of the "
```
tozka, line 394: 'Something happened' is very vague. Let's say what happened. 

Also ValueError could be thrown from other places potentially . if you want to make sure datetime is parsed without exception have a function `datatime_from_iso_no_except` 

Generally a good principle for exceptions is to catch an exception as close as possible to its source

projects/vdk-heartbeat/tests/vdk/internal/test_job_controller.py
tozka: That does not make sense to me. 

 The API must return either 
```
{"status": "finished"}
```

where the "end_time" key is not present 

```
{"status": "finished", "end_time": "1988-11-12..."}
```
where end_time is valid timestamp . 

Are you saying the API (or vdk-control-cli) returns something else . Anything else seems like a bug either in vdk-control-cli or in the API...

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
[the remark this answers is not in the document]
doks5: Fair point. Removed the general exception handling and refactored the code, so that the exception handling deals only with issues with datetime format. Thanks for the review

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
doks5: Done, thanks

projects/vdk-heartbeat/tests/vdk/internal/test_job_controller.py
doks5: Well, if the format was `{"status": "finished"}`, then it would not throw a ValueError, but rather a KeyError. It looks like sometimes the format is `{"status": "finished", "end_time": "<some-value>"}`, where `<some-value>` is not a date string.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -458,3 +457,15 @@ def run(job_input):
             os.path.join(heartbeat_job_dir, "06_override_properties.py"), "w"
         ) as pyfile:
             pyfile.write(python_script)
+
+    @staticmethod
+    def _datetime_from_iso_format(execution: dict):
```
tozka, line 462: From an interface (abstraction) point of view, it would be better the method to accept string and not dictionary (where special key need to be set) , what if we want to use it for start_time . 

Think separation of concerts (https://en.wikipedia.org/wiki/Separation_of_concerns) 
This method deals with parsing dates, while "end_time" is part of the business "layer" and its separate concern.

projects/vdk-heartbeat/tests/vdk/internal/test_job_controller.py
tozka: can you give an example?

tozka (approved):

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -458,3 +459,15 @@ def run(job_input):
...
         ) as pyfile:
             pyfile.write(python_script)
+
+    @staticmethod
+    def _datetime_from_iso_format(datetime_string: str):
+        try:
+            return datetime.fromisoformat(datetime_string)
+        except ValueError as e:
+            log.info(
+                "An exception occurred while converting end_time to "
```
tozka, line 469: let's log the value as well. datetime_string

doks5: Done, thanks

PR #751, opened 2022-03-02
vdk-core: lowercase env variables are inferred as configuration

tozka:
vdk config-help prints all configs in lower-case so it's easy to copy
paste it and set it and it's error prone to allow only upper-cased.
Spring does the same thing when inferring applicaiton properties from
environment variables.

Added also better testing around use-cases with configuration in the
functional test suite

Testing Done: the functional tests


YanaZhivkova (approved):
LGTM

PR #752, opened 2022-03-07
vdk-control-cli: set vdk version and enabled when deploying new job

tozka:
Currently we cannot create new deployment and at the same time enable it
or set other vdk version in a single operation. This means we have to do
it in two different cli calls which may create race condition (where job
deployment starts with previous configuration before new is applied)

In this change we enable passing --vdk-version and --enabled when
creating deployment
with vdk deploy --create e.g
```
vdk deploy --create --vdk-version 1.2 --disabled
```
would create a new deplloyment which uses vdk version 1.2 and is
disabled.

Testing Done: automated tests included

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
@@ -162,12 +162,12 @@ vdk deploy --update --job-version <job-version-here> -n example-job -t job-team
 def deploy(
     name: str,
     team: str,
-    job_version: str,
-    vdk_version: str,
+    job_version: Optional[str],
+    vdk_version: Optional[str],
     enabled: Optional[bool],
-    job_path: str,
+    job_path: Optional[str],
     operation: str,
-    reason: str,
+    reason: Optional[str],
     rest_api_url: str,
     output: str,
 ):
@@ -178,7 +178,13 @@ def deploy(
         # default to ask for job version to update
         if not vdk_version and not job_version and enabled is None:
             job_version = get_or_prompt("Job Version", job_version)
-        return cmd.update(name, team, enabled, job_version, vdk_version, output)
+        if job_path:
+            reason = get_or_prompt("Reason", reason)
+            return cmd.create(
+                name, team, job_path, reason, output, vdk_version, enabled
+            )
+        else:
+            return cmd.update(name, team, enabled, job_version, vdk_version, output)
     if operation == DeployOperation.REMOVE:
         name = get_or_prompt("Job Name", name)
         team = get_or_prompt("Job Team", team)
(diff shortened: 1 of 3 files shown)


gageorgiev (approved):
lgtm

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_deploy.py
```diff
@@ -22,98 +22,69 @@
 
 
 def test_deploy(httpserver: PluginHTTPServer, tmpdir: LocalPath):
-    rest_api_url = httpserver.url_for("")
-
-    mock_base_requests(httpserver)
+    job_version, deploy_args = prepare_new_deploy(httpserver)
```
gageorgiev, line 25:
```suggestion
    _, deploy_args = prepare_new_deploy(httpserver)
```

YanaZhivkova (approved):
LGTM

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli.py
```diff
@@ -178,7 +178,13 @@ def deploy(
         # default to ask for job version to update
         if not vdk_version and not job_version and enabled is None:
             job_version = get_or_prompt("Job Version", job_version)
-        return cmd.update(name, team, enabled, job_version, vdk_version, output)
+        if job_path:
+            reason = get_or_prompt("Reason", reason)
+            return cmd.create(
+                name, team, job_path, reason, output, vdk_version, enabled
```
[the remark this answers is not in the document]
tozka: Until now you can do `vdk deploy -p xxx` and you'd be promoted for reason (handled by `if operation == CREATE` below)

But you could not do `vdk deploy -p xxx --disable` . You could either do `vdk deploy --disable` or `vdk deploy -p xxx` 

And now I want both to be possible. I am trying to make the command as smart as possible so whatever the user passes it would work .

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_deploy.py
```diff
@@ -22,98 +22,69 @@
...
-    open(f"{test_job_path}.zip", "w")
+    posted_data = json.loads(httpserver.log[3][0].data)
+    assert posted_data["job_version"] is not None
 
-    assert os.path.exists(f"{test_job_path}.zip") is True
+
+def test_deploy_with_vdk_version_disable(
+    httpserver: PluginHTTPServer, tmpdir: LocalPath
+):
+    job_version, deploy_args = prepare_new_deploy(httpserver)
```
tozka, line 41:
```suggestion
    _, deploy_args = prepare_new_deploy(httpserver)
```

PR #753, opened 2022-03-07
vdk-trino: fix ingesting value with bool type failing

tozka:
vdk-trino ingestion was failing if `send_object_for_ingestion` is passed
value with type bool (as it was trying to cast it as if it's stirng.
Error was

"We could not convert the value to that type. Error is bool cast accept
only True/true/False/false values."

Testing Done: unit tests covered the new case

### DIFF
--- a/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
+++ b/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
@@ -90,7 +90,9 @@ class IngestToTrino(IIngesterPlugin):
             )
 
     @staticmethod
-    def __to_bool(value: str) -> bool:
+    def __to_bool(value: Any) -> bool:
+        if isinstance(value, bool):
+            return value
         if value == "true" or value == "True":
             return True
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
```diff
@@ -91,6 +91,8 @@ def _ingest_payload(
 
     @staticmethod
     def __to_bool(value: str) -> bool:
```
YanaZhivkova, line 93: I know that arguments type is not enforced in Python, but do we need the str type (does it help us somehow), since now we can also accept bool?

YanaZhivkova (approved):
LGTM

projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/ingest_to_trino.py
tozka: I will fix it. It's not enforeced by IDE shows warnings if there's type mismatch so it's important to be accurate.

PR #754, opened 2022-03-07
vdk-server: status returns if server is not running

tozka:
When running `vdk server --status` command it was returning that the
control service was installed but it was not running (the process was
down). So added a check to explicitly verify that.
Added a few more debug messages.

Testing Done: vdk server -s locally and verified message.


YanaZhivkova (approved):
LGTM

PR #755, opened 2022-03-07
control-service: add kerberos auth provider

mrMoZ1:
what: Added a new AuthProviderConfig which should work with kerberos.

why: users requested Kerberos authentication in parallel with OAuth2

todo: Expose configuration properties against env variables/helm charts
we need to decide which ones to expose and on default values.

testing: Locally ran control service, made sure enable/disable flag works.
Ran requests on a locally running instance. Verified that: both OAuth2 and
SPNEGO filters are used when making requests and that users can authenticate
with either OAuth2 or Kerberos credentials. Requests were made with an
authenticated vdkcli:
`vdk execute --list -t supercollider -n job -u http://localhost:8092`
and with a curl command for the kerberos authentication:
`curl -vvv --negotiate -u : "http://localhost:8092/data-jobs/for-team/supercollider/jobs/smoke-test/executions"
`


projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -143,3 +143,9 @@ mail.smtp.host=smtp.vmware.com
 # Correspond to those defined by kubernetes
 # See https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy
 datajobs.deployment.builder.imagePullPolicy=IfNotPresent
+
+#Settings for Kerberos authorization provider.
+datajobs.properties.kerberosPrincipal=""
```
mrMoZ1, line 148: We need to decide which properties to expose as an env variable/helm chart and property names.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/authorization/config/KerberosAuthProviderConfig.java
```diff
@@ -0,0 +1,106 @@
...
+
+   @Value("${datajobs.properties.keytabFileLocation}")
+   private String keytabFileLocation;
+
+
+   @Override
+   protected void configure(HttpSecurity http) throws Exception {
+      http
+            .authorizeRequests()
+            .antMatchers("/**").authenticated()
```
mrMoZ1, line 48: Are those the correct paths? I'm not sure.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/authorization/config/KerberosAuthProviderConfig.java
```diff
@@ -0,0 +1,106 @@
...
+import org.springframework.security.kerberos.authentication.sun.SunJaasKerberosTicketValidator;
+import org.springframework.security.kerberos.web.authentication.SpnegoAuthenticationProcessingFilter;
+import org.springframework.security.web.authentication.www.BasicAuthenticationFilter;
+
+@Configuration
+@EnableWebSecurity
+@EnableGlobalMethodSecurity(prePostEnabled = true)
+@Order(101)
+@ConditionalOnProperty(value = "datajobs.properties.enableKerberosAuthentication")
+public class KerberosAuthProviderConfig extends WebSecurityConfigurerAdapter {
```
tozka, line 35: I think it might be better if this is in base project 

https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/base/src/main/java/com/vmware/taurus/security 

for consistency.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/authorization/config/KerberosAuthProviderConfig.java
```diff
@@ -0,0 +1,106 @@
...
+
+   @Value("${datajobs.properties.kerberosPrincipal}")
```
tozka, line 37: They keys likely need to be something like datajobs.security.keberos.principal 
since they are not just for properties here.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/authorization/config/KerberosAuthProviderConfig.java
```diff
@@ -0,0 +1,106 @@
...
+   private String kerberosPrincipal;
+
+   @Value("${datajobs.properties.keytabFileLocation}")
```
tozka, line 40: `datajobs.security.kerberos.keytabFileLocation` ?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/authorization/config/KerberosConfig.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.kerberos.authentication.sun.GlobalSunJaasKerberosConfig;
+
+@Configuration
+@ConditionalOnProperty(value = "datajobs.properties.enableKerberosAuthentication")
+public class KerberosConfig {
+
+   @Value("${datajobs.properties.krb5ConfigLocation}")
```
tozka, line 18: `datajobs.security.kerberos.krb5ConfigLocation`

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
tozka: We'd need to extend [security section of the helm chart](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml#L306)

We can and should discuss this in separate PR.

tozka: My suggestion is to create PR with modified values.yaml with the new options described (and their description) only without any implementation (yet).

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/authorization/config/KerberosAuthProviderConfig.java
```diff
@@ -0,0 +1,106 @@
...
+import org.springframework.security.kerberos.authentication.KerberosServiceAuthenticationProvider;
+import org.springframework.security.kerberos.authentication.sun.SunJaasKerberosTicketValidator;
+import org.springframework.security.kerberos.web.authentication.SpnegoAuthenticationProcessingFilter;
+import org.springframework.security.web.authentication.www.BasicAuthenticationFilter;
+
+@Configuration
+@EnableWebSecurity
+@EnableGlobalMethodSecurity(prePostEnabled = true)
+@Order(101)
+@ConditionalOnProperty(value = "datajobs.properties.enableKerberosAuthentication")
```
tozka, line 34: `datajobs.security.kerberos.enabled` ?

tozka (changes requested):

tozka:
Can you provide a bit more details around testing done 
"When making a request it is evident that the spnego filter is used:" ... What was the request - did you use vdk cli (with kerberos / properties) or curl (curl --negotiate -u : ) ? 

How did you verify the authentication passes successfully ?

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -59,6 +69,28 @@
...
+          // but until we can expose them in swagger they are very hard to use with Auth.
+          "/data-jobs/debug/**"};
+
+    @Value("${datajobs.security.kerberos.kerberosPrincipal}")
+    private String kerberosPrincipal;
+
+    @Value("${datajobs.security.kerberos.keytabFileLocation}")
+    private String keytabFileLocation;
+
+    @Value("${datajobs.security.kerberos.enabled:false}")
```
tozka, line 92: Isn't better to pass them in the constructor as the rest of the options?

tozka: Now that we have multiple authentication mechanisms, we should have a flag for oauth2 authentication as well (before it was switched on/off by the general security flag).

tozka: But please add the flag in a separate (short) PR. This one is about Kerberos.

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -179,4 +204,67 @@ public JwtDecoder jwtDecoder() {
...
+
+    @Bean
+    @ConditionalOnProperty(value = "datajobs.security.kerberos.enabled")
+    public SecurityConfiguration.DataJobsUserDetailsService dataJobsUserDetailsService() {
+        return new SecurityConfiguration.DataJobsUserDetailsService();
+    }
+
+    @Override
+    @Bean
+    @ConditionalOnProperty(value = "datajobs.security.kerberos.enabled")
```
tozka, line 255: Extract datajobs.security.kerberos.enabled as contant. 

Also there's class called FeatureFlags, I think those kinda (feature on/off) flags should be defined there

mrMoZ1: ConditionalOnProperty is effectively doing that. It will return false if the property is not present in the environment by default.
 I think using a Conditional annotation is the best way to configure those Beans (it's managed by spring) I'm worried that if KRB auth is disabled those effectively blank configurations could interfere with other authenticators plus it's neater than having a bunch of if statements in every bean.
 I don't think there is an option like ConditionalOnBoolean and the other option of using a boolean in those Beans doesn't look ideal to me.

tozka: I mean you can put it as a constant there so that the string is re-used

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
mrMoZ1: https://github.com/vmware/versatile-data-kit/pull/764

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
mrMoZ1: For some reason the code won't compile when I put use a constant from the FeatureFlags class in an annotation in the SecurityConfiguration class so I put the constant in the same class.

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
mrMoZ1: https://github.com/vmware/versatile-data-kit/pull/765 But we should review this when we finish with this commit since it depends on code introduced here.

projects/control-service/projects/base/src/main/java/com/vmware/taurus/base/FeatureFlags.java
```diff
@@ -26,4 +26,7 @@
    @Value("${featureflag.authorization.enabled:false}")
    boolean authorizationEnabled = false;
 
+   @Value("${datajobs.security.kerberos.enabled:false}")
+   boolean KRBAuthEnabled = false;
```
tozka, line 30: variable names should start with lower-case letter
```suggestion
   boolean krbAuthEnabled = false;
```

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -59,6 +69,22 @@
...
+          "/data-jobs/webjars/**",
+          // There should not be sensitive data in prometheus, and it makes
+          // integration with the monitoring system easier if no auth is necessary.
+          "/data-jobs/debug/prometheus",
+          // TODO: likely /data-jobs/debug is too permissive
+          // but until we can expose them in swagger they are very hard to use with Auth.
+          "/data-jobs/debug/**"};
+    private final String kerberosPrincipal;
+    private final String keytabFileLocation;
+    private final String KERBEROS_AUTH_ENABLED_PROPERTY = "datajobs.security.kerberos.enabled";
```
tozka, line 87: my suggest was to add this constatnt in FeatureFlags so you can also reuse it in [KerberosConfig.java](https://github.com/vmware/versatile-data-kit/pull/755/files#diff-ef66e0bc2235c9ce8dabdbd1622fdd3e76efa28513ffcb028c114022f5082d80)

Also it should be `static final` I think.

tozka (changes requested):

tozka (approved):
Looks good to me. I made some minor comment you should look.
 But I am ok with merging it.

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
mrMoZ1: Yes well if it's in the FeatureFlags class the project won't compile - some gimmick of how annotations work I guess or some issue with lombok. It fails with [attribute must be a constant expression error] and if I move the variable to the SecurityConfiguration class it compiles ( I tried with a static final variable from FeatureFlags).

PR #757, opened 2022-03-08
versatile-data-kit: set automatic java formatter

tozka:
We have settled on using Google Java coding style . This is adding
github action to automatically format the code.

I nitially wanted to use pre-commit hook
(https://github.com/macisamuele/language-formatters-pre-commit-hooks)
but it requires to have installed locally java and I could not figuore
out how to make it optional for local runs.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,15 @@
+name: Google Java Format
+
+on:
+  push:
+    branches: [ "**" ]
+
+jobs:
+
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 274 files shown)


gageorgiev (approved):

ivakoleva (approved):
lgtm, seems the hook will autoupdate to use latest google java formatter, unless pinned to a specific one

tozka:
This is on hold for now because I have not figure out how to apply the pre-commit hook without requiring java to be installed . Java is only required for control service.  There are people that have contributed only to the SDK. We should not require them to have java.

PR #759, opened 2022-03-09
vdk-server: bugfix in running status check

tozka:
The check if the service is up was using bad URL (which was returning
404) so `vdk server -s` is always returning that it is down

Testing Done: installed locally vdk server (`vdk server -i`) and check
its status was correctly returned with `vdk server -s`


doks5 (approved):
LGTM

PR #761, opened 2022-03-09
control-service: Adopt kubernetes-client 14.0.1

gageorgiev:
To take advantage of the V1CronJob API and enable automatic
job cleanup for the builder jobs and manually executed data jobs,
we need to adopt the latest version of kubernetes-client.

Further steps are replacing the currently used CronJob object
with the new V1CronJob, and including the TTL parameter for
builder jobs and manually executed data jobs.

Testing done: project builds successfully and unittests pass


tozka:
> replacing the currently used CronJob object with the new V1CronJob,

As we need to support both older and newer versions of CronJob it's not straight up replacement, right?

PR #767, opened 2022-03-15
control-service: Adopt use of the V1CronJob API

gageorgiev:
To allow for the automatic job clean up of manually ran data jobs
and builder jobs, the VDK control service must use the V1CronJob
API. However, V1CronJobs are available from Kubernetes 1.21 onwards,
and VDK must be able to run on older versions of Kubernetes.
This change manages this by duplicating relevant methods, and relying
on a feature flag to determine whether the Kubernetes cluster supports
V1CronJobs, and defaulting back to V1beta1CronJobs in the case where
it does not.

Testing done: all existing unit and intergration tests pass, added unit tests


projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
```diff
@@ -309,12 +416,19 @@ public void testQuantityToMbConversion(int expectedMb, String providedResources)
 
     private KubernetesService mockCronJobFromTemplate() {
        KubernetesService mock = Mockito.mock(KubernetesService.class);
-       Mockito.when(mock.cronJobFromTemplate(
+       Mockito.when(mock.getK8sSupportsV1CronJob()).thenReturn(false);
```
ivakoleva, line 419: it seems the getter is mocked, however implementation seems to use the field directly,
for example `this.k8sSupportsV1CronJob` usages in KubernetesService.java?
This reference is also redundant btw

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -370,43 +448,136 @@ public void deleteNamespace(String namespaceName) throws ApiException {
...
+        }
+    }
 
+    private void startNewV1beta1CronJobExecution(String cronJobName, String executionId, Map<String, String> annotations,
+                                         Map<String, String> envs, Map<String, Object> extraJobArguments, String jobName) throws ApiException {
+        var cron = initBatchV1beta1Api().readNamespacedCronJob(cronJobName, namespace, null);
         Optional<V1beta1JobTemplateSpec> jobTemplateSpec = Optional.ofNullable(cron)
-              .map(v1beta1CronJob -> v1beta1CronJob.getSpec())
-              .map(v1beta1CronJobSpec -> v1beta1CronJobSpec.getJobTemplate());
+                .map(v1beta1CronJob -> v1beta1CronJob.getSpec())
```
ivakoleva, line 529: all of those lambdas could be optimised to direct method reference, for example
`v1beta1CronJob -> v1beta1CronJob.getSpec()` becomes `V1beta1CronJob::getSpec`

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
```diff
@@ -5,8 +5,10 @@
 
 package com.vmware.taurus.service.kubernetes;
 
+import com.vmware.taurus.base.FeatureFlags;
 import com.vmware.taurus.service.KubernetesService;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
```
ivakoleva, line 11: what is the problem with autowiring feature flags config?
btw, not sure if base module needs to be released first in case modified;

gageorgiev: I was getting errors due to the fact that autowiring requires an instance of the class, whereas I was using the autowired object inside the constructor. I went with the current approach as it seemed simpler (and it worked). I'm open to suggestions though.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
gageorgiev: Thanks for the note, I fixed the references to this.k8sSupportsV1CronJob in KubernetesService and replaced them with the getter.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
gageorgiev: noted

tozka (approved):
Looks ok to me.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -235,38 +243,69 @@ public void afterPropertiesSet() throws Exception {
...
         if(cronjobTemplate == null) {
-            return loadInternalCronjobTemplate();
+            return loadInternalV1beta1CronjobTemplate();
         }
 
         return cronjobTemplate;
     }
 
-    private V1beta1CronJob loadInternalCronjobTemplate() {
+    private V1beta1CronJob loadInternalV1beta1CronjobTemplate() {
```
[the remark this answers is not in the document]
gageorgiev: Seems like a massive restructuring, maybe something we can consider in the future?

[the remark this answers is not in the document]
tozka: It's not that massive. It's pretty basic "Extract method/class" re-factoring . And when in the future? One of the best times to do refactoring is around the time of doing functional changes.

 Of course, you should take into account what your planning and other tasks are.

PR #772, opened 2022-03-18
airflow-provider-vdk: Initial Airflow provider structure

gageorgiev:
Versatile Data Kit allows users to schedule job executions using
a cron-like interface, however there is no way to express job
dependencies through VDK. For example, a user might want to
run two different ingestion jobs, and when both pass successfully,
to run a transformation job on the ingested data.
Apache Airflow is an open-source workload scheduling framework
which allows exactly that sort of job dependency specification.
To take advantage of this, this PR sets the initial structure of the
VDK Airflow provider, which will integrate Airflow's scheduling
capabilities with VDK.

README is left short as much of the implementation and usage
is still somewhat up in the air AFAIK.
Interested in feedback on the way authentication is approached
in VDKHook.
Also interested in opinions on what would be good default values
for timeout and the retry params for VDKHook.

Testing done: package installs successfully


tozka:
Let's put it in the plugins directory. Technically it can be thought as of a plugin (extension) to vdk regardless if it uses vdk hooks or not.

tozka (approved):

support/airflow-provider-vdk/setup.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from setuptools import setup
+
+__version__ = "0.0.1"
+
+with open("README.md") as fh:
+    long_description = fh.read()
+
+"""
```
doks5, line 10: I think that we need to be careful to avoid failing static code checks.

support/airflow-provider-vdk/setup.py
```diff
@@ -0,0 +1,42 @@
...
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    entry_points={
+        "apache_airflow_provider": [
+            "provider_info=vdk_airflow.__init__:get_provider_info"
+        ]
+    },
+    license="Apache License 2.0",
+    packages=[
+        "vdk_airflow",
```
doks5, line 26: Not sure if the packages (or the provider itself for that matter) should include the word airflow. No official providers seem to include it https://github.com/apache/airflow/tree/main/airflow/providers.

projects/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2021 VMware, Inc.
```
tozka, line 1: This seems wrong ?

projects/vdk-plugins/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2021 VMware, Inc.
```
tozka, line 1: This seems wrong. __init__.py does not have place here. vdk-plugins is not a python module. It's just 
a directory

gageorgiev: seems these two got added accidentally, I'll get rid of them

PR #774, opened 2022-03-20
vdk-kerberos-auth: support kerberos auth for all CLI commands

tozka:
vdk-control-cli commands also may require Kerberos authentication so
here we are adding support. Currently, we only authenticate during vdk
run (part of initialize_job). But with support for Kerberos for the
Control Service we'd need it for all operations. E.g `vdk properties`

Now authentication is always attempted as long as it's explicitly
enabled (KRB_AUTH is set to some valid value). 
There's new flag "KRB_AUTH_FAIL_FAST" (false by default) 
which would cause the plugin to fail if auth fails but otherwise only a warning would be logged. 
It's necessary because only small subsets of features would use kerberos. 
And vdk should work even if those feature do not.

Alternatives considered: 

I was considering adding a new decorator/annotation to vdk-kerberos-auth 
which "users" can use to denote that their method needs Kerberos authentication
 (and would be authenticated on demand) 

e.g  
```
@kerberos 
def get_property(team, name): 
```

when get_property is called then Authenticator.authenticate is triggered. But for now I've discarded 
a) it's more expensive 
b) How do we deal with multiple auth mechanisms (as weh ave Oauth2 and kerberos) - 
e.g when do we fail?, do we attempt both , etc. 
c) caching of authentication
But it's nice idea and I am reserving it for the future.

Testing Done: extended the automated test of the plugin (see
test_kerberos).

### DIFF
--- a/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
+++ b/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
@@ -1,8 +1,11 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+import logging
 import os
 from pathlib import Path
+from typing import Optional
 
+from vdk.internal.core import errors
 from vdk.internal.core.config import Configuration
 from vdk.internal.core.config import ConfigurationBuilder
 
@@ -13,10 +16,16 @@ KEYTAB_FILENAME = "KEYTAB_FILENAME"
 KEYTAB_PRINCIPAL = "KEYTAB_PRINCIPAL"
 KEYTAB_REALM = "KEYTAB_REALM"
 KERBEROS_KDC_HOST = "KERBEROS_KDC_HOST"
+KRB_AUTH_FAIL_FAST = "KRB_AUTH_FAIL_FAST"
 
 
 class KerberosPluginConfiguration:
-    def __init__(self, job_name: str, job_directory: str, config: Configuration):
+    def __init__(
+        self,
+        job_name: Optional[str],
+        job_directory: Optional[str],
+        config: Configuration,
+    ):
         self.__job_name = job_name
         self.__job_directory = job_directory
         self.__config = config
@@ -26,25 +35,40 @@ class KerberosPluginConfiguration:
 
     def keytab_folder(self):
         keytab_folder = self.__config.get_value(KEYTAB_FOLDER)
-        if keytab_folder is None:
+        if keytab_folder is None and self.__job_directory:
             keytab_folder = Path(self.__job_directory).parent
         return keytab_folder
 
     def keytab_filename(self):
         keytab_filename = self.__config.get_value(KEYTAB_FILENAME)
-        if keytab_filename is None:
+        if keytab_filename is None and self.__job_name:
             keytab_filename = f"{self.__job_name}.keytab"
         return keytab_filename
 
     def keytab_pathname(self):
-        return os.path.join(self.keytab_folder(), self.keytab_filename())
+        keytab_folder = self.keytab_folder()
+        keytab_filename = self.keytab_filename()
+        if not keytab_filename:
+            errors.log_and_throw(
+                to_be_fixed_by=errors.ResolvableBy.CONFIG_ERROR,
+                log=logging.getLogger(__name__),
+                what_happened="Cannot find keytab file location.",
+                why_it_happened="Keytab filename cannot be inferred from configuration.",
+                consequences=errors.MSG_CONSEQUENCE_DELEGATING_TO_CALLER__LIKELY_EXECUTION_FAILURE,
+                countermeasures="Provide configuration variables KEYTAB_FILENAME KEYTAB_FOLDER. "
+                "During vdk run they are automatically inferred from data job location "
+                "but for other commands they need to be explicitly set.",
+            )
+        if keytab_folder:
+            return os.path.join(keytab_folder, keytab_filename)
+        return keytab_filename
 
     def krb5_conf_filename(self):
         return self.__config.get_value(KRB5_CONF_FILENAME)
 
     def keytab_principal(self):
         keytab_principal = self.__config.get_value(KEYTAB_PRINCIPAL)
-        if keytab_principal is None:
+        if keytab_principal is None and self.__job_name:
             keytab_principal = f"pa__view_{self.__job_name}"
         return keytab_principal
 
@@ -53,6 +77,9 @@ class KerberosPluginConfiguration:
 
     def kerberos_host(self):
         return self.__config.get_value(KERBEROS_KDC_HOST)
+
+    def auth_fail_fast(self) -> bool:
+        return self.__config.get_value(KRB_AUTH_FAIL_FAST)
 
 
 def add_definitions(config_builder: ConfigurationBuilder) -> None:
@@ -100,4 +127,13 @@ def add_definitions(config_builder: ConfigurationBuilder) -> None:
         default_value="localhost",
         description="Specifies the name of the Kerberos KDC (Key Distribution Center) host. "
         "This value is used only with the 'minikerberos' authentication type.",
+    )
+    config_builder.add(
+        key=KRB_AUTH_FAIL_FAST,
+        default_value=False,
+        description="Specify if the authenticator must raise exception "
+        "if it fails to successfully authenticate with kerberos (basically - kinit). "
+        "If set to false, only warning will be logged on authentication failure. "
+        "Subsequent kerberos related requests may fail but that'd fail lazily (on demand) "
+        "that makes it possible for non-kerberos related features to work.",
     )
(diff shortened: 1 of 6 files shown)


gageorgiev:
>There's new flag "KRB_AUTH_FAIL_FAST" (false by default)
>which would cause the plugin to fail if auth fails but otherwise only a warning would be logged.

I'm not clear I understand this in the desc. Do you mean that if the flag is true, an auth fail causes the whole execution of whatever VDK command is being ran to fail, and if it is false, it logs a warning but otherwise it proceeds normally? Perhaps rephrase it for clarity.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
```diff
@@ -26,7 +33,7 @@ def authentication_type(self):
 
     def keytab_folder(self):
         keytab_folder = self.__config.get_value(KEYTAB_FOLDER)
-        if keytab_folder is None:
+        if keytab_folder is None and self.__job_directory:
             keytab_folder = Path(self.__job_directory).parent
         return keytab_folder
```
gageorgiev, line 38: What happens if keytab_folder is None but also self.__job_directory doesn't exist? There might be reason why this isn't possible that I'm missing, but otherwise it seems the whole function will return None.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
```diff
@@ -101,3 +111,12 @@ def add_definitions(config_builder: ConfigurationBuilder) -> None:
...
     )
+    config_builder.add(
+        key=KRB_AUTH_FAIL_FAST,
+        default_value=False,
+        description="Specify if the authenticator must raise exception "
+        "if it fails to successfully authenticate with kerberos (basically - kinit). "
+        "If set to false, only warning will be logged on authentication failure. "
+        "Subsequent kerberos related requests may fail but that'd fail lazily (on demand) "
+        "that makes it possible for non-kerberos related features to work.",
+    )
```
gageorgiev, line 122: I don't understand the need for this env flag. If not all commands require Kerberos auth, and thus would fail, in what use case would a user want the app to fail when auth fails?

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_plugin.py
```diff
@@ -18,32 +19,56 @@ class KerberosPlugin:
...
+    def vdk_initialize(self, context: CoreContext) -> None:
+        kerberos_configuration = KerberosPluginConfiguration(
+            None, None, context.configuration
+        )
+        self.__attempt_kerberos_authentication(kerberos_configuration)
+
     @hookimpl(tryfirst=True)
     def initialize_job(self, context: JobContext) -> None:
+        """
+        This is called during vdk run (job execution) and here we know hte job directory
```
gageorgiev, line 65:
```suggestion
        This is called during vdk run (job execution) and here we know the job directory
```
nit

gageorgiev (commented):
Looks good overall; I want clarity on some details before I can hit approve on this.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
doks5: I think I have the opposite question: Is it possible for `__job_directory` to not exist?

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_plugin.py
```diff
@@ -18,32 +19,56 @@ class KerberosPlugin:
...
+                else:
+                    log.warning(
+                        "Kerberos was enabled but we failed to authenticate (kinit) and generate TGT."
+                        f"The error was: {e} "
+                        "This means potential future operations requiring "
+                        "valid kerberos authentication will fail. "
+                        "You may need to inspect the above error and try to fix it if that happens. "
+                        "If no kerberos related operations are done, ignore this warning."
+                        "If you prefer to fail fast (instead of this warning), enable krb_auth_fail_fast flag."
+                        "See vdk config-help for more info."
```
doks5, line 43:
```suggestion
                        "See vdk config-help for more information."
```

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_plugin.py
```diff
@@ -18,32 +19,56 @@ class KerberosPlugin:
...
+                "no authentication for kerberos is attempted as it's not enabled."
+                "To enable set KRB_AUTH to the correct type. See vdk config-help for more info."
+            )
+
     @staticmethod
     @hookimpl
     def vdk_configure(config_builder: ConfigurationBuilder) -> None:
         add_definitions(config_builder)
 
+    def vdk_initialize(self, context: CoreContext) -> None:
```
doks5, line 56: Shouldn't this have `@hookimpl`?

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
tozka: When trying to debug issues - making it possible to fail fast would make it easier to debug since they'd see the original error quickly. Or in some cloud runs where we know that job should not continue without authentication successful.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
tozka: Added tests to cover some of those scenarios.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_plugin.py
tozka: fixed

doks5 (approved):

ivakoleva (commented):
I am missing the exact problem description, seems to me the problem was `__attempt_kerberos_authentication` was only invoked during `initialize_job`, and now it's also invoked during `vdk_initialize` also. So it now applies to all vdk commands, not only vdk run.


Is it true that this feature is purposed to enable VDK-CLI-requesting data-job-specific endpoints (for example properties) using a keytab only? 

I see two use-cases:
1. a user who does not have valid oauth2 credentials, yet has gotten the data job keytab, and is now able to read data-job details, to operate/update/delete this data job;
2. a data job now is able to manage itself via VDK CLI invocations (including reschedule, update it's own sources).

Use-case 1 is very rare - keytab is generated on vdk create, that runs after oauth2 authentication; the only way is for the keytab secret to be retrieved from someone having oauth2 credentials in the first place.
Use-case 2 does involve executing a vdk command, is that a recommendation and/or possibility? Should we document this?

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/kerberos_configuration.py
```diff
@@ -13,10 +16,16 @@
 KEYTAB_PRINCIPAL = "KEYTAB_PRINCIPAL"
 KEYTAB_REALM = "KEYTAB_REALM"
 KERBEROS_KDC_HOST = "KERBEROS_KDC_HOST"
+KRB_AUTH_FAIL_FAST = "KRB_AUTH_FAIL_FAST"
```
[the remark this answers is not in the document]
tozka: > Is it a correct statement that kerberos auth has precedence over oauth2 when using VDK CLI? Should that be configurable, or described?

That will be a separate change in vdk-control-cli which would decide how would both work.

This change just make it sure to kinit during vdk_initialize so that other commands (not just vdk run) can re-use the generated Kerberos ticket (TGT) if they need to.

tozka:
A use-case is to enable Control Service deployments that use only Kerberos as authentication (without Oauth2). 

If the Control Service supports Kerberos but the vdk cli does not, it is not complete. It's important to keep the client (vdk-control-cli) and server (Control Service REST API) in sync.

ivakoleva:
> A use-case is to enable Control Service deployments that use only Kerberos as authentication (without Oauth2).
> 
> If the Control Service supports Kerberos but the vdk cli does not, it is not complete. It's important to keep the client (vdk-control-cli) and server (Control Service REST API) in sync.
@tozka 

In general I agree, and in favour of consistency. Yet, Kerberos auth is mainly targeted for an actor being a data job (in case of a person, that is a very rare corner case). Does that mean that VDK CLI command is now purposed for evaluation from within a data job context? For example, a data job would be evaluating VDK CLI commands? 
And if yes, maybe we need to document this design/use-case/guideline? 

Also, isn't another approach more suitable for facilitating such an use case (data job requesting the REST API), rather than the CLI? Perhaps the vdk-plugin-control-cli instead?

What I'm missing, is the use-cases first, then the implementation

PR #777, opened 2022-03-28
vdk-core: Verify payload after pre-processing it

YanaZhivkova:
Ingestion allows pre-processing payloads. If the payload
is curated during pre-processing, currently, verification
would fail.

Verify payload after pre-processing it, since this
pre-processing might be responsible for making it serializable.
Also, add VDK_ prefix to configs so that ingestion pre/post
processors are properly initialized.

Tested by: locally run ingestion job which curates json
during pre-processing

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
@@ -112,8 +112,6 @@ class IngesterBase(IIngester):
         """
         See parent doc
         """
-        self.__verify_payload_format(payload_dict=payload)
-
         if collection_id is None:
             collection_id = "{data_job_name}|{execution_id}".format(
                 data_job_name=self._data_job_name, execution_id=self._op_id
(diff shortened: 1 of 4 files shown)


ivakoleva (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
```diff
@@ -226,10 +226,10 @@ def __initialize_ingester(self, method) -> IngesterBase:
         else:
             # Initialize the pre- and post- processors.
             initialized_pre_processors = self.__get_initialized_processors(
-                "INGEST_PAYLOAD_PREPROCESS_SEQUENCE"
+                "VDK_INGEST_PAYLOAD_PREPROCESS_SEQUENCE"
```
tozka, line 229: It should be without vdk_

YanaZhivkova: Yes, VDK_ prefix is now removed in supercollider-vdk, which is the correct way to fix this mismatch which lead to never loading pre-processors in our test case.

tozka (approved):

PR #782, opened 2022-04-04
control-service: fix concatAddresses NPE

mivanov1988:
This change aims to fix the NPE when null is passed to EmailNotification.concatAddresses().

Testing done: unit tests


gageorgiev (approved):

mivanov1988:
> This is crying for unit tests, I think :)
> 
> And I find it a good practice to have a test for each bug we fix. Bug tend to re-surface (unless we check for them). I'd usually even write the test before I make the fix (thus helping me write the fix faster and making sure I have a good test).

The whole class is crying for unit tests, but such have not been implemented :)

mivanov1988:
> This is crying for unit tests, I think :)
> 
> And I find it a good practice to have a test for each bug we fix. Bug tend to re-surface (unless we check for them). I'd usually even write the test before I make the fix (thus helping me write the fix faster and making sure I have a good test).

Added tests.

PR #784, opened 2022-04-05
control-service: fix builder security context

mivanov1988:
Converted builder security context properties to String, because "if 0"
returns false and as a result the property is not present in control
service envs.

Testing Done: helm template

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -29,9 +29,9 @@ deploymentBuilderImage:
 
 deploymentBuilder:
   securityContext:
-    runAsUser: 0
-    runAsGroup: 1000
-    fsGroup: 1000
+    runAsUser: "0"
+    runAsGroup: "1000"
+    fsGroup: "1000"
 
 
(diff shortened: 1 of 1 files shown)


gageorgiev (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -29,9 +29,9 @@ deploymentBuilderImage:
 
 deploymentBuilder:
   securityContext:
-    runAsUser: 0
-    runAsGroup: 1000
-    fsGroup: 1000
+    runAsUser: "0"
```
YanaZhivkova, line 32: Where is the "if 0" check? If possible, is it not better to convert values to string there, before the check, so that cases like this are avoided in the future?

mivanov1988: https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml#L219

mivanov1988: It's debatable

PR #786, opened 2022-04-06
VEP-554: Extend Security and Permissions section

doks5:
As part of the development of the Versatile Data Kit Airflow Provider, we need to
ensure that all connections to the Control Service are properly authenticated. For
this to happen, we need to define the authentication mechanisms and flows that would
be used.

This change extends the `Security and Permissions` section of the enhancement proposal
to include more details on how authentication would be facilitated in the Airflow Provider.

Testing Done: Documentation change, no testing needed.

### DIFF
--- a/specs/vep-554-apache-airflow-integration/README.MD
+++ b/specs/vep-554-apache-airflow-integration/README.MD
@@ -417,7 +417,7 @@ Connection(
 ```
 where the `conn_type` would be "vdk" to identify that this is a Versatile Data Kit Connection.
 For the authentication to be successful, users would need to provide credentials as shown in the tables
-above, and an `extra` prameter pointing to the Authorization Server that would be used.
+above, and an `extra` parameter pointing to the Authorization Server that would be used.
 
 In terms of security, Airflow uses [Fernet](https://github.com/fernet/spec/) to encrypt passwords in
 connection configurations, which should be sufficient. However, this is where `vdk-control-cli` falls
--- a/specs/vep-554-apache-airflow-integration/README.MD
+++ b/specs/vep-554-apache-airflow-integration/README.MD
@@ -377,7 +377,12 @@ Unit and integration tests will be introduced by following the official Airflow 
 
 The communication between VDK Provider and Control Service will be secured through the
 use of HTTPS and OAuth 2.0. The most commonly used OAuth 2.0 flows will be supported (
-e.g. Client Credentials Flow, Refresh Token Flow, etc.).
+e.g. Client Credentials Flow, Refresh Token Flow, etc.) by using `vdk-control-cli`'s authentication
+functionality. This would make `vdk-control-cli` a hard dependency of the Provider, though
+it is possible for the functionality to be moved as a separate library.
+
+Currently, `vdk-control-cli` fully supports the Refresh Token Flow, and provides partial support for
+the Client Credentials Flow.
 
 Secrets will be stored as part of the Airflow Connection like follows:
 
@@ -394,6 +399,31 @@ Secrets will be stored as part of the Airflow Connection like follows:
 |---------------------|---------------|
 | connection.login    | Client ID     |
 | connection.password | Client Secret |
+
+For the connection itself, a standard Airflow Connection object object would be used,
+
+```python
+Connection(
+  conn_id="",
+  conn_type="vdk",
+  description=None,
+  login="<client-id-or-empty>",
+  password="<password-or-oauth-token>",
+  host="<Control-Service-host-URL>",
+  port=,
+  schema=None,
+  extra=json.dumps(dict(auth_endpoint="<oauth2-authentication-endpoint>"))
+)
+```
+where the `conn_type` would be "vdk" to identify that this is a Versatile Data Kit Connection.
+For the authentication to be successful, users would need to provide credentials as shown in the tables
+above, and an `extra` prameter pointing to the Authorization Server that would be used.
+
+In terms of security, Airflow uses [Fernet](https://github.com/fernet/spec/) to encrypt passwords in
+connection configurations, which should be sufficient. However, this is where `vdk-control-cli` falls
+short, as it caches the credentials/tokens in plain text. Therefore, some code changes would be required
+to better secure credentials.
+
 
 <!--
 How is access control handled?
--- a/specs/vep-554-apache-airflow-integration/README.MD
+++ b/specs/vep-554-apache-airflow-integration/README.MD
@@ -400,7 +400,7 @@ Secrets will be stored as part of the Airflow Connection like follows:
 | connection.login    | Client ID     |
 | connection.password | Client Secret |
 
-For the connection itself, a standard Airflow Connection object object would be used,
+For the connection itself, a standard Airflow Connection object would be used,
 
 ```python
 Connection(
--- a/specs/vep-554-apache-airflow-integration/README.MD
+++ b/specs/vep-554-apache-airflow-integration/README.MD
@@ -1,7 +1,7 @@
 # VEP-554: Apache Airflow Integration
 
 * **Author(s):** Miroslav Ivanov (miroslavi@vmware.com), Antoni Ivanov (
-  aivanov@vmware.com)
+  aivanov@vmware.com), Andon Andonov (andonova@vmware.com)
 * **Status:** draft
 
 - [Summary](#summary)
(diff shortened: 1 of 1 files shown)


specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -377,7 +377,12 @@ Unit and integration tests will be introduced by following the official Airflow
 
 The communication between VDK Provider and Control Service will be secured through the
 use of HTTPS and OAuth 2.0. The most commonly used OAuth 2.0 flows will be supported (
-e.g. Client Credentials Flow, Refresh Token Flow, etc.).
+e.g. Client Credentials Flow, Refresh Token Flow, etc.) by using `vdk-control-cli`'s authentication
+functionality. This would make `vdk-control-cli` a hard dependency of the Provider, though
+it is possible for the functionality to be moved as a separate library.
```
tozka, line 382:
```suggestion
it is possible for the functionality to be moved as a separate library.
```

Let's be more definite. Please make a recommendation should it be moved or not? 
I personally think yes. But we can discuss pros and cons

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -377,7 +377,12 @@ Unit and integration tests will be introduced by following the official Airflow
...
+
+Currently, `vdk-control-cli` fully supports the Refresh Token Flow, and provides partial support for
```
tozka, line 384: vdk-control-cli provides support for [Authorization Code Flow with Proof Key for Code Exchange (PKCE)](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce).


What do you mean by partial support for the Client Credentials Flow ?


Also What is refresh token flow ? I assume you mean either refresh token used as part of [Autorization Code flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) or the custom API token flow we have .. ? 

Please include links to new terms you use (in a similar way as I did above) or define them in the Terms section if they are our own terminology. Otherwise, there's more room for ambiguity and miscommunication.

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -395,6 +400,31 @@ Secrets will be stored as part of the Airflow Connection like follows:
...
+  extra=json.dumps(dict(auth_endpoint="<oauth2-authentication-endpoint>"))
+)
+```
+where the `conn_type` would be "vdk" to identify that this is a Versatile Data Kit Connection.
+For the authentication to be successful, users would need to provide credentials as shown in the tables
+above, and an `extra` prameter pointing to the Authorization Server that would be used.
+
+In terms of security, Airflow uses [Fernet](https://github.com/fernet/spec/) to encrypt passwords in
+connection configurations, which should be sufficient. However, this is where `vdk-control-cli` falls
+short, as it caches the credentials/tokens in plain text. Therefore, some code changes would be required
```
tozka, line 424: What changes exactly? Please be precise.

tozka (approved):

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -395,6 +400,31 @@ Secrets will be stored as part of the Airflow Connection like follows:
 | connection.login    | Client ID     |
 | connection.password | Client Secret |
 
+For the connection itself, a standard Airflow Connection object object would be used,
```
gageorgiev, line 403:
```suggestion
For the connection itself, a standard Airflow Connection object would be used,
```

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -395,6 +400,31 @@ Secrets will be stored as part of the Airflow Connection like follows:
...
+  password="<password-or-oauth-token>",
+  host="<Control-Service-host-URL>",
+  port=,
+  schema=None,
+  extra=json.dumps(dict(auth_endpoint="<oauth2-authentication-endpoint>"))
+)
+```
+where the `conn_type` would be "vdk" to identify that this is a Versatile Data Kit Connection.
+For the authentication to be successful, users would need to provide credentials as shown in the tables
+above, and an `extra` prameter pointing to the Authorization Server that would be used.
```
gageorgiev, line 420:
```suggestion
above, and an `extra` parameter pointing to the Authorization Server that would be used.
```

specs/vep-554-apache-airflow-integration/README.MD
doks5: Good point. Refactored, thanks

specs/vep-554-apache-airflow-integration/README.MD
doks5: Reworded. I don't think we need new terms when it comes to Oauth2.

tozka: Please include a link to the specification - so it's clear what you mean. 
In case of Refresh Token flow link to https://datatracker.ietf.org/doc/html/rfc6749#section-1.5 (I assume)

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -395,6 +402,31 @@ Secrets will be stored as part of the Airflow Connection like follows:
...
 
+For the connection itself, a standard Airflow Connection object would be used,
+
+```python
+Connection(
+  conn_id="",
+  conn_type="vdk",
+  description=None,
+  login="<client-id-or-empty>",
+  password="<password-or-oauth-token>",
```
tozka, line 413: user(login) and password are not the same as client id / client secretor oauth token.
In the future we may support some type of Basic Authentication with user and password in which case those field would be used. 

please look at what other similar hooks/connectoins that support oauth2 do - https://github.com/airflow-plugins/marketo_plugin/blob/master/hooks/marketo_hook.py, https://github.com/airflow-plugins/google_sheets_plugin/blob/master/hooks/google_hook.py 

I suggest we put the required security configuration in extras.

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -395,6 +402,31 @@ Secrets will be stored as part of the Airflow Connection like follows:
...
+  host="<Control-Service-host-URL>",
+  port=,
+  schema=None,
+  extra=json.dumps(dict(auth_endpoint="<oauth2-authentication-endpoint>"))
```
tozka, line 417: what is auth endpoint ? There are a couple of options.

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -377,7 +377,14 @@ Unit and integration tests will be introduced by following the official Airflow
 
 The communication between VDK Provider and Control Service will be secured through the
 use of HTTPS and OAuth 2.0. The most commonly used OAuth 2.0 flows will be supported (
-e.g. Client Credentials Flow, Refresh Token Flow, etc.).
+e.g. Client Credentials Flow, Refresh Token Flow, etc.) by using `vdk-control-cli`'s authentication
+functionality. The functionality would be extracted as a stand-alone
+library, which would make it more concise and lightweight, avoiding the
+installation of unneeded dependencies.
+
+Having a stand-alone authentication library would also allow for the
```
tozka, line 385: Please describe what functionality the library would provide? What is the interface of the library (in other words What is the expected input and output)
Let's finalize it here. 

I am thinking something along those lines: 

It would provide an Authentication class to which we can pass configuration - we will reuse https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/src/vdk/internal/control/auth/auth.py#L84 class here. 

We will extend the class also to support necessary features: 
 A) Because in airflow we may not have local disk access always or it may not be considered secure we will introduce the option to keep the Autheorization Cache in memory 
B) THis is a question for you. do we need to encrypt the keys? You mentioned below it's necessary ?

specs/vep-554-apache-airflow-integration/README.MD
```diff
@@ -377,7 +377,14 @@ Unit and integration tests will be introduced by following the official Airflow
 
 The communication between VDK Provider and Control Service will be secured through the
 use of HTTPS and OAuth 2.0. The most commonly used OAuth 2.0 flows will be supported (
-e.g. Client Credentials Flow, Refresh Token Flow, etc.).
+e.g. Client Credentials Flow, Refresh Token Flow, etc.) by using `vdk-control-cli`'s authentication
```
tozka, line 380: don't say 'e.g' (for example) . Be specific. 
We do not support Client Credentials Flow as defined in https://datatracker.ietf.org/doc/html/rfc6749#section-4.4  ? (or described more simply here - https://oauth.net/2/grant-types/client-credentials/)  . Unless I am having a brain check out now (which happens to me). 

--
We do support Refresh token Grant (or flow) - it's implemented [here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/src/vdk/internal/control/auth/auth.py#L180)

We do support  Authorization Code Flow with Proof Key for Code Exchange (PKCE) - it's implemented [here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/src/vdk/internal/control/auth/redirect_auth.py) 

And we do support this (nonofficial - in other words, it does not conform to any RFC spec) API Key  flow - implemented [here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/src/vdk/internal/control/auth/apikey_auth.py) - which is somewhat similar to refresh token grant (and might be used for that actually). 

We also support kerberos. But as we will not use it in Airflow , let's not think about it.

We definitely need to document those. I don't believe we have good documentation. I can certainly help with that. 

--
Now that it's clear what we support currently in VDK. I believe we want to support this API key login type , right? 
This is probably what you mean by refresh token flow ? Is that right?

tozka (changes requested):

mivanov1988:
Don't forget to add your name to the Authors section :)

specs/vep-554-apache-airflow-integration/README.MD
tozka: I don't think that's a concern. It will also mask extras if they contain certain keys
See https://airflow.apache.org/docs/apache-airflow/stable/security/secrets/mask-sensitive-values.html#sensitive-field-names

```
When masking is enabled, Airflow will always mask the password field of every Connection that is accessed by a task.
It will also mask the value of a Variable, or the field of a Connection's extra JSON blob 
if the name contains any words in ('access_token', 'api_key', 'apikey','authorization', 'passphrase', 'passwd', 'password', 'private_key', 'secret', 'token').
```

specs/vep-554-apache-airflow-integration/README.MD
mivanov1988: But it is a core configuration, right?

mivanov1988: For example, Databrick provider uses password filed for client_sercret.

mivanov1988: I see no problem in reusing the password field, but I am ok with both approaches.

specs/vep-554-apache-airflow-integration/README.MD
mivanov1988: But it is a core configuration, right?

mivanov1988: For example, the Databrick provider uses a password field for client_sercret.

mivanov1988: I see no problem in reusing the password field, but I am ok with both approaches.

specs/vep-554-apache-airflow-integration/README.MD
doks5: Added links.

specs/vep-554-apache-airflow-integration/README.MD
doks5: Updated the spec.

specs/vep-554-apache-airflow-integration/README.MD
doks5: Fixed, thanks

specs/vep-554-apache-airflow-integration/README.MD
doks5: Extended the description of the expected functionality.

tozka (approved):
Looks good. Thanks.

PR #791, opened 2022-04-07
control-service: configurable builder job service account

mivanov1988:
Currently, the builder job is based on Kaniko and requires root privileges.
In some cases, Pod Security Policy can be configured based on namespace and service account.
However, Control Service does not provide the ability to set a service account to the Builder Job.

This change aims to provide an ability to set a service account through the Helm Chart.

spec.template.spec.serviceAccountName: "account_name"

Testing done: unit tests and helm template

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -32,6 +32,8 @@ deploymentBuilder:
     runAsUser: "0"
     runAsGroup: "1000"
     fsGroup: "1000"
+  ## The name of the ServiceAccount to use.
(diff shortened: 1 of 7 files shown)


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -32,6 +32,8 @@ deploymentBuilder:
     runAsUser: "0"
     runAsGroup: "1000"
     fsGroup: "1000"
+  ## The name of the ServiceAccount to use.
+  # serviceAccountName: ""
```
doks5, line 36: Should this be commented out?

doks5 (approved):

PR #793, opened 2022-04-07
vdk-core: New feature: StandaloneDataJob

mrdavidlaing:
The PR contains a WIP implementation of a vdk-core extension that addresses #791

`StandaloneDataJob` - a context manager that:
- can be instantiated from code rather than via the VDK CLI
- can be run without needing any data job files
- gives access to an instantiated `job_input` object before it is finalized

Example usage:

```
 with StandaloneDataJobFactory.create(datajob_directory) as job_input:
      #... use job_input object to interact with SuperCollider
```

### DIFF
--- a/projects/vdk-core/tests/functional/run/test_run_standalone_data_job.py
+++ b/projects/vdk-core/tests/functional/run/test_run_standalone_data_job.py
@@ -1,0 +1,64 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import collections
+from pathlib import Path
+
+import pytest
+from functional.run import util
+from vdk.internal.builtin_plugins.run.standalone_data_job import (
+    StandaloneDataJobFactory,
+)
+
+ExecutedStandaloneDataJobDetails = collections.namedtuple(
+    "ExecutedStandalongDataJob", "hook_tracker job_input_name"
+)
+
+
+@pytest.fixture(scope="module")
+def executed_standalone_data_job():
+    hook_tracker = util.HookCallTracker()
+    job_input_name = "unknown"
+
+    with StandaloneDataJobFactory.create(
+        data_job_directory=Path(__file__), extra_plugins=[hook_tracker]
+    ) as job_input:
+        job_input_name = job_input.get_name()
+
+    return ExecutedStandaloneDataJobDetails(
+        hook_tracker=hook_tracker,
+        job_input_name=job_input_name,
+    )
+
+
+def test_standalone_data_job_named_after_data_job_directory(
+    executed_standalone_data_job,
+):
+    assert executed_standalone_data_job.job_input_name == Path(__file__).name
+
+
+def test_standalone_data_job_executes_hooks_in_expected_order(
+    executed_standalone_data_job,
+):
+    assert (
+        ", ".join(executed_standalone_data_job.hook_tracker.calls)
+        == "vdk_start, vdk_configure, vdk_initialize, initialize_job, run_job, run_step, finalize_job, vdk_exit"
+    )
+
+
+def test_standalone_data_job_does_not_call_command_line_hook(
+    executed_standalone_data_job,
+):
+    assert "vdk_command_line" not in executed_standalone_data_job.hook_tracker.calls
+
+
+def test_standalone_data_job_does_calls_exception_hook_after_exception_thrown():
+    hook_tracker = util.HookCallTracker()
+
+    with pytest.raises(ValueError) as exc_info:
+        with StandaloneDataJobFactory.create(
+            data_job_directory=Path(__file__), extra_plugins=[hook_tracker]
+        ):
+            raise ValueError("Test Error")
+
+    assert "vdk_exception" in hook_tracker.calls
+    assert str(exc_info.value) == "Test Error"
--- a/projects/vdk-core/src/vdk/api/data_job.py
+++ b/projects/vdk-core/src/vdk/api/data_job.py
@@ -1,0 +1,82 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import pathlib
+from abc import abstractmethod
+
+from vdk.api.job_input import IJobInput
+
+
+class IStandaloneDataJob:
+    """
+    A contextmanager that initialises VDK and its plugins,
+    triggers plugin lifecycle hooks needed to expose an initialised IJobInput on entry
+    and finalization hooks when the context is exited
+
+    NB: The following plugin hooks are not triggered:
+        * CoreHookSpecs.vdk_command_line
+
+    Sample usage::
+
+        with StandaloneDataJobFactory.create(datajob_directory) as job_input:
+            #... use job_input object to interact with SuperCollider
+    """
+
+    @abstractmethod
+    def __init__(
+        self,
+        data_job_directory: pathlib.Path | None,
+        name: str | None = None,
+        job_args: dict | None = None,
+    ):
+        """
+        Initialises:
+            * PluginRegistry
+                * load_plugins_from_setuptools_entrypoints
+                * core-plugins
+                * ExecutionTrackingPlugin
+            * Configuration
+            * CoreContext
+            * StateStore
+        Triggers plugin lifecycle hooks:
+            * CoreHookSpecs.vdk_start.call_historic
+            * CoreHookSpecs.vdk_configure
+            * CoreHookSpecs.vdk_initialize
+
+        Arguments:
+            data_job_directory: Optional[pathlib.Path]
+                The source code of the data job that will be executed.
+            name: Optional[str]
+                The name of the job.  If omitted will be inferred from the director name.
+            job_args: Optional[dict]
+                Allows for users to pass arguments to data job run.
+                Data Job arguments are also used for parameter substitution in queries, see execute_query docstring.
+        """
+
+    @abstractmethod
+    def __enter__(self) -> IJobInput:
+        """
+        Executes on entry to the context block:
+            * Adds a noop step DataJob hook to prevent any logic being dynamically executed from step files
+            * Initialises the JobContext
+
+        And then triggers plugin lifecycle hooks:
+            * initialize_job
+            * run_job
+
+        Returns:
+            An initialised IJobInput object
+        """
+        pass
+
+    @abstractmethod
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        """
+        Executes on exit of the context block
+
+        Triggers plugin lifecycle hooks:
+            * finalize_job
+            * vdk_exit
+        """
+        pass
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/standalone_data_job.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/standalone_data_job.py
@@ -1,0 +1,186 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import pathlib
+import sys
+from pathlib import Path
+from typing import cast
+
+from click import ClickException
+from vdk.api.data_job import IStandaloneDataJob
+from vdk.api.job_input import IJobInput
+from vdk.api.plugin.core_hook_spec import CoreHookSpecs
+from vdk.api.plugin.core_hook_spec import JobRunHookSpecs
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins import builtin_hook_impl
+from vdk.internal.builtin_plugins.internal_hookspecs import InternalHookSpecs
+from vdk.internal.builtin_plugins.run.data_job import DataJobDefaultHookImplPlugin
+from vdk.internal.builtin_plugins.run.data_job import JobArguments
+from vdk.internal.builtin_plugins.run.execution_tracking import ExecutionTrackingPlugin
+from vdk.internal.builtin_plugins.run.file_based_step import TYPE_PYTHON
+from vdk.internal.builtin_plugins.run.job_context import JobContext
+from vdk.internal.builtin_plugins.run.step import Step
+from vdk.internal.core.config import ConfigurationBuilder
+from vdk.internal.core.context import CoreContext
+from vdk.internal.core.statestore import StateStore
+from vdk.internal.plugin.plugin import PluginRegistry
+
+
+class NoOpStepDataJobHookImplPlugin(DataJobDefaultHookImplPlugin):
+    """
+    Implementation that overrides of the functionality of the DataJobDefaultHookImplPlugin
+    so that it runs a single NoOp step
+    """
+
+    @hookimpl
+    def initialize_job(self, context: JobContext):
+        step = Step(
+            name="NoOpStep",
+            type="noop",
+            runner_func=lambda _step, _job_input: True,  # Intentional noop
+            file_path=Path(__file__),
+            job_dir=context.job_directory,
+        )
+        context.step_builder.add_step(step)
+
+
+class StandaloneDataJob(IStandaloneDataJob):
+    """
+    A contextmanager that executes all the necessary "setup" logic needed to return an initialised IJobInput
+    object and finalize logic when the context is exited
+
+    Sample usage::
+
+        with StandaloneDataJob(datajob_directory) as job_input:
+            #... use job_input object to interact with SuperCollider
+    """
+
+    def __init__(
+        self,
+        data_job_directory: pathlib.Path | None,
+        name: str | None = None,
+        job_args: dict | None = None,
+        extra_plugins: list = [],
+    ):
+        if data_job_directory is None and name is None:
+            raise ValueError(
+                "Cannot initialize DataJob. "
+                "At least one of data job directory or data job name need to be set. "
+            )
+        self._name = data_job_directory.name if name is None else name
+        self._data_job_directory = data_job_directory
+        if job_args is None:
+            job_args = {}
+        self._job_args = job_args
+
+        self._plugin_registry = PluginRegistry()
+        self._plugin_registry.add_hook_specs(InternalHookSpecs)
+        for plugin in extra_plugins:
+            self._plugin_registry.load_plugin_with_hooks_impl(plugin)
+        self._plugin_registry.load_plugins_from_setuptools_entrypoints()
+        self._plugin_registry.add_hook_specs(CoreHookSpecs)
+        self._plugin_registry.load_plugin_with_hooks_impl(
+            builtin_hook_impl, "core-plugin"
+        )
+
+        cast(CoreHookSpecs, self._plugin_registry.hook()).vdk_start.call_historic(
+            kwargs=dict(plugin_registry=self._plugin_registry, command_line_args=[])
+        )
+
+        conf_builder = ConfigurationBuilder()
+        cast(CoreHookSpecs, self._plugin_registry.hook()).vdk_configure(
+            config_builder=conf_builder
+        )
+        configuration = conf_builder.build()
+
+        core_context = CoreContext(self._plugin_registry, configuration, StateStore())
+        core_context.plugin_registry.load_plugin_with_hooks_impl(
+            ExecutionTrackingPlugin()
+        )
+        cast(CoreHookSpecs, self._plugin_registry.hook()).vdk_initialize(
+            context=core_context
+        )
+
+        self._core_context = core_context
+        self._job_context = None
+
+    def __enter__(self) -> IJobInput:
+        try:
+            self._core_context.plugin_registry.load_plugin_with_hooks_impl(
+                NoOpStepDataJobHookImplPlugin(), NoOpStepDataJobHookImplPlugin.__name__
+            )
+
+            from vdk.internal.builtin_plugins.templates.template_impl import (
+                TemplatesImpl,
+            )
+
+            self._job_context = JobContext(
+                name=self._name,
+                job_directory=self._data_job_directory,
+                core_context=self._core_context,
+                job_args=JobArguments(self._job_args),
+                templates=TemplatesImpl(
+                    job_name=self._name, core_context=self._core_context
+                ),
+            )
+
+            # We need to call both the initialize_job and run_job hooks to get a fully initialized JobInput
+            cast(JobRunHookSpecs, self._plugin_registry.hook()).initialize_job(
+                context=self._job_context
+            )
+            cast(JobRunHookSpecs, self._plugin_registry.hook()).run_job(
+                context=self._job_context
+            )
+
+            return self._job_context.job_input
+        except:
+            # Ensure __exit__ gets called so the vdk_exception hook can be triggered
+            self.__exit__(*sys.exc_info())
+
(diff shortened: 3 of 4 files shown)


tozka:
Thanks. I think it looks in right direction. 

But first what would integration with dagster look like using this?
If you can post a snippet, pseudo-code something so we can imagine it better. 

I am asking because 
 a) I want us to make sure this would work as an integration point
 b) I would like to check if similar integration with other tools (like prefect or flyte) is possible.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/noop_datajob.py
```diff
@@ -0,0 +1,123 @@
...
+    Sample usage:
+            datajob_directory = Path(abspath(file_relative_path(__file__, '../')))
+            datajob = NoOpStepDataJob(datajob_directory)
+            try:
+                datajob.initialize_job()
+                vdk_job_input = datajob.run_and_return_job_input()
+
+                ... use vdk_job_input object to interact with SuperCollider
+            finally:
+                datajob.finalize_job()
```
tozka, line 58: I am not sure about NoOpStepDataJob name but I can't think of a better one.  

I would prefer to hide initailzie_job and finalize_job because 
A) I'd like to minimize the new interfaces we expose . 
B) Make it simpler to use (without requiring to call manually methods)
C) The implementation of Data Job run(execution) cycle algorithm [here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py#L255) is something that in the future may change and even become a plugin hook itself (so customizable) so it's better it's a bit hidden.

In python often context manager may be used to achieve that (https://docs.python.org/3/library/contextlib.html )

So maybe something like 

```
class NoOpDataJob: 
   def __enter__(self)
     # effectively this is hidden from user and we can change it if 
      self.job.initialize_job 

  def __exit__(self): 
     self.job.finalize_job()
```

and would be used like 
```
with NoOpDataJob() as job:
    job.get_job_input() 
```

mrdavidlaing:
> Thanks. I think it looks in right direction.

Thanks for the quick feedback!

> But first what would integration with dagster look like using this? If you can post a snippet, pseudo-code something so we can imagine it better.

The way we've approached it is to integrate into [Dagster's Resource system](https://docs.dagster.io/concepts/resources#resources); which basically stores an instance of the  VDK `job_input` object on an object that is passed to every `op()` executed by Dagster

It looks roughly like this:

```
@resource(...snip...)
def supercollider_datawarehouse_resource(init_context):
    datajob_directory = Path(abspath(file_relative_path(__file__, '../')))

    datajob = NoOpStepDataJob(datajob_directory)
    try:
        datajob.initialize_job()
        vdk_job_input = datajob.run_and_return_job_input()

        yield SuperColliderDwhResource(
            vdk_job_input=vdk_job_input,
            ...snip...
        )
    finally:
        datajob.finalize_job()
```

The `SuperColliderDwhResource()` is a fairly thin wrapper around `vdk_job_input.get_managed_connection()`, `vdk_job_input.load()` and  `vdk_job_input.send_tabular_data_for_ingestion()` internally.

We have a similar `SQLiteDwhResource()` that we use during local development & testing.

For full details see the internal implementation & discussion at https://gitlab.eng.vmware.com/tanzu-portfolio-insights/tanzu-dm/-/merge_requests/24, specifically:

* [data-jobs/tanzu-restricted-dm-daily/tanzu-restricted-dm-daily/resources/datawarehouse.py#L88-114](https://gitlab.eng.vmware.com/tanzu-portfolio-insights/tanzu-dm/-/blob/64d1e5e81460b3f1665e5e1b9e0b870036a3c72e/data-jobs/tanzu-restricted-dm-daily/tanzu-restricted-dm-daily/resources/datawarehouse.py#L88-114) and 
* [lib/vdk.py#L30-131](https://gitlab.eng.vmware.com/tanzu-portfolio-insights/tanzu-dm/-/blob/64d1e5e81460b3f1665e5e1b9e0b870036a3c72e/lib/vdk.py#L30-131)
 
> I also made some small comments on the code but we probably should clear the interfaces and expected usage first.

Agreed.  I'll loop back and address the points mentioned above once we've got the high level structure sorted out

projects/vdk-core/src/vdk/internal/builtin_plugins/run/noop_datajob.py
mrdavidlaing: @tozka I took a stab at hiding the `initialize_job()` and `finalize_job()` in f01eaf6 using a contextmanager as suggested, enabling usage like this:

```
with InitialisedJobInput(datajob_directory) as vdk_job_input:
    #... use vdk_job_input object to interact with SuperCollider
```

Notes:
1.  I really like how this cleans up usage & reduces the API surface area
2. Renamed `NoOpDataJob` to `InitialisedJobInput` - still not sure about this name

WDYT?

tozka:
> The way we've approached it is to integrate into [Dagster's Resource system](https://docs.dagster.io/concepts/resources#resources); which basically stores an instance of the VDK job_input object on an object that is passed to every op() executed by Dagster

Sorry for taking my time. I was a bit busy and I wanted to research a bit more Dagster. 
I like the approach. It's really simple and it's very much inline with how other somewhat similar tools have integrated with dagster. I took [dagster-airbyte](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-airbyte/dagster_airbyte) ,  [dagster-pyspark](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-pyspark) and [daster-mlflow](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-mlflow/dagster_mlflow) 

It will also make it easier to integrate with other frameworks (like Jupyter notebook, e.g to be able to execute data jobs though a notebook). Something I am planning to integrate soon so I'd like to re-use this. 



I would love if we can contribute to dagster to have[dagster-vdk](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries) integration library. I don't know if it's something you can consider.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/noop_datajob.py
tozka: I didn't like the "NoOp" part because using plugins you can provide actual steps to the job so it might not be NoOpStep. You simply add a noOp step to avoid the validation for empty steps but that's an implementation detail. 


I still think it's a DataJob though not JobInput. We simply expose job input for further usage. Maybe InitializedDataJob or StandaloneDataJob. Something like that.

Here we are inverting the flow of control really. From more declarative (where the job is executed by framework (vdk run)) to more imperative - the user decides when the job would be executed.

tozka:
> We have a similar set of datawarehouse_resources backed by SQLite that we use during local development & testing.

Oh btw, there's now vdk-sqlite plugin - if you install it and set in db_default_type = sqlite - you'd get sqlite based data job. 
Note sure if it might help here but I wanted to pointed out.

mrdavidlaing:
> I would love if we can contribute to dagster to have [dagster-vdk](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries) integration library. I don't know if it's something you can consider.

That seems like a good long term goal - although I'm not sure when I'd have time to contribute towards that.

Since having a way to instantiate a DataJob from code (the subject of this PR) is a necessary building block to enable a `dagster-vdk` library; could we keep this PR focussed on that narrower objective?

tozka:
> > I would love if we can contribute to dagster to have [dagster-vdk](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries) integration library. I don't know if it's something you can consider.
> 
> That seems like a good long term goal - although I'm not sure when I'd have time to contribute towards that.
> 
> Since having a way to instantiate a DataJob from code (the subject of this PR) is a necessary building block to enable a `dagster-vdk` library; could we keep this PR focussed on that narrower objective?

Most certainly yes. I didn't mean to imply otherwise. It's definitely a separate effort.

tozka:
@mrdavidlaing 

To avoid ambiguity. To close the PR I think we need the following things 


1. Finalize naming discussion
2. Take care of comments on the code - the code duplication particularly is something for me that I'd like we fix.
3. Create the public interface in vdk.api.
   One simple way to do that is to create vdk/api/data_job.py something like
```
class SomeNameWeAgreeOnDataJob:

    def __init__(self):
        from vdk.internal.builtin_plugins.run import DataJobImpl
        self.__job = DataJobImpl()

    def __enter__(self):
        self.__job.__enter__(self)

    def __exit__(self):
        self.__job.__exit()

    def get_job_input(self):
        return self.__job.get_job_input()
```

4. Tests. I suggest some kind of "functional" test similar to any of those https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-core/tests/functional 

We have not yet set up our CICD to trigger from a fork. So as long as the tests pass locally (from vdk-core folder ./cicd/build.sh succeeds) I think that's enough and I'll merge the change.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/datajob_initializer.py
```diff
@@ -41,15 +41,15 @@ def initialize_job(self, context: JobContext):
...
+    A contextmanager that executes all the necessary "setup" logic needed to return an initialised JobContext
+    and finalize logic when the context is exited
 
     Sample usage::
 
-        datajob_directory = Path(abspath(file_relative_path(__file__, '../')))
-        with InitialisedJobInput(datajob_directory) as vdk_job_input:
-            #... use vdk_job_input object to interact with SuperCollider
+        with InitialisedJobContext(datajob_directory) as vdk_job_context:
+            #... use vdk_job_context.job_input object to interact with SuperCollider
```
mrdavidlaing, line 52: What do you think about returning the `JobContext` object rather than just the `JobInput` ?

Related, what do you think about the name `InitialisedJobContext`?

tozka: Are there any use-cases that require JobContext? It's too broad interface. I think we should expose the minimum necessary interface surface otherwise we are in hard time to maintain it. 

Also it's not very mature.  Unlike JobInput it was created fairly recently. In the future, it's possible to make it Immutable (once initialize_job phase is finished) to prevent changes that might lead to unpredictable behaviour. In that case it might be fine.

mrdavidlaing: I like `StandaloneDataJob`.  I'll switch to that

mrdavidlaing: > Are there any use-cases that require JobContext?

I found one (although perhaps this is possible via the IJobInput interface?)

After calling `send_tabular_data_for_ingestion` I want to wait for the ingestor to finish sending the messages.

The only "cleanish" way I've found to do this is via the ingester attribute on JobContext.

```
  self.vdk_job_context.job_input.send_tabular_data_for_ingestion(
            rows=...snip...
        )
        # Ensure the ingestor send queue is emptied before we exit
        self.log.info(f"Waiting for VDK ingestor to finish sending queued messages")
        cast(IngesterRouter, self.vdk_job_context.ingester).close()
```       

Now that I think of it however, I could perhaps put a call to `ingester.close()` during the `__exit__` method, ala:

```
def __exit__(self):
        cast(IngesterRouter, self.__job.job_context.ingester).close()
        self.__job.__exit()
```       

WDYT?

tozka: This is possible by using [`wait_to_finish_after_every_send`](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration.py#L108) configuration. 

Here is an example job where it is used : https://github.com/vmware/versatile-data-kit/blob/main/examples/online-exhibition/online-exhibition/config.ini#L33

mrdavidlaing: Aha!  Thats a much better way to do it.

I'll switch that on and revert to just exposing the `JobInput`

mrdavidlaing:
Still working on this - just got sidetracked a bit.

Will hopefully push up some new code early next week

projects/vdk-core/src/vdk/internal/builtin_plugins/run/datajob_initializer.py
mrdavidlaing: I've addressed all the points in this conversation.  Resolving

mrdavidlaing:
TODO:

[x] Finalize naming discussion
[ ] Take care of comments on the code - the code duplication particularly is something for me that I'd like we fix.
[ ] Create the public interface in vdk.api.
[ ] Tests.

tozka (commented):
That looks good to me. My suggestion is to postpone the refactoring (de-duplication) for separate PR. We just need some automated tests and we can merge it.

mrdavidlaing:
> That looks good to me. My suggestion is to postpone the refactoring (de-duplication) for separate PR. We just need some automated tests and we can merge it.

Sounds good.

I've been fighting a bit trying to get the build working on an Mac M1 (which doesn't officially support Python 3.7).  Should hopefully have something to commit later in the week

tozka:
> 
> I've been fighting a bit trying to get the build working on an Mac M1 (which doesn't officially support Python 3.7). Should hopefully have something to commit later in the week

You don't need to use Python 3.7 . This VDK works with all recent python versions.

mrdavidlaing:
Is there a way you can disable/skip the use of python 3.7 in the commit process?

```
❯ git commit
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/jorisroovers/gitlint.
[INFO] Initializing environment for https://github.com/jorisroovers/gitlint:./gitlint-core[trusted-deps].
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://github.com/pycqa/pydocstyle.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-pylint.
[INFO] Initializing environment for https://github.com/asottile/reorder_python_imports.
[INFO] Initializing environment for https://github.com/asottile/pyupgrade.
[INFO] Initializing environment for https://github.com/Lucas-C/pre-commit-hooks.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/Users/mrdavidlaing/.pyenv/versions/3.8.13/bin/python3.8', '-mvirtualenv', '/Users/mrdavidlaing/.cache/pre-commit/repoej7q_26p/py_env-python3.7', '-p', 'python3.7')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'
    
stderr: (none)
Check the log at /Users/mrdavidlaing/.cache/pre-commit/pre-commit.log
```

```
❯ cat /Users/mrdavidlaing/.cache/pre-commit/pre-commit.log
### version information

pre-commit version: 2.18.1
git --version: git version 2.32.0 (Apple Git-132)
sys.version:
    3.8.13 (default, May  4 2022, 18:37:14) 
    [Clang 13.1.6 (clang-1316.0.21.2.3)]
sys.executable: /Users/mrdavidlaing/.pyenv/versions/3.8.13/bin/python3.8
os.name: posix
sys.platform: darwin

### error information

An unexpected error has occurred: CalledProcessError: command: ('/Users/mrdavidlaing/.pyenv/versions/3.8.13/bin/python3.8', '-mvirtualenv', '/Users/mrdavidlaing/.cache/pre-commit/repoej7q_26p/py_env-python3.7', '-p', 'python3.7')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'
    
stderr: (none)


  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/error_handler.py", line 73, in error_handler
    yield
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/main.py", line 343, in main
    return hook_impl(
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/commands/hook_impl.py", line 237, in hook_impl
    return retv | run(config, store, ns)
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/commands/run.py", line 414, in run
    install_hook_envs(to_install, store)
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/repository.py", line 221, in install_hook_envs
    _hook_install(hook)
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/repository.py", line 79, in _hook_install
    lang.install_environment(
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/languages/python.py", line 202, in install_environment
    cmd_output_b(*venv_cmd, cwd='/')
  File "/Users/mrdavidlaing/.pyenv/versions/3.8.13/lib/python3.8/site-packages/pre_commit/util.py", line 146, in cmd_output_b
    raise CalledProcessError(returncode, cmd, retcode, stdout_b, stderr_b)
pre_commit.util.CalledProcessError: command: ('/Users/mrdavidlaing/.pyenv/versions/3.8.13/bin/python3.8', '-mvirtualenv', '/Users/mrdavidlaing/.cache/pre-commit/repoej7q_26p/py_env-python3.7', '-p', 'python3.7')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'
    
stderr: (none)
```

tozka:
> Is there a way you can disable/skip the use of python 3.7 in the commit process?
>  

Interesting, that seems some configuration issue with our pre-commit hooks. I opened PR to fix it:  https://github.com/vmware/versatile-data-kit/pull/826

projects/vdk-core/tests/functional/run/test_run_standalone_data_job.py
```diff
@@ -0,0 +1,15 @@
...
+from vdk.internal.builtin_plugins.run.standalone_data_job import (
+    StandaloneDataJobFactory,
+)
+
+
+def test_standalone_data_job():
+    data_job_directory = Path(util.job_path("simple-job"))
+
+    with StandaloneDataJobFactory.create(data_job_directory) as job_input:
+        assert job_input.get_name() == "simple-job"
```
mrdavidlaing, line 15: @tozka I'd like to add a bunch of tests that validate that all the correct vdk hooks have been called in the right order.

Is there an example of a test already doing this?  Thx

tozka: I don't think we have a test just for the order. That seems like an omission that we have not realized until now. 
Certainly, some tests would fail if the order is not kept (e.g vdk_initialize hook often relies on some configuration from vdk_configure hook). 

But not all - I switched the order of execution of "vdk_command_line" and "vdk_configure" in the code and re-run all tests and nothing failed !!! (Something should have failed). 

I can suggest implementing a few hooks and checking the order.
The order is documented at https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-plugins#types-of-plugins 
Something like : 
```
class OrderedHooks:
    calls = []

    @hookimpl
    def vdk_configure(...):
        self.calls.append("vdk_configure")

 ....

```


CliEntryBasedTestRunner might be helpful also: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-test-utils/src/vdk/plugin/test_utils/util_funcs.py#L81

tozka: CliEntryBasedTestRunner may not be as helpful for your case. I guess you can inject PluginRegistry and use             plugin_registry.load_plugin_with_hooks_impl(plugin) to load your testing plugins

projects/vdk-core/tests/functional/run/test_run_standalone_data_job.py
```diff
@@ -1,15 +1,62 @@
...
+        == "vdk_start, vdk_configure, vdk_initialize, initialize_job, run_job, run_step, finalize_job"
+    )
+
+
+@pytest.mark.skip(
+    reason="Not sure whether vdk_exit should be called, given we're not executing from the command line"
+)
+def test_standalone_data_job_executes_vdk_exit_hooks(executed_standalone_data_job):
+    assert "vdk_exit" in executed_standalone_data_job.hook_tracker.calls
+
```
mrdavidlaing, line 53: @tozka, Do you think the `StandaloneDataJob` should call the `vdk_exit` hook? 

I'm unsure from the [docs](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L113) where this is CLI specific behaviour

projects/vdk-core/tests/functional/run/test_run_standalone_data_job.py
mrdavidlaing: Thanks for the pointers.  [517e334](https://github.com/vmware/versatile-data-kit/pull/793/commits/517e3349838f0c6e245cd1856c5b76e68ab47018) contains my first attempt at these kinds of tests

projects/vdk-core/tests/functional/run/test_run_standalone_data_job.py
tozka: Yes, I bet there are (and will be) plugins providing some functionality for data jobs that rely on vdk_exit.

mrdavidlaing:
@tozka With 54f4a09 done I'm happy that we have a basic set of functional tests.  Are there any other tests you'd like to see added?

tozka (approved):
It looks good to me. Have you run ./cicd/build.sh locally from vdk-core folder ? 

We have not configured yet to run CI from forks, unfortunately. So I'd need to merge your change manually. I will run the test myself but it would be good to get confirmation first.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/noop_datajob.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    Implementation that overrides of the functionality of the DataJobDefaultHookImplPlugin
+    so that it runs a single NoOp step
+    """
+
+    @hookimpl
+    def initialize_job(self, context: JobContext):
+        step = Step(
+            name=f"NoOpStep",
+            type=TYPE_PYTHON,
```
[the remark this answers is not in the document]
mrdavidlaing: Addressed by 210ec2a

tozka:
I've merged your change. Congrats and thanks for your first contribution (I hope one of many ;) ) 

The CI has kicked in - https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/536648262 hopefully it will pass and your change would be automatically released.

tozka:
> @tozka Am I correct in deducing that this feature landed in https://pypi.org/project/vdk-core/0.2.536648262/ ?

Yes

PR #798, opened 2022-04-11
control-service: Kerberos authentication IT

mrMoZ1:
why: As a part of recently introduced functionality to provide kerberos authentication
we needed to add an integration test with a running control service.

what: Added Kerberos authentication tests against a running control service. The test
creates it's own Kerberos server (using apache kerby) and creates as unique keytab
used to authenticate against the server each time. The test verifies that authentication
works as expected.

tests: n/a


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/KerberosAuthenticationIT.java
```diff
@@ -0,0 +1,172 @@
...
+   private static String principal = "pa__view_foo";
+
+
+//   @Autowired
+//   protected MockMvc mockMvc;
+
+   @LocalServerPort
+   int randomPort;
+
+   static {
```
tozka, line 65: Shouldn't this be in [beforeAll](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/BeforeAll.html)

mrMoZ1: No since this is the only way I could think of to start the KDC before spring startup code...

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/KerberosAuthenticationIT.java
```diff
@@ -0,0 +1,172 @@
...
+//      KerberosTicket kerberosTicket = null;
+//
+//      for (var object : subject.getPrivateCredentials()) {
+//         if (object instanceof KerberosTicket) {
+//            kerberosTicket = (KerberosTicket) object;
+//         }
+//      }
+
+      KerberosRestTemplate kerberosRestTemplate = new KerberosRestTemplate(keytab.getPath(), principal);
+      var response = kerberosRestTemplate.getForObject("http://localhost:" + randomPort + "/data-jobs/for-team/supercollider/jobs", String.class);
```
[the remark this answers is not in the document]
mrMoZ1: Currently this doesn't work (the request that the rest template sends is without authentication header) and throws Unauthorized exception even though KerberosRestTemplate calls it's own login method. I have been following this example : https://docs.spring.io/spring-security-kerberos/docs/current/reference/html/ssk-resttemplate.html , I have also tried different libraries (kerb4j) etc but there is some issue with the authentication. The whole test is based on this example: https://www.baeldung.com/spring-security-kerberos-integration but it just doesn't work.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/KerberosAuthenticationIT.java
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright 2021 VMware, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
ivakoleva, line 5:
```suggestion
/*
 * Copyright 2021 VMware, Inc.
 * SPDX-License-Identifier: Apache-2.0
 */

package com.vmware.taurus.datajobs.it;

import com.vmware.taurus.ServiceApp;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.boot.web.server.LocalServerPort;
import org.springframework.core.Ordered;
import org.springframework.security.kerberos.client.KerberosRestTemplate;
import org.springframework.security.kerberos.test.KerberosSecurityTestcase;
import org.springframework.test.context.TestContext;
import org.springframework.test.context.TestExecutionListener;
import org.springframework.test.context.TestExecutionListeners;
import org.springframework.test.context.TestPropertySource;
import org.springframework.test.context.junit.jupiter.SpringExtension;

import java.io.File;
import java.util.Collections;

@TestPropertySource(properties = {
        "test.dir=target",
        "featureflag.security.enabled=true",
        "datajobs.security.kerberos.enabled=true",
        "datajobs.security.kerberos.kerberosPrincipal=HTTP/localhost",
        "datajobs.security.kerberos.keytabFileLocation=target/foo.keytab",
        "datajobs.security.kerberos.krb5ConfigLocation=target/krb5.conf",
        "java.security.krb5.conf=target/krb5.conf",
        "sun.security.krb5.debug=true"
})
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = ServiceApp.class)
@ExtendWith(SpringExtension.class)
@TestExecutionListeners(value = {KerberosAuthenticationIT.KdcExecutionListener.class}, mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
public class KerberosAuthenticationIT {
    // composition in favour of inheritance, due to KDC instance is used via TestExecutionListeners and Test both (shared)
    public static final KerberosSecurityTestcase testcase = new KerberosSecurityTestcase();
    public static final File keytab = new File("target/foo.keytab");

    public static class KdcExecutionListener implements TestExecutionListener, Ordered {
        @Override
        public int getOrder() {
            return 0;
        }

        @Override
        public void beforeTestClass(final TestContext testContext) throws Exception {
            testcase.startMiniKdc();
            testcase.getKdc().createPrincipal(keytab, "client/localhost", "HTTP/localhost");
        }
    }

    @LocalServerPort
    int randomPort;

    @Test
    public void test() throws Exception {
        System.out.println();
        KerberosRestTemplate kerberosRestTemplate = new KerberosRestTemplate(keytab.getPath(), "client/localhost", Collections.singletonMap("isInitiator", "true"));
        var response = kerberosRestTemplate.getForObject("http://localhost:" + randomPort + "/data-jobs/for-team/supercollider/jobs", String.class);
    }
}
```
That's a suggestion with bootstrapping KDC before the test using the Spring way. 

Nothing else to configure from KerberosRestTemplate's side - did not find any headers setting. ResponseHeaderSettingKerberosAuthenticationSuccessHandler seems to set some headers to the response upon successful authentication. 
The issue with the test is, that the workflow does not reach KerberosServiceAuthenticationProvider's `authenticate()`.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/KerberosAuthenticationIT.java
mivanov1988: What about the Spring Extension approach that I mentioned before?

mrMoZ1: Tried it, but I didn't find a reliable way to run the test without a static block.

support/gitlab-runners/values.yaml
```diff
@@ -75,7 +75,7 @@ runners:
     cpuLimit: 1000m
     memoryLimit: 4Gi
     cpuRequests: 500m
-    memoryRequests: 2Gi
+    memoryRequests: 3Gi
```
tozka, line 78: I think you should increase the limit "memoryLimit" as well . memoryRequests is a "soft" limit and you can go over it and nothing would happen. If you go over the memoryLimit , then you'd get Out of Memory error. 

I recommend reading https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run

tozka: Also open a new PR for this change only (new branch) do not change existing PR.

tozka: You also need to apply the limits manually. Ther are instructions in https://github.com/vmware/versatile-data-kit/tree/main/support/gitlab-runners but ping me if you need help.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/authorization/KerberosAuthenticationIT.java
```diff
@@ -0,0 +1,146 @@
...
+import org.springframework.boot.web.server.LocalServerPort;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.io.File;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * This test is in the unit tests directory, because
```
tozka, line 33: I would say the test is in the unit test because 
A) it's fast (it takes a few seconds)
B) It mocks and manages all its dependencies and does not create side effects . 
C) It does not work well with MiniKDC 

The Mini KDC related issues - I would add as a TODO in teh MniKDC class in the integration test suite

But even if we replace MiniKDC with Kerby it's fine to leave the test here. for reasons A and B.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/authorization/KerberosAuthenticationIT.java
```diff
@@ -0,0 +1,146 @@
...
+
+   @Test
+   public void testAuthenticatedCall() throws Exception {
+
+      SpnegoClient spnegoClient = SpnegoClient.loginWithKeyTab(CLIENT_PRINCIPAL, KEYTAB.getPath());
+      URL url = new URL("http://127.0.0.1:" + randomPort + "/data-jobs/for-team/test-team/jobs/test-name");
+      var krb5 = simpleKdcServer.getKrbClient().getKrbConfig();
+      SpnegoContext context = spnegoClient.createContext(new URL("http://" + krb5.getKdcHost() + "/" + krb5.getKdcPort()));
+      HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+      conn.setRequestProperty("Authorization", context.createTokenAsAuthroizationHeader());
```
tozka, line 121: extract this in separate Utility class (KerberosUtility.requestWithKerberosAuth(httpUri, princpak, keytab). 

It's very clear we'd want to reuse this . And generally, this kind of helper code does not belong in the test method itself.  It might be a helper method in the same class. But I think it's better to have a separate utility class since I think there's a very good chance to be reused.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/KerberosUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import com.kerb4j.client.SpnegoContext;
+import org.ietf.jgss.GSSException;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.security.PrivilegedActionException;
+
+public class KerberosUtil {
+   public static HttpURLConnection requestWithKerberosAuth(URL requestURL, URL kdcAuthURL,
```
tozka, line 18: Please explain what the method does and what is each parameter.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/authorization/KerberosAuthenticationIT.java
```diff
@@ -0,0 +1,106 @@
...
+      "datajobs.security.kerberos.keytabFileLocation=target/foo.keytab",
+      "datajobs.security.kerberos.krb5ConfigLocation=target/krb5.conf",
+      "java.security.krb5.conf=target/krb5.conf",
+      "sun.security.krb5.debug=true",
+      "datajobs.authorization.authorized-roles="
+})
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = ControlplaneApplication.class)
+@ExtendWith(SpringExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class KerberosAuthenticationIT extends TestKerberosServer {
```
tozka, line 57: I prefer composition over inheritance. I'd have written it as Junit extension. But that's also fine. We've spent enough time on this.

tozka (approved):
LGTM. You can write a bit more documentation if you wish and merge it.

PR #803, opened 2022-04-14
vdk-core: minor refactoring in managed_cursor to reduce long method

tozka:
Testing Done: unit tests


projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -78,51 +78,14 @@ def execute(
...
-                        errors.ResolvableBy.PLATFORM_ERROR,
-                        self._log,
-                        what_happened="Decorating query FAILED.",
-                        why_it_happened=errors.MSG_WHY_FROM_EXCEPTION(e),
-                        consequences=errors.MSG_CONSEQUENCE_DELEGATING_TO_CALLER__LIKELY_EXECUTION_FAILURE,
-                        countermeasures=errors.MSG_COUNTERMEASURE_FIX_PARENT_EXCEPTION,
-                        exception=e,
-                    )
+            self._validate_operation(operation, parameters)
+            self._decorate_operation(managed_operation, operation)
```
gageorgiev, line 82:
```suggestion
            if self.__connection_hook_spec.db_connection_decorate_operation.get_hookimpls():
                    self._validate_operation(operation, parameters)
            if self.__connection_hook_spec.db_connection_validate_operation.get_hookimpls():
                    self._decorate_operation(managed_operation, operation)
```

Personally I think it makes more sense semantically to have the wrapping if's outside the actual methods. This is a tiny detail and I understand if you ignore it.

gageorgiev (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
tozka: It depends on the purpose of the method. 

In this case, it's better to be inside because it's really part of decorate invocation - it's some kind of optimization - we check if there's decorate hook implementation and invoke it if there is.  Technically the if should not be necessary as pluggy should take care . But as it's purely refactoring change I didn't want to make a semantic change to remove it.

PR #804, opened 2022-04-14
vdk-core: print query duration

mrMoZ1:
why: Users requested clearer examples on how long given operations took: 
"- In concrete case problems was with SQL query , it would be nice if we can
print profile details (especially if query fails)"

what: Added new log statements that show how long an SQL query took.

testing: Added unit tests.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
@@ -89,13 +89,15 @@ class ManagedCursor(PEP249Cursor):
             result = super().execute(
                 *managed_operation.get_operation_parameters_tuple()
             )
-            self._log.info(f"Executing query SUCCEEDED. Query {self._get_query_duration(query_start_time)}")
+            self._log.info(
+                f"Executing query SUCCEEDED. Query {_get_query_duration(query_start_time)}"
+            )
             return result
         except Exception as e:
             try:
                 self._recover_operation(e, managed_operation)
                 self._log.info(
-                    f"Recovered query {self._get_query_duration(query_start_time)}"
+                    f"Recovered query {_get_query_duration(query_start_time)}"
                 )
             except Exception as e:
                 # todo: error classification
@@ -103,9 +105,7 @@ class ManagedCursor(PEP249Cursor):
                 blamee = errors.ResolvableBy.USER_ERROR
                 # else:
                 #     blamee = errors.ResolvableBy.PLATFORM_ERROR
-                self._log.info(
-                    f"Failed query {self._get_query_duration(query_start_time)}"
-                )
+                self._log.info(f"Failed query {_get_query_duration(query_start_time)}")
                 errors.log_and_rethrow(
                     blamee,
                     self._log,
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -121,18 +123,28 @@ def execute(
                     )
 
         self._log.info("Executing query:\n%s" % managed_operation.get_operation())
+        query_start_time = timer()
         try:
             super().execute(*managed_operation.get_operation_parameters_tuple())
             self._log.info("Executing query SUCCEEDED.")
+            self._log.info(
```
tozka, line 130: 2 lines seems unnecssary. Let's merge the in one.

projects/vdk-core/tests/vdk/internal/builtin_plugins/connection/test_managed_cursor.py
```diff
@@ -163,3 +164,54 @@ def mock_recover(recovery_cursor: RecoveryCursor):
...
+    (
+        mock_native_cursor,
+        mock_managed_cursor,
+        _,
+        _,
+        mock_connection_hook_spec,
+    ) = populate_mock_managed_cursor()
+    # set logging level to info
+    with mock.patch.object(
+        mock_managed_cursor.__class__, "_recover_operation"
```
tozka, line 208: It's a bad practice to be using any private or protected fields or methods. 
This makes the test more brittle (a small change in teh implementation would break it) 

We want our tests to be as black box as possible (know little of what's going on inside and use only the public interfaces)


There is  `mock_connection_hook_spec.db_connection_recovery_operation` which we can mock around . Wouldn't that work?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -121,18 +123,28 @@ def execute(
...
+                f"Successful query {self._get_query_duration(query_start_time)}"
+            )
         except Exception as e:
             try:
                 self._recover_operation(e, managed_operation)
+                self._log.info(
+                    f"Recovered query {self._get_query_duration(query_start_time)}"
```
tozka, line 137: Can you copy paste the log. How would it look like? 

Let's make it debug.

mrMoZ1: 'Recovered query duration H:M:S 0:01:02.103319'

> Let's make it debug.
Why?

tozka: We have too many unnecessary logs. 

Let's see more user friendly formatting.

mrMoZ1: New format is 'Recovered query duration 00h:00m:09s'

tozka: We talked offline that for now we can leave them info and we will revisit the full default logging verbosity in a separate effort.

tozka (approved):

PR #805, opened 2022-04-18
vdk-core: add new managed db_connection_execute_operation hook

tozka:
Currently we cannot track the full execution of a query (before and
after it) . This is necessary in order to be able to generate proper
lineage events (query start and end) or to track its duration (we had to
currently had to add this to vdk-core
(https://github.com/vmware/versatile-data-kit/pull/804) but that's
really better implemented as plugin (think aspect-oriented programming).

In the future we should consider adding hook for fetch (fetchMany,
fetchAll) as well as it's there where the data is returned (plugins can
take stats - count number of rows, validate sensitive columns, do result
validation tests, etc.)

Testing Done: unit tests, functional test.


projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
```diff
@@ -68,6 +70,54 @@ def db_connection_decorate_operation(
         """
         pass
 
+    @hookspec(firstresult=True)
+    def db_connection_execute_operation(
+        self, raw_cursor: Any, managed_operation: ManagedOperation
```
ivakoleva, line 75: `raw_cursor: Any` contrasts with `managed_operation: ManagedOperation` abstraction,
I'll elaborate  further

projects/vdk-core/tests/vdk/internal/builtin_plugins/connection/test_connection_router.py
```diff
@@ -45,7 +53,7 @@ def test_router_raw_connection():
     router.add_open_connection_factory_method("RAW_DB", lambda: mock_conn)
 
     conn = router.open_connection("RAW_DB")
-    assert mock_conn == conn.connect()
+    assert mock_conn == conn._connect()
```
ivakoleva, line 56: why going for non-PEP249 interface here?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -29,16 +45,41 @@ def __init__(
...
+        if self.__plugin_registry:
+            self.__connection_hook_spec = cast(
+                ConnectionHookSpec, self.__plugin_registry.hook()
+            )
+
+            if not self.__plugin_registry.has_plugin(
+                DefaultConnectionHookImpl.__name__
+            ):
+                self.__plugin_registry.load_plugin_with_hooks_impl(
+                    DefaultConnectionHookImpl(), DefaultConnectionHookImpl.__name__
```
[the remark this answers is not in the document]
tozka: I moved it 3 times. 

Initially I initialized it in a plugin but that means ManagedCursor does not stand on its own (it's a dependency because the default implementation is necessary for it to work). 

Then I moved it ManagedConnectionBase. But I felt there also it's not a good place for similar reasons. 

I could have created ManagedCursorFactory and initialized it there and used the factory to created new managed cursors but I thought that's too much changes. And also was not sure if it's really good fit. It's the best option perhaps if not this. 

I prefer to merge it this way now and think of refactoring it later when we have better idea on what would be best.

projects/vdk-core/tests/vdk/internal/builtin_plugins/connection/test_connection_router.py
tozka: The test was wrong - the public .connect() method returns Managed Cursor 

The internal _connect method return a raw cursor, and still it's good idea to test it (since it is used by other internal vdk modules).

projects/vdk-plugins/vdk-test-utils/src/vdk/plugin/test_utils/util_funcs.py
```diff
@@ -212,6 +215,17 @@ def populate_mock_managed_cursor(
             mock_connection_hook_spec.db_connection_decorate_operation
         )
 
+    def stub_db_connection_execute_operation(raw_cursor, managed_operation):
+        if managed_operation.get_parameters():
+            return raw_cursor.execute(
+                managed_operation.get_operation(), managed_operation.get_parameters()
+            )
+        return raw_cursor.execute(managed_operation.get_operation())
```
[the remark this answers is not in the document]
ivakoleva: I think `plugin_registry` needs to be propagated no further than the router at best. An option is to add an abstraction before routing, and invoke it during router initialisation -> outer scope (from job_context), or within the router itself. 
In all cases, I don't see the plugin_registry being field-stored within the router, or propagated further to user-deliverables like connection or cursor.
I think the router initialized should be aware of the relevant ConnectionHookSpec.

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
ivakoleva: raw_cursor exposed is an anti-pattern in a use-case where we strive being on the data path

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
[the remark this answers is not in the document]
tozka: Ok. I fixed it by introducing ConnectionHookSpecFactory which is responsible for providing properly initialized ConnectionHookSpec .

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
tozka: I've created new ExecutionCursor and documented there how it should be used. It has much more clear contract this way.

projects/vdk-core/tests/vdk/internal/builtin_plugins/connection/test_connection_router.py
tozka: The protected method `_connect` returns native (or raw) cursor.
The public method `connect` returns managed cursor.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -29,16 +45,41 @@ def __init__(
         self,
         cursor: Any,
         log: logging.Logger = None,
-        connection_hook_spec: ConnectionHookSpec = None,
+        plugin_registry: IPluginRegistry = None,
```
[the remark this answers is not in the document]
tozka: Fixed

projects/vdk-core/tests/vdk/internal/builtin_plugins/connection/test_connection_router.py
ivakoleva: So what's the reason for that change from a public interface, to invoking a protected native (raw) cursor?

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
```diff
@@ -5,6 +5,10 @@
```
ivakoleva, line 5: you have force-pushed new changes, and yet commits on this PR seem still 1?
let me review the latest changes,

ivakoleva (changes requested):

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
```diff
@@ -68,6 +72,70 @@ def db_connection_decorate_operation(
...
+    ) -> Optional[ExecuteOperationResult]:
+        """
+        The method that executes the actual SQL query using the raw_cursor.
+
+        The default implementation is basically this:
+        .. code-block::
+
+            if parameters:
+                return raw_cursor.execute(operation, parameters)
+            return raw_cursor.execute(operation)
```
ivakoleva, line 87: The `default implementation` mentioned/code-blocked is not linked; don't understand why the implementation specifics that execute the query are important and documented - is it expected to be used for implementing the hooks spec, why, and how? missing any example on that

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
```diff
@@ -68,6 +72,70 @@ def db_connection_decorate_operation(
...
+            @hookimpl(hookwrapper=True)
+            db_connection_execute_operation(execution_cursor: ExecutionCursor) -> Optional[int]:
+                # let's track duration of the query
+                start = time.time()
+                log.info(f"Starting query: {execution_cursor.get_managed_operation().get_operation()}")
+                outcome: pluggy.callers._Result
+                outcome = yield # we yield the execution to other implementations (including default one)
+                is_success: bool = outcome.excinfo is None
+                log.info(f"Query finished. duration: {time.time() - start}. is_success: {is_success}")
+                # no return!
```
ivakoleva, line 109: this is a very complex simple example, how error-prone is that?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/connection_hooks.py
```diff
@@ -0,0 +1,71 @@
...
+from vdk.internal.builtin_plugins.connection.execution_cursor import ExecutionCursor
+from vdk.internal.core.errors import ErrorMessage
+from vdk.internal.core.errors import PlatformServiceError
+
+
+class DefaultConnectionHookImpl:
+    """
+    The default implementation of execute operation.
+    Generally it should not be overridden.
+    To "override" with new implementation, make a new hook which returns non-None ExecuteOperationResult.
```
ivakoleva, line 21: In such a case, does the default implementation remain evaluated, and only the result is augmented?
Having a "default hook implementation" differs majorly from other hook specs so far, what's the benefit here (pros & cons)? What are the alternatives to be compliant with the already established approaches?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/execution_cursor.py
```diff
@@ -0,0 +1,67 @@
...
+        Get the underlying native cursor. Used to actually execute the operation.
+        Generally should not be accessed directly.
+        But it's useful to be exposed in case the native library used provide some extra features.
+        For example Impala (impyla driver) provides a way to get the profile of a query after it's executed
+        which can be printed to aid debugging.
+
+        Check the example usages in docstring of connection_hook_spec#db_connection_execute_operation .
+
+        :return: PEP249Cursor
+            the underlying native cursor being managed.
```
ivakoleva, line 54: what is the use-case for publicly exposing the native cursor? execution cursor should be itself able to handle everything the native does, and more. Why running away from being on the data path design?

projects/vdk-core/tests/functional/run/test_run_sql_queries.py
```diff
@@ -241,3 +244,55 @@ def test_run_job_with_properties_and_sql_substitution_priority_order():
...
+        self.log = []
+
+    @hookimpl(hookwrapper=True)
+    def db_connection_execute_operation(
+        self, execution_cursor: ExecutionCursor
+    ) -> Optional[int]:
+        self.log.append("start")
+        out: pluggy.callers._Result
+        out = yield
+        self.log.append(("end", out.excinfo is None))
```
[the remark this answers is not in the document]
tozka: Because with two hooks you have less control and it's harder to use. 
We'd like to be able to write a plugin that tracks the duration of a query (for example(- the easiest way to do this is to use hookwrapper feature of pluggy (https://pluggy.readthedocs.io/en/stable/#wrappers)

This is not a new pattern. It is used in [run_job](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L145) and [run_step](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L171) hooks. 

See similar example in pytest (which is inspiration for this): https://docs.pytest.org/en/6.2.x/_modules/_pytest/hookspec.html#pytest_cmdline_main

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
```diff
@@ -68,6 +72,70 @@ def db_connection_decorate_operation(
...
+        .. code-block::
+
+            if parameters:
+                return raw_cursor.execute(operation, parameters)
+            return raw_cursor.execute(operation)
+
+
+        Hook executions will stop at first hook implementation that returns non-None result.
+        Hence if you want to override the default implementation return a non-None result.
+        But in most cases you would not need to overwrite the default implementation.
```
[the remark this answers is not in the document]
tozka: How "firstresult" works is also documented here: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/hook_markers.py#L16

I am not introducing a new feature here.

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
tozka: Beucase of the way "firstresult" hooks work - https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/hook_markers.py#L16 

You enable users to override "default" implementation. Writing the default implementation in the doc is for clarity so users know what to expect. It's also inspired of pytest documentation on similar plugins (https://docs.pytest.org/en/6.2.x/_modules/_pytest/hookspec.html#pytest_cmdline_main)

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
tozka: What is so complex here - we track duration of a query ...

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/connection_hooks.py
tozka: This is not a new pattern. It is used in [run_job](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L145) and [run_step](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L171) hooks.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -156,6 +155,14 @@ def _validate_operation(self, operation, parameters):
                     exception=e,
                 )
 
+    def _execute_operation(self, managed_operation: ManagedOperation):
+        self._log.info("Executing query:\n%s" % managed_operation.get_operation())
+        execution_cursor = ExecutionCursor(self._cursor, managed_operation, self._log)
+        result = self.__connection_hook_spec.db_connection_execute_operation(
+            execution_cursor=execution_cursor
```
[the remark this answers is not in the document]
tozka: I disagree. And this is not a new pattern. It is used in [run_job](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L145) and [run_step](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/core_hook_spec.py#L171) hooks.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/execution_cursor.py
tozka: I think I explained the use-cases before: 
 

There are compelling use-cases for access to native/raw cursor by users as it provides extra features. 

Impala (impyla driver) provides a way to get the profile of a query after it's executed - this is something that vdk-impala plugin can use to extract lineage and statistics
Trino driver provides similar way to get information about the query from the raw cursor - again for similar reasons - to extract lineage and stats.

Those are implementation-specific but necessary to achieve feature-rich database plugins.

Those are important use-cases, for example, I want to get extra debug info this with Impala driver:
```
@hookimpl(hookwrapper=True)
            db_connection_execute_operation(execution_cursor: ExecutionCursor) -> Optional[int]:
                yield # let the query execute first
                c = cast(impala.interface.Cursor, execution_cursor.get_native_cursor())
                log.info(f"Query {execution_cursor.get_managed_operation().get_operation()} debug info:"
                         f"summary: {c.get_summary()}, profile: {c.get_profile()}")
```

For now this is the best way I have figured out how to do that. 
I have a another idea - If we reuse the functionality in [ManagedCursor](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py#L39) for  Dynamic interception and delegation , then we may not need get_native_cursor method. 

I will do that in separate PR (after merging this one) though.

projects/vdk-core/tests/vdk/internal/builtin_plugins/connection/test_connection_router.py
tozka: There's no change in interfaces. It's a bug fix in the test. But I will remove it from this PR. I should have opened a separate PR (as it is per [our recommendation](https://github.com/vmware/versatile-data-kit/wiki/How-to-prepare-a-new-PR#break-your-code-commits-into--small-self-contained-units) ) . 

See PR https://github.com/vmware/versatile-data-kit/pull/841 for the change and more details.

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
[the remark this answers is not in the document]
ivakoleva: well, it's still not clear, what is the difference for the workflow - between raising an exception, and returning a result? those are both valid scenarios, and I don't see them described in the api pydocs

projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py
ivakoleva, line 87: wondering, if there is a default implementation pasted within the API spec,
shouldn't it be placed directly within the spec file then, for visibility?
And yet, no implementation is a part of the API, so .. why is the API concerned with any default implementations?

My point is, if any such explanation is needed, maybe the hook(s) itself should be before or after the execute query implementation, depending on when the user wants to intercept the workflow.
A before-query-execution hook is validate & decorate hooks. What was missing, is after-query-execution hook, is that the problem solved by the PR?

ivakoleva, line 87: wondering, if there is a default implementation pasted within the API spec,
shouldn't it be placed directly within the spec file then, for visibility?
And yet, no implementation is a part of the API, so .. why is the API concerned with any default implementations?

My point is, if any such explanation is needed, maybe the hook(s) itself should be before or after the execute query implementation, depending on when the user wants to intercept the workflow.
A before-query-execution hook is validate & decorate hooks. What was missing, is after-query-execution hook, is that the problem solved by the PR?

ivakoleva (approved):
Could be simpler for API usage, yet local context with yielding is possible this way,
we agreed to merge

tozka:
> Could be simpler for API usage

I don't think we have an idea for simpler API usage than this: 

The other possible API could be to have two hooks 

```
db_connection_execute_before_operation(cursor, query)
db_connection_execute_after_operation(cursor, query)
```

But in that case, a simple use case of tracking the duration of query becomes very hard to implement 
Maybe like this

```
class QueryDuratinPlugin:
     queries: Dict # keep state of queries

     def db_connection_execute_before_operation(self, cursor) 
            # we do not have id because query is not yet executed so we have to generate it
            id = generate_id(query) 
            query = update_query_comments_with_id(query)
            self.queries[id] = time.time() 
   
    def db_connection_execute_after_operation(self, cursor) 
            id = parse_id_from_comments(query) # we do not have id because query is not yet executed 
            start = self.queries[id]
            end = time.time()
            log.info("Duration: {end - start}")
   
```

while now with this hook it is: 

```
@hookimpl(hookwrapper=True)
            db_connection_execute_operation(execution_cursor: ExecutionCursor) -> Optional[int]: 
                start = time.time()
                outcome = yield # we yield the execution to other implementations (including default one) 
                log.info(f" duration: {time.time() - start}. ")
```

PR #815, opened 2022-04-26
vdk-control-api-auth: Add core auth logic

doks5:
As part of the work on extracting the authentication logic of vdk-control-cli
and moving it into a stand-alone library, we need to have a generalized
approach to authentication which is compliant with RFC-6749 while being able
to be integrated with the existing components of Versatile Data Kit, which
require authentication.

This change introduces the core authentication logic, which will be used and
extended to facilitate the different authentication workflows (api/refresh token
flow, pkce flow, etc.).

This is the first Pull Request of a series that will build the complete authentication library (plugin). The
plugin will not be released before all work is done.


projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_config.py
```diff
@@ -0,0 +1,157 @@
...
+                    pathlib.Path(self.vdk_config_folder).mkdir(
+                        parents=True, exist_ok=True
+                    )
+                except Exception as e:
+                    raise VDKAuthException(
+                        what="Configuration folder was not created and user is not logged in",
+                        why=f"Cannot create: {self.vdk_config_folder} configuration folder: {str(e)}",
+                        consequence="User won't be able to access Control Service",
+                        countermeasure=f"Check if the client has write access to: {base_dir}",
+                    )
```
gageorgiev, line 110:
```suggestion
                    raise VDKAuthException(
                        what="Configuration folder was not created and user is not logged in",
                        why=f"Cannot create: {self.vdk_config_folder} configuration folder: {str(e)}",
                        consequence="User won't be able to access Control Service",
                        countermeasure=f"Check if the client has write access to: {base_dir}",
                    ) from e
```
Is including the exception string inside the why parameter just as good as raising from the original exception? It might be, and if so you can ignore this suggestion.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
```diff
@@ -0,0 +1,284 @@
...
+                self.__vdk_config.api_token_authorization_url
+            )
+            self.__exchange_api_for_access_token()
+        else:
+            log.debug(
+                "No authentication mechanism found. Will not cache access token."
+                "If Control Service authentication is enabled, API calls will fail."
+            )
+
+    def __exchange_refresh_for_access_token(self):
```
gageorgiev, line 194: nit but maybe move __exchange_refresh_for_access_token up so it's next to __exchange_api_for_access_token

gageorgiev (approved):
Looks solid; it's a big PR so maybe wait for an additional approve before merging in case I've missed something.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
doks5: Moved. Thanks

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_config.py
doks5: I am hesitant to excessively clutter the logs, but if others think the whole exception should be logged, as well, I am fine to go with the suggested change.

projects/vdk-plugins/vdk-control-api-auth/setup.py
```diff
@@ -25,14 +25,9 @@
     "authentication.",
     long_description=pathlib.Path("README.md").read_text(),
     long_description_content_type="text/markdown",
-    install_requires=[],
+    install_requires=["requests"],
```
ivakoleva, line 28: why not reusing vdk-core's configuration mechanism?

projects/vdk-plugins/vdk-control-api-auth/setup.py
```diff
@@ -25,14 +25,9 @@
...
     package_dir={"": "src"},
     packages=setuptools.find_namespace_packages(where="src"),
-    entry_points={
```
ivakoleva, line 31: looking at vdk-control-api-auth README.md, is it meant to be used by other plugins like "library";
maybe imported & explicitly instantiated? like common reusable code? is it concurrent? it is not clear by the README, so I'm just guessing; maybe provide an example snippet in the readme?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_utils.py
```diff
@@ -0,0 +1,39 @@
...
+from configparser import ConfigParser
+from configparser import MissingSectionHeaderError
+
+from vdk.plugin.control_api_auth.auth_exception import VDKAuthException
+
+
+log = logging.getLogger(__name__)
+
+
+def read_config_ini_file(
```
ivakoleva, line 13: what's the intended usage? didn't find any usages (except in test utils, where a duplicate's in usage )

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_request_values.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from enum import Enum
+
+
+class AuthRequestValues(Enum):
```
ivakoleva, line 6: maybe mark with #deprecated the duplicates (e.g. `auth_request_values.py` in vdk-control-cli's auth package), so it's easier to diff current code with prior, and to locate what needs to be cleaned up afterwards?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
```diff
@@ -0,0 +1,284 @@
...
+from requests import post
+from requests.auth import HTTPBasicAuth
+from requests_oauthlib import OAuth2Session
+from vdk.plugin.control_api_auth.auth_config import AuthConfig
+from vdk.plugin.control_api_auth.auth_config import AuthConfigFolder
+from vdk.plugin.control_api_auth.auth_config import InMemAuthConfiguration
+from vdk.plugin.control_api_auth.auth_exception import VDKAuthException
+from vdk.plugin.control_api_auth.auth_request_values import AuthRequestValues
+from vdk.plugin.control_api_auth.login_types import LoginTypes
+
```
ivakoleva, line 18: you mention pkce within the PR description, and I don't find any such related changes?
so `auth_pkce.py` from vdk-control-cli remains?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_request_values.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    An enum used to the common values used in the redirect logic.
```
ivakoleva, line 8: are we extracting also RedirectAuthentication for reusing?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
```diff
@@ -0,0 +1,284 @@
...
+    caching is enabled.
+    """
+
+    def __init__(
+        self,
+        authorization_url=None,
+        refresh_token=None,
+        access_token=None,
+        access_token_expiration_time=0,
+        client_id=None,
```
ivakoleva, line 35: why None in favour of priorly used string placeholder by `auth.py`?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
```diff
@@ -0,0 +1,284 @@
...
+
+    @staticmethod
+    def deserialize(content: str) -> AuthenticationCache:
+        if content:
+            return AuthenticationCache(**json.loads(content))
+        else:
+            return AuthenticationCache()
+
+
+class BaseAuth:
```
ivakoleva, line 94: why `BaseAuth` in favour of the old `Authentication`? is it abstract and meant for extension? don't see any prior extensions;
also, transitioning the imports only is going to be easier

projects/vdk-plugins/vdk-control-api-auth/tests/test_core_auth.py
```diff
@@ -0,0 +1,219 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+"""
```
ivakoleva, line 3: why file named core? is there something in addition to test?

projects/vdk-plugins/vdk-control-api-auth/tests/test_core_auth.py
```diff
@@ -0,0 +1,219 @@
...
+        "token_type": "bearer",
+        "expires_in": 1799,
+        "scope": "csp:support_user",
+        "access_token": "axczfe12casASDCz",
+        "refresh_token": "refresh",
+    }
+    return json_response_mock
+
+
+def allow_oauthlib_insecure_transport():
```
ivakoleva, line 218: why not reusing test utils?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_request_values.py
```diff
@@ -0,0 +1,17 @@
...
+    """
```
ivakoleva, line 9: Do you think it's a good idea to transfer also `ApiKeyAuthentication`, since it does make use of the Authentication class to instantiate it?
It contains the usage workflow; is it concurrent?

ivakoleva (commented):
Since the library plugin is meant to be used by other plugins, maybe add functional tests that do exactly that and verify operability?
for example, access (read) the freshest access token generated;

If I recall correctly, based on the provider specifics - old tokens may be invalidated (after a certain period of time, or even immediately) - on token renewal; 
so multiprocessing for scalability has to do with concurrent Authentication support.

projects/vdk-plugins/vdk-control-api-auth/setup.py
doks5: To not rely on vdk-core and its dependencies. I am hoping for the library to be as light as possible, as we will use it for the airflow provider.

projects/vdk-plugins/vdk-control-api-auth/setup.py
doks5: Yes, it is a library, I'll update the README with subsequent PRs. This is only the first PR, and currently the package is not yet being released.
It is not currently concurrent, do we have a requirement for libraries to be concurrent?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_exception.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+
+class VDKAuthException(Exception):
```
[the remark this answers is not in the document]
doks5: The vdk-control-cli's VdkException extends `click.UsageError` in order to allow click to properly display the error message in terminal.
I did not re-use the exceptions from vdk-core/vdk-control-cli to avoid circular dependencies.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_utils.py
doks5: Oopsie, good catch. This one was supposed to be part of a separate PR. Removed.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_request_values.py
doks5: Will do as part of a separate PR.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
doks5: Just updated the PR description. The pkce authentication flow will be added as part of a separate PR. I didn't want to put everything in a single PR, as the change would be really hard to review properly, so I opted for a multi-PR change.
The release step in the CI/CD will not be enabled until the work on the library is complete.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_request_values.py
doks5: Yes, we will. See https://github.com/vmware/versatile-data-kit/pull/815#discussion_r860761342

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
doks5: To get rid of the codacy warning messages.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/base_auth.py
doks5: There will be a different `Authentication` api with more generalized logic.

projects/vdk-plugins/vdk-control-api-auth/tests/test_core_auth.py
doks5: Yes, there will be additional tests for the different authentication flows.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_request_values.py
doks5: Yes, `ApiKeyAuthentication` will also be transferred, alongside the other authentication flows, as separate PRs to make the changes easier to review.

projects/vdk-plugins/vdk-control-api-auth/setup.py
ivakoleva: the airflow provider is not vdk-core-dependant?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_exception.py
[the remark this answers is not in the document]
doks5: Not exactly. There will be a few more PRs for the different authentication flows. Then I'll enable the release step of the librry, and only after that there will be a PR with the vdk-control-cli integration with this library.

projects/vdk-plugins/vdk-control-api-auth/setup.py
doks5: Not currently, https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/airflow-provider-vdk/setup.py#L28
Not sure if there are plans to have vdk-core as a dependency.

projects/vdk-plugins/vdk-control-api-auth/.plugin-ci.yml
```diff
@@ -3,29 +3,60 @@
...
-#build-py310-vdk-control-api-auth:
-#  extends: .build-vdk-control-api-auth
-#  image: "python:3.10"
+.retry:
+  retry_options:
+    max: 1
+    when:
+      - always
+
+.build-vdk-control-api-auth:
```
tozka, line 12: Why not reuse .build-plugin from plugin-common ?

tozka (approved):

projects/vdk-plugins/vdk-control-api-auth/requirements.txt
```diff
@@ -1,4 +1,5 @@
 pytest
-vdk-core
-
+pytest-httpserver
+requests
+requests_oauthlib
```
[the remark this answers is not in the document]
doks5: Good catch. Will do

projects/vdk-plugins/vdk-control-api-auth/.plugin-ci.yml
doks5: @tozka Updated the main plugin CI/CD script to ignore the entry-point of the plugin, similar to how `quickstart-vdk` and `vdk-test-utils` are handled. We may need to think about a different way to handle such plugins, though, as this does not seem manageable in the long run.

tozka: Yes, I agree, it's not ideal. 2 options come to mind: 

A) We remove the check. If we think what the check is doing (testing the "wiring") is covered by other tests . 
B) Make the check option but reverse - that is expose variable (CHECK_VDK_PLUGIN_ENTRYPOINT=false/true) and those 3 plugins/libs would set it to false.

PR #817, opened 2022-04-27
vdk-core: encapsulate router-specific properties logic

ivakoleva:
This MR attempts to simplify outer usages of PropertiesRouter,
in equivalent to IngesterRouter. The reason is IngesterRouter does
support multiple ingesters evaluated in a sequence, and PropertiesRouter
supports singleton properties back-end implementation yet.

PropertiesRouter interface surface no more exposes a singleton
implementation, instead implements the IProperties interface and
does the routing internally.
Also IngesterRouter already implements IIngester methods, so elaborated
the class definition on that explicitly, so it is effectively coupled
with the interfaces evolution.

Testing Done: ci/cd. No functional changes made, except interface
surface improvements and encapsulation.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/properties_router.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/properties_router.py
@@ -1,7 +1,9 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import logging
+from typing import Any
 
+from vdk.api.job_input import IProperties
 from vdk.api.plugin.plugin_input import IPropertiesFactory
 from vdk.api.plugin.plugin_input import IPropertiesRegistry
 from vdk.internal.builtin_plugins.config.job_config import JobConfigKeys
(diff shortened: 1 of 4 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/properties_router.py
```diff
@@ -42,7 +44,26 @@ def set_properties_factory_method(
...
+    def get_all_properties(self) -> dict:
+        return self.__get_properties_impl().get_all_properties()
+
+    def set_all_properties(self, properties: dict):
+        self.__get_properties_impl().set_all_properties(properties)
+
+    def has_properties_impl(self):
+        """
+        If any properties implementation backend available
+        :return: bool
```
doks5, line 59: nit: Docstring mentions return type. Maybe update method signature?

doks5 (approved):
LGTM

ivakoleva:
> The change looks good to me.
> 
> But I am wondering what prompted the change now?

A prerequisite for multi properties client invocation support, like mentioned in the description,
see https://github.com/vmware/versatile-data-kit/pull/819

PR #818, opened 2022-04-27
airflow-provider-vdk: VDKSensor poke method

gageorgiev:
This change implements the poke method of the VDKSensor,
which checks the status of a particular job execution, returns
True if the execution is successful, false if it is not yet complete,
and raises an exception if the job execution has been cancelled,
skipped or has failed.

Also added an exception class for exceptions raised by the poke
method, and dropped vdk-core as a dependency as it is not needed.

Testing done: added 7 unit tests for each status a job execution
can have


projects/vdk-plugins/airflow-provider-vdk/vdk_provider/sensors/vdk.py
```diff
@@ -45,4 +50,21 @@ def poke(self, context: Dict[Any, Any]) -> bool:
...
+            raise VDKJobExecutionException(
+                f"Job execution {self.job_execution_id} has been {job_status}."
+            )
+        elif job_status == "user_error" or job_status == "platform_error":
+            raise VDKJobExecutionException(
+                f"Job execution {self.job_execution_id} has failed due to a {job_status.replace('_', ' ')}. "
+                f"Check job execution logs for more information."
+            )
+
+        return job_status == "succeeded"
```
mivanov1988, line 70: In case of success, we should print execution details (version, logsUrl, etc.) as well as the execution log.

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/sensors/vdk.py
```diff
@@ -45,4 +50,30 @@ def poke(self, context: Dict[Any, Any]) -> bool:
...
+        vdk_hook = VDKHook(self.conn_id, self.job_name, self.team_name)
+
+        job_execution = vdk_hook.get_job_execution_status(self.job_execution_id)
+        job_status = job_execution.status
+
+        log.info(
+            f"Current status of job execution {self.job_execution_id} is: {job_status}."
+        )
+
+        if job_status == "succeeded":
```
mivanov1988, line 62: I think we should print the execution log event if the job fails.

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/sensors/vdk.py
```diff
@@ -45,4 +50,30 @@ def poke(self, context: Dict[Any, Any]) -> bool:
...
+            log.info(f"Job status: {job_execution}")
+            log.info(
+                f"Job logs: {vdk_hook.get_job_execution_log(self.job_execution_id)}"
+            )
+
+            return True
+        elif job_status == "cancelled" or job_status == "skipped":
```
mivanov1988, line 69: Can we get these statuses from vdk-control-service-api as constants?

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/sensors/vdk.py
gageorgiev: Makes sense.

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/sensors/vdk.py
gageorgiev: I tried to but this is how they're listed in the DataJobExecution object:
```
@status.setter
    def status(self, status):
        """Sets the status of this DataJobExecution.

        The current status  # noqa: E501

        :param status: The status of this DataJobExecution.  # noqa: E501
        :type: str
        """
        allowed_values = ["submitted", "running", "succeeded", "cancelled", "skipped", "user_error", "platform_error"]  # noqa: E501
        if self.local_vars_configuration.client_side_validation and status not in allowed_values:  # noqa: E501
            raise ValueError(
                "Invalid value for `status` ({0}), must be one of {1}"  # noqa: E501
                .format(status, allowed_values)
            )

        self._status = status
```

mivanov1988: Ok, let's define them in the hook in order to reuse them in both operators.

gageorgiev: noted

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/sensors/vdk.py
```diff
@@ -45,4 +51,35 @@ def poke(self, context: Dict[Any, Any]) -> bool:
...
+            log.info(
+                f"Job logs: {vdk_hook.get_job_execution_log(self.job_execution_id)}"
+            )
+
+            return True
+        elif job_status == JobStatus.CANCELLED or job_status == JobStatus.SKIPPED:
+            raise VDKJobExecutionException(
+                f"Job execution {self.job_execution_id} has been {job_status}."
+            )
+        elif (
```
tozka, line 74: What if a new status is introduced in the future?

PR #819, opened 2022-04-27
vdk-core: properties write pre-processing support

ivakoleva:
This change introduces `IPropertiesServiceClient` `write_properties`
chaining, enabling use-cases like decorate/curate properties
prior write, parallel properties storage, and so on.
In addition, autonomous `IPropertiesServiceClient` write properties may
now be invoked in a sequence.

A backwards-compatible API change is introduced, so now
`write_properties` does return the properties written, that unblocks
the chaining abstraction. Obsolete method signatures would be already
updated. Pydocs updated.
Preprocessors configured in `PROPERTIES_WRITE_PREPROCESS_SEQUENCE` are
verified and available before any properties writing initiated.
`Propertiesnotavailable.py` file renamed, unused imports cleaned;
a few typos fixed in ingestion processing user-facing documentation;

Testing Done: introduced 3 properties preprocessing functional tests,
covering success/error/misconfiguration cases.


projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
```diff
@@ -63,8 +69,29 @@ def get_all_properties(self) -> Dict[str, PropertyValue]:
 
     def set_all_properties(self, properties: Dict[str, PropertyValue]) -> None:
         """
-        Persists the passed properties overwriting all previous properties
+        Invokes the write pre-processors if any such configured.
+        Persists the passed properties overwriting all previous properties.
         """
+        if self._write_preprocessors:
+            properties = deepcopy(
```
tozka, line 76: Why not name this variable properties_copy ? Having 2 variables with the same name but different scope is confusing .

tozka (commented):
As it's a introducing a new user-facing options, we should include a functional test under https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-core/tests/functional/run to test the PROPERTIES_WRITE_PREPROCESS_SEQUENCE with vdk run

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
```diff
@@ -1,17 +1,21 @@
...
 from typing import Dict
+from typing import List
+from typing import Optional
 from typing import Union
 
 from vdk.api.job_input import IProperties
 from vdk.api.plugin.plugin_input import IPropertiesServiceClient
 
+from ...core.errors import log_and_throw
+from ...core.errors import ResolvableBy
```
doks5, line 14: Why dots `...` instead of full path?

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
```diff
@@ -63,8 +69,29 @@ def get_all_properties(self) -> Dict[str, PropertyValue]:
...
+                properties
+            )  # keeps the outer scope originally-referenced dictionary intact
+            for client in self._write_preprocessors:
+                try:
+                    properties = client.write_properties(
+                        self._job_name, self._team_name, properties
+                    )
+                except Exception as e:
+                    log_and_throw(
+                        to_be_fixed_by=ResolvableBy.USER_ERROR,
```
doks5, line 86: Are we sure this would always be a `User Error`? If something happens due to an issue with a plugin or a backend, this may not be in user's control.

doks5 (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
ivakoleva: Because in case of potential preprocessing modification, we don't want to change user's outer scope `set_all_properties()` parameter, the value modified is expected upon return.
I've added a unit test to cover/visualize this scenario

ivakoleva: In details - not introducing a new variable, because I want to dereference the param with a deep copy if we're about to modify it, then proceed further with handling;

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
ivakoleva: also spotted that, it's due IDE auto-import; didn't deep-dive into details and presumed it's some syntax sugar.
Do you have something in mind?

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
ivakoleva: Properties write preprocessors are designed for user-specific customisations. I believe the case is the same with ingestion preprocessors, where error-handling approach matches

ivakoleva:
> As it's a introducing a new user-facing options, we should include a functional test under https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-core/tests/functional/run to test the PROPERTIES_WRITE_PREPROCESS_SEQUENCE with vdk run

the functional test is implemented; should I comment it out however, and create a PR with vdk-test-utils changes to merge first?

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
ivakoleva: in case we introduce platform-level properties/ingestion preprocessors (excluding vdk-test-utils for test purposes), 
we may add in both - an additional except cases for PlatformError raised to simply rethrow it

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
doks5: Ok.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
doks5: Not sure if the last comment is for this thread.

doks5 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
tozka: Maybe right more detailed comment "keeps the outer scope originally-referenced dictionary intact" was not enough for me to get it.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
tozka: The dots indicate relative import. 

Per PEP 8 (https://peps.python.org/pep-0008/) Absolute imports are recommended over relative imports.

tozka (approved):
Looks good to me. I made some small suggestions that you should look over but I am ok if you decide not to change anything.

PR #822, opened 2022-05-03
[vdk-plugins] vdk-control-api-auth: Add api-token flow

doks5:
As part of the ongoing work to extract the authentication login of
Versatile Data Kit into a stand-alone library and make it available
to all components (core and plugins) to use, we need to ensure that
the interface surface is as generic as possible.

This change introduces an`Authentication` class, which will act as
the entry point for using the library and authenticating.

Additionally, the api token authentication flow from vdk-control-cli
is also adapted and added to vdk-control-api-auth.

Testing Done: Unit tests.

### DIFF
--- a/projects/vdk-plugins/vdk-control-api-auth/README.md
+++ b/projects/vdk-plugins/vdk-control-api-auth/README.md
@@ -6,7 +6,29 @@ utilities used by vdk-control-cli and other plugins.
 ## Usage
 
 This is a library plugin, not a runnable plugin, and it is intended to be
-used as a dependency for other plugins which need to authenticate users.
+used as a dependency for other plugins, which need to authenticate users.
+
+To use the library within a plugin or another Versatile Data Kit component,
+just import the `Authentication` class, and create an instance of it. The
+different authentication flows require different parameters to be specified.
+
+Once everything is done, in order to authenticate, call `.authenticate()` on
+the `Authentication` instance.
+
+Example Usage:
+```python
+from vdk.plugin.control_api_auth.authentication import Authentication
+
+auth = Authentication(
+    token="<oauth-api-token>",
+    authorization_url="https://some-authorization-endpoint",
+    auth_type="api-token",
+)
+
+auth.authenticate()   # authenticate
+
+auth.read_access_token()   # fetch the cached access token
+```
 
 ## Build and testing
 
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -0,0 +1,93 @@
...
+            The URL which exchanges token for access token.
+        :param auth_type:
+            What type of authentication should be used (e.g., refresh_token,
+            basic_auth, etc.).
+        :param cache_locally:
+            A flag, indicating if credentials should be cached locally (in a
+            file).
+        """
+        self._username = username
+        self._password = password
```
gageorgiev, line 46: Are we missing an authentication method? These fields aren't used anywhere; maybe add an empty method to signify that there's code to be added.

projects/vdk-plugins/vdk-control-api-auth/README.md
```diff
@@ -6,7 +6,27 @@ utilities used by vdk-control-cli and other plugins.
...
+```python
+from vdk.plugin.control_api_auth.authentication import Authentication
+
+auth = Authentication(
+    token="<oauth-api-token>",
+    authorization_url="https://some-authorization-endpoint",
+    auth_type="api-token",
+)
+
+auth.authenticate()
```
tozka, line 28: How do I get the access token ?

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
doks5: There will be additional PR-s with the other authentication workflows. I've added a placeholder method.

projects/vdk-plugins/vdk-control-api-auth/README.md
doks5: Good question. I've added a method to get the access token, though I wonder if we could just return the access token as part of the `.authenticate` call?

tozka: It's better to be separate and explicit call (get oauth2 access token). As we'd have another type of authentication mechanism working at teh same time. As part of https://github.com/vmware/versatile-data-kit/pull/809/ I will be adding support for kebreros where authentication is using "Authentication: Negotiate" header. So probably I can add get_kerberos_auth_header method ...

doks5: I see, ok.

tozka (approved):

PR #827, opened 2022-05-09
vdk-control-cli: Make schedule_cron config param optional

gageorgiev:
This change allows for jobs to be deployed without
the schedule_cron parameter specified.

Testing done: deployed job and verified warning is printed


projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
```diff
@@ -124,7 +124,9 @@ def __update_data_job_deploy_configuration(
     ) -> None:
         job: DataJob = self.__read_data_job(name, team)
         local_config = JobConfig(job_path)
-        schedule = self.__check_value("schedule_cron", local_config.get_schedule_cron())
+        schedule = (
```
tozka, line 127: I think we agreed we will do that part of the Control Service on in Control CLI. Because API users (e.g UI users) would have different experience.

gageorgiev: Yes, but:
1. I thought the redundancy is beneficial;
2. I haven't dug deep enough to say conclusively but I think the control-service can already handle jobs with no cron_schedule, see https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/ToModelApiConverter.java#L42

I might be wrong on both points though so we can scrap this change in such a case.

tozka: I am not sure about 1. It seems more of a duplication of logic violating DRY principle.

 If we decide to change the how we handle missing cron in Control Service , we'd mask the change by passing specific cron in vdk-control-cli unless we change it here as well (hence DRY is violated)

2. That seem irrelevant for this change. But that would be super if true - less work . But we need to test it (and have a automated test in Control Service tests that verifies this use case).

doks5 (approved):
LGTM

PR #833, opened 2022-05-10
vdk-control-cli: Validate cron schedule before deploying job

gageorgiev:
This change ensures a job's cron schedule is a valid cron
expression before deploying it to the control-service, and
prints an appropriate message if it is not.

Testing done: attempted to deploy a job with a malformed
cron schedule and observed error; verified a job with a
correct cron schedule can still be deployed

### DIFF
--- a/projects/vdk-control-cli/setup.cfg
+++ b/projects/vdk-control-cli/setup.cfg
@@ -54,6 +54,7 @@ install_requires =
     tabulate
     requests_oauthlib>=1.0
     urllib3>=1.26.5
+    croniter==1.3.4
 
 # Require a specific Python version
 python_requires = >=3.7, <3.11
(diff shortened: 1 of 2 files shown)


projects/vdk-control-cli/setup.cfg
```diff
@@ -54,6 +54,7 @@ install_requires =
     tabulate
     requests_oauthlib>=1.0
     urllib3>=1.26.5
+    croniter==1.3.4
```
tozka, line 57: Do we need to pin the version ?

tozka: Yes, It's a good practice to pin versions in Command-Line interface - because 
A) it's more secure as you mention 
b) it's more stable - CLI won't break if dependency library releases a new version with breaking change. 

It's less of a good practice in a library because it restricts what the users of the library can do - users would want to install their own dependencies and likely will use newer version and will conflict with pinned one.

While vdk-control-cli is a library it's often packaged alongside vdk-core and plgins (which are libraries) and is used in same environment so I prefer to leave them unpinned for now.

gageorgiev:
Closing this PR as as we agreed the cron validation should happen in the control-service.

mivanov1988:
Do we have a separate issue for validation on the API level?

PR #834, opened 2022-05-12
[vdk-plugins] vdk-control-api-auth: Add authorization code flow

doks5:
As part of the ongoing work to build a stand-alone authentication plugin
library, which is to be used by all components of Versatile Data Kit that
rely on authentication, we need to move and generalize the existing logic
from vdk-control-cli.

This change introduces the authorization code flow as defined in
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1 The implementation
of this flow for the purposes of Versatile Data Kit also relies on the use
of Proof Key for Code Exchange or pkce.

Testing Done: Unit tests.

### DIFF
--- a/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
+++ b/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from vdk.plugin.control_api_auth.auth_config import InMemAuthConfiguration
 from vdk.plugin.control_api_auth.auth_exception import VDKAuthException
+from vdk.plugin.control_api_auth.autorization_code_auth import RedirectAuthentication
 from vdk.plugin.control_api_auth.base_auth import BaseAuth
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -101,6 +108,21 @@ def __authenticate_with_authorization_code(self):
...
+        This type of authentication relies on pkce
+        https://datatracker.ietf.org/doc/html/rfc7636
         :return:
         """
-        raise NotImplementedError("This method is yet to be implemented.")
+        if not self._client_id or not self._auth_discovery_url:
+            print(
+                f"Login type: {self._auth_type} requires arguments: "
+                f"client-secret, client-id, auth-discovery-url"
+            )
```
mivanov1988, line 119: shouldn't we throw the exception in case of missing properties?

doks5: Indeed. Changed to throw exception, instead of just printing a message. Thanks for the review

PR #840, opened 2022-05-19
vdk-control-cli: Adopt new vdk-control-api-auth library

doks5:
With the release of the vdk-control-api-auth library, we now have a
stand-alone authentication package that is to be used by all Versatile
Data Kit components that need to authenticate against a third-party
service.

This change adopts the new authentication package in vdk-control-cli,
and is now used when issuing `vdk login` commands.

Additionally, authentication logic in vdk-control-cli which is overlapping
with such logic in vdk-control-api-auth is removed.

Testing Done: Installed the new changes locally and ran `vdk login` to verify
that the browser redirecs work as expected. Also, ran a docker container with
the latest changes to verify that authentication is possible even when there
is no access to a browser.


gageorgiev (approved):

projects/vdk-control-cli/src/vdk/internal/control/command_groups/login_group/login.py
```diff
@@ -112,17 +110,25 @@ def login(
...
             )
-        else:
-            redirect_auth = RedirectAuthentication(
-                client_id, client_secret, oauth2_discovery_url, oauth2_exchange_url
-            )
-            redirect_auth.authentication_process()
+            auth.authenticate()
             click.echo("Login Successful")
+        except VDKAuthException as e:
+            if "requires client_id and auth_discovery_url to be specified" in e.message:
```
ivakoleva, line 125: looking for test coverage this exception+message is raised in case of `not client_id or not oauth2_discovery_url or not oauth2_exchange_url`

ivakoleva (approved):

projects/vdk-control-cli/src/vdk/internal/control/command_groups/login_group/login.py
doks5: Good point. Test added. Thanks

projects/vdk-control-cli/src/vdk/internal/control/command_groups/login_group/login.py
```diff
@@ -112,17 +110,28 @@ def login(
...
             )
-        else:
-            redirect_auth = RedirectAuthentication(
-                client_id, client_secret, oauth2_discovery_url, oauth2_exchange_url
-            )
-            redirect_auth.authentication_process()
+            auth.authenticate()
             click.echo("Login Successful")
+        except VDKAuthException as e:
+            if (
```
tozka, line 125: Parsing error messages  is not a good practice. You should not look at what we do in Impala as a good example. We have no control over what Expcetion Impala raises so we have to parse them there. We do within vdk-control-api-auth library. We should use error codes or enums or exception hierarchy to handle this .

projects/vdk-control-cli/src/vdk/internal/control/command_groups/login_group/login.py
```diff
@@ -134,8 +143,16 @@ def login(
...
+            apikey_auth = Authentication(
+                authorization_url=api_token_authorization_url,
+                token=api_token,
+                auth_type=auth_type,
+                cache_locally=True,
+            )
+            try:
+                apikey_auth.authenticate()
+            except VDKAuthException as e:
+                click.echo(f"Login failed. Error was: {e.message}")
```
tozka, line 155: That does not seem correct. If auth fails we should exit with non 0 exit code. This will only print the failure and will exit with 0 . Why not let the exception propagate or wrap it in [UserError](https://click.palletsprojects.com/en/7.x/api/#click.UsageError)

doks5: I agree, though I believe we didn't exit with non-0 status before either.

doks5: I opted out of propagating the exception as that would be displayed in the user's terminal, and I was not sure if this would be good UX. Also, we have never showed exception messages in terminal for failed login, so doing it now seems like a regression.

PR #841, opened 2022-05-21
vdk-core: fix buggy (false positive) connection unit test

tozka:
We have a unit test that checks the raw (or native) connection is being
saved properly - but the test was buggy since they are qual in a sense
they have the same attribute values (as there's only one "is_valid") .
But we care if it's the same instance. So we switch to "is"

Testing Done: After changing == to "is" ran the test and it failed (as
it should) and then fix the test.


gageorgiev:
The PR desc mentions fixing the test, but I only see replacement of `==` with `is` in the changes. Am I missing something?

tozka:
> The PR desc mentions fixing the test, but I only see replacement of `==` with `is` in the changes. Am I missing something?

No, this is the fix in the test. It was false positive (it was passing even though it should not have) because native_conn.__eq__(managed_conn) (native_conn == managed_conn. Instead using "is" operator now would return true only if both vars refer to the same object

gageorgiev (approved):

PR #844, opened 2022-05-25
versatile-data-kit: fix description

tozka:
It was using internal (unknown to most people) terms. 


projects/vdk-core/src/vdk/api/data_job.py
```diff
@@ -20,7 +20,7 @@ class IStandaloneDataJob:
     Sample usage::
 
         with StandaloneDataJobFactory.create(datajob_directory) as job_input:
-            #... use job_input object to interact with SuperCollider
+            #... use job_input object to intreact with underlying data systems.
```
gageorgiev, line 23:
```suggestion
            #... use job_input object to interact with underlying data systems.
```

gageorgiev (approved):

PR #853, opened 2022-06-01
Person/zverulacis/vdk documentation

zverulacis:


README.md
```diff
@@ -21,25 +21,18 @@
...
-Everything exposed by Versatile Data Kit provides built-in monitoring, troubleshooting,
-and smart notification capabilities. For example, tracking both code and data modifications and the relations between them
-enables engineers to troubleshoot more quickly and provides an easy revert to a stable version.
-
-Versatile Data Kit consists of:
-
-* Control Service which enables creating, deploying, managing and executing Data Jobs in a Kubernetes runtime environment.
-  It offers multitenancy support, SSO, Access Control and auditing capabilities. It exposes CLI.
-* A development Kit to develop, test and run Data Jobs on your machine. It comes with common functionality for data ingestion and processing.
+* A **Data SDK**, which provides all tools for data extraction, transformation and loading, as well as a plugin framework, which permits to extend the framework according to specific requirements of data application.
```
gageorgiev, line 28:
```suggestion
* A **Data SDK**, which provides all tools for the automation of data extraction, transformation and loading, as well as a plugin framework which allows users to extend the framework according to their specific requirements.
```

`data application` feels like a very specific term which we don't define anywhere (and shouldn't) so I think it's better to avoid it.

gageorgiev (commented):
Small notes; I haven't approved it since I'd like someone more senior to take a look as well.

tozka (approved):
Looks good to me. Merge it after addressing Gabriel's comments.

I think we need a picture with perhaps a marketecture of VDK. What do you think?

tozka:
I will merge it on your behalf.

PR #856, opened 2022-06-07
[vdk-plugins] airflow-provider-vdk: Adopt auth plugin

doks5:
After the release of the `vdk-control-api-auth` plugin, which acts as
the de-facto authentication package for Versatile Data Kit, we no longer
need to depend on `vdk-control-cli` and all its unnecessary dependencies.

With this change, we adopt the `vdk-control-api-auth` package and its
authentication capabilities, and drop any dependency on `vdk-control-cli`.

Testing Done: Unit tests.

### DIFF
--- a/projects/vdk-plugins/airflow-provider-vdk/vdk_provider/hooks/vdk.py
+++ b/projects/vdk-plugins/airflow-provider-vdk/vdk_provider/hooks/vdk.py
@@ -2,12 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 import logging
 import os
+import sys
 import time
 import uuid
 from enum import Enum
+from typing import Any
+from typing import Dict
 from typing import Optional
 
 from airflow.exceptions import AirflowException
+from airflow.models import Connection
 from airflow.providers.http.hooks.http import HttpHook
 from taurus_datajob_api import ApiClient
 from taurus_datajob_api import Configuration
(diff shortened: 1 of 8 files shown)


tozka (approved):

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/hooks/vdk.py
```diff
@@ -41,18 +50,29 @@ class JobStatus(str, Enum):
 
 
 class VDKHook(HttpHook):
+    """
+    Interact with Verstile Data Kit (VDK) Control Service
+    """
+
+    conn_name_attr = "vdk_conn_id"
+    default_conn_name = "vdk_default"
+    conn_type = "vdk"
```
mivanov1988, line 59: isn't it 'http'?

mivanov1988 (approved):

projects/vdk-plugins/airflow-provider-vdk/vdk_provider/hooks/vdk.py
doks5: The HttpHook has `conn_type="http"` https://github.com/apache/airflow/blob/047a6162b0b4cbf07fe2fd978e335839a7d3900b/airflow/providers/http/hooks/http.py#L41, so it didn't make sense to me to assign the same connection type.

PR #860, opened 2022-06-09
control-service: Atomic job cancellation

gageorgiev:
Currently, there is a possibility for a job to be set to be cancelled,
and have it complete before the cancellation can complete. This
leads to a 500 error with no explanation.

This change introduces an additional check that the cancellation does
not raise a NullPointerException, and if it does, it logs an appropriate
message and raises and appropriate Exception.

Testing done: unit tests


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -11,9 +11,7 @@
 import com.google.common.collect.Iterables;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.reflect.TypeToken;
-import com.vmware.taurus.exception.JsonDissectException;
-import com.vmware.taurus.exception.KubernetesException;
-import com.vmware.taurus.exception.KubernetesJobDefinitionException;
+import com.vmware.taurus.exception.*;
```
tozka, line 14: Yes, let's not use wildcard imports. 

If you recall we agree to use Google Java Style 
See https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

I have not yet figured out how to automated properly so https://github.com/vmware/versatile-data-kit/pull/757 remains open :(

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -639,10 +637,16 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
...
-            //Status of the operation. One of: "Success" or "Failure"
-            if (operationResponse.getStatus().equals("Failure")) {
-                log.warn("Failed to delete K8S job. Reason: {} Details: {}", operationResponse.getReason(), operationResponse.getDetails().toString());
-                throw new ApiException(operationResponse.getCode(), operationResponse.getMessage());
+
+            try {
+                //Status of the operation. One of: "Success" or "Failure"
+                if (operationResponse.getStatus().equals("Failure")) {
+                    log.warn("Failed to delete K8S job. Reason: {} Details: {}", operationResponse.getReason(), operationResponse.getDetails().toString());
+                    throw new ApiException(operationResponse.getCode(), operationResponse.getMessage());
```
tozka, line 645: We prefer KubernetsException because it's more user friendly error message

tozka:
LGTM as a draft.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
gageorgiev: The wildcard import was included by the pre-commit hook - that or Intellij IDEA's formatter since I haven't manually changed the imports. I assumed it's okay if it was done by an automatic formatter but I'll change it back.

mivanov1988 (approved):

PR #861, opened 2022-06-13
control-service: improve api message on oom job execution errors

mrMoZ1:
why: Recently users of control-service have been complaining that jobs
which fail with User Error and have failed with an OOM error didn't
have a sufficiently clear enough message that classifies such types of errors.

what: Changed internal implementation so that when a data job fails due
to K8S OOM errors the message stored in the database contains a hint as
to the cause of failure - out of memory issues. We now explicitly check the
container termination message and in case of User Errors with an OOM error
we set an appropriate message.

testing: Refactored existing tests. Added new unit tests.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMonitor.java
```diff
@@ -195,7 +196,7 @@ void recordJobExecutionStatus(KubernetesService.JobExecution jobStatus) {
         final DataJob dataJob = dataJobOptional.get();
 
         // Update the job execution and the last execution state
-        jobExecutionService.updateJobExecution(dataJob, jobStatus, executionResult)
+        jobExecutionService.updateJobExecution(dataJob, jobStatus, executionResult, containerTerminationReason)
```
mivanov1988, line 199: You don't need an additional parameter since we already have a parameter jobStatus that contains the termination reason.

```suggestion
        jobExecutionService.updateJobExecution(dataJob, jobStatus, executionResult)
```

mivanov1988 (approved):

PR #863, opened 2022-06-14
control-service: fix graphql team filter not retrieveing special chars

mrMoZ1:
why: users of control-service complained that graphQL API calls
for a team with special characters such as '(' don't yeild results
even though the team exists.

what: investigating the issue concluded that the team filter strategy
uses a regex method - String.matches("regex") string to
match the teams and would get the results tripped by special chars.

testing: added unit test.


gageorgiev:
This entire issue begs the question, should special characters even be allowed in team names? IMO team names should be only comprised of lowercase letters and dashes, or more generally letters and a separator. Maybe we should discuss this more thoroughly.

tozka (approved):

PR #864, opened 2022-06-14
vdk-core: Fix destination_table referenced early

doks5:
In the IngesterBase class, we currently pass `destination_table` to the
call to `_queue_payload_for_posting` in cases when the objects queue is empty.
However, at this point, the destination_table is not defined, which causes an
UnboundLocalError to be raised.

This change fixes the issue by replacing `destination_table` with `current_destination_table`
(set to None by default) which is defined before the call to `_queue_payload_for_posting`.

Testing Done: All existing tests pass


gageorgiev (approved):

tozka:
Don't forget tests, please. All bug fixes should generally have a test covering them (e.i. the test fails before the fix and passes afterwards). This way we make sure that the bug is properly fixed, that we won't regress again in the future (as bugs tend to resurface).

PR #867, opened 2022-06-21
vdk-core: Split execution summary into chunks

doks5:
At the end of a data job execution, we log execution summary, which describes the
python and sql steps that have been executed, their status, etc. Sometimes,
when a job has a large number of steps, the execution summary will be logged as a
single large string message.

This is not an issue in most cases. However, on some operating systems, there are
limits as to how large a logging message can be when printed to terminal, or send
through a socket, and in such situations jobs fail with `OSError: [Errno 40] Message too long`
errors triggered by the logging module.

This change checks if the message is too long, and if that is the case, splits it into
chunks. The chunks are then logged as separate messages to avoid OS-specific errors
related to logging.

Testing Done: Locally ran a data job with a large number of steps and verified that
the execution summary is split into multiple log messages.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
@@ -2,10 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 import json
 import logging
+import math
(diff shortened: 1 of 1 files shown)


gageorgiev (approved):
lgtm

PR #873, opened 2022-06-22
control-service: add default jwt jwk uri

mrMoZ1:
what: Added a default value for one of our properties.

why: Recently we discovered that if this variable is not set
will prevent the control-service from starting up. This was
manifested on a manual local deploy.

testing: trivial.

Signed-off-by: mrMoZ1 <mzhivkov@vmware.com>


gageorgiev (approved):

gageorgiev:
Is this sufficient to resolve the vdk-server installation issue?

PR #874, opened 2022-06-22
vdk-core: detect non vdk_ prefixed environment values for config

tozka:
Currently user need to set VDK_ + config name as environment variable so
VDK  can recognize it. The idea was to avoid conflicts with similarly
named env variables. But this proved to be very error prone where users
tend to forget or get confused on where vdk_ prefix is needed and where
it is not.

This is solved by allowing non-prefixed env variable to be set.

For for configuration "team"

Now if "VDK_TEAM" is set as environment variable is set it is used (as
before)
If it is not set but "TEAM" is set as env variable, it is now used.

Prefixed variables do have higher priority.

Testing Done: unit test


projects/vdk-core/tests/vdk/internal/builtin_plugins/config/test_vdk_config.py
```diff
@@ -0,0 +1,55 @@
...
+
+    assert configuration_builder.build().get_value("option.with.dots") == "value1"
+    assert configuration_builder.build().get_value("no.prefix.option.dots") == "value2"
+
+
+@mock.patch.dict(
+    os.environ,
+    {"VDK_OPTION_WITH_DASH": "value1", "no_prefix_option_dash": "value2"},
+)
+def test_vdk_config_env_variable_dots_replaced():
```
tozka, line 46:
```suggestion
def test_vdk_config_env_variable_dash_replaced():
```

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -134,10 +137,17 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
         )
         upper_cased_env = {k.upper(): v for k, v in os.environ.items()}
         for key in config_keys:
-            normalized_key = VDK_ + key.replace(".", "_").upper()
+            normalized_key_with_vdk_prefix = VDK_ + self.__normalize_key(key)
+            normalized_key = self.__normalize_key(key)
             value = upper_cased_env.get(normalized_key)
-            if value is not None:
-                log.debug(f"Found environment variable for key {key}")
```
doks5, line 140: Maybe we can add a log message at the info level with a statistic of what environment variables were found and used to make it easier for the users to understand what happened, in case they end up in a situation, where they have environment variables for other tools with the same names as the ones we use?

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -108,6 +108,9 @@ class EnvironmentVarsConfigPlugin:
     Configuration loaded from environment variables.
     """
 
+    def __normalize_key(self, key):
```
doks5, line 111: Why not replace with a lambda?
```python
__normalize_key = lambda key: key.replace("-", "_").replace(".", "_").upper()
```

doks5 (approved):
Overall, LGTM

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
tozka: I think already our logs are way too verbose. For somebody new it's hard to get around them. I think we need to minimize INFO logs to the essential. Also I think we need to have them more structured (but that's separate topic)

But the problem you are pointing out (easier for the users to understand what happened and how something was configured) is real. 

But I am thinking of perhaps extending vdk config-help to show the current value and how it was configured.
Maybe printing stats may not be too much logging.  I'll merge this now and think about this problem in separate PR. Thanks for pointing it out.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
tozka: Seems the same. The function style may be slightly more readable. I void lambda unless there are really really short

PR #876, opened 2022-06-24
vdk-core: add option to disable version check

tozka:
Version check is not needed always. Perhaps in cloud runs for example
and we can skip it. We need option to do so.

Testing Done: added unit tests


gageorgiev (approved):

doks5:
Maybe update the description to mentioned how it was tested?

PR #885, opened 2022-06-30
[vdk-plugins] vdk-kerberos-auth: Separate async event loop

doks5:
The minikerberos authenticator relies on an async kerberos client to speed up some of its
processes. This client uses `asyncio.run()` to execute async operations. As can be seen from the source
code of asyncio, https://github.com/python/cpython/blob/3.10/Lib/asyncio/runners.py, the `run()` method
creates a new event loop and sets it as the [default loop](https://github.com/python/cpython/blob/3.10/Lib/asyncio/runners.py#L41).
When all async operations are finished, the loop it terminated and all tasks are cleaned. This is fine
in general, as long as other components don't rely on the main event loop.
    
If they do, however, we could end up in a situation, where the kerberos plugin closes the asyncio
main event loop, and breaks the other components that rely on the same thread.
    
To avoid issues with multiple components relying on the asyncio main event loop, this change introduces
a separate event loop for the minikerberos authenticator, thus keeping kerberos operations away of the main
event loop.
    
Testing done: Functional test added.
    
### DIFF
--- a/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
+++ b/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
@@ -75,7 +75,22 @@ class MinikerberosGSSAPIAuthenticator(BaseAuthenticator):
             async def get_tgt():
                 await krb_client.get_TGT()
 
-            asyncio.run(get_tgt())
+            # Create a separate event loop and run minikerberos coroutines in it
+            # to avoid interfering with other plugins or with data jobs that
+            # rely on asyncio functionality, as well. The steps of the process
+            # are as follows:
+            # 1) Create a new event loop which is different from the default one
+            # 2) Execute the `get_tgt()` coroutine and wait until it is
+            #    finished.
+            # 3) After the kerberos TGT is retrieved and the coroutine has
+            #    finished, close the event loop. As it is a single coroutine,
+            #    we don't really need to do anything else.
+            loop = asyncio.new_event_loop()
+            try:
+                loop.run_until_complete(get_tgt())
+            finally:
+                loop.close()
+
             krb_client.ccache.to_file(self._ccache_file)
             log.info(
                 f"Got Kerberos TGT for {self._kerberos_principal}@{self._kerberos_realm} "
--- a/projects/vdk-plugins/vdk-kerberos-auth/tests/jobs/test-job/20_async_test.py
+++ b/projects/vdk-plugins/vdk-kerberos-auth/tests/jobs/test-job/20_async_test.py
@@ -1,0 +1,13 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import asyncio
+import logging
+
+from vdk.api.job_input import IJobInput
+
+log = logging.getLogger(__name__)
+
+
+def run(job_input: IJobInput):
+    log.info("Test Async step")
+    _ = asyncio.Semaphore(2)


projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
```diff
@@ -75,7 +75,12 @@ def _kinit(self) -> None:
             async def get_tgt():
                 await krb_client.get_TGT()
 
-            asyncio.run(get_tgt())
+            # Create a separate event loop to avoid interfering with users'
+            # data jobs utilizing asyncio, as well.
+            loop = asyncio.new_event_loop()
+            loop.run_until_complete(get_tgt())
+            loop.close()
```
tozka, line 82: Can you please explain a bit more? I am not sure whether this is the correct root cause and the solution. But it's possible I am not understanding something.

asyncio.run (https://github.com/python/cpython/blob/main/Lib/asyncio/runners.py#L150) inside basically does this - create a new event loop and closes it. So I cannot see what changes here. It does a bit more robust cleaning but I would find it surprising if that is the issue.

projects/vdk-plugins/vdk-kerberos-auth/tests/jobs/test-job/20_async_test.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import asyncio
+import logging
+
+from vdk.api.job_input import IJobInput
+
```
tozka, line 7: Did you run the modified unit tests with the source code version before the fix ? 

I very much recommend when writing a test that verifies a fix  to make sure you run it with the previous version and make certain the test *fails* and after the fix is applied the test passes. This gives you very high certainty that the root cause, the problem and the fix have been correctly identified. 

I suggest you revert your change and try to run the test. It should fail if the root cause was correct.

doks5: Very good catch. Changed the test, and verified it breaks without the changes.
Thank you

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
doks5: Updated the description. Hopefully, it is better worded now.
That being said, the issue with using `asyncio.run()` in our case is, that the run method sets the default event loop https://github.com/python/cpython/blob/3.10/Lib/asyncio/runners.py#L41, and when it closes it, other components that rely on it break.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
```diff
@@ -75,7 +75,12 @@ def _kinit(self) -> None:
             async def get_tgt():
                 await krb_client.get_TGT()
 
-            asyncio.run(get_tgt())
+            # Create a separate event loop to avoid interfering with users'
+            # data jobs utilizing asyncio, as well.
```
tozka, line 79: Explain in the code a bit more why we are doing this. Some future maintainer may wonder as well.

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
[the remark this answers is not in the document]
tozka: Also it calls 
```
 _cancel_all_tasks(loop)
            loop.run_until_complete(loop.shutdown_asyncgens())
            loop.run_until_complete(loop.shutdown_default_executor())
```

Is it necessary ?

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
doks5: Updated and expanded the comments. Thanks

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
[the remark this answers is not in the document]
doks5: Updated the code to explicitly close the loop whatever happens with `get_tgt()`.
Also, after looking at the source of asyncio, it does not seem like we need to make additional calls, as `run_until_complete()` waits until the coroutine has finished, and at the top level, we call a single coroutine -- `get_tgt()`.

tozka (approved):

PR #886, opened 2022-07-01
Add data lifecycle image

zverulacis:


tozka:
I think it would be better to have svg e.g : 
data-journey


 I am a bit more married to data-journey than data-cycle as naming. If you google both data journey images appear much more similar to this than those of data cycle.

(hint: btw you can drag and drop images in comments and github would automatically upload them to user-images.githubusercontent)

tozka:
I still think there's value in putting some common important images in support folder. 

Because the above link (in images.githubusercontent)
A) Cannot be updated, so we need to generate a new link if we update the image . 
B) It's harder to reuse since, how do you know you share it ? 

My suggestion was to use svg image instead of png .svg images are vector images and they resize better for different screens, and are consistent quality regardless of resolution.

PR #896, opened 2022-07-06
vdk-lineage-model: Extract Lineage Model in separate plugin

VladimirPetkov1:
Currently there is code duplication in vdk-trino and vdk-linage plugins and
also any new plugin that wants to emit lineage data will have to copy-paste the
lineage model and logger interface.
Now that the lineage model is spearate plugin/lib it can be consumed by
different plugins like vdk-trino, vdk-lineage and for e.g. vdk-impala without
introducing code duplication.
After this change gets released will remove code duplication in vdk-lineage and
vdk-trino plugins in separate change.

Signed-off-by: Vladimir Petkov <petkovv@vmware.com>


projects/vdk-plugins/build-plugin.sh
```diff
@@ -30,4 +34,10 @@ fi
 
 pip install pytest-cov
 
-pytest --junitxml=tests.xml --cov vdk --cov-report term-missing --cov-report xml:coverage.xml
+# List exceptions to below check here.
+# Those do not contain any meaningful logic so they do not have any tests defined
```
tozka, line 38: We will have tests in the future perhaps. 

 I prefer to just write some test and not exclude it. You can test the __str__ method :) or just that it can be imported without exception.

VladimirPetkov1: Ok, I have added a test that checks the __str__ implementation

tozka (approved):

PR #905, opened 2022-07-18
vdk-core: show log_level_vdk in help

tozka:
Due to small bug in defining LOG_LEVEL_VDK the description was not
showin in `vdk config-help`

Testing Done: Ran `vdk config-help` and saw it shows up correctly

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
@@ -64,6 +64,7 @@ class CoreConfigDefinitionPlugin:
         config_builder.add(
             LOG_LEVEL_VDK,
             None,
+            True,
             "Logging verbosity of VDK code can be controlled from here. "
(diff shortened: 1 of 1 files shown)


doks5 (approved):
LGTM

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -64,6 +64,7 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
         config_builder.add(
             LOG_LEVEL_VDK,
             None,
+            True,
```
ivakoleva, line 67: wasn't the default value true?

```
def add(
        self,
        key: ConfigKey,
        default_value: ConfigValue,
        show_default_value=True,
        description=None,
    ) 
```

tozka: Yes, but, but I was setting show_default_value = "description" due to order of parameters.

PR #906, opened 2022-07-19
control-service: Introduce job termination status counter

doks5:
Currently, we expose "gauge" metrics for data job termination statuses, which
we can then use to moonitor the operability of data jobs deployed in kubernetes
clusters. This works fine for simple monitoring when we are looking for the current
execution status of a job or its change over time.

However, if we want to aggregate data for all data jobs that we have deployed, for
example check what percentage of all jobs fail with user or platform error compared
to all job executions, things can get complicated.

This change introduces "counter" metrics that measure the number of specific statuses
observed for each data job to help with situations when high-level picture of data job
executions is needed.

Testing Done: Unit tests and existing tests.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
@@ -15,11 +15,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
```diff
@@ -173,6 +172,33 @@ public void updateTerminationStatusGauge(final DataJob dataJob) {
...
     }
 
+    /**
+     * Creates a "taurus.datajob.termination.status.counter" counter for the specified
+     * data job and execution status, if one does not exist. If a counter already
+     * exists, its value is incremented.
+     *
+     * @param dataJob The data job for which to increment the counter.
+     * */
+    public void incrementTerminationStatusCounter(final DataJob dataJob) {
```
ivakoleva, line 182: Looking at the PR description

> This change introduces "counter" metrics that measure the number of specific statuses
> observed for each data job to help with situations when high-level picture of data job
> executions is needed.

How practical would that change be? If we monitor the count only, then how do we establish what data jobs do indicate those specific statuses?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
```diff
@@ -173,6 +172,33 @@ public void updateTerminationStatusGauge(final DataJob dataJob) {
...
+        Objects.requireNonNull(dataJob);
+
```
ivakoleva, line 184: Also from PR description

> However, if we want to aggregate data for all data jobs that we have deployed, for
> example check what percentage of all jobs fail with user or platform error compared
> to all job executions, things can get complicated.

Are the other data job status metrics usable prior this change? 
Is the reason for the change is like "introducing one additional metric that is easier to use yet not that informative"?
Is such a counter applicable for monitoring & data jobs troubleshooting? What is the use-case?

tozka: I think a good answer to that question is to give an example query of how it would be used. We need it for our amon dashboards

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
```diff
@@ -173,6 +172,33 @@ public void updateTerminationStatusGauge(final DataJob dataJob) {
...
+            var dataJobStatus = dataJob.getLastExecutionStatus().getAlertValue().toString();
+            var jobStatusKey = dataJobName + "__" + dataJobStatus;
+            var counter = statusCounters.getOrDefault(jobStatusKey, null);
+            var tags = createStatusCounterTags(dataJob);
+
+            if (counter == null) {
+                statusCounters.computeIfAbsent(jobStatusKey, name -> createTerminationStatusCounter(dataJobName, tags, dataJobStatus));
+            } else {
+                counter.increment();
+            }
```
tozka, line 196: Seems like off-by one error - are we incrementing the counter the first time ?

 But I may not be seeing something. The DataJobTerminationStatusIT should be able to verify this case

tozka (changes requested):
I think it's a good idea to extend DataJobTerminationStatusIT test (it should be adding a new assertion for the counter I think. 
 
 
 How are we going to ensure that different pods are showing the same number? 
And if they are showing the different number, would that be a problem and what kind of problem? 
 What happens if a pod is restarted? Its counter would be reset but the other pod would not be so Prometheus would scrape what?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
doks5: An example query would be
```
sum(increase(data_jobs_termination_status_counter{status=2}[24h]  / sum(increase(data_jobs_termination_status_counter[24h]))
```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
doks5: The change is more for statistic purposes. It would allow the user to get a high-level picture of the overall job executions. Also, because the counter has label for job_name and status, one could possibly have a query to filter by job_name.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
doks5: Good catch. Yes, the counter was off-by one.

ivakoleva (approved):
do we keep documentation of enlisted/summarized metrics?

tozka:
> do we keep documentation of enlisted/summarized metrics?

Yes, please let's update https://github.com/vmware/versatile-data-kit/tree/main/projects/control-service/projects/helm_charts/pipelines-control-service#metrics

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMonitor.java
```diff
@@ -141,12 +141,15 @@ public void clearDataJobsGaugesNotIn(final Iterable<DataJob> dataJobs) {
...
+     * Creates metrics (a gauge and counters) that expose termination status
+     * for the specified data job and count the number of instances of the said
+     * status.
+     * If a gauge or counters already exist for the data job, they are updated
+     * if necessary.
      *
      * @param dataJob The data job for which to create or update a gauge.
      */
-    void updateDataJobTerminationStatusGauge(final DataJob dataJob) {
+    void updateDataJobTerminationStatusMetrics(final DataJob dataJob) {
```
tozka, line 152: Can this function be called more than once. It did not matter when we were updating a gauge. But now it does. Since we can increment counter more than once by mistake.

tozka (changes requested):
I am still waiting for answer to 
```
How are we going to ensure that different pods are showing the same number (assuming it's a problem)?
And if they are showing the different number, would that be a problem and what kind of problem?
What happens if a pod is restarted? Its counter would be reset but the other pod would not be so Prometheus would scrape what?

```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMonitor.java
doks5: Logically speaking, yes it can be called more than once. I'm trying to figure out a way to test it, so I can implement some safeguard. Without a test, anything implemented would be as good as a bug.

For your first question:
> How are we going to ensure that different pods are showing the same number (assuming it's a problem)?
> And if they are showing the different number, would that be a problem and what kind of problem?

It would be a reporting problem, as when queried, it would skew reality showing a different number of executions and different statuses than expected.
Again, I need to figure out a way to test it.

> What happens if a pod is restarted? Its counter would be reset but the other pod would not be so Prometheus would scrape what?

I am not sure I understand completely the question. In general the rules for the termination gauges would apply to the counters, as well. If a pod is restarted, new termination gauge is created, as well as a new counter. If I understand correctly (though I might as well be wrong in my understanding) how the metrics work, they are time-based, so Prometheus would scrape the counters for the restarted pods with a different timestamp.

doks5:
Change got corrupted after rebase. Closing PR, as won't fix.

PR #907, opened 2022-07-20
vdk-server: fix for vdk server crashing on startup

mrMoZ1:
what: Added a missing env variable that was causing the control-service to crash on startup, because even though the authentication/authorization is disabled the variable is used in a bean (constructor) initialization and it crashes on spring boot initialization (before it enters our code). The env variable is found in some of our application.properties files, but it varies between prod, testing etc (not all have default values) and there are comments that say setting it is mandatory.

why: the vdk-server's vdk server --install command was failing.

testing: Tested locally with pip install -e option of a locally editable vdk-server plugin. Server was failing to start before the change. After the change - server starts up normally.
 
### DIFF
--- a/projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
+++ b/projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
@@ -693,6 +693,8 @@ class Installer:
             "--set",
             "extraEnvVars.DATAJOBS_DEPLOYMENT_BUILDER_EXTRAARGS=--insecure",
             "--set",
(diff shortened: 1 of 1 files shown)


doks5 (approved):

tozka:
Why is this necessary ?

mrMoZ1:
> Why is this necessary ? Should not this #873 have fixed the issue?

Different variable which has a very similar name.

tozka:
> Member

tozka:
> > Why is this necessary ? Should not this #873 have fixed the issue?
> 
> Different variable which has a very similar name.

Ok. 

Can't we fix it the same way ? It's better to make our control service helm chart easy to install as it will not be used only by vdk-server by other directly.

PR #914, opened 2022-07-26
Extending vdk-csv by adding the export-csv command

duyguHsnHsn:
resolves the issue: https://github.com/vmware/versatile-data-kit/issues/370

### DIFF
--- a/projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
+++ b/projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
@@ -3,21 +3,21 @@
 """
 VDK CSV plugin script.
 """
+import csv
 import json
 import logging
 import os
-import csv
 import pathlib
 
 import click
 from click import Context
+from tabulate import tabulate
 from vdk.api.plugin.hook_markers import hookimpl
 from vdk.internal.builtin_plugins.run.cli_run import run
-from vdk.plugin.csv.csv_ingest_job import csv_ingest_step
-from tabulate import tabulate
 from vdk.internal.builtin_plugins.run.job_context import JobContext
 from vdk.internal.core.config import ConfigurationBuilder
 from vdk.internal.util.decorators import closing_noexcept_on_close
+from vdk.plugin.csv.csv_ingest_job import csv_ingest_step
 from vdk.plugin.sqlite import sqlite_configuration
 from vdk.plugin.sqlite.ingest_to_sqlite import IngestToSQLite
 from vdk.plugin.sqlite.sqlite_configuration import SQLiteConfiguration
@@ -115,7 +115,8 @@ def initialize_job(context: JobContext) -> None:
 
 
 @click.command(
-    name="export-csv", help="Execute a SQL query against a local SQLite database and export to a local CSV file."
+    name="export-csv",
+    help="Execute a SQL query against a local SQLite database and export to a local CSV file.",
 )
 @click.option("-q", "--query", type=click.STRING, required=True)
 @click.option("-n", "--name", type=click.STRING, default="result.csv", required=False)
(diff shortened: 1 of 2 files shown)


mrMoZ1:
Please have a look at our commit template: https://github.com/vmware/versatile-data-kit/blob/main/support/git-commit-template.txt . Usually pull requests start with the component that is being modified e.g 'vdk-csv: add export-csv' command and then in the description we have the what, why, testing paragraphs.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+)
+@click.option("-q", "--query", type=click.STRING, required=True)
+@click.option("-n", "--name", type=click.STRING, default="result.csv", required=False)
+@click.option("-p", "--path", type=click.STRING, default="", required=False)
+@click.pass_context
+def export_csv(ctx: click.Context, query, name: str, path: str):
+    conf = SQLiteConfiguration(ctx.obj.configuration)
+    conn = SQLiteConnection(conf.get_sqlite_file())
+
+    with closing_noexcept_on_close(conn.new_connection().cursor()) as cursor:
```
gageorgiev, line 129: Similar to the previous comment, we'd like this command to work with any configured database. The conf and conn variables should be dropped, and you'd need to get the database cursor through an alternative way; I believe it's possible to do it through the click Context object, but @tozka might be able to clarify.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+        column_names = (
+            [column_info[0] for column_info in cursor.description]
+            if cursor.description
+            else ()  # same as the default value for the headers parameter of the tabulate function
+        )
+        res = cursor.fetchall()
+        if not path:
+            path = os.path.abspath(os.getcwd())
+        fullpath = os.path.join(path, name)
+        if os.path.exists(fullpath):
```
gageorgiev, line 140: This check should happen earlier in the command execution. Imagine you've ran some massive query and pulled the result in memory just to find out that a file with that name exists in the specified location; you'd then need to run the query ran, which is inefficient.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+            click.echo(
+                f"Already existing path: {fullpath}. Try again with another path or file name."
+            )
+        else:
+            with open(fullpath, "w", encoding="UTF8", newline="") as f:
+                writer = csv.writer(f, lineterminator="\n")
+                for row in res:
+                    writer.writerow(row)
+            click.echo(f"You can find the result here: {fullpath}")
+            click.echo(tabulate(res, headers=column_names))
```
gageorgiev, line 150: I don't think we need to print out the query results, given we've just saved them to a file.

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -1,10 +1,16 @@
...
 import os
 from unittest import mock
 
 from click.testing import Result
 from vdk.plugin.csv import csv_plugin
+from vdk.plugin.sqlite import sqlite_configuration
+from vdk.plugin.sqlite import sqlite_plugin
+from vdk.plugin.sqlite.ingest_to_sqlite import IngestToSQLite
+from vdk.plugin.sqlite.sqlite_configuration import SQLiteConfiguration
+from vdk.plugin.sqlite.sqlite_connection import SQLiteConnection
```
gageorgiev, line 13: unused import

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -76,3 +82,47 @@ def test_ingestion_csv_with_options():
 
     assert len(ingest_plugin.payloads) > 0
     assert ingest_plugin.payloads[0].payload[0]["US Zip"] == "08056"
+
+
```
gageorgiev, line 86: Maybe add a test for your error handling logic, for example create a file, then try to save a query result to a file with the same name, and verify that it fails correctly, i.e. check that a correct error message is output.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+    )
+
+    context.ingester.add_ingester_factory_method(
+        "sqlite", (lambda: IngestToSQLite(conf))
+    )
+
+
+@click.command(
+    name="export-csv",
+    help="Execute a SQL query against a local SQLite database and export to a local CSV file.",
```
gageorgiev, line 119:
```suggestion
    help="Execute a SQL query against a configured database and export the result to a local CSV file.",
```

gageorgiev (changes requested):

PR #916, opened 2022-07-26

duyguHsnHsn:

### DIFF
--- a/projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
+++ b/projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
@@ -7,19 +7,14 @@ import csv
 import json
 import logging
 import os
-import pathlib
 
 import click
 from click import Context
-from tabulate import tabulate
+from vdk.internal.core import errors
 from vdk.api.plugin.hook_markers import hookimpl
 from vdk.internal.builtin_plugins.run.cli_run import run
-from vdk.internal.builtin_plugins.run.job_context import JobContext
-from vdk.internal.core.config import ConfigurationBuilder
 from vdk.internal.util.decorators import closing_noexcept_on_close
 from vdk.plugin.csv.csv_ingest_job import csv_ingest_step
-from vdk.plugin.sqlite import sqlite_configuration
-from vdk.plugin.sqlite.ingest_to_sqlite import IngestToSQLite
 from vdk.plugin.sqlite.sqlite_configuration import SQLiteConfiguration
 from vdk.plugin.sqlite.sqlite_connection import SQLiteConnection
 
@@ -29,7 +24,7 @@ log = logging.getLogger(__name__)
 @click.command(
     name="ingest-csv",
     help="Ingest CSV file."
-    """
+         """
 The ingestion destination depends on how vdk has been configured.
 See vdk config-help  - search for "ingest" to check for possible ingestion configurations.
 
@@ -71,7 +66,7 @@ vdk ingest-csv -f revenue.csv --options='{"names": ["gender", "os", "visits", "a
     default=None,
     type=click.STRING,
     help="The table in which the csv file will be ingested into."
-    "If not specified, it will default to the csv file name without the extension",
+         "If not specified, it will default to the csv file name without the extension",
 )
 @click.option(
     "-o",
@@ -79,8 +74,8 @@ vdk ingest-csv -f revenue.csv --options='{"names": ["gender", "os", "visits", "a
     default="{}",
     type=click.STRING,
     help="""Pass extra options when reading CSV file formatted as json. For example {"sep": ";", "verbose": true} """
-    "Those are the same options as passed to pandas.read_csv. See "
-    "https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html.",
+         "Those are the same options as passed to pandas.read_csv. See "
+         "https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html.",
 )
 @click.pass_context
 def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
@@ -93,27 +88,6 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
     click.echo(f"Ingesting csv file {file} finished.")
 
 
-@hookimpl
-def vdk_configure(config_builder: ConfigurationBuilder) -> None:
-    """
-    Here we define what configuration settings are needed for sqlite with reasonable defaults
-    """
-    sqlite_configuration.add_definitions(config_builder)
-
-
-@hookimpl
-def initialize_job(context: JobContext) -> None:
-    conf = SQLiteConfiguration(context.core_context.configuration)
-    context.connections.add_open_connection_factory_method(
-        "SQLITE",
-        lambda: SQLiteConnection(pathlib.Path(conf.get_sqlite_file())).new_connection(),
-    )
-
-    context.ingester.add_ingester_factory_method(
-        "sqlite", (lambda: IngestToSQLite(conf))
-    )
-
-
 @click.command(
     name="export-csv",
     help="Execute a SQL query against a configured database and export the result to a local CSV file.",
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+                f"Already existing path: {fullpath}. Try again with another path or file name."
+            )
+        else:
+            with open(fullpath, "w", encoding="UTF8", newline="") as f:
+                writer = csv.writer(f, lineterminator="\n")
+                for row in res:
+                    writer.writerow(row)
+            click.echo(f"You can find the result here: {fullpath}")
+            click.echo(tabulate(res, headers=column_names))
+            raise Exception(f"You can find the result here: {fullpath}")
```
mrMoZ1, line 151: I would advise against using the base Exception class. Have a look at our errors module: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/core/errors.py and choose an appropriate exception type.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+            if cursor.description
+            else ()  # same as the default value for the headers parameter of the tabulate function
+        )
+        res = cursor.fetchall()
+        if not path:
+            path = os.path.abspath(os.getcwd())
+        fullpath = os.path.join(path, name)
+        if os.path.exists(fullpath):
+            click.echo(
+                f"Already existing path: {fullpath}. Try again with another path or file name."
```
mrMoZ1, line 142: I would probably throw an exception here. Make sure to read the https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/CONTRIBUTING.md#error-handling error handling docs.

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -76,3 +80,46 @@ def test_ingestion_csv_with_options():
 
     assert len(ingest_plugin.payloads) > 0
     assert ingest_plugin.payloads[0].payload[0]["US Zip"] == "08056"
+
+
+def test_csv_export(tmpdir):
```
mrMoZ1, line 85: Does the test cover the full 'export_csv' method e.g do we hit the 'Already existing path: {fullpath}. Try again with another path or file name.' log call? I would try to cover as much of the method as possible (tests hit all lines)

mrMoZ1:
The title of the pull request should be:
and not the name of the branch.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+def initialize_job(context: JobContext) -> None:
+    conf = SQLiteConfiguration(context.core_context.configuration)
+    context.connections.add_open_connection_factory_method(
+        "SQLITE",
+        lambda: SQLiteConnection(pathlib.Path(conf.get_sqlite_file())).new_connection(),
+    )
+
+    context.ingester.add_ingester_factory_method(
+        "sqlite", (lambda: IngestToSQLite(conf))
+    )
```
gageorgiev, line 114: We'd like the export-csv command to work with any database that we have configured, not just SQLite; this entire chunk should be dropped I believe.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+)
+@click.option("-q", "--query", type=click.STRING, required=True)
+@click.option("-n", "--name", type=click.STRING, default="result.csv", required=False)
+@click.option("-p", "--path", type=click.STRING, default="", required=False)
+@click.pass_context
+def export_csv(ctx: click.Context, query, name: str, path: str):
+    conf = SQLiteConfiguration(ctx.obj.configuration)
+    conn = SQLiteConnection(conf.get_sqlite_file())
+
+    with closing_noexcept_on_close(conn.new_connection().cursor()) as cursor:
```
gageorgiev, line 129: Similar to the previous comment, we'd like this command to work with any configured database. The conf and conn variables should be dropped, and you'd need to get the database cursor through an alternative way; I believe it's possible to do it through the click Context object, but @tozka might be able to clarify.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+        column_names = (
+            [column_info[0] for column_info in cursor.description]
+            if cursor.description
+            else ()  # same as the default value for the headers parameter of the tabulate function
+        )
+        res = cursor.fetchall()
+        if not path:
+            path = os.path.abspath(os.getcwd())
+        fullpath = os.path.join(path, name)
+        if os.path.exists(fullpath):
```
gageorgiev, line 140: This check should happen earlier in the command execution. Imagine you've ran some massive query and pulled the result in memory just to find out that a file with that name exists in the specified location; you'd then need to run the query ran, which is inefficient.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+            click.echo(
+                f"Already existing path: {fullpath}. Try again with another path or file name."
+            )
+        else:
+            with open(fullpath, "w", encoding="UTF8", newline="") as f:
+                writer = csv.writer(f, lineterminator="\n")
+                for row in res:
+                    writer.writerow(row)
+            click.echo(f"You can find the result here: {fullpath}")
+            click.echo(tabulate(res, headers=column_names))
```
gageorgiev, line 150: I don't think we need to print out the query results, given we've just saved them to a file.

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -76,3 +80,46 @@ def test_ingestion_csv_with_options():
 
     assert len(ingest_plugin.payloads) > 0
     assert ingest_plugin.payloads[0].payload[0]["US Zip"] == "08056"
+
+
```
gageorgiev, line 84: Maybe add a test for your error handling logic, for example create a file, then try to save a query result to a file with the same name, and verify that it fails correctly, i.e. check that a correct error message is output.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +93,65 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+    )
+
+    context.ingester.add_ingester_factory_method(
+        "sqlite", (lambda: IngestToSQLite(conf))
+    )
+
+
+@click.command(
+    name="export-csv",
+    help="Execute a SQL query against a local SQLite database and export to a local CSV file.",
```
gageorgiev, line 119:
```suggestion
    help="Execute a SQL query against a configured database and export the result to a local CSV file.",
```

gageorgiev (changes requested):

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
tozka: Gabi is correct. It needs to work for any configured database. 

The idea is to do the same thing as in ingest-csv. Copy paste the code and change it to export (instead of ingest) . ingest-csv starts a "data job" in order to ingest the data. So we need to create new data job dir that exports (instead of ingest) data and call it from `export-csv` command.

This way user can configure  DB_DEFAULT_TYPE to decide to what database they want to export. 
You can see doc of  db_default_type using `vdk config-help`

tozka:
You will not be able to run the gitlab CI tests from fork as this is not enabled yet, :( 

You need to create branch instead of open PR using the branch.

doks5 (changes requested):
Change seems broken, includes merge conflict dividers.

projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +90,39 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
...
+            "Use another name or choose another location for the file",
+        )
+    args = dict(query=query, fullpath=fullpath)
+    try:
+        ctx.invoke(
+            run,
+            data_job_directory=os.path.dirname(csv_export_step.__file__),
+            arguments=json.dumps(args),
+        )
+    except UserCodeError as e:
```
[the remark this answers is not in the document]
duyguHsnHsn: Not really, I used it to check something earlier and forgot to delete it. I will delete it now.

PR #917, opened 2022-07-27
vdk-core: BaseVdkError exception propagation flaw fix

ivakoleva:
There was an issue, where a particular exceptions were not propagating
to the respective except(catch) clauses. The process was suddenly
blindly exiting with code 1.

Did troubleshoot the flaw to vdk-core's error framework, where
BaseVdkError was extending ClickException without any purpose
known/documented.
That was causing all the BaseVdkError-based DomainErrors like
VdkConfigurationError, UserCodeError, PlatformServiceError, and so on,
to stop propagating to handlers. Did fix the errors framework design
flaw.

Testing Done: added the vdk_exception exit code test coverage,
that relies on proper exception propagation and handling; ci/cd


doks5 (approved):
LGTM

PR #918, opened 2022-07-28
control-service: add thread.sleep to integration test

mrMoZ1:
what: Attempted to fix a flaky test by adding an interval between
job execution writes and spacing out the end time of executions the
test writes to the database. The particular test was failing because
there were more statuses written in the database than we expect.

why: Integration tests have been failing with various errors. I believe
the data job watch task might be interfering with storing job executions
during the test. The data job watch task is a task that periodically
monitors and updates data job executions. Example test failure:
https://vmware-analytics.gitlab.io/-/versatile-data-kit/-/jobs/2788641321/artifacts/projects/control-service/projects/pipelines_control_service/build/reports/tests/integrationTest/classes/com.vmware.taurus.graphql.it.GraphQLJobExecutionsStatusCountIT.html#testExecutionStatusCount_expectOneSuccessfulOneFailed(String,%20String)

testing: Integration tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/graphql/it/GraphQLJobExecutionsStatusCountIT.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/graphql/it/GraphQLJobExecutionsStatusCountIT.java
@@ -109,10 +109,10 @@ public class GraphQLJobExecutionsStatusCountIT extends BaseDataJobDeploymentIT {
    @Test
    public void testExecutionStatusCount_expectOneSuccessfulOneFailed(String jobName, String username) throws Exception {
       var expectedEndTimeLarger = OffsetDateTime.now();
-
-      addJobExecution(expectedEndTimeLarger, ExecutionStatus.SUCCEEDED, jobName);
-      addJobExecution(expectedEndTimeLarger, ExecutionStatus.PLATFORM_ERROR, jobName);
-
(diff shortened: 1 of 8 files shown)


ivakoleva (commented):
Can we reset/cleanup the test environment, and retry the integrationTest?
If problem still exists, then,

[Seems](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:how_much_memory_does_the_daemon_use_and_can_i_give_it_more) gradle is using default 512MB of heap size; maybe double it? Also, the initial heap size would be 256MB, maybe double as well?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/graphql/it/GraphQLJobExecutionsStatusCountIT.java
```diff
@@ -109,10 +109,10 @@ public void testExecutionStatusCount_expectTwoSuccessfulTwoFailed(String jobName
    @Test
    public void testExecutionStatusCount_expectOneSuccessfulOneFailed(String jobName, String username) throws Exception {
       var expectedEndTimeLarger = OffsetDateTime.now();
-
-      addJobExecution(expectedEndTimeLarger, ExecutionStatus.SUCCEEDED, jobName);
```
ivakoleva, line 113: I thought cleaning up the test environment did solve the issue, yes?

Please describe, what is the problem that this change solves?

ivakoleva (changes requested):

PR #920, opened 2022-07-28
vdk-core: step loading failure misclassified as Platform error fix

ivakoleva:
Cases of any data job step source cannot be loaded, due to syntax error,
or issue with importing dependency, were misclassified as
Platform error.

Failure to load step source, is now classified correctly as User error.
 
Testing Done: added a test to cover the use-case.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
@@ -74,8 +74,27 @@ class StepFuncFactory:
             sys.path.insert(0, str(step.job_dir))
             filename = step.file_path.name
             namespace = "step_" + str(filename).lower().strip(".py")
-            python_module = imp.load_source(namespace, str(step.file_path))
             success = False
+
+            try:
+                log.debug("Loading %s ..." % filename)
+                python_module = imp.load_source(namespace, str(step.file_path))
+                log.debug("Loading %s SUCCESS" % filename)
+            except BaseException as e:
+                log.info("Loading %s FAILURE" % filename)
+                errors.log_and_rethrow(
+                    to_be_fixed_by=errors.ResolvableBy.USER_ERROR,
+                    log=log,
+                    what_happened="Failed loading job sources of %s" % filename,
+                    why_it_happened=errors.MSG_WHY_FROM_EXCEPTION(e),
+                    consequences=errors.MSG_CONSEQUENCE_TERMINATING_APP,
+                    countermeasures=errors.MSG_COUNTERMEASURE_FIX_PARENT_EXCEPTION
+                    + " Most likely importing a dependency or data job step failed, see"
+                    + " logs for details and fix the failed step (details in stacktrace).",
+                    exception=e,
+                    wrap_in_vdk_error=True,
+                )
+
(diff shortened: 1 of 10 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
```diff
@@ -74,8 +74,27 @@ def run(job_input)
...
             filename = step.file_path.name
             namespace = "step_" + str(filename).lower().strip(".py")
-            python_module = imp.load_source(namespace, str(step.file_path))
             success = False
+
+            try:
+                log.debug("Loading %s ..." % filename)
+                python_module = imp.load_source(namespace, str(step.file_path))
+                log.debug("Loading %s SUCCESS" % filename)
+            except BaseException as e:
```
doks5, line 83: Is there a specific reason for catching BaseException and not just Exception?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
```diff
@@ -74,8 +74,27 @@ def run(job_input)
             sys.path.insert(0, str(step.job_dir))
             filename = step.file_path.name
             namespace = "step_" + str(filename).lower().strip(".py")
-            python_module = imp.load_source(namespace, str(step.file_path))
             success = False
+
+            try:
+                log.debug("Loading %s ..." % filename)
```
doks5, line 80: nit: Why not using f strings?

doks5 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
ivakoleva: for consistency with the same usage, several lines bellow:
`log.info("Entering %s#run(...) ..." % filename)`

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
ivakoleva: saw same approach in `cli_run.py` and `data_job.py` on the way the exception propagates

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
doks5: Gotcha, ok.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
doks5: Maybe we need to consider how we construct exceptions. According to the documentation, BaseException is not meant to be inherited from https://docs.python.org/3/library/exceptions.html#BaseException, so in theory, we shouldn't expect to need to catch any exception at that low level

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
tozka: Note that some of those lines were written 4 years ago or more before "f-strings". So we should switch them all over to f-strings when we see opportunity.

PR #921, opened 2022-07-29
vdk-core: add memory properties client

tozka:
In case Control Service is not available, for demo purposes, it would be
good to be able to use some simple client so that jobs that stil use
properties would work.
To make it clear that it's not recommended for production (after all
properties are supposed to be used to store state, so in memory client
may be misleading) we write warning each  time write proerties is
called.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/inmemproperties.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/inmemproperties.py
@@ -22,5 +22,9 @@ class InMemPropertiesServiceClient(IPropertiesServiceClient):
         return res
 
     def write_properties(self, job_name: str, team_name: str, properties: Dict) -> Dict:
+        log.warning(
+            "You are using In Memory Properties client. "
+            "That means the properties will not be persisted past the Data Job run."
+        )
         self._props = deepcopy(properties)
         return self._props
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/inmemproperties.py
```diff
@@ -22,5 +22,9 @@ def read_properties(self, job_name: str, team_name: str) -> Dict:
         return res
 
     def write_properties(self, job_name: str, team_name: str, properties: Dict) -> Dict:
+        log.warning(
+            "You are using In Memory Properties client. "
```
doks5, line 26:
```suggestion
            "You are using In-Memory Properties client. "
```

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/properties_api_plugin.py
```diff
@@ -1,7 +1,11 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+from src.vdk.internal.builtin_plugins.job_properties.inmemproperties import (
```
doks5, line 3:
```suggestion
from vdk.internal.builtin_plugins.job_properties.inmemproperties import (
```

doks5 (approved):
Overall, LGTM

PR #924, opened 2022-07-30
vdk-airflow: populate readme

tozka:
Add more meaningful readme that links to important resources related to the airflow provider and shows examples/demo. 

### DIFF
--- a/projects/vdk-plugins/airflow-provider-vdk/README.md
+++ b/projects/vdk-plugins/airflow-provider-vdk/README.md
@@ -2,14 +2,14 @@
 
 A set of Airflow operators, sensors and a connection hook intended to help schedule Versatile Data Kit jobs using Apache Airflow.
 
-# Usage 
+# Usage
 
-To install it simply run: 
+To install it simply run:
 ```
 pip install airflow-provider-vdk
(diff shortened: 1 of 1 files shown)


projects/vdk-plugins/airflow-provider-vdk/README.md
```diff
@@ -1,3 +1,62 @@
...
 A set of Airflow operators, sensors and a connection hook intended to help schedule Versatile Data Kit jobs using Apache Airflow.
+
+# Usage
+
+To install it simply run:
+```
+pip install airflow-provider-vdk
+```
+
+Then you can create workflow of data jobs like this:
```
doks5, line 12: Maybe add a note that the jobs need to be deployed before running the DAG, in case users miss this part in the linked example.

doks5 (approved):

projects/vdk-plugins/airflow-provider-vdk/README.md
tozka:
```suggestion
Then you can create a workflow of data jobs (deployed by VDK Control Service) like this:
```

PR #926, opened 2022-08-01

duyguHsnHsn:

### DIFF
--- a/projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
+++ b/projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
@@ -92,37 +92,43 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
 Examples:
 
 \b
-# This will execute the given query and save the data 
-# in a CSV file with name result.csv in the current directory 
-# if there is no file with the same name there 
+# This will execute the given query and save the data
+# in a CSV file with name result.csv in the current directory
+# if there is no file with the same name there
 vdk export-csv -q  "SELECT * FROM test_table"
 
 \b
-# This will execute the given query and save the data 
+# This will execute the given query and save the data
 # in a CSV file with name result1.csv in User/Documents/csv
-# if there is no file with the same name there 
+# if there is no file with the same name there
 vdk export-csv -q  "SELECT * FROM test_table" -p User/Documents/csv -n result1.csv
 
     """,
     no_args_is_help=True,
 )
-@click.option("-q",
-              "--query",
-              type=click.STRING,
-              required=True,
-              help="The query that will be executed against the specified database.")
-@click.option("-n",
-              "--name",
-              type=click.STRING,
-              default="result.csv",
-              required=False,
-              help="The name of the CSV file that will be created.")
-@click.option("-p",
-              "--path",
-              type=click.STRING,
-              default="",
-              required=False,
-              help="Path to the directory where the CSV file will be saved.")
+@click.option(
+    "-q",
+    "--query",
+    type=click.STRING,
+    required=True,
+    help="The query that will be executed against the specified database.",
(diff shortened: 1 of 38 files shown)


projects/vdk-plugins/vdk-csv/src/vdk/plugin/csv/csv_plugin.py
```diff
@@ -83,6 +85,36 @@ def ingest_csv(ctx: Context, file: str, table_name: str, options: str) -> None:
     click.echo(f"Ingesting csv file {file} finished.")
 
 
+@click.command(
+    name="export-csv",
+    help="Execute a SQL query against a configured database and export the result to a local CSV file.",
+)
+@click.option("-q", "--query", type=click.STRING, required=True)
+@click.option("-n", "--name", type=click.STRING, default="result.csv", required=False)
```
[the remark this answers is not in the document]
duyguHsnHsn: I will do the changes today.

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -76,3 +80,111 @@ def test_ingestion_csv_with_options():
...
+        output = []
+        with open(os.path.join(os.path.abspath(os.getcwd()), "result.csv")) as file:
+            reader = csv.reader(file, delimiter=",")
+            for row in reader:
+                output.append(row)
+        assert output[0] == ["some_test_data", "more_test_data"]
+        assert output[1] == ["some_test_data_copy", "more_test_data_copy"]
+
+
+def test_export_csv_with_already_existing_file():
```
doks5, line 126: This test case seems to fail. Have you tried patching the environment similar to the case above?
```
with mock.patch.dict(
        os.environ,
        {
            "VDK_DB_DEFAULT_TYPE": "SQLITE",
            "VDK_SQLITE_FILE": <some-file>,
        },
    ):
```

duyguHsnHsn: I tried it, still fails in ci/cd but works locally.

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -76,3 +81,114 @@ def test_ingestion_csv_with_options():
...
+                "SELECT * FROM test_table",
+                "--name",
+                "result2.csv",
+            ]
+        )
+        assert isinstance(result.exception, UserCodeError)
+        cli_assert_equal(1, result)
+
+
+def test_csv_export_with_no_data(tmpdir):
```
tozka, line 152: I don't think export-csv should fail if there's no data. 

A) If the SQL query returns empty result (or teh table is empty) , no rows -> we should return empty file. It's normal for people to write query that may return empty result. If this is an issue or not is not something we can tell. 


B) if the table doesn't exist, then we should fail  

Let's split this test into 2 to handle both cases and fix code accordingly.

tozka (changes requested):

projects/vdk-plugins/vdk-csv/tests/functional/test_csv_plugin.py
```diff
@@ -76,3 +81,114 @@ def test_ingestion_csv_with_options():
...
+        },
+    ):
+        runner = CliEntryBasedTestRunner(sqlite_plugin, csv_plugin)
+        runner.invoke(
+            [
+                "sqlite-query",
+                "--query",
+                "DROP TABLE IF EXISTS test_table",
+            ]
+        )
```
tozka, line 168: Also Are you aware of  DRY principle - https://en.wikipedia.org/wiki/Don%27t_repeat_yourself 

In this case you have multiple places you are doing runner.invoke (sqlite-query ..) you can extract this with a method with arguments the query. This would simplify the method.

PR #930, opened 2022-08-03
vdk-core: Improve ingestion error logging

gageorgiev:
Currently, after ingestion completes, a user is presented
with the following message:
```
Successful uploads:1
Failed uploads:0
ingesting plugin errors:defaultdict(<class 'vdk.internal.builtin_plugins.ingestion.ingester_utils.AtomicCounter'>, {})
```

This is not ideal, as printing internal Python representations
in user logs might be misleading, as a user might believe
some error has occured when none did.

This change fixes this by printing `None` instead when no
errors occur, and printing the dictionary result instead of
the defaultdict representation when errors did occur.
```
Successful uploads: 1
Failed uploads: 0
Ingesting plugin errors: None
```

Testing done: tested locally


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -543,9 +543,9 @@ def close_now(self):
 
             log.info(
```
doks5, line 544: Can you update the description of the PR with a snippet how the message looks like after the change?

doks5 (approved):

PR #932, opened 2022-08-04
vdk-impala: Add support for lineage in vdk-impala

VladimirPetkov1:
Add support for calculating data lineage for jobs that execute SQL queries
against impala.

Lineage is calculated by examining the query plan of the executed query. The
query plan is accessible from the cursor that executed the query, hence no
additional load is present on the target query engine. As this is not very
official way to obtain the query plan there is a risk that this implentation
may break in future impala releases.However the only option was to execute
additional explain statements that would practically double the amount of
executed queries, which could be problematic for loaded environments. Lineage
is calculated only for successfully executed queries that have actually read
or written data from/to the underlying storage.

Implemented functional tests that verify the behaviour with all templates and
simple jobs. As I hit problems starting impala with pytest, I have manually
run the docker compose file and run the tests against that instance.

Signed-off-by: Vladimir Petkov <petkovv@vmware.com>

### DIFF
--- a/projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
+++ b/projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
@@ -22,6 +22,7 @@ from vdk.plugin.impala.impala_configuration import ImpalaPluginConfiguration
 from vdk.plugin.impala.impala_connection import ImpalaConnection
 from vdk.plugin.impala.impala_error_classifier import is_impala_user_error
 from vdk.plugin.impala.impala_error_handler import ImpalaErrorHandler
+from vdk.plugin.impala.impala_lineage_plugin import ImpalaLineagePlugin
 
 
 def _connection_by_configuration(configuration: ImpalaPluginConfiguration):
@@ -149,6 +150,9 @@ class ImpalaPlugin:
 @hookimpl
 def vdk_start(plugin_registry: IPluginRegistry, command_line_args: List):
     plugin_registry.load_plugin_with_hooks_impl(ImpalaPlugin(), "impala-plugin")
+    plugin_registry.load_plugin_with_hooks_impl(
+        ImpalaLineagePlugin(), "impala-lineage-plugin"
+    )
 
 
 def get_jobs_parent_directory() -> pathlib.Path:
--- a/projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
+++ b/projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
@@ -1,0 +1,369 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import json
+import os
+import sys
+from unittest import mock
+from unittest import TestCase
+from unittest.mock import patch
+
+import pytest
+from click.testing import Result
+from vdk.api.lineage.model.logger.lineage_logger import ILineageLogger
+from vdk.api.lineage.model.sql.model import LineageData
+from vdk.api.lineage.model.sql.model import LineageTable
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.plugin.impala import impala_plugin
+from vdk.plugin.impala.impala_lineage_plugin import LINEAGE_LOGGER_KEY
+from vdk.plugin.test_utils.util_funcs import cli_assert_equal
+from vdk.plugin.test_utils.util_funcs import CliEntryBasedTestRunner
+from vdk.plugin.test_utils.util_funcs import jobs_path_from_caller_directory
+
+VDK_DB_DEFAULT_TYPE = "VDK_DB_DEFAULT_TYPE"
+VDK_IMPALA_HOST = "VDK_IMPALA_HOST"
+VDK_IMPALA_PORT = "VDK_IMPALA_PORT"
+VDK_IMPALA_SYNC_DDL = "VDK_IMPALA_SYNC_DDL"
+VDK_IMPALA_QUERY_POOL = "VDK_IMPALA_QUERY_POOL"
+
+
+class TestConfigPlugin:
+    def __init__(self, lineage_logger: ILineageLogger):
+        self.lineage_logger = lineage_logger
+
+    @hookimpl
+    def vdk_initialize(self, context):
+        # register a lineage logger which will be used to validate if correct lineage data was emitted
+        context.state.set(LINEAGE_LOGGER_KEY, self.lineage_logger)
+
+
+def execute_query(runner, query: str):
+    result: Result = runner.invoke(["impala-query", "--query", query])
+    cli_assert_equal(0, result)
+
+
+@pytest.mark.skipif(
+    sys.version_info >= (3, 10),
+    reason="Lineage collection is not supported in python 3.10."
+    "See impala_lineage_plugin.py for more details",
+)
+@pytest.mark.usefixtures("impala_service")
+class ImpalaLineageTest(TestCase):
+    @patch.dict(
+        os.environ,
+        {
+            VDK_DB_DEFAULT_TYPE: "IMPALA",
+            VDK_IMPALA_HOST: "localhost",
+            VDK_IMPALA_PORT: "21050",
+        },
+    )
+    def test_lineage_not_collected_for_invalid_query(self):
+        mock_lineage_logger = mock.MagicMock(ILineageLogger)
+        runner = CliEntryBasedTestRunner(
+            TestConfigPlugin(mock_lineage_logger), impala_plugin
+        )
+        result: Result = runner.invoke(
+            ["run", jobs_path_from_caller_directory("sql-job-syntax-error")]
+        )
+        cli_assert_equal(1, result)
+        assert not mock_lineage_logger.send.called
+
+    @patch.dict(
+        os.environ,
+        {
+            VDK_DB_DEFAULT_TYPE: "IMPALA",
+            VDK_IMPALA_HOST: "localhost",
+            VDK_IMPALA_PORT: "21050",
+        },
+    )
+    def test_lineage_for_create_and_insert_queries(self):
+        mock_lineage_logger = mock.MagicMock(ILineageLogger)
+        runner = CliEntryBasedTestRunner(
+            TestConfigPlugin(mock_lineage_logger), impala_plugin
+        )
+        result: Result = runner.invoke(
+            ["run", jobs_path_from_caller_directory("sql-job")]
+        )
+        cli_assert_equal(0, result)
+
+        class InsertLineageDataMatcher:
+            def __eq__(self, lineage_data: LineageData):
+                # the query is different on every run of the test as it has the opid (which is unique) prepended
+                assert "INSERT INTO stocks VALUES" in lineage_data.query
+                assert lineage_data.query_type is None
+                assert lineage_data.query_status == "OK"
+                assert lineage_data.output_table.catalog is None
+                assert lineage_data.output_table.schema == "default"
+                assert lineage_data.output_table.table == "stocks"
+                assert len(lineage_data.input_tables) == 0  # no inputs for insert query
+                return True
+
+        # create table and select 1 queries should not result in sending lineage, hence test with called_once to verify
+        # that only one lineage payload has been emitted
+        mock_lineage_logger.send.assert_called_once_with(InsertLineageDataMatcher())
+
+    @patch.dict(
+        os.environ,
+        {
+            VDK_DB_DEFAULT_TYPE: "IMPALA",
+            VDK_IMPALA_HOST: "localhost",
+            VDK_IMPALA_PORT: "21050",
+        },
+    )
+    def test_lineage_for_scd1_template(self):
+        mock_lineage_logger = mock.MagicMock(ILineageLogger)
+        runner = CliEntryBasedTestRunner(
+            TestConfigPlugin(mock_lineage_logger), impala_plugin
+        )
+        schema = "default"
+        source = "scd1_source"
+        target = "scd1_target"
+        args = {
+            "source_schema": schema,
+            "source_view": source,
+            "target_schema": schema,
+            "target_table": target,
+        }
+        result: Result = runner.invoke(
+            [
+                "run",
+                jobs_path_from_caller_directory("load_dimension_scd1_template_job"),
+                "--arguments",
+                json.dumps(args),
+            ]
+        )
+        cli_assert_equal(0, result)
+
+        class SCD1LineageDataMatcher:
+            def __init__(self, outer_instance: ImpalaLineageTest):
+                self.outer_instance = outer_instance
+
+            def __eq__(self, lineage_data: LineageData):
+                assert (
+                    "INSERT OVERWRITE TABLE default.scd1_target" in lineage_data.query
+                )
+                assert "SELECT * FROM default.scd1_source" in lineage_data.query
+                assert lineage_data.query_type is None
+                assert lineage_data.query_status == "OK"
+                assert lineage_data.output_table.catalog is None
+                assert lineage_data.output_table.schema == schema
+                assert lineage_data.output_table.table == target
+                expected_inputs = [LineageTable(None, schema, source)]
+                self.outer_instance.assertCountEqual(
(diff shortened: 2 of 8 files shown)


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -53,6 +65,9 @@ def impala_query(ctx: click.Context, query):
 
 
 class ImpalaPlugin:
+    def __init__(self, lineage_logger=None):
```
tozka, line 68: My preference would be to be a bit more encapsulated and have "LineagePlugin" class which contains all the lineage related hooks . Better separation of responsibilities and decoupled code. 
Otherwise this file is becoming dumpster for everything. 

The only change then in this file would be in vdk_start hook to register the new Lineage Plugin 

```
    plugin_registry.load_plugin_with_hooks_impl(LineagePlugin())
```

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -3,25 +3,37 @@
...
 from vdk.plugin.impala.impala_configuration import add_definitions
 from vdk.plugin.impala.impala_configuration import ImpalaPluginConfiguration
 from vdk.plugin.impala.impala_connection import ImpalaConnection
 from vdk.plugin.impala.impala_error_classifier import is_impala_user_error
 from vdk.plugin.impala.impala_error_handler import ImpalaErrorHandler
+from vdk.plugin.impala.lineage_utils import get_lineage_data
+
+LINEAGE_LOGGER_KEY = StoreKey[ILineageLogger]("impala-lineage-logger")
+
+_log = logging.getLogger(__name__)
```
tozka, line 36: nit: we do not add "_" prefix for log variable.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
```diff
@@ -0,0 +1,55 @@
...
+        query_status="OK",
+        query_type=None,  # TODO specify query_type once there is clear spec for it
+        input_tables=[
+            _get_lineage_table_from_table_name(table_name) for table_name in inputs
+        ],
+        output_table=_get_lineage_table_from_table_name(output),
+    )
+
+
+def _parse_inputs_outputs(query_profile: str) -> Tuple[list, str]:
```
tozka, line 39: For which versions of impala would that work ? 
You are using the text profile instead of the thrift but how fixed the schema is for the text profile ? Can we rely on it for future versions?

tozka (changes requested):
Thanks for making the change. 

I made some more high level questions and suggestions  which we need to clear up first. I will take a more detailed look at the code later.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
```diff
@@ -0,0 +1,55 @@
...
+from impala.hiveserver2 import HiveServer2Cursor
+from vdk.api.lineage.model.sql.model import LineageData
+from vdk.api.lineage.model.sql.model import LineageTable
+
+_log = logging.getLogger(__name__)
+
+
+def get_lineage_data(cursor: HiveServer2Cursor) -> Optional[LineageData]:
+    """
+    Calculates lineage information from the query plan for the last executed query from the provided cursor. Lineage
```
ivakoleva, line 18: https://docs.cloudera.com/runtime/7.2.10/impala-reference/topics/impala-profile.html docs state:

> It shows physical details such as the number of bytes read, maximum memory usage, and so on for each node.

How is TGetRuntimeProfileReq->TGetRuntimeProfileResp evaluated after each query, in terms of load? What nodes do handle that request? 
What's the overhead, and is there a correlation with executor / coordinator nodes quantity (is handling a profile request heavier when having more cluster nodes)?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -63,6 +78,37 @@ def vdk_configure(config_builder: ConfigurationBuilder) -> None:
...
+        yield  # let the query execute first
+        try:
+            if (
+                self._lineage_logger
+            ):  # calculate and send lineage data only if there is a registered lineage logger
+                hive_cursor = cast(HiveServer2Cursor, execution_cursor)
+                lineage_data = get_lineage_data(hive_cursor)
+                if (
+                    lineage_data
+                ):  # some queries do not have data lineage - select 1, create table, compute stats, etc
```
ivakoleva, line 102: Do we still request a profile in case of `select 1`, and if yes - what's the overhead? For example, how many select 1 is the driver sending per minute, and does it make sense to exclude such `select 1` beforehand, to reduce the pressure on Impala?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
```diff
@@ -0,0 +1,55 @@
...
+        return None  # no lineage was present in the query plan, e.g. select 1 query
+    return LineageData(
+        query=cursor._cursor.query_string,  # a bit hacky but otherwise it is not trivial to parse the profile string
+        query_status="OK",
+        query_type=None,  # TODO specify query_type once there is clear spec for it
+        input_tables=[
+            _get_lineage_table_from_table_name(table_name) for table_name in inputs
+        ],
+        output_table=_get_lineage_table_from_table_name(output),
+    )
```
ivakoleva, line 36: does it make sense to add a debug level message that contains the query profile retrieved, so it's troubleshooted easier?

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
```diff
@@ -0,0 +1,363 @@
...
+                assert "INSERT INTO stocks VALUES" in lineage_data.query
+                assert lineage_data.query_type is None
+                assert lineage_data.query_status == "OK"
+                assert lineage_data.output_table.catalog is None
+                assert lineage_data.output_table.schema == "default"
+                assert lineage_data.output_table.table == "stocks"
+                assert len(lineage_data.input_tables) == 0  # no inputs for insert query
+                return True
+
+        # create table and select 1 queries should not result in sending lineage, hence test with called_once to verify
```
ivakoleva, line 94: How many `select 1` queries are to be expected in this test case?
Aren't the quantities of `select 1` correlated to the session length, the time window a data job does active DB work?

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
```diff
@@ -0,0 +1,363 @@
...
+        mock_lineage_logger = mock.MagicMock(ILineageLogger)
+        runner = CliEntryBasedTestRunner(
+            TestConfigPlugin(mock_lineage_logger), impala_plugin
+        )
+        result: Result = runner.invoke(
+            ["run", jobs_path_from_caller_directory("sql-job")]
+        )
+        cli_assert_equal(0, result)
+
+        # the query is different on every run of the test
```
ivakoleva, line 82: don't quite understand this comment, what about the query is different on every run?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
```diff
@@ -0,0 +1,55 @@
...
+_log = logging.getLogger(__name__)
+
+
+def get_lineage_data(cursor: HiveServer2Cursor) -> Optional[LineageData]:
+    """
+    Calculates lineage information from the query plan for the last executed query from the provided cursor. Lineage
+    is calculated only if the query has finished successfully and has resulted actual data read or write from/to the
+    underlying storage.
+    """
+    query_profile = cursor.get_profile(profile_format=TRuntimeProfileFormat.STRING)
```
ivakoleva, line 22: So, my understanding is the driver is doing `select 1` in the background, in the scope of the active session.
At the same time, `profile` is using the latest query metadata.
Is there a possibility, that the latest query is in fact `select 1`, and what happens in such a case?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
VladimirPetkov1: Initially I started with using the thrift format, but that does not seem to be available with the version of impala that is used for testing. Furthermore the interesting part for the lineage is the query plan which is a text field in both string and thrift format, so no difference there.

VladimirPetkov1: Can we rely on it for future versions? Both can be equally unreliable, but this seems the best we could to.

tozka: What does the Impala documentation say? If it doesn't you can ask dev@impala.apache.org or user@impala.apache.org Asking never hurts. Worse case they won't answer. 

Even though we cannot predict the future, a good indication we can rely on it is if it has not changed (in breaking way) across multiple major versions of Impala. 

Anyhow this is not an issue to block the change, but let's find out so we know more.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
VladimirPetkov1: Umm, I do not agree on 100% here as, but I do not disagree also 😄 . Currently the linage logic is encapsulated in separate module - lineage_utils and only the hook is defined in impala_plugin. This way you can see all impala related hooks in a single place which makes it easier to understand what is happening.
Anyhow I will perform the suggested refactoring mainly because I dislike how complex the impala_plugin module has become. For example vdk-trino seems much more streamlined.

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
VladimirPetkov1: every query begins with a comment that contains the op id of the job which is different for every execution, hence not validating the full query.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
VladimirPetkov1: Although I have not seen any official documentation, my observations are that select 1 queries are executed right before the actual query is executed. Moreover we are hooking before the result set is consumed, so there should not be any other query executed in the mean time.

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
VladimirPetkov1: I think the amount of select 1 queries is linearly correlated to the amount of sql queries executed.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
VladimirPetkov1: I though about it, but for complex queries it is really long like >2k lines. I was thinking to log at least the query id or something if an error occurs, but it does not improve much debug-ability as the query itself is logged and it can be used to retrieve the profile from impala.
What is your opinion?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
VladimirPetkov1: Yep we request a profile for select 1 queries. The overhead should be minimal, the profile is already calculated by the query engine. The amount of select 1 queries is the same as the actual queries being executed.
It is a good suggestion to exclude them as it should be fairly straightforward to do so. Will update the implementation,

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
VladimirPetkov1: The query engine builds a profile for every query that it executes regardless of if it is requested or not. TGetRuntimeProfileReq just fetches the already calculated profile so in terms of load it is mainly I/O. I believe this is handled by the coordinator that is responsible for executing the query. So the number of executors nodes in the impala cluster should not affect in any way this operation. The number of coordinator nodes in the cluster should not affect the overhead of a **single** get profile operation.

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
ivakoleva: Is it documented like a common knowledge somewhere? Did you read and where from / test it ?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
ivakoleva: I don't mind

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
ivakoleva: The expectations regarding load and performance I imagine being documented in the plugin README.md, since it evaluates Impala cluster requests practically upon each query

VladimirPetkov1: Ok, will do. But they cannot be very detailed as the performance aspect depends a lot on the actual load on the cluster we could not know.

projects/vdk-plugins/vdk-impala/tests/functional/test_impala_lineage.py
VladimirPetkov1: Nope. It is my assumption based on the logs and that I observe them before queries are being executed (apart from the very first query executed on fresh DB connection).
I was hoping that the VDK team have better understanding of it.

tozka (approved):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/lineage_utils.py
ivakoleva: I looked at the README.md,
So we don't have any performance measurements made in advance? And, at the same time, I don't understand what request is made to Impala, and what do we ask to be performed exactly? What is the scope/workflow of the cluster components to be involved?
for example is `get_profile` handled by an executor that makes an RPC request to the coordinator, that keeps a profile forest, for how long, is it serialized, is it overwritten by keep alive queries performed ?

Now with the README.md, I don't have any more information at hand, just a brief statement that Impala is requested that I knew before as it's too obvious

ivakoleva: The README is too generic for me, and I don't have any impression nor understanding about what's going on under the hood; however we can ignore that, and merge so we have this capability available, right?

VladimirPetkov1: I have linked Impala docs that answer some of the questions that you have and clarified the README.md to share a bit more details.
I have tested this functionality only against local containerised Impala and there was no measurable performance impact. However this is not representative as it is not run against big loaded environment. Once we have a release of the plugin will measure against more representable environment and share if there are any meaningful findings. 

Is get_profile handled by an executor that makes an RPC request to the coordinator, that keeps a profile forest, for how long, is it serialized?
The RPC originates from  the data job not Impala executor. How many profiles are kept is a cluster configuration, it is explained in the linked documents. They are serialized, again explained in the linked documents.

Is it overwritten by keep alive queries performed? No as the RPC is executed right after actual query is executed, keepalive queries are executed before actual queries are executed. This is elaborated in the README.md
The profile

VladimirPetkov1: Let me know if this is enough?

ivakoleva: I guess

PR #943, opened 2022-08-18
vdk-ingest-http: reduce verbosity of ingestion logs

tozka:
Logging each payload at INFO is extremely verbose, we can have thousands
and thousands of payloads being sent. So it also slows down teh
ingestion.

### DIFF
--- a/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
+++ b/projects/vdk-plugins/vdk-ingest-http/src/vdk/plugin/ingest_http/ingest_over_http.py
@@ -128,7 +128,7 @@ class IngestOverHttp(IIngesterPlugin):
     ) -> Optional[IngestionResult]:
         header = {"Content-Type": "application/octet-stream"}  # TODO: configurable
 
-        log.info(
+        log.debug(
             f"Ingesting payloads for target: {target}; "
             f"collection_id: {collection_id}"
         )


gageorgiev (approved):

ivakoleva (commented):
ltgm;

On the logging topic, we have ~50 vdk built-in/internal loggers; maybe we may consider a plugin for (text-)configurable logging level per logger. The use case is, a data analyst would like to increase `vdk.internal.builtin_plugins.ingestion` logging when developing on local machine. So when configured, the vdk config will be stored, and applied to each data job local run.

tozka:
> ltgm;
> 
> On the logging topic, we have ~50 vdk built-in/internal loggers; maybe we may consider a plugin for (text-)configurable logging level per logger. The use case is, a data analyst would like to increase `vdk.internal.builtin_plugins.ingestion` logging when developing on local machine. So when configured, the vdk config will be stored, and applied to each data job local run.

Yes. I've thought about this. we currently wehave LOG_LEVEL_VDK config - https://github.com/vmware/versatile-data-kit/blob/1469b93952fa8409e17cde481231e2a3daa70c32/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py#L64 

But maybe we can make it more genric -e.g  LOG_LEVEL=vdk.api=INFO,vdk.internal=WARNING
(Similar to how in spring you can set logging.level.com.domain=INFO

PR #946, opened 2022-08-22
vdk-core: Fix issue with serializing Decimal values during payload check

gageorgiev:
As part of the ingestion process, the payload passes a
check to verify it is JSON serializable. However,
`json.dumps` does not natively support serializing
Decimal objects, which leads to the check failing when
it should not. This change fixes this problem by amending
the JsonEncoder object to convert Decimals to floats
during the check.
Note that this does not affect the actual ingestion of
Decimal objects, merely the serialization check.

Testing done: existing test, added a unit test for the new class


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -669,7 +669,15 @@ def __verify_payload_format(payload_dict: dict):
         # Check if payload dict is valid json
         # TODO: optimize the check - we should not need to serialize the payload every time
         try:
-            json.dumps(payload_dict)
+            from decimal import Decimal
```
tozka, line 672: Tests? Any change must have an automated test written or a good reason not to have one.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -669,7 +669,15 @@ def __verify_payload_format(payload_dict: dict):
...
+
+            class DecimalJsonEncoder(json.JSONEncoder):
+                def default(self, obj):
+                    if isinstance(obj, Decimal):
+                        return float(obj)
```
tozka: Why does this work? Should we not get rounding errors ?

tozka:
I know the PR is about decimal, but it might be a good moment to think about other common types - e.g. timestamps and such . If they'd break something.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
gageorgiev: What rounding errors would we get?

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
gageorgiev: The __verify_payload_format changes are covered by the existing ingestion tests (I verified this in the coverage), added a unittest for the DecimalJsonEncoder class.

tozka: > The __verify_payload_format changes are covered by the existing ingestion tests

The fact that we have a bugfix in progress shows that the existing tests do *not* cover the error. Otherwise we'd not be fixing it :) 

The test coverage is misleading. It simply shows that we have visited that line of code. But we have visited with another argument (clearly not decimal). We need tests that verifies the behaviour of the ingest when it's passed decimal values.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
tozka: It depends on behaviour of float. 
Decimal is a  fixed point representation of a double (real number). While float is a floating point representation of a number. 
Sometimes when you cast integer or decimal to float it may be different e.g: cast(1.5 as float) == 1.500000000000001 
It depends - see this problem in c# https://stackoverflow.com/questions/8284275/casting-float-to-decimal-loses-precision-in-c-sharp 

So my question for you is it possible to have 

```
string_json = json.dumps(dict_with_decimal, DecimalJsonEncoder)
json_back = json.laods(string_json). 
```
Could some of the above situation happens? If yes, we need to document it in DecimalJsonEncoder

gageorgiev: I'll include a comment; I don't think this issue would be a problem as we only use this class for the check, not for any actual serialization.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
gageorgiev: This is correct, I've amended of the tests to include a Decimal value.

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
```diff
@@ -14,7 +15,7 @@
 
 shared_test_values = {
     "test_payload1": {"key1": "val1", "key2": "val2", "key3": "val3"},
-    "test_payload2": {"key1": 1, "key2": 2, "key3": 3},
+    "test_payload2": {"key1": Decimal(1), "key2": 2, "key3": 3},
```
tozka, line 18: If you revert your change, will the modified test fail? This is sure way to 
A) check that the test is indeed verifying the bug you have. 
B) that you have really fixed the bug properly

I suspect it should fail in this case but let's check.

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
```diff
@@ -14,7 +15,7 @@
 
 shared_test_values = {
     "test_payload1": {"key1": "val1", "key2": "val2", "key3": "val3"},
```
tozka, line 17: You should do this in another PR. But I would like to extend the test to cover all possible types . Currently only int, string and now decimal are covered.

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
```diff
@@ -14,7 +15,7 @@
...
-    "test_payload2": {"key1": 1, "key2": 2, "key3": 3},
+    "test_payload2": {"key1": Decimal(1), "key2": 2, "key3": 3},
     "test_expected_payload1": [{"key1": "val1", "key2": "val2", "key3": "val3"}],
     "test_expected_payload2": [{"key1": 1, "key2": 2, "key3": 3}],
```
tozka, line 20: That seems wrong. SHouldn't key1 be "Decimal" as well ? We are changing the payload ?

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
gageorgiev: I've checked this, it does indeed fail.

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
gageorgiev: amended

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
gageorgiev: will create story for this

tozka (approved):

PR #1112, opened 2022-08-28
vdk-impala: Add validation for queries that doesn't provide lineage info

kostoww:
# Why?
Checking for lineage information is gathered from the profile making one more request to the impala server, which in some point the number of requests might be huge, and there might not be any lineage data at all.

# What?
Previously there was just a validation for keep-a-live connections. Current commit is adding fine validation also if query is of DCL or some DDL type (REFRESH, COMPUTE, EXPLAIN, etc) which could be frequently used in a data jobs and leading to a huge load on the impala server.

# How has this been tested?

Improvement


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py
```diff
@@ -111,8 +111,19 @@ def _is_query_have_lineage(query_statement: str) -> bool:
             if line.endswith("select 1 -- testing if connection is alive."):
```
tozka, line 111: "testing if connection is alive" - this is not standard comment - it's custom comment made by one user of VDK . It cannot be part of the open source project. 

Quire reminder toward the contributing guide: https://github.com/vmware/versatile-data-kit/wiki/How-to-prepare-a-new-PR#self-evaluate-changefeature-proposal 

Particularly 

> Before suggesting a change/feature, think if this change only serves your needs, or it is broader, that is good for the project itself because it serves multiple users. If it is a very specific change limited in its usefulness, can it be generalised to meet the needs of more users? If yes, propose the more broadly useful feature.

tozka: That will not catch comments made with "/* comment */" style, right ? 

Which might be ok, if you say so explicitly in the documentation -that we do not want 100% coverage just to reduce the chance of unnecessary calls to db. 

Also 
"\n" is just line ending should we handle "\r" or "\r\n" ?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py
```diff
@@ -111,8 +111,19 @@ def _is_query_have_lineage(query_statement: str) -> bool:
...
                 return False  # Queries for keep alive connections does not provide meaningful lineage data
             if line.startswith(
-                    ("compute", "refresh", "describe", "set", "show", "explain", "invalidate", "revoke", "grant",
-                     "use")):
+                (
+                    "compute",
```
tozka, line 115: How exhaustive is this list ?  Also can you sorted alphabetically ? It would be easier to see what's in. I decided to compare with https://docs.cloudera.com/documentation/enterprise/6/6.3/topics/impala_langref_sql.html#langref_sql and see if we should add something more but it's a bit of headache . So I will just ask you if you've checked the docs and evaluated all possible commands.

tozka (changes requested):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py
kostoww: i think keep alive comment is from this OSS repo in the core: 
https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_connection_base.py#L212
and its executed when you open a connection https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py#L80

tozka: OK. Thanks. I've forgotten. Explain in the code with a comment that it's from <link-here> that's why we are checking it.

Another option would be not to rely on the comment. Someone may change it. If we could change it to just check for "select 1" (which is a frequently used connectivity check not just in vdk) might be better.

kostoww: I agree for the comment, but wouldn't select 1 pass also queries like:
SELECT 1 FROM table;

tozka: That depends on the implementation, right?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py
```diff
@@ -93,6 +93,42 @@ def _get_lineage_data(self, cursor: HiveServer2Cursor) -> Optional[LineageData]:
...
+    def _is_query_have_lineage(query_statement: str) -> bool:
+        """
+        This method checks the query type because not every query
+        has information that could be classified as lineage data
+        and making an extra query for the profile is unnecessary
+        """
+        if query_statement is None:
+            return False
+
+        statement_lines = query_statement.split("\n")
```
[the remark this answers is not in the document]
kostoww: I've already checked the sqlparse but for now it seems over optimization. I will leave it as todo

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py
kostoww: List update, documentation noted

tozka (approved):

kostoww:
The Codacy static code analysis error seems unrelated, impala support COMPUTE stats

kostoww:
@tozka Could you merge this PR? 
Thanks!

tozka:
I had not noticed that htis is a fork. Currently we do not accept PR from forks unless they do not require tests (e.g doc or examples contribution). 

Please join VMware organization in Github. You can do it form here: https://upstreamcontrib.eng.vmware.com/oss 

Then I will add you as contributor. And you will need to report the PR from a branch instead of a fork.

tozka:
> I've joined the vmware organization

I added you as a contributor. Create a new PR from a branch, and you will be able to merge it.

PR #1114, opened 2022-08-29
control-service: refactor job cancellation method due to 404 errors

mrMoZ1:
what: Added a new interface method in the job execution repository that retrieves the executions by comparing ID, data job name and data job team. Switched the Spring data interface methods used in the KubernetesService class' cancelJobExecution method. Fixed a few log calls which had reversed variables for printing e.g it would pring the job instead of the team name and vice versa.

why: The previous implementation would only compare for execution ID, allowing users to cancel job executions of other team/job by providing the correct execution. Moreover the default method would produce a query with INNER JOIN data_job datajob1_ on the DataJob table which in some cases would return 404 not found errors when attempting to cancel job executions. After debugging we noticed that the listExecutions method is not affected by this pitfall and it's query uses a LEFT OUTER join, which the new method also uses. Note below is the old findByID method's produced query and the new method's query for comparison:
```
SELECT datajobexe0_.id                                  AS id1_1_0_,
       datajobexe0_.job_name                            AS job_nam18_1_0_,
       datajobexe0_.end_time                            AS end_time2_1_0_,
       datajobexe0_.job_schedule                        AS job_sche3_1_0_,
       datajobexe0_.job_version                         AS job_vers4_1_0_,
       datajobexe0_.last_deployed_by                    AS last_dep5_1_0_,
       datajobexe0_.last_deployed_date                  AS last_dep6_1_0_,
       datajobexe0_.message                             AS message7_1_0_,
       datajobexe0_.op_id                               AS op_id8_1_0_,
       datajobexe0_.resources_cpu_limit                 AS resource9_1_0_,
       datajobexe0_.resources_cpu_request               AS resourc10_1_0_,
       datajobexe0_.resources_memory_limit              AS resourc11_1_0_,
       datajobexe0_.resources_memory_request            AS resourc12_1_0_,
       datajobexe0_.start_time                          AS start_t13_1_0_,
       datajobexe0_.started_by                          AS started14_1_0_,
       datajobexe0_.status                              AS status15_1_0_,
       datajobexe0_.type                                AS type16_1_0_,
       datajobexe0_.vdk_version                         AS vdk_ver17_1_0_,
       datajob1_.NAME                                   AS name1_0_1_,
       datajob1_.enabled                                AS enabled2_0_1_,
       datajob1_.db_default_type                        AS db_defau3_0_1_,
       datajob1_.description                            AS descript4_0_1_,
       datajob1_.enable_execution_notifications         AS enable_e5_0_1_,
       datajob1_.generate_keytab                        AS generate6_0_1_,
       datajob1_.name_deprecated                        AS name_dep7_0_1_,
       datajob1_.notification_delay_period_minutes      AS notifica8_0_1_,
       datajob1_.notified_on_job_deploy                 AS notified9_0_1_,
       datajob1_.notified_on_job_failure_platform_error AS notifie10_0_1_,
       datajob1_.notified_on_job_failure_user_error     AS notifie11_0_1_,
       datajob1_.notified_on_job_success                AS notifie12_0_1_,
       datajob1_.schedule                               AS schedul13_0_1_,
       datajob1_.team                                   AS team14_0_1_,
       datajob1_.last_execution_duration                AS last_ex15_0_1_,
       datajob1_.last_execution_end_time                AS last_ex16_0_1_,
       datajob1_.last_execution_status                  AS last_ex17_0_1_,
       datajob1_.latest_job_deployment_status           AS latest_18_0_1_,
       datajob1_.latest_job_execution_id                AS latest_19_0_1_,
       datajob1_.latest_job_termination_status          AS latest_20_0_1_
FROM   data_job_execution datajobexe0_
       INNER JOIN data_job datajob1_
               ON datajobexe0_.job_name = datajob1_.NAME
WHERE  datajobexe0_.id = ?
```
And this is the query produced by the new method in the interface:
```
SELECT datajobexe0_.id                       AS id1_1_,
       datajobexe0_.job_name                 AS job_nam18_1_,
       datajobexe0_.end_time                 AS end_time2_1_,
       datajobexe0_.job_schedule             AS job_sche3_1_,
       datajobexe0_.job_version              AS job_vers4_1_,
       datajobexe0_.last_deployed_by         AS last_dep5_1_,
       datajobexe0_.last_deployed_date       AS last_dep6_1_,
       datajobexe0_.message                  AS message7_1_,
       datajobexe0_.op_id                    AS op_id8_1_,
       datajobexe0_.resources_cpu_limit      AS resource9_1_,
       datajobexe0_.resources_cpu_request    AS resourc10_1_,
       datajobexe0_.resources_memory_limit   AS resourc11_1_,
       datajobexe0_.resources_memory_request AS resourc12_1_,
       datajobexe0_.start_time               AS start_t13_1_,
       datajobexe0_.started_by               AS started14_1_,
       datajobexe0_.status                   AS status15_1_,
       datajobexe0_.type                     AS type16_1_,
       datajobexe0_.vdk_version              AS vdk_ver17_1_
FROM   data_job_execution datajobexe0_
       LEFT OUTER JOIN data_job datajob1_
                    ON ( datajobexe0_.job_name = datajob1_.NAME )
WHERE  datajobexe0_.id = ?
       AND datajob1_.NAME = ?
       AND datajob1_.team = ?  
```

testing: We already have a unit test that covers job cancellations, which contained no changed tests locally.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobExecutionRepository.java
```diff
@@ -53,4 +53,13 @@ List<DataJobExecution> findDataJobExecutionsByStatusInAndStartTimeBefore(
           + "GROUP BY dje.status, dje.dataJob")
   List<DataJobExecutionStatusCount> countDataJobExecutionStatuses(
       @Param("statuses") List<ExecutionStatus> statuses, @Param("dataJobs") List<String> dataJobs);
+
+  @Query(
+      "SELECT dje from DataJobExecution dje "
+          + "LEFT JOIN DataJob dj ON dje.dataJob = dj.name "
```
mivanov1988, line 59: Why LEFT JOIN?

mrMoZ1: That's because otherwise the query will produce an INNER JOIN which is the same join that is produced by the findById method. On the other hand when debugging the listJobExecutions method (the method that isn't affected by 404 errors) I noticed that the query it produces uses a LEFT JOIN. Therefore in an attempt to create a query as close as possible to the one used in the listJobExecutions method I've used a LEFT JOIN. All existing tests on job cancellation have passed.

tozka: I may be stupid (still drinking my first coffee). But the change does not make sense to me. I understand that the listJobExecutions is not affected. I don't think it answer the why question. We need to understand the root cause of the problem. 

Why is it not affected. If we get 404 when using inner join - does that mean that we have data job execution for a job that has not corresponding job name in the data_job table ? That seems like a bug , doesn't it?

mrMoZ1: You might not have all the context since we discussed this during one of our dailies. We think this might be a bug related to the following stackoverflow threads:
https://stackoverflow.com/questions/57049480/jpa-repository-findbyid-returns-null-but-the-value-is-exist-on-db
https://stackoverflow.com/questions/62745294/why-findbyid-response-is-id-not-found-but-findall-returns-object-with-this-id

What I did was test this against a running instance of control service and input the followig commands in that same order:

`vdk execute --list -t <existing_team_name> -n <existing_data_job>`

The above mentioned command then returned a list of the data job executions as expected. The latest execution listed was in Running state and had the execution id as is expected. Immediately after running this command I then attempted to cancel the running execution by running:
`vdk execute --cancel -n <existing_data_job> -t <existing_team_name>`
And providing the obtained execution id from the list command to the cancel command. However the --cancel command then returned:
```
Error: ¯\_(ツ)_/¯

what: The Data Job execution '<execution_id_returned_by_list_command> cannot be cancelled.
why: Specified data job execution does not exist.
consequences: API request to cancel Data Job Execution failed.
countermeasures: The provided data job execution does not exist. Make sure the provided execution id is correct. List all data job executions of the data job and make sure you specify one that exists.
```
The same cancel command was then re-tried several times and on the fourth attempt if was successful (meaning that the execution was cancelled successfully). After further investigation I determined that the findById method that is refactored in this merge request was the culprit of the 404 error message. This is when I found the above mentioned stack overflow threads. After this I also checked the list method's produced query and it was different from the findById in that it uses the LEFT JOIN clause, that's why in the refactored query I'm using the LEFT JOIN clause (attempting to be as close as possible to the query generated by the listJobExecutions method)

tozka: That still doesn't answer 
A) Why is the inner join causing a problem ? Inner join selects records that have matching values in both tables. That means we have no match ... Why?
B) Why does it succeed on the 4th try ? It's still the same query right?

tozka (changes requested):
Tests are missing. I assume you've planned some tests?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobExecutionRepository.java
mrMoZ1: It is assumed to be a bug in spring boot's findById method. It was agreed that we should at least try to fix it this way and if it doesn't work we can always revert. I have already spent a lot of time trying to debug this issue but this is what I've come up with. If you can provide any better solution/insight will be much appreciated, but otherwise I think I've reached my limit on this.

mrMoZ1:
> Tests are missing. I assume you've planned some tests?

No since we already have a test that covers cancellation functionality and all tests pass.

tozka:
> No since we already have a test that covers cancellation functionality and all tests pass.

Well of course not. One can never have a bug that is covered by our tests. By definition, our tests would have caught it before it became a bug. Clearly,  we have a gap in the tests. 

We need to reproduce the issue in a test  . Otherwise we have no insurance against regression - that the issue will re-surface again in the future. It also gives us more confidence that the buf fix is working (especially when we are not even sure). 
In this case, most likely some kind of integration test would be meaningful.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobExecutionRepository.java
tozka: Ok. If you write up what steps you took when troubleshooting maybe I can give ideas. 

PS:  My advice is to take notes when troubleshooting (be it in a incident or Root cause analysis). It's good idea because one can review the steps and help. It's also helpful for your own review later. I often spot things I've missed when I do it.

mrMoZ1: I've tried to reproduce it but cannot do so in a test environment. Like I already described the debugging method is as follows:

* Attempt to cancel an execution - cancellation fails, retry a few times it eventually suceeds.
* Inspecting the logs It is evident that the cancellation call fails on line 147 of the JobExecutionServiceClass https://github.com/vmware/versatile-data-kit/blob/61e010960a5a089a4adbf02778e447b17848c943/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java#L145-L147
which is right after the default generated findById method is called and we check the returned optional.
* However if we immediately attempt to call the vdk list method which calls the following custom method in our repository: https://github.com/vmware/versatile-data-kit/blob/61e010960a5a089a4adbf02778e447b17848c943/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java#L213-L215
the job execution that we attempt to cancel is found and returned with status Running. If we try to cancel it again it fails again (with the exact same id returned by the list method) until it succeeds after a few retries. Which tells me that the Execution is present and stored in the database but the findById method doesn't return it. Now I believe this might be an issue with spring boot as is discussed in some of the stackoverfow threads that are linked above. The threads mention similar issues with repository methods returning empty values when data is present. Moreover, I have trouble reproducing this locally (I've tried to populate the test database with some dummy data, however the findById method works locally) The issue only manifests in a live deployment. Any suggestion would be welcome, however my recommendation is to try the refactored method which also fixes users being able to provide any job id for cancellation.

tozka (approved):

PR #1115, opened 2022-08-30
versatile-data-kit: Airflow feature comparison

gageorgiev:
This PR contains a section of the Airflow comparison blogpost,
aimed specifically at comparing the two on features which
VDK shines at. It is made for review only and should not be merged.

### DIFF
--- a/airflow-comparison.md
+++ b/airflow-comparison.md
@@ -1,15 +1,25 @@
-## Data Lineage
+## Intro
+
+What are the differences between these two workload automation engines - and how do they fare in typical data engineering cases? This blog post serves to answer these questions and help you decide which one you should use.
+
+## Functionality Comparison
+
+
+
+## Use Case Comparison
+
+### Data Lineage
 
 Before comparing VDK and Airflow, let's first briefly discuss what data lineage is. Data lineage is the process of monitoring, recording, and visualization of the flow of data through a particular data system. It also broadly refers to a collection of tools, practices and approaches which allow the tracking of the dataflow. Data lineage tools maintain a record of data within its lifecycle, including its source and any transformations applied to it. The main purpose of data lineage is to provide a platform for ensuring data quality, tracing the flow of data from source to destination while tracking any transformations made to it, as well as a way of troubleshooting any errors which arise within it.
 
 OpenLineage is ['an open platform for collection and analysis of data lineage’](https://openlineage.io/). It is the defacto industry standard for lineage collection, and it will be the basis for our comparison of lineage capabilities between Airflow and VDK.
 
-### Lineage in Apache Airflow
+#### Lineage in Apache Airflow
 
 Apache Airflow offers support for OpenLineage compliant lineage data collection through the `airflow-openlineage` package; however, out-of-the-box support is only offered for PostgreSQL, Snowflake, MySQL, BigQuery, and the Great Expectations data quality validation framework. Users are expected to create their own solution or borrow an existing one should they want to collect lineage for queries ran against a different database.
 The `airflow-openlineage` package allows the collection of lineage data for individual jobs within a DAG run. However, no DAG-wide lineage can be collected.
 
-### Lineage in Versatile Data Kit
+#### Lineage in Versatile Data Kit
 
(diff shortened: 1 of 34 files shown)


airflow-comparison.md
```diff
@@ -0,0 +1,47 @@
+## Lineage
+
+Data lineage broadly refers to a collection of tools, practices and approaches which allow the tracking of the flow of data within a data pipeline. Data lineage tools maintain a record of data within its lifecycle, including its source and any transformations applied to it. The main purpose of data lineage is to provide a platform for ensuring data quality, as well as a way of troubleshooting any errors which arise within the data.
```
ivakoleva, line 3: > The main purpose of data lineage is to provide a platform for ensuring data quality, as well as a way of troubleshooting any errors which arise within the data.

I would add data flow & movement traceability from source to destination, with a focus on the transformations made. Data lineage reflects the data transformation journey.

ivakoleva, line 3: > Data lineage tools maintain a record of data within its lifecycle, including its source and any transformations applied to it.

Ideally, data lineage capability e2e test would be data transformation flow replay, I think. If we agree, we may consider additionally elaborating the vdk-lineage README.md. 

Yet the current VDK lineage state of the art described in it:
```
At POC level currently.
Currently, lineage data is collected:
* For each data job run/execution both start and end events including the status of the job (failed/succeeded)
* For each execute query we collect input and output tables.
```

airflow-comparison.md
```diff
@@ -0,0 +1,47 @@
...
+
+OpenLineage is 'an open platform for collection and analysis of data lineage’. It is the defacto industry standard for lineage collection, and it will be the basis for our comparison of lineage capabilities between Airflow and VDK.
```
ivakoleva, line 5: maybe a link to

airflow-comparison.md
ivakoleva, line 3: This section's purpose is to describe the term Data Lineage, yes? Maybe start with indicating you're about to explain it (so a person could skip if now needed and read faster), like adding a question or a statement about intention to describe it

airflow-comparison.md
ivakoleva, line 5: Seems this section could be "optimized" by saying it's defacto the open standard and providing a link, maybe would be sufficient.

airflow-comparison.md
```diff
@@ -0,0 +1,47 @@
...
+
+Apache Airflow offers support for OpenLineage compliant lineage data collection through the `airflow-openlineage` package; however, out-of-the-box support is only offered for PostgreSQL, Snowflake, MySQL, BigQuery, and the Great Expectations data quality validation framework. Users are expected to implement their own solution or borrow an existing one should they want to collect lineage for queries ran against a different database.
```
ivakoleva, line 7: This is the catchy part for me, the actual comparison (purpose) of the article. Maybe start the sentence with the verb "Comparing ..., ..." or similar. Also, comparison usually includes a table-like structure, where you cover different aspects / points you test. It's easier to compare one next to the other.

airflow-comparison.md
```diff
@@ -0,0 +1,47 @@
...
+
+Versatile Data Kit offers a built-in mechanism for sending data to a configured destination. It provides an API for asynchronous and thread-safe automatic serialization, packaging and sending of data collected from a PEP249-compliant database cursor, a API response call, or any serializable Python iterable object. Additionally, users can configure the parallelism of and frequency at which data is sent, allowing very high throughput of sending data.
+
+Apache Airflow does not provide any out-of-the-box solution for ingesting data; providing support for this is offloaded onto the user.
+
+
+
+## Summary
+
+To summarize, Versatile Data Kit and Apache Airflow’s differences arise out of the different abstraction levels they attempt to provide the user with. Airflow is much lower level, more generalized, offering a wider set of capabilities, whereas VDK offers more restricted functionality aimed specifically at providing data engineers with higher level abstractions tailored to their needs.
```
ivakoleva, line 41: I like the overall article vibe, and I like the summary opening statement. I think we should add to it native extensibility, due to the built-in features are also implemented leveraging the extensibility mechanism, therefore it's proven to be very usable.

ivakoleva: Maybe "more restricted" could be transformed to "more opinionated"

Issue #580, opened 2021-11-30
Improve Error classifier logic testing


antoniivanov:
_Originally posted by @tozka in https://github.com/vmware/versatile-data-kit/pull/577#discussion_r758794539_

I think we are missing test cases for error classifying logic :

Possible stack traces (last one is assumed to be the exception)

```
1. vdk.internal
2. job.module
3. vdk.internal (job_input.execute_query for example raises exception)
```
Should be platform error, right? That's already covered by PLATFORM_ERROR_STACKTRACE test

```
1. vdk.internal
2. external library (sqlalchemy, pandas)
3. job.module (called by the external library)
```
This is covered by USER_ERROR_STACKTRACE

```
1. vdk.internal
2. job.module
3. vdk.internal (job_input)
4. external_library
```
Should be platform error . We should add a test. I suspect it will pass.

```
1. vdk.internal
2. job.module
3. vdk.internal (job_input)
4. job.module
```
This is mainly possible in send_xx_for_ingestion as they pass (user provided) iterator which can fail  .
Should be user error.  I suspect the detection won't work.

```
1. vdk.internal
2. job.module
3. vdk.interanl (job_input.ingest)
4. external_library_provided_by_job 
```
Should be user error similar to above but the issue here is harder to detect because we do not know if external_library_provided_by_job is provided by job or by vdk. 

Can you think of other scenarios?

_Originally posted by @tozka in https://github.com/vmware/versatile-data-kit/pull/577#discussion_r758794539_

sabadzhiev:
Triaged. Closed as stale.

PR #1173, opened 2022-09-04
vdk-core: fix error classification of vdk code

tozka:
In case the User code is indirectly responsible (yet closest to the
error in the stack trace), for example using a dependency incorrectly,
the error cause (vdk, or user code) is not recognized correctly.

See more details in https://github.com/vmware/versatile-data-kit/issues/1118 and https://github.com/vmware/versatile-data-kit/issues/580

Testing Done: added functional test covering the scenario

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
@@ -99,7 +99,7 @@ class StepFuncFactory:
                 if func.__name__ == "run":
                     try:
                         log.info("Entering %s#run(...) ..." % filename)
-                        StepFuncFactory.invoke_run_function(func, job_input)
+                        StepFuncFactory.invoke_run_function(func, job_input, step.name)
                         success = True
                         return True
                     finally:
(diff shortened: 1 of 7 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
```diff
@@ -133,7 +134,21 @@ def invoke_run_function(func: Callable, job_input: IJobInput):
             if arg_name in parameter_names
         }
         if actual_arguments:
-            func(**actual_arguments)
+            try:
+                func(**actual_arguments)
+            except BaseException as e:
```
doks5, line 139: Why not catching `Exception`?

doks5 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
tozka: To be able to classy properly things lie All Errors - including things like GeneratorExit, MemoryError, etc.

tozka: Actually only GenertorExit,SystemExit and KeyboardInterrupt inherit BaseException (from the built in exceptions). If user code raises them how should we handle them?

tozka: users will do whatever they want. 

Here we catch it, pass it through classifier to detect the type of error and then re-raise it wrap in the appropriate VDK exception. 

The question is do we want to handle GenertorExit,SystemExit and KeyboardInterrupt the same way?

PR #1175, opened 2022-09-06
vdk-impala: Add validation for queries that doesn't provide lineage info 

kostoww:
Checking for lineage information is gathered from the profile making one more request to the impala server, which in some point the number of requests might be huge, and there might not be any lineage data at all.

# What?
Previously there was just a validation for keep-a-live connections. Current commit is adding fine validation also if query is of DCL or some DDL type (REFRESH, COMPUTE, EXPLAIN, etc) which could be frequently used in a data jobs and leading to a huge load on the impala server.

# How has this been tested?

Improvement


tozka:
Please fix the title of the PR

tozka (approved):

PR #1176, opened 2022-09-07
vdk-core: avoid circular references in print results

tozka:
If there's a circular reference in the data printed by execution_results, json.dumps will fail.
So we are failing back to pprint which can handle them.

pprint would sort the keys for us which leads to more confusing message (e.g the name of the stpe is not first for each step, etc.) That's why we are not using it by default.

Sorting can be disabled but only from 3.8+

Testing Done: extended unit tests

### DIFF
--- a/projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_execution_results.py
+++ b/projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_execution_results.py
@@ -116,3 +116,31 @@ def test_serialization_non_serializable():
 
     result_as_string = result.__repr__()
     assert json.loads(result_as_string) is not None
+
+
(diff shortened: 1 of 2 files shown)


projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_execution_results.py
```diff
@@ -116,3 +116,31 @@ def test_serialization_non_serializable():
 
     result_as_string = result.__repr__()
     assert json.loads(result_as_string) is not None
+
+
+def test_serialization_circular_referecen():
```
gageorgiev, line 121:
```suggestion
def test_serialization_circular_reference():
```

gageorgiev (approved):

PR #1182, opened 2022-09-13
vdk-core: errors occurred and the state (handled or not) context missing

ivakoleva:
User code may trigger a SDK feature in a way that raises a user error. In such a use case, the error could be handled by the cautious user, in runtime. The VDK platform should be aware if user errors are handled in runtime or not, so that affects the data job termination status.

The old errors.BLAMEES (experimental data structure for keeping track of responsibles for errors priorly logged) should be refactored:
* a resolvable context abstraction introduced
* it is memory-managed (singleton, reasoning documented)
* populated by the 3 user-facing `errors.log_*()` methods - in favour of a method designed to populate ErrorMessage entries
* keeps track if the actual exception, along with all properties like who is the actual respobsible persona (user or platform, that was free-form str before), resolved by a data job step or not, etc.

Testing Done: added a functional test covering the
 `simple-query-failed-handled` scenario, added singleton verification
 for the resolvable context; ci/cd

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
@@ -11,6 +11,7 @@ from typing import List
 from vdk.api.job_input import IJobInput
 from vdk.internal.builtin_plugins.run.step import Step
 from vdk.internal.core import errors
+from vdk.internal.core.errors import ResolvableByActual
 
 log = logging.getLogger(__name__)
 
@@ -105,6 +106,15 @@ class StepFuncFactory:
                     finally:
                         if success:
                             log.info("Exiting  %s#run(...) SUCCESS" % filename)
+                            user_errors_in_step_handled = [
+                                i
+                                for i in errors.resolvable_context().resolvables.get(
+                                    ResolvableByActual.USER, []
+                                )
+                                if str(step.file_path) == i.caused_by_step_file
+                            ]
+                            for e in user_errors_in_step_handled:
+                                e.resolved = True
                         else:
                             log.error("Exiting  %s#run(...) FAILURE" % filename)
             log.warn(
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
@@ -29,9 +29,10 @@ class TerminationMessageWriterPlugin:
 
     @hookimpl
     def vdk_exit(self, context: CoreContext, exit_code: int):
+        not_resolved_predicate = lambda e: not e.resolved
         self.write_termination_message(
-            errors.get_blamee_overall(),  # TODO: get this from context
-            errors.get_blamee_overall_user_error(),  # TODO: get this from context
+            errors.get_blamee_overall(not_resolved_predicate),
+            errors.get_blamee_overall_user_error(not_resolved_predicate),
             context.configuration,
         )
 
--- a/projects/vdk-core/src/vdk/internal/core/errors.py
+++ b/projects/vdk-core/src/vdk/internal/core/errors.py
@@ -10,6 +10,7 @@ It defines classes and methods for handling exceptions, and ensuring that there 
 """
 from __future__ import annotations
 
+import enum
 import logging
 import re
 import sys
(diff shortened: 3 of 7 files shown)


projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -146,116 +238,92 @@ def to_html(self) -> str:
         return self._to_string(self._get_template("<br />"))
 
 
-def get_blamee_overall() -> str | None:
+def get_blamee_overall(
```
tozka, line 241: I find the method over-complicated. Can we simplify it. Do we need a predicate ?  

I don't understand its purpose.

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
```diff
@@ -29,9 +29,10 @@ def vdk_configure(self, config_builder: ConfigurationBuilder):
 
     @hookimpl
     def vdk_exit(self, context: CoreContext, exit_code: int):
+        not_resolved_predicate = lambda e: not e.resolved
         self.write_termination_message(
-            errors.get_blamee_overall(),  # TODO: get this from context
-            errors.get_blamee_overall_user_error(),  # TODO: get this from context
+            errors.get_blamee_overall(not_resolved_predicate),
+            errors.get_blamee_overall_user_error(not_resolved_predicate),
```
tozka, line 35: The purpose of those TODOs is to figure out a solution which doesn't require keeping global state - which I think it's very problematic. We can probably refactor it in a way that the error state is kept in the JobContext and that would simplify the codebase. 

This change doesn't address this problem though so I'd keep the TODOs.

projects/vdk-core/src/vdk/internal/core/errors.py
ivakoleva: the reason for introducing the predicate, is backwards-compatibility with older usages, that do not need any object pre-filtering, before returning the object property, simply overloading. I think it's the simplest way, since it does not require any rewriting of old code and usages, yet brings new functionality. The alternative is, to introduce multiple methods that call eachother -> to achieve just the same with one optional predicate argument

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
ivakoleva: @tozka I've described in the PR summary, that I went for resolving those todos, in the scope of this PR. 

So the todos suggest to make errors registry inner non-static of *Context instead), and the proposal in the scope of this PR is to keep the errors context static. The reason is, that errors registry needs to be ready and available at earlier stages of the VDK life-cycle (than any context initialisation) - due to errors could occur earlier, so we need to handle those.

Looking at the `core_hook_spec.py` API, most of the hooks are actually happening before CoreContext and especially JobContext initialization. I think the todos are wrong to move, and my proposal is to cleanup those.

What do you think, do you have anything else in mind?

tozka: > The reason is, that errors registry needs to be ready and available at earlier stages of the VDK life-cycle (than any context initialisation) - due to errors could occur earlier, so we need to handle those.

That makes a lot of sense. Ok. Since I need to shift my thinking I will take a new look at the review. I will do that tomorrow though. Sorry I didn't notice the comment earlier today.

ivakoleva: @tozka ping

projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
```diff
@@ -29,9 +29,10 @@ def vdk_configure(self, config_builder: ConfigurationBuilder):
 
     @hookimpl
     def vdk_exit(self, context: CoreContext, exit_code: int):
+        not_resolved_predicate = lambda e: not e.resolved
         self.write_termination_message(
-            errors.get_blamee_overall(),  # TODO: get this from context
-            errors.get_blamee_overall_user_error(),  # TODO: get this from context
+            errors.get_blamee_overall(not_resolved_predicate),
```
murphp15, line 34: To be it makes intuitive sense to filter based on the not resolved errors.
Why is it in the file 
`projects/control-service/versatile-data-kit/projects/vdk-core/src/vdk/internal/builtin_plugins/notification/notification.py`. 
that you would include resolved errors too?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
```diff
@@ -105,6 +106,15 @@ def run(job_input)
...
+                            user_errors_in_step_handled = [
+                                i
+                                for i in errors.resolvable_context().resolvables.get(
+                                    ResolvableByActual.USER, []
+                                )
+                                if str(step.file_path) == i.caused_by_step_file
+                            ]
+                            for e in user_errors_in_step_handled:
+                                e.resolved = True
                         else:
```
tozka, line 118: The whole "resolved" logic based on step file seems error prone. Note that in the future we can have steps that aren to produced out of files. I actually have a POC on my computer for SQL splitter which splits SQL file into multiple steps using delimiter. 


Why not do "errors.resolvable_context().clear()" on success. Success means the errors from teh stpe are not relevant and can be cleared ?

projects/vdk-core/src/vdk/internal/core/errors.py
tozka: What older usages? This should not be used outside of this git repo ... And we can fix all usages in this repo.

tozka (changes requested):

projects/vdk-core/tests/functional/run/test_run_errors.py
```diff
@@ -33,7 +37,7 @@ def tmp_termination_msg_file(tmpdir) -> pathlib.Path:
 
 
 def test_initialize_step_user_error(tmp_termination_msg_file):
-    errors.clear_intermediate_errors()
+    errors.resolvable_context().clear()
```
murphp15, line 40: I think this needs to be done at the start of absolutely every test. 
And not just tests in this file.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
tozka: We  synced face to face we agree that we should keep the history of the errors (for troubleshooting purposes) but 
A) we should mark all errors as resolved up to this point (regardless of step or user/platform error)

tozka (approved):

PR #1184, opened 2022-09-14
versatile-data-kit: Without and with VDK image

zverulacis:
This PR adds an image of before and after VDK to the README file and has a minor edit on the wording.


gageorgiev (approved):
Wonderful graphics.

PR #1185, opened 2022-09-19
vdk-core: added default values to write termination message method

mivanov1988:
Currently, we have an internal customer experiencing concurrent executions of the same data job. The issue is in the execution logic because of missing default values of the write_termination_message method.

Exception:
```
execution_skip.py:130  _skip_job_if_nec[id:moonshine-stage-cost-optimization-strategies-1663579200-qfxw2]- Error while checking for another data job excecution: write_termination_message() missing 2 required positional arguments: 'error_overall' and 'user_error' 
```

This change aims to fix the issue ASAP since the customer may have duplicate data because of it.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py
@@ -36,7 +36,11 @@ class TerminationMessageWriterPlugin:
         )
 
     def write_termination_message(
-        self, error_overall, user_error, configuration, execution_skipped=False
+        self,
+        error_overall=None,
+        user_error=None,
+        configuration=None,
+        execution_skipped=False,
     ):
         termination_message_writer_cfg = WriterConfiguration(configuration)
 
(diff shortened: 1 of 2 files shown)


dakodakov (approved):

tozka:
I don't understand the issue? Why does setting them explicitly fixes it ? Don't they have defaults ? 

There are tests missing.

mivanov1988:
> I don't understand the issue? Why does setting them explicitly fixes it ? Don't they have defaults ?
> 
> There are tests missing (tests would make it easier to understand the problem)

Currently, they don't have default values.

mivanov1988:
> I don't understand the issue? Why does setting them explicitly fixes it ? Don't they have defaults ?
> 
> There are tests missing (tests would make it easier to understand the problem)

Do you think it is better to add defaults like that?

```
def write_termination_message(
        self, error_overall=None, user_error=None, configuration=None, execution_skipped=False
    ):
```

mrMoZ1 (approved):

tozka:
> > I don't understand the issue? Why does setting them explicitly fixes it ? Don't they have defaults ?
> > There are tests missing (tests would make it easier to understand the problem)
> 
> Do you think it is better to add defaults like that?
> 
> ```
> def write_termination_message(
>         self, error_overall=None, user_error=None, configuration=None, execution_skipped=False
>     ):
> ```

Yes,   the methods should set sensible defaults for optional arguments IMO.

PR #1188, opened 2022-09-23
vdk-core: add ability to cancel remaining job steps

mrMoZ1:
what: Added the ability to cancel the remaining job/template steps through the job_input interface.

why: This feature is required by users in certain cases e.g: if a data job depends on processing data from a source which has indicated no new entries since last run, then we can skip the execution.

testing: Tested locally with data job/templates and added functional tests.


projects/vdk-core/src/vdk/api/job_input.py
```diff
@@ -361,3 +361,18 @@ def get_execution_properties(self) -> dict:
                 SELECT id, {pa__job_start_ts_expr} FROM mytable
         """
         pass
+
+    @abstractmethod
+    def cancel_job_execution(self) -> None:
```
tozka, line 366: We should not that in case of a template, it will cancel the template being executed but the parent job (or template) will not be canceled).

tozka, line 366: Let's name it "cancel_remaining_steps" ? Since canceling template execution will not cancel the whole job and it may confuse people.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -76,6 +77,9 @@ def run_step(context: JobContext, step: Step) -> StepResult:
                 if step_executed
                 else ExecutionStatus.NOT_RUNNABLE
             )
+        except CancelJobExecutionException as e:
+            # We need to re-raise this exception so it can be handled in the run_job level
+            raise e
```
tozka, line 82: I am not sure about this isn't it better to mark the StepResult with status "canceled" . 

or rather "cancel_requested" as this is the accurate state at this point. 

Then in run_job it will actually cancel everything beloynd.

Otherwise other hook implementation will think it has failed or they'd need to check explicitly for this exception.

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -107,6 +107,17 @@ class UserCodeError(DomainError):
     pass
 
 
+class CancelJobExecutionException(DomainError):
```
tozka, line 110: We have an API for canceling job execution. as this is not doing that, it would confuse users. 

Better names I think are "cancel remaining steps" or "skip remaining steps"

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -399,6 +410,8 @@ def find_whom_to_blame_from_exception(exception: Exception) -> ResolvableBy:
         )  # TODO find out if this is a local or platform deployment and fix this line.
     if issubclass(type(exception), PlatformServiceError):
         return ResolvableBy.PLATFORM_ERROR
+    if issubclass(type(exception), CancelJobExecutionException):
+        return ResolvableBy.USER_ERROR
```
tozka, line 414: I don't like this. CancelJobExecution (or cancel remaining steps) is supposed to be "success" as written in the doc. 

We are using exception for control flow which is anti pattern as very well argumented [here](https://dzone.com/articles/exceptions-as-controlflow-in-java) or [here](https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why) or [here](https://web.archive.org/web/20140430044213/http://c2.com/cgi-bin/wiki?DontUseExceptionsForFlowControl).  And the bad side effects show up here.

But in our case I don't think we have another "lean" solution. But we can reduce the scope. 

I think this check might not be necessary if we handle it in "run_step" and return simply status (as I wrote in above comment).  Then nobody else but run_step would know the exception is raised ...

projects/vdk-core/tests/functional/run/test_run_templates.py
```diff
@@ -59,3 +58,38 @@ def test_run_job_with_template_error(tmpdir: py.path.local):
...
+    Test checks:
+    template execution is canceled and all job steps are executed.
+    template steps prior to the cancel_job_execution() call are executed.
+    proper exit code is set.
+    proper message is logged confirming template was cancelled from cancel_job_execution() method.
+    :return:
+    """
+    test_runner = CliEntryBasedTestRunner(
+        TemplatePlugin(
+            "template-cancel", pathlib.Path(util.job_path("template-cancel"))
```
tozka, line 80: nit: though. this is also fine , you can also reuse "job-cancel" as a template here instead of creating new job "template-cancel"

tozka (changes requested):

tozka:
I think this is  a cool new feature. I've talked with some users in the past who have complained they cannot stop a job in the middle. We should popularize it after releasing it. Thanks for coming up with it.

projects/vdk-core/src/vdk/api/job_input.py
```diff
@@ -361,3 +361,22 @@ def get_execution_properties(self) -> dict:
...
+
+    @abstractmethod
+    def skip_remaining_steps(self) -> None:
+        """
+        :return:
+
+        Usage:
+            job_input.skip_remaining_steps()
+
+            Raises an internal exception that signalises to the VDK runtime that all remaining steps should be skipped
```
tozka, line 373: nit: "Raises an internal exception" seems like epxosing an implementation detail. I think say "Singalizes to the VDK runtime that all remaining steps should be skipped.
```suggestion
            Signalizes to the VDK runtime that all remaining steps should be skipped
```

projects/vdk-core/src/vdk/internal/core/errors.py
tozka: Did you forget to rename it ?

tozka (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
```diff
@@ -146,7 +146,7 @@ def invoke_run_function(func: Callable, job_input: IJobInput, step_name: str):
                     what_happened=f"Data Job step {step_name} completed with error.",
                     why_it_happened=errors.MSG_WHY_FROM_EXCEPTION(e),
                     consequences="I will not process the remaining steps (if any), "
-                    "and this Data Job execution will be marked as failed.",
```
tozka, line 149: I didn't notice this. That's why I dislike the exception approach for control flow (as i mentioned before). 

Isn't it better to catch teh exception and handle it, re-raise it. I think this will print an error in the logs - but we have not error ?

PR #1189, opened 2022-09-23
vdk-impala: fix impala template empty source view usr err 

mrMoZ1:
what: The template arguments validator now makes use of the cancel_job_execution() method when an empty source view is returned.

why: Users of the template requested that empty source view stopped resulting in a User Error, after a discussion it was agreed that this is the correct and desired behaviour.

testing: Trivial, however this change depends on https://github.com/vmware/versatile-data-kit/pull/1188

Signed-off-by: Momchil Zhivkov [mzhivkov@vmware.com](mailto:mzhivkov@vmware.com)


tozka:
I will never tire asking this :) Tests ? 

Any bug we fix should have tests. I am surprised that no tests fail (clearly we have not added tests covering this scenario - what happens if SQL view returns 0 rows) but now it's a good opportunity  to do so

mrMoZ1:
> I will never tire asking this :) Tests ?
> 
> Any bug we fix should have tests. I am surprised that no tests fail (clearly we have not added tests covering this scenario - what happens if SQL view returns 0 rows) but now it's a good opportunity to do so

Tests will likely fail. And I will update the MR accordingly.

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
```diff
@@ -417,9 +417,12 @@ def test_load_fact_snapshot_empty_source(self) -> None:
             },
         )
         # Expecting data job not to finish due to empty source.
-        assert res.exception
-        assert "Source view returned no results." in res.exception.args[0]
-        assert isinstance(res.exception, errors.UserCodeError)
+        assert not res.exception
+        assert "Source view returned no results." in res.output
+        assert (
```
tozka, line 422: Let's not get used to verifying by logs. That's not a good practice as we can change the logs to make them more user friendly or changing their log level and their format is not fixed . It's a last option in case there's no other way. 

Another reason is we can log "execution was skipped" but it may not be skipped (maybe we have a bug and the steps are not actually skipped) . The log is misleading but the test is false positive. 

I think in this case checking that exist status is success and that there's no exception is  enough. The template with empty view has succeeded.

tozka (approved):

tozka:
Some small comment around test verification but looks good to me. Thanks.

Issue #1187, opened 2022-09-21
[vdk-server] kind error - resource mapping not found for name: "vdk-cockroachdb-budget"


zverulacis:
**Describe the bug**
When I'm installing vdk server on Windows WSL 2 following the instructions in [wiki](https://github.com/vmware/versatile-data-kit/wiki/Installation#install-versatile-data-kit-control-service)
`vdk server --install` command fails with an error: 

```
v-error: Stderr output: Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "vdk-cockroachdb-budget" namespace: "default" from "": no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"
error: ensure CRDs are installed first
```

**Steps To Reproduce**
1. Install and set up WSL 
2. Install helm, docker, kind
3. run `vdk server --install`

Full output: 
```
vdk server --install
Starting installation of Versatile Data Kit Control Service (this may take several minutes)
Creating Kind cluster "vdk"...
Kind cluster "vdk" created
Creating Docker registry container "vdk-docker-registry"...
Docker registry container "vdk-docker-registry" created
Creating Git server container "vdk-git-server"...
Git server container "vdk-git-server" created
Git server configured successfully
Installing prerequisites...
Done
Installing Control Service in Kind cluster...
-error: Stderr output: Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "vdk-cockroachdb-budget" namespace: "default" from "": no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"
error: ensure CRDs are installed first
```

**Expected behavior**
Installation should complete successfully

**Version  (please complete the following information):**
 - OS: Windows
 - WSL version 2
 - VDK Version

```
Versatile Data Kit (VDK)
Version: 0.3.572339461
Build details: RELEASE_VERSION=0.3.572339461, BUILD_DATE=Fri Jun 24 14:57:27 UTC 2022, BUILD_MACHINE_INFO=Linux runner-opyxurxq-project-28359933-concurrent-0wcg7j 5.4.181-99.354.amzn2.x86_64 #1 SMP Wed Mar 2 18:50:46 UTC 2022 x86_64 GNU/Linux, GITLAB_CI_JOB_ID=2637062491, GIT_COMMIT_SHA=e8e2359b407acbef5982e190e38c9a9af51fa2b2, GIT_BRANCH=main
[...]

PR #1190, opened 2022-09-26
vdk-properties-fs: new plugin for local FS properties storage

ivakoleva:
We need a simplistic plugin, that allows a developer or presenter to quickly store secrets on the local FS during a dev/demo session, that does not require a prerequisite of the entire Control Service installed and running.

Added a FS properties client, that uses a customizable dir and filename on the local file system to store the properties key-values. It does support multiple teams and data jobs, the properties of those are placed within a dedicated section for keys deduplication.

Testing Done: added `test_fs_properties_plugin.py`, that covers multiple data jobs using matching keys, and default/customized directory and filename for file storage on the local FS.

### DIFF
--- a/projects/vdk-plugins/vdk-properties-fs/README.md
+++ b/projects/vdk-plugins/vdk-properties-fs/README.md
@@ -1,6 +1,6 @@
 This plugin allows vdk-core to read/write properties on the local FS. Mainly for development purposes,
 to simplify a use-case with local Properties API usage, that would otherwise require a Control Service instance prerequisite.
-For example, to quick demo a data job that uses a secret, so that secret is quickly stored on the presenter local FS.
+For example, to quickly demo a data job that uses a secret, so that secret is quickly stored on the presenter's local FS.
 
 # Usage
 
--- a/projects/vdk-plugins/vdk-properties-fs/README.md
+++ b/projects/vdk-plugins/vdk-properties-fs/README.md
@@ -1,0 +1,36 @@
+This plugin allows vdk-core to read/write properties on the local FS. Mainly for development purposes,
+to simplify a use-case with local Properties API usage, that would otherwise require a Control Service instance prerequisite.
+For example, to quick demo a data job that uses a secret, so that secret is quickly stored on the presenter local FS.
+
+# Usage
+
+Run
+```bash
+pip install vdk-properties-fs
+```
+
+After this, data jobs will be able to read and write properties stored on the local file system.
+
+For example
+
+```python
+    def run(job_input: IJobInput):
+        my_props = {"key": "value"}
+        job_input.set_all_properties(my_props)
+        assert job_input.get_all_properties() == my_props
+        assert job_input.get_property("key") == "value"
+```
+
+# Configuration
+
+To enable the plugin, set "PROPERTIES_DEFAULT_TYPE" to "fs-properties-client".
+Run vdk config-help - search for those prefixed with "FS_PROPERTIES_" to see what configuration options are available.
+
+# Testing
+
+Testing this plugin locally requires installing the dependencies listed in vdk-plugins/vdk-properties-fs/requirements.txt
+
+Run
+```bash
+pip install -r requirements.txt
+```
--- a/projects/vdk-plugins/vdk-properties-fs/setup.py
+++ b/projects/vdk-plugins/vdk-properties-fs/setup.py
@@ -1,0 +1,34 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import pathlib
+
+import setuptools
+
+
+__version__ = "0.0.1"
+
+setuptools.setup(
+    name="vdk-properties-fs",
+    version=__version__,
+    url="https://github.com/vmware/versatile-data-kit",
+    description="Versatile Data Kit SDK plugin provides support for Properties API client that uses local FS storage.",
+    long_description=pathlib.Path("README.md").read_text(),
+    long_description_content_type="text/markdown",
+    install_requires=["vdk-core"],
+    package_dir={"": "src"},
+    packages=setuptools.find_namespace_packages(where="src"),
+    include_package_data=True,
+    entry_points={
+        "vdk.plugin.run": [
+            "vdk-properties-fs = vdk.plugin.properties_fs.fs_properties_plugin"
+        ]
+    },
+    classifiers=[
+        "Development Status :: 4 - Beta",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
+    ],
+)
--- a/projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
+++ b/projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
@@ -1,9 +1,10 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+import json
 import logging
 import os
-from configparser import ConfigParser
 from typing import Dict
+from typing import Optional
 
 from vdk.api.plugin.plugin_input import IPropertiesServiceClient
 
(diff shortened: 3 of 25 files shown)


ivakoleva:
Useful for workaround in cases of https://github.com/vmware/versatile-data-kit/issues/1187 for example

projects/vdk-plugins/vdk-properties-fs/README.md
```diff
@@ -0,0 +1,36 @@
+This plugin allows vdk-core to read/write properties on the local FS. Mainly for development purposes,
+to simplify a use-case with local Properties API usage, that would otherwise require a Control Service instance prerequisite.
+For example, to quick demo a data job that uses a secret, so that secret is quickly stored on the presenter local FS.
```
gageorgiev, line 3:
```suggestion
For example, to quickly demo a data job that uses a secret, so that secret is quickly stored on the presenter's local FS.
```

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
```diff
@@ -0,0 +1,60 @@
...
+    A single properties file store, supporting multiple teams and data jobs properties,
+    via placing the key-values in a dedicated team and job name section to deduplicate.
+    """
+
+    def __init__(self, directory: str, filename: str):
+        self._file_path = os.path.join(directory, filename)
+
+    def read_properties(self, job_name: str, team_name: str) -> Dict:
+        config_parser = ConfigParser()
+        config_parser.read(self._file_path)
```
gageorgiev, line 25: Wouldn't this fail if you try to read properties before writing any? Not sure if ConfigParser.read fails if the provided file path doesn't exist, maybe my comment is wrong.

gageorgiev (approved):
Looks good, just one small comment about the properties client.

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
```diff
@@ -0,0 +1,60 @@
...
+        # filter properties by prefix for this particular team and data job
+        prefix = FileSystemPropertiesServiceClient._prefix(team_name, job_name)
+        if config_parser.has_section(prefix):
+            return dict(config_parser.items(prefix))
+        return {}
+
+    def write_properties(self, job_name: str, team_name: str, properties: Dict) -> Dict:
+        log.warning(
+            "You are using local FS Properties client. "
+            "That means the properties will not be stored using the Control Service."
```
tozka, line 36: nit: I think this line is unnecessary as Control Service is just one potential backend. There can be another properties backends in the future that are remote . We should not that the props are stored only locally so it will not affect a deployed job perhaps.

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
```diff
@@ -0,0 +1,60 @@
...
+            "The properties will be read and written to the local file system only."
+        )
+        config_parser = ConfigParser()
+        config_parser.read(self._file_path)
+
+        # filter properties by prefix for this particular team and data job
+        prefix = FileSystemPropertiesServiceClient._prefix(team_name, job_name)
```
tozka, line 43: I think we should keep that as JSON and not as in file. 

ini structure is much more limited - no types support (which properties have) no nested support (properties can be arbitrary nested) . 

And it's pretty easy to implement: `json.dumps(properties)` and `json.loads(properteis)`. 

20 lines become 1 :)

tozka (changes requested):

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
murphp15: Yeah it shouldn't fail.
from the docs 
```
        Files that cannot be opened are silently ignored; this is
        designed so that you can specify an iterable of potential
        configuration file locations (e.g. current directory, user's
        home directory, systemwide directory), and all existing
        configuration files in the iterable will be read.
```

murphp15 (approved):
LGTM

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
tozka: I am ok to merge it as it is. But let's create a story/github issue for future improvement

projects/vdk-plugins/vdk-properties-fs/setup.py
```diff
@@ -0,0 +1,34 @@
...
+    package_dir={"": "src"},
+    packages=setuptools.find_namespace_packages(where="src"),
+    include_package_data=True,
+    entry_points={
+        "vdk.plugin.run": [
+            "vdk-properties-fs = vdk.plugin.properties_fs.fs_properties_plugin"
+        ]
+    },
+    classifiers=[
+        "Development Status :: 4 - Beta",
```
tozka, line 27: I'd mark it as alpha, especially if we are leaving TODOs for the future

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
ivakoleva: kept the comment, however added info about stateful storage (oppositely to vdk-core's in-memory one)

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
ivakoleva: I've decided to use JSON, for fully supporting nested data structures.

projects/vdk-plugins/vdk-properties-fs/src/vdk/plugin/properties_fs/fs_properties_client.py
```diff
@@ -0,0 +1,66 @@
...
+            "all properties data will survive VDK process restarts."
+            "The properties will be read and written to the local file system only."
+        )
+        props: Optional[Dict] = None
+
+        # init storage file
+        if not os.path.isfile(self._file_path) or os.stat(self._file_path).st_size == 0:
+            with open(self._file_path, "w+") as props_file:
+                props = {}
+                json.dump(props, props_file)
```
tozka, line 49: nit: You can ignore now but consider using https://docs.python.org/3/library/pathlib.html#pathlib.Path , it has methods for reading and writing into files from text that is single line and handles closing/opening file streams

projects/vdk-plugins/vdk-properties-fs/tests/test_fs_properties_plugin.py
```diff
@@ -0,0 +1,97 @@
...
+@patch.dict(
+    os.environ,
+    {
+        PROPERTIES_DEFAULT_TYPE: "fs-properties-client",
+    },
+)
+def test_multiple_properties_read_write_config_default():
+    runner = CliEntryBasedTestRunner(fs_properties_plugin)
+    # default file path expected
+    file_path = os.path.join(tempfile.gettempdir(), "vdk_data_jobs.json")
```
tozka, line 38: tempfile doesn't work very well since it does not actually delete the files ,. It simply places them in temp directory.

For python tests it's recommended to use tmp path fixture - https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html  

Probably something like this: 
```suggestion
def test_multiple_properties_read_write_config_default(tmp_path):
    runner = CliEntryBasedTestRunner(fs_properties_plugin)
    # default file path expected
    file_path = os.path.join(str(tmp_path), "vdk_data_jobs.json")
```

tozka (approved):
Looks good to me. Some small comments/suggestions

projects/vdk-plugins/vdk-properties-fs/tests/test_fs_properties_plugin.py
ivakoleva: I'm finalising the test file resources (except in case of reusing the default file path); will be merging since we need now the plugin released for demo prep

PR #1196, opened 2022-09-28
Person/zverulacis/GitHub stars example

zverulacis:
This PR will add the GitHub history app to the examples


examples/github-stars-example/README.md
```diff
@@ -0,0 +1,307 @@
...
+Data Job with name local-github-stars created locally in /home/gita/vdk/local-github-stars.
+```
+4. Clear the new directory and copy example files to it
+```
+rm local-github-stars/*
+cp versatile-data-kit/examples/github-stars-example/github-stars-example-job/* local-github-stars/
+```
+5. Add your GitHub token and repo name to the file 00_properties.py
+```
+vim local-github-stars/00_properties.py
```
murphp15, line 41: IMO it is strange that you would be expected to add a secret in a file. 
As opposed to it reading it from an env_variable or something similar to that.

examples/github-stars-example/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+Job Name: local-github-stars
+Job Team: test
+Path to where sample data job will be created locally [/home/gita/vdk]:
+Data Job with name local-github-stars created locally in /home/gita/vdk/local-github-stars.
+```
+4. Clear the new directory and copy example files to it
+```
+rm local-github-stars/*
+cp versatile-data-kit/examples/github-stars-example/github-stars-example-job/* local-github-stars/
```
murphp15, line 37: why don't they just run it from the examples folder? 
what is the advantage of copying it somewhere else?

examples/github-stars-example/README.md
zverulacis: Is it an issue? Both ways are ok. I would prefer having it in the file.

examples/github-stars-example/README.md
zverulacis: If I remove that part, I can also remove this. Maybe second opinion could be good.

murphp15: Ah I see!

examples/github-stars-example/README.md
murphp15: I think it is an issue. 
Even though this is a toy example it is supposed to guide others in how to write production jobs and you would never be expected to put a secret in a file in a production job. 

I think you should probably load it from the filesystem using @ivakoleva's plugin?

examples/github-stars-example/github-stars-example-job/00_properties.py
```diff
@@ -0,0 +1,15 @@
...
+# SPDX-License-Identifier: Apache-2.0
+import logging
+
+from vdk.api.job_input import IJobInput
+
+log = logging.getLogger(__name__)
+
+
+def run(job_input: IJobInput):
+    properties = job_input.get_all_properties()
```
ivakoleva, line 11: no need to explicitly load the properties before setting them, in case the loaded (properties var) are not used;
you can drop this line

examples/github-stars-example/README.md
tozka: Storing credentials as part of the source code is a [bad](https://pvs-studio.com/en/docs/warnings/v5601/) [practice](https://medium.com/twodigits/keep-passwords-out-of-source-code-why-and-how-e84f9004815a) so a) we definitely should not encourage it b) it will make a bad impression for our project if someone who knows about this stuff is looking through those. 

The way our users store credentials is to use Properties API but they do it using CLI or API/UI : 

So in the 00_properties.py file, we should only set the API URL perhaps and instruct users to use 
```
export properties_default_type=fs-properties-client
vdk properties --set-secret "token" # will be prompted for the actual token
```
See vdk properties --help for more info.

ivakoleva: For me it makes sense the way it is implemented now. However a fresh view is probably more valuable in terms of presenting examples to new users.

maybe (1) mark vdk-properties-fs for local dev/demo purposes only (requirements.txt, config.ini); 
and (2) instead of having a file with placeholder -> add instructions how to initialize the properties before data job run/deploy - using the VDK CLI commands, so that the data job reuses those secrets after? that includes toggling the properties client config (between the new and default plugin)

ivakoleva (approved):

examples/github-stars-example/README.md
tozka: As I mentioned in the other comment let users use CLI command `vdk properteis` and they should not modify the source code.

examples/github-stars-example/github-stars-example-job/requirements.txt
```diff
@@ -0,0 +1,6 @@
+pandas
+plotly
+pygithub
+streamlit
+vdk-core
```
tozka, line 5: vdk-core is not necessary here. AND is not advisable to add it as vdk version is managed by the platform when deployed

examples/github-stars-example/README.md
tozka: Another option is to use arguments : 
```
vdk run --arguments '{"token":"xxx", "repo_path":"versatiledatakit/foo"}'
``` 

Arguments are then accessed in the job step using `job_input.get_arguments()` method 


Properties is better because `vdk properties --set-secret `  prompts you for the secret. While in arguments you set it in the command line.  Supplying passwords via command line argument is considered bad security practices because the password itself is leaked into the process table and/or in the shell history file. If curious you can read more details here - https://www.netmeister.org/blog/passing-passwords.html 

Another reason it's better is that you set it once only no need to supply them every time. 
Now of course local fs properties do store them in a file. But if one uses Control Service it won't. Besides in the future we may encrypt the file if FS properties are really used more frequently. 

In any case, the passing arguments approach is still better than setting them in the source file. 

This is more of FYI sharing what all the options are. I think we should use the properties CLI for this example.

examples/github-stars-example/README.md
```diff
@@ -0,0 +1,307 @@
...
+├── 00_properties.py
+├── 10_drop_table.sql
+├── 20_create_table.sql
+├── 30_stargazers.py
+├── 40_build_streamlit_dashboard.py
+├── config.ini
+├── requirements.txt
+```
+
+<details>
```
tozka, line 130: Are those details needed ? They are just repeating the source code but now we need to keep then in sync ... ?

ivakoleva:
Should we merge the example?

ivakoleva:
I think we should merge the example, it was demoed on the public channel and there is a chance people are looking for it in the code base? cc @zverulacis

examples/github-stars-example/README.md
zverulacis: When I use properties CLI, the properties get set, but the job fails with an error: None
I don't know how to fix it, maybe someone else can test? I need some help with this. 

```
    {
      "name": "30_stargazers.py",
      "type": "python",
      "start_time": "2022-11-09T10:44:25.686497",
      "end_time": "2022-11-09T10:44:25.715008",
      "status": "error",
      "details": "An exception occurred, exception message was: An error in data job code  occurred. The error should be resolved by User Error. Here are the details:\n  WHAT HAPPENED : Data Job step 30_stargazers.py completed with error.\nWHY IT HAPPENED : An exception occurred, exception message was: None\n   CONSEQUENCES : I will not process the remaining steps (if any), and this Data Job execution will likely be marked as failed.\nCOUNTERMEASURES : See exception and fix the root cause, so that the exception does not appear anymore.",
      "exception": {
        "is_logged": true,
        "cause_exception_name": "AssertionError",
        "cause_exception": {
          "is_logged": true
        }
      },
      "blamee": "User Error",
      "exception_name": "UserCodeError"
    }
  ],
  "exception": null,
  "blamee": "User Error",
  "exception_name": "UserCodeError"
}

Comment (tozka) [reply examples/github-stars-example/README.md:41]: In config,ini there is 

`properties_default_type=fs-properties-client` 

This controls what is the source for the properties of your data job. 

When you are setting them manually with `vdk properties --set`, if the same is not set as env variable you may be setting them somewhere else (by default control service perhaps)> 

Would something like this work ?
```
export properties_default_type=fs-properties-client
vdk properties --set ...
```

examples/github-stars-example/github-stars-example-job/00_properties.py
zverulacis: file removed

examples/github-stars-example/README.md
zverulacis: I wish to keep this part for  2 reasons: 
1. to show how easy it is to create a job
2. to keep the structure of the file.

I'll remove this if you find it necessary

examples/github-stars-example/README.md
zverulacis: I fixed the code, maybe someone else can test it. I can't run this job, it still fails with error; None even after setting the properties_default_type to fs-properties-client with export. 
I don't have the control service installed where I'm running the job.

ivakoleva: Is this resolved?

ivakoleva (changes requested):
I've also tested myself as requested,
I've met setup issues that had to solve, then an issue with 30_stargazers.py GitHub lib:
`github.GithubException.UnknownObjectException: 404 {"message": "Not Found", "documentation_url": "https://docs.github.com/rest"}`

My suggestion is, please create a new python environment, and replay the steps in README.md to verify. For example, if the intention is to use properties-fs, then you should install it first (`pip instal -r requirements.txt` from step 6.) and set as default client -> and only afterwards set the properties values (step 5. currently).

examples/github-stars-example/README.md
zverulacis: No, it doesn't work for me

murphp15: Looking at this now

tozka:
Closing the PR for now as there has not been activity for more than a month, and it's currently not in progress. 
Feel free to re-open it when it's back on the table

PR #1202, opened 2022-09-30
control-service: remove executables from secure job builder

mivanov1988:
Why: In order to reduce the attack surface, we need to remove all unnecessary system executables after the job building.

What: The change aims to remove all left and unnecessary executables after job building - chown, uname and pip.

Testing done: built image locally.


gageorgiev (approved):
What about chmod? Also, fix the name of your PR.

mivanov1988:
> What about chmod? Also, fix the name of your PR.

The base image contains only chwon, uname, sh and python.

PR #1204, opened 2022-10-03
vdk-plugins: fix broken link

tozka:
The link to the cookiecutter repo was wrong.


projects/vdk-plugins/README.md
```diff
@@ -32,7 +32,7 @@ pip install -U cookiecutter
 Generate a VDK Plugin package project:
 
 ```
-cookiecutter https://github.com/tozkata/cookiecutter-vdk-plugin.git
+cookiecutter https://github.com/tozka/cookiecutter-vdk-plugin.git
```
ivakoleva, line 35: this slipped last time https://github.com/vmware/versatile-data-kit/pull/942/files#r949007999
is cookiecutter able to work with subdirectories, so that we can integrate this into the current repository?

ivakoleva (approved):

projects/vdk-plugins/README.md
tozka: It may work. Not sure. But our repo is very big, it will be very slow. I toyed with the idea to use a branch (with only cookie-cutter code inside) but for now, we can use my repo. I generously would maintain it for VDK sake :)

PR #1210, opened 2022-10-04
vdk-plugins: run tests on release of vdk-core

tozka:
Changes in vdk-core can break some plugins. We need some functionality that enables us to catch such regression quicker.

The proposoed change is to introduce in .plugin-common.yml new generic "job" template `.build-plugin-on-vdk-core-release` which can be used by plugins CI.

As it should not be necessary to run all plugin tests, we can select some more important

Testing Done: CI

### DIFF
--- a/projects/vdk-plugins/vdk-impala/.plugin-ci.yml
+++ b/projects/vdk-plugins/vdk-impala/.plugin-ci.yml
@@ -4,15 +4,9 @@
 image: "python:3.7"
 
 .build-vdk-impala:
-  image: docker:19.03.8
-  services:
-    - docker:19.03.8-dind
   variables:
-    DOCKER_HOST: tcp://localhost:2375
-    DOCKER_DRIVER: overlay2
-    DOCKER_TLS_CERTDIR: ""
     PLUGIN_NAME: vdk-impala
-  extends: .build-plugin
+  extends: .build-plugin-dind
 
 build-py37-vdk-impala:
   extends: .build-vdk-impala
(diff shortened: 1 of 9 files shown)


projects/vdk-plugins/vdk-impala/.plugin-ci.yml
```diff
@@ -4,15 +4,9 @@
...
-  image: docker:19.03.8
-  services:
-    - docker:19.03.8-dind
   variables:
-    DOCKER_HOST: tcp://localhost:2375
-    DOCKER_DRIVER: overlay2
-    DOCKER_TLS_CERTDIR: ""
     PLUGIN_NAME: vdk-impala
-  extends: .build-plugin
+  extends: .build-plugin-dind
```
murphp15, line 9: what does `dind` stand for here?

tozka: Docker in Docker (also known as dind) is, as the name implies, running Docker on top of a Docker container. We are starting docker containers inside a docker container. 

All Gitlab runners and jobs are docker containers (Kubernetes pods really) .

dakodakov:
Hey @tozka ,
What about the vdk-csv and vdk-lineage plugins? I believe they are quite important as well?

tozka:
> Hey @tozka , What about the vdk-csv and vdk-lineage plugins? I believe they are quite important as well?

vdk-lineage is explicitly said to be POC level yet. But I will add vdk-csv

dakodakov:
Thanks!

dakodakov (approved):

PR #1212, opened 2022-10-04
vdk-core: errors occurred and the state (handled or not) context missing

tozka:
Re submit the change -
https://github.com/vmware/versatile-data-kit/pull/1182 - again. The only difference is that I renamed _get_eception_message to get_exception_message as not to break backwords compatability. I will merge this after

--

User code may trigger a SDK feature in a way that raises a user error. In such a use case, the error could be handled by the cautious user, in runtime. The VDK platform should be aware if user errors are handled in runtime or not, so that affects the data job termination status.

The old errors.BLAMEES (experimental data structure for keeping track of responsibles for errors priorly logged) should be refactored:
* a resolvable context abstraction introduced
* it is memory-managed (singleton, reasoning documented)
* populated by the 3 user-facing `errors.log_*()` methods - in favour of a method designed to populate ErrorMessage entries
* keeps track if the actual exception, along with all properties like who is the actual respobsible persona (user or platform, that was free-form str before), resolved by a data job step or not, etc.

Testing Done: added a functional test covering the
 `simple-query-failed-handled` scenario, added singleton verification
 for the resolvable context; ci/cd

### DIFF
--- a/projects/vdk-core/src/vdk/internal/core/errors.py
+++ b/projects/vdk-core/src/vdk/internal/core/errors.py
@@ -10,6 +10,7 @@ It defines classes and methods for handling exceptions, and ensuring that there 
 """
 from __future__ import annotations
 
+import enum
 import logging
 import re
 import sys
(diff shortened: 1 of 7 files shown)


projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -410,28 +454,15 @@ def find_whom_to_blame_from_exception(exception: Exception) -> ResolvableBy:
...
-MSG_CONSEQUENCE_TERMINATING_APP = (
-    "The provided Data Job will not be executed. Terminating application."
-)
-MSG_COUNTERMEASURE_FIX_PARENT_EXCEPTION = (
-    "See contents of the exception and fix the problem that causes it."
-)
-
-
-def get_exception_message(exception: Exception) -> str:
+def _get_exception_message(exception: Exception) -> str:
```
ivakoleva, line 460: > The only difference is that I renamed _get_eception_message to get_exception_message as not to break backwords compatability.
Didn't get your intention on access modifier

Out of curiosity, what was the usage of `get_exception_message`? And shouldn't we update the usage instead, so that we improve the interface surface in early stages while possible, or is it late already?

ivakoleva (approved):

dakodakov (approved):

PR #1214, opened 2022-10-05
vdk: update contributing.md

tozka:
Clarify the current status around PR using forks so new users are not getting confused.


CONTRIBUTING.md
```diff
@@ -35,7 +35,8 @@ We welcome many different types of contributions and not all of them need a Pull
 This is a rough outline of what a contributor's workflow looks like:
 
 * Make a fork of the repository within your GitHub account
-* Create a topic branch in your fork from where you want to base your work - name your branch according to our naming convention - person/<github-username>/<feature-name>
+   * Currently we accept contribution from forks only for examples and documentation changes until PR 854 is fixed. Until then please request write priveleges and create a branch in the main repo.
```
ivakoleva, line 38: I had to go and read the PR 854, to understand this statement.
Maybe first the guideline, then the explanation?
`Use forks to only contribute changes on examples and documentation. Currently we accept..`
And emphasise those changes are manually applied, is that correct?

CONTRIBUTING.md
```diff
@@ -35,7 +35,8 @@ We welcome many different types of contributions and not all of them need a Pull
...
+* Create a topic branch from where you want to base your work - name your branch according to our naming convention - person/<github-username>/<feature-name>
```
ivakoleva, line 39: Move `Create a topic branch` as primary way of contributing, above the fork statement.

ivakoleva (changes requested):

CONTRIBUTING.md
tozka: OK

CONTRIBUTING.md
tozka: Since we can accept some PRs using forks , Creating a fork for that case it is still a first step.

dakodakov (approved):

PR #1219, opened 2022-10-05

BoBaeva:


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert-snapshot/README.md
```diff
@@ -0,0 +1,51 @@
+### Purpose:
+
+This template can be used to load raw data from Data Lake to target 'Snapshot Periodic Fact Table' in Data Warehouse.
```
tozka, line 3: I think this is misleading a bit. The template doesn't happen snapshot of records. It simply appends records. 

I would also rename it to "insert" or "append"

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert-snapshot/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+- "periodic_snapshot"
+
+### Template Parameters (template_args):
+
+- target_schema   - SC Data Warehouse schema, where target data is loaded
+- target_table    - SC Data Warehouse table of DW type 'Snapshot Periodic Fact Table', where target data is loaded
+- source_schema   - SC Data Lake schema, where source raw data is loaded from
+- source_view     - SC Data Lake view, where source raw data is loaded from
+- last_arrival_ts - Timestamp column, on which increments to target_table are done
```
tozka, line 20: I think last_arrival_ts is not needed here. It's not used anywhere in the tempalte.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert-snapshot/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2021 VMware, Inc.
```
tozka, line 1: Please remove this file. It's not needed.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert-snapshot/README.md
```diff
@@ -0,0 +1,51 @@
...
+This template can be used to load raw data from Data Lake to target 'Snapshot Periodic Fact Table' in Data Warehouse.
+In summary, it appends a snapshot of records observed between time t1 and t2 from the source table to the target table. After that inserts the rows from the source table.
+
+### Details:
+
+<https://www.kimballgroup.com/data-warehouse-business-intelligence-resources/kimball-techniques/dimensional-modeling-techniques/periodic-snapshot-fact-table/>
+
+### Template Name (template_name):
+
+- "periodic_snapshot"
```
tozka, line 12: As noted , "append" or "insert" seems more accurate name

tozka:
The title of the PR should be "component name: words" 

For example "vdk-impala: insert only impala load template"

tozka:
Hi, 

The control-service tests will pass after you update your branch with the latest changes. 

You can do it by 
* Option 1: clicking the Update Branch button.  (then locally you will need to fetch them as well)
* Option 2:  Or locally with `git pull --rebase origin main` command 



vdk-impala tests is something you need to fix. I am not sure why they fail - it appears to be some kind of syntax error from https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3142950082#L380 

You would need to execute them locally. I can help with that if you need to, ping me.

PR #1221, opened 2022-10-05
vdk-plugins: introduced vdk-audit-hook plugin

mivanov1988:
Currently, the VDK users have the ability to execute any kind of system commands through the data job. This increases the risk to the system as data jobs contain arbitrary user code.

Visibility into the actions provides opportunities for test frameworks, logging frameworks, and security tools to monitor and optionally limit actions taken by the runtime.
This plugin provides the ability to audit and potentially limit user actions. In order to reduce the attack surface, we will limit the user actions related to the interaction with the system like os commands (e.g. `os.system("ls")`). Also, to have a better understanding of what precisely the job does, we will log not permitted job actions. The plugin is based on the Python Audit Hook and you can find more information here - https://peps.python.org/pep-0578/.

Testing Done: local job execution and unit tests

### DIFF
--- a/projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_config.py
+++ b/projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_config.py
@@ -31,9 +31,12 @@ class AuditHookConfiguration:
 
 
 def add_definitions(config_builder: ConfigurationBuilder) -> None:
+<<<<<<< HEAD
     """
     Here we define what configuration settings are needed for Impala with reasonable defaults
     """
+=======
+>>>>>>> 239d7165 (vdk-plugins: introduced vdk-audit-hook plugin)
     config_builder.add(
         key=AUDIT_HOOK_ENABLED,
         default_value=True,
--- a/projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_config.py
+++ b/projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_config.py
@@ -1,0 +1,51 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.internal.core.config import ConfigurationBuilder
+
+AUDIT_HOOK_ENABLED = "AUDIT_HOOK_ENABLED"
+NOT_PERMITTED_EVENTS_LIST = "NOT_PERMITTED_EVENTS_LIST"
+EXIT_ON_NOT_PERMITTED_EVENT = "EXIT_ON_NOT_PERMITTED_EVENT"
+NOT_PERMITTED_EVENTS_LIST_DEFAULT = (
+    "os.system;os.chdir;os.chflags;os.chmod;os.chown;os.fork;"
+    "os.forkpty;os.getxattr;os.kill;os.killpg;os.link;os.listxattr;"
+    "os.lockf;os.posix_spawn;os.putenv;os.removexattr;os.rmdir;"
+    "os.scandir;os.setxattr;os.spawn;os.startfile;os.symlink;"
+    "os.truncate;os.unsetenv;os.utime;pty.spawn"
+)
+
+
+class AuditHookConfiguration:
+    def __init__(self, config):
+        self.__config = config
+
+    def enabled(self):
+        return self.__config.get_value(AUDIT_HOOK_ENABLED)
+
+    def not_permitted_events_list(self):
+        return self.__config.get_value(NOT_PERMITTED_EVENTS_LIST)
+
+    def exit_on_not_permitted_event(self):
+        return self.__config.get_value(EXIT_ON_NOT_PERMITTED_EVENT)
+
+
+def add_definitions(config_builder: ConfigurationBuilder) -> None:
+    """
+    Here we define what configuration settings are needed for Impala with reasonable defaults
+    """
+    config_builder.add(
+        key=AUDIT_HOOK_ENABLED,
+        default_value=True,
+        description="Set to false if you want to disable audit hook plugin entirely.",
+    )
+    config_builder.add(
+        key=NOT_PERMITTED_EVENTS_LIST,
+        default_value=NOT_PERMITTED_EVENTS_LIST_DEFAULT,
+        description="List of NOT permitted actions.",
+    )
+    config_builder.add(
+        key=EXIT_ON_NOT_PERMITTED_EVENT,
+        default_value=True,
+        description="Set to false if you want to disable termination of data job on not permitted action.",
+    )
--- a/projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_plugin.py
+++ b/projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_plugin.py
@@ -41,7 +41,11 @@ class AuditHookPlugin:
 
                 if self._config.exit_on_not_permitted_event():
                     logging.getLogger(__name__).error(
+<<<<<<< HEAD
                         f'[Audit Hook] Exiting the data job due to the NOT permitted operation "{event}" with arguments "{args}"'
+=======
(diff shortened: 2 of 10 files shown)


tozka:
Let's name the plugin `vdk-audit`? Hook is a specific single method that is implemented. A plugin is a collection of hooks implemented together to solve a common problem. At least that's the differentiation in the terms that I want to establish. 

And while I understand that this is referring to PEP 578 audit hooks, I still think vdk-audit is clear enough and I will be grateful if you change it

projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_config.py
```diff
@@ -0,0 +1,48 @@
...
+def add_definitions(config_builder: ConfigurationBuilder) -> None:
+    config_builder.add(
+        key=AUDIT_HOOK_ENABLED,
+        default_value=True,
+        description="Set to false if you want to disable audit hook plugin entirely.",
+    )
+    config_builder.add(
+        key=NOT_PERMITTED_EVENTS_LIST,
+        default_value=NOT_PERMITTED_EVENTS_LIST_DEFAULT,
+        description="List of NOT permitted actions.",
```
tozka, line 42: Let's be more specific, What is the definition of "action" . Let's specify that is semi-colon separated values. Give an example.

projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_plugin.py
```diff
@@ -0,0 +1,53 @@
...
+            ):
+                logging.getLogger(__name__).warning(
+                    f'[Audit Hook] Detected NOT permitted operation "{event}" with arguments "{args}"'
+                )
+
+                if self._config.exit_on_not_permitted_event():
+                    logging.getLogger(__name__).error(
+                        f'[Audit Hook] Terminating the data job due to the NOT permitted operation "{event}" with arguments "{args}"'
+                    )
+                    os._exit(0)
```
tozka, line 46: This is a hard exit. .. I imagine it's because we do not want users to catch Exception and swallow it. 
 
 We should document it that if exit_on_not_permitted_event it's fully terminated - no clean up ,no per attempt notifications, etc.
 
 But this seems an error scenario. Shouln't this fail with non zero status?

projects/vdk-plugins/vdk-audit-hook/tests/functional/test_audit_hook_plugin.py
```diff
@@ -0,0 +1,76 @@
...
+
+        runner.invoke(["run", jobs_path_from_caller_directory("os-system-command-job")])
+
+        assert os._exit.called
+
+
+def test_audit_hook_enabled_and_permitted_action():
+    with mock.patch.dict(
+        os.environ,
+        {"VDK_AUDIT_HOOK_ENABLED": "True", "NOT_PERMITTED_EVENTS_LIST": "os.system"},
```
tozka, line 34: We should have a test where NOT_PERMITTED_EVENTS_LIST is a list (not 1 element)

projects/vdk-plugins/vdk-audit-hook/tests/functional/test_audit_hook_plugin.py
```diff
@@ -0,0 +1,76 @@
...
+
+        result: Result = runner.invoke(
+            ["run", jobs_path_from_caller_directory("os-system-command-job")]
+        )
+
+        cli_assert_equal(0, result)
+        assert not os._exit.called
+
+
+def test_audit_hook_disabled_and_permitted_action():
```
tozka, line 63: We shoudl have test verifying exit_on_not_permitted_event

tozka (changes requested):

projects/vdk-plugins/vdk-audit-hook/tests/functional/test_audit_hook_plugin.py
mivanov1988: We do have such test: `test_audit_enabled_and_not_permitted_action`

projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_plugin.py
mivanov1988: > This is a hard exit. .. I imagine it's because we do not want users to catch Exception and swallow it.
> 
> We should document it that if exit_on_not_permitted_event it's fully terminated - no clean up ,no per attempt notifications, etc.
> 

You are absolutely correct about that.

> But this seems an error scenario. Shouln't this fail with non zero status?

In my opinion, we should not retry the job in case of not permitted operation. As far as I remember If I use os._exit(1) (or something) the job will be retried, right? I remember that there was a discussion around that.

tozka: Yes, but we will change that in the future. Especially with [Pod failure policy](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-failure-policy) being introduced in Kubernetes, that would be much easier. 

 I'd prefer to at least make the exit status configurable . 

We can leave the default to zero. But we should explain why it's set to zero (so the job is not re-tried)

projects/vdk-plugins/vdk-audit/src/vdk/plugin/audit/audit_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+            if any(
+                event in not_permitted_event
+                for not_permitted_event in not_permitted_events_list
+            ):
+                logging.getLogger(__name__).warning(
+                    f'[Audit] Detected NOT permitted operation "{event}" with '
+                    f'arguments "{args}" '
+                )
+
+                if self._config.exit_on_not_permitted_event():
```
murphp15, line 45: I think this line looks a bit weird. 
If something is in the not permitted events you would think it should always terminate the job.

murphp15 (approved):
This is a really cool feature.

projects/vdk-plugins/vdk-audit/src/vdk/plugin/audit/audit_plugin.py
mivanov1988: Agreed, I put it just in case we want only to monitor the job.

projects/vdk-plugins/vdk-audit-hook/src/vdk/plugin/audit_hook/audit_hook_plugin.py
mivanov1988: Ok, I will change it.

projects/vdk-plugins/vdk-audit/src/vdk/plugin/audit/audit_plugin.py
mivanov1988: @tozka What is your opinion?

tozka: I think it's a very good idea. It's a required feature for using the audit functionality in any pre-existing (brownfield) environment. I imagine an adoption would be 

1. Enable the audit hook with exit_on_forbidden_event == false 
2. Analyze how many jobs are violating the rules and which events
3. Fix the jobs or amend the forbidden events and back to 2 
4. Eventually enable the hook to terminate jobs when you know it will work safely for all jobs.

Also, it's useful when debugging/fixing a single job to be compliant. So that you can see all violations at once (when there is more than one) and not: fix the first one, re-run to see the second one, fix the second, re-run ... .

tozka (approved):
LGTM.

projects/vdk-plugins/vdk-audit/src/vdk/plugin/audit/audit_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+    @staticmethod
+    @hookimpl
+    def vdk_configure(config_builder: ConfigurationBuilder) -> None:
+        add_definitions(config_builder)
+
+    @hookimpl
+    def initialize_job(self, context: JobContext) -> None:
+        self._config = AuditConfiguration(context.core_context.configuration)
+
+        logging.getLogger(__name__).error("Start:")
```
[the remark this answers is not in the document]
mivanov1988: Fixed

PR #1223, opened 2022-10-05
vdk-core: platform error no longer logged when skipping execution steps

mrMoZ1:
what: Changed the call to the errors.log_and_rethrow method for
a simple log statement and re-raising the exception when the
CancelRemainingSteps exception is raised.

why: Users reading the logs might be confused by the statement,
which assumes a platform error occurred. Recently reverted changes
to the error handling also made jobs using this feature fail with
platform errors since the log_and_rethrow method tripped the
internal logic and made the writer plugin write a platform error
termination message.

testing: Expanded unit test to cover the termination plugin when using
the skip_remaining_steps() functionality.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
@@ -137,7 +137,11 @@ class StepFuncFactory:
             try:
                 func(**actual_arguments)
             except SkipRemainingStepsException as e:
-                log.info(e)
+                # Catch and rethrow exception here since it is handled at
+                # the run_step() hook level. We also don't want to use the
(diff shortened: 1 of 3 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py
```diff
@@ -135,6 +136,9 @@ def invoke_run_function(func: Callable, job_input: IJobInput, step_name: str):
         if actual_arguments:
             try:
                 func(**actual_arguments)
+            except SkipRemainingStepsException as e:
+                log.info(e)
```
tozka, line 140: Let's log it at debug level. Write a comment on why are we catching it here.

tozka (approved):

gageorgiev (approved):
agreed with Toni's comment

PR #1224, opened 2022-10-06
vdk-impala: stop using errors.get_exception_message

tozka:
This will allow us to make it private in vdk-core. It's basically to string of the exception message. Seem overkill to have a vdk core inteface that shallow.

Teseting Done: vdk-impala tests


doks5 (approved):
LGTM

dakodakov (approved):

PR #1227, opened 2022-10-07
vdk-control-service: fix race condition in test

murphp15:
# Why?
There is an intermittent failure in the integration tests for the control plane. 
The stacktrace can be found here https://github.com/vmware/versatile-data-kit/issues/1129. 
It happens because the job watcher sees a job in kubernetes before a row exists in the database. 
So it creates a row from nothing instead of updating the existing row. 
When the watcher creates a row from scratch it doesn't have the startedBy value. 
Because deploying the job and updating the database can't be an atomic operation we need to make sure that the job isn't submitted unless the row exists in the database.

# What
We now add the row into the database before deploying to k8s to prevent the race condition. 
If there is a failure during the deployment we roll back the commit to the database.

# How has this been tested?
I added a test to make sure that we rollback correctly. 
I also added back in the assertion to check that the started by is set correctly.

bug fix. 


tozka:
I will try to provide a review by end of the day or Monday morning. 

But please fix the title of the PR to follow the recommended format: https://github.com/vmware/versatile-data-kit/blob/main/support/git-commit-template.txt (add component and move the link to the ticket somewhere else). 

That format is used later when making release documentation so it's kinda important.

murphp15:
> But please fix the title of the PR to follow the recommended format: https://github.com/vmware/versatile-data-kit/blob/main/support/git-commit-template.txt (add component and move the link to the ticket somewhere else).

Done.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
```diff
@@ -120,6 +111,24 @@ public String startDataJobExecution(
...
+        dataJobsKubernetesService.startNewCronJobExecution(
+            jobDeploymentStatus.getCronJobName(),
+            executionId,
+            annotations,
+            envs,
+            extraJobArguments,
+            jobName);
+      } catch (Exception e) {
+        // rollback data job execution
+        jobExecutionRepository.deleteDataJobExecutionByIdAndDataJobAndStatusAndType(
```
mivanov1988, line 125:
```suggestion
        jobExecutionRepository.deleteById(executionId);
```
isn't it enough?

mivanov1988 (approved):

mivanov1988:
Good catch.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
murphp15: AFAIK, it would be enough. 
Given that I don't know the codebase too well I wanted to err on the side of caution. 
I only wanted to delete the line if I'm sure no one has changed it since I added it.
WDYT will I go with your suggested change?

mivanov1988: Ok, leave it as it is.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/datajobs/DataJobsExecutionControllerIT.java
```diff
@@ -44,6 +53,20 @@ public void testDataJobExecutionStartNotFound() throws Exception {
...
+  @WithMockUser
+  @DirtiesContext
+  public void testWhenADeploymentFailsNoEntryIsSavedInTheDatabase() throws Exception {
+    // arrange
+    doThrow(new RuntimeException())
+        .when(dataJobsKubernetesService)
+        .startNewCronJobExecution(any(), any(), any(), any(), any(), any());
+    // act
+    TestUtils.createDataJob(mockMvc, TEST_TEAM_NAME, TEST_JOB_NAME);
+    TestUtils.createDeployment(mockMvc, TEST_TEAM_NAME, TEST_JOB_NAME);
```
tozka, line 66: I am not sure the idea of the test but without starting execution, the changed code would never be called. 

Maybe there's some misunderstanding in terminology: 

See https://github.com/vmware/versatile-data-kit/wiki/User-Guide#data-jobs-development-workflow or https://raw.githubusercontent.com/wiki/vmware/versatile-data-kit/vdk-data-job-lifecycle-state-diagram.png 

Basically the user workflow is like this 
1. User creates a job using [Control Service Job API ](https://app.swaggerhub.com/apis-docs/aivanovorg/taurus-data-pipelines/1.1#/Data%20Jobs)
   Then they develop it and test it locally 
2. Eventually they would deploy it to run in our managed by Control Service runtime using CS [Deployment API](https://app.swaggerhub.com/apis-docs/aivanovorg/taurus-data-pipelines/1.1#/Data%20Jobs%20Deployment) .  Deployment will build, test, version/release and deploy the job in Kubernetes. The job is ready to be executed but it' not yet.
3. If schedule is specified the job would be executed every so often and Exectuion Servie will track it. Users can request recent executions or start a manual execution using [Execution API](https://app.swaggerhub.com/apis-docs/aivanovorg/taurus-data-pipelines/1.1#/Data%20Jobs%20Execution) 

So in reality our Control Service is a collection of a few (fairly) independent services (Job, Deployment and Execution among them). 

In this mock tests, the mock deployment only deploys a job without any schedule. So to test anything done by Execution Service we need to explicitly call some method like TestUtils.startMockExecution .


Of course it's possible that I am the one missing something in this particular case. But still want to find out what.

tozka (approved):
Looks ok to me. If you can explain the unit test.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/datajobs/DataJobsExecutionControllerIT.java
murphp15: Nope your not missing anything. 
This was a mistake on my side. 
Fixed now.

PR #1231, opened 2022-10-10

BoBaeva:


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```python
+def run(job_input):
+    # . . .
+    template_args = {
+        'source_schema': 'default',
+        'source_view': 'vw_fact_vmc_utilization_cpu_mem_every5min_daily',
+        'target_schema': 'history',
+        'target_table': 'fact_vmc_utilization_cpu_mem_every5min_daily',
+        'last_arrival_ts': 'updated_at',
```
murphp15, line 36: Should this property be here? 
its not in the template args section.

BoBaeva: Thanks, @murphp15! I removed the argument from template_args.

projects/vdk-plugins/vdk-impala/tests/functional/jobs/insert_template_partition_job/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2021 VMware, Inc.
```
tozka, line 1: Please remove the empty `__init__.py` files. I think those are created accidentally by VSCode

tozka (approved):
Looks good to me. Only remove the empty __init__.py files.

PR #1233, opened 2022-10-11

BoBaeva:

### DIFF
--- a/projects/vdk-plugins/vdk-impala/tests/functional/jobs/insert_template_job/01_prepare_input_data.py
+++ b/projects/vdk-plugins/vdk-impala/tests/functional/jobs/insert_template_job/01_prepare_input_data.py
@@ -1,0 +1,88 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.job_input import IJobInput
+
+
+__author__ = "VMware, Inc."
+__copyright__ = (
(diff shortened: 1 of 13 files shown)


tozka (approved):

projects/vdk-plugins/vdk-impala/tests/functional/jobs/insert_template_job/01_prepare_input_data.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.job_input import IJobInput
+
+
+__author__ = "VMware, Inc."
+__copyright__ = (
+    "Copyright 2019 VMware, Inc.  All rights reserved. -- VMware Confidential"
+)
```
doks5, line 9: @tozka, should these copyrights stay? If yes, maybe @BoBaeva can bump them to 2022 instead of 2019?

doks5 (approved):
Overall, looks ok to me.

projects/vdk-plugins/vdk-impala/tests/functional/jobs/insert_template_job/01_prepare_input_data.py
tozka: We can remove them in follow-up PR from everywhere (they are not just here - https://github.com/vmware/versatile-data-kit/search?q=VMware+Confidential) . It's seems when migrating the code when open sourcing it was forgotten

PR #1239, opened 2022-10-13
vdk-impala: Refactor query classifier for data lineage

kostoww:
# Why?
Analysing the SQL query is not so trivial task, as there are number of variations of the DDL, DML and DDL with DML combinations. Current avoidance of single line comments does not work properly and skips queries that are eligible for lineage collection.

# What?
Added sqlparse library in order to externalise comment-cleaning task, now it leaves us only with the query and there isn't any need to analyse whole query. What we do now is we check if query starts with blacklisted keyword, if so - query is not eligible for lineage. Although some keywords blacklisted might contain lineage data, like CREATE statement, when table is created from another with SELECT statement. In this case blacklisted word needs to be whitelisted in some conditions.

As sql statements are complex, we need to get which queries definitely has data lineage, instead of searching which query have for sure and for sure not data lineage. False positives queries for lineage, ie queries that does not have lineage information provided from the profile, but are marked as having one, will not break the logic at all, instead we optimize load on the impala server with a simple approach

# How has this been tested?

Bug-fixing


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py
```diff
@@ -108,35 +109,46 @@ def _is_query_have_lineage(query_statement: str) -> bool:
...
-                )
-            ):
-                # these commands are not providing lineage data in the profile at the moment
-                return False
+        query_statement = query_statement.lower()
+        # managed_connection has a way to open and check connections with keep alive query
+        if "select 1 -- testing if connection is alive." in query_statement:
+            return False
+
+        query_statement = sqlparse.format(
```
tozka, line 117: Why are you not using `sqlparse.parse` ?  

This will not work for CTEs : 
for example 
```
with x as (
   select * from y
)
select * from z
```
for example

kostoww: What will be the benefit of sqlparse.parse?
Mentioned example will work, already has this unit test

VladimirPetkov1 (approved):
You can add some docs in _does_query_have_linage method explaining why the method exists.
LGTM

tozka (approved):

PR #1240, opened 2022-10-16
vdk-pipelines-controls-service: upgrade to 3.9 in the vdk init container

murphp15:
# Why?
The newest version of quickstart-vdk docker image was changed to use python 3.9 instead of 3.7. https://hub.docker.com/r/versatiledatakit/quickstart-vdk/tags

This caused a test to fail because it was looking for python 3.7. 
This change points it to 3.9 instead. 

# What?
Change to version 3.9

# How has this been tested?
integration tests

Bug fix


tozka:
This would be a breaking change as jobs relying on 3.7 would stop working. And we should avoid/minimize breaking changes. 

We need to make it 
A) either configurable with default being 3.7
B) Make it so that it will work with any python version (e.g do not hard code the python path with a version)

PR #1241, opened 2022-10-17
vdk-core: resolve library error classification on startup

mrMoZ1:
what: Added functionality which handles ImportErrors, logs the error and writes
termination message via the appropriate internal plugin implementation and exits.

Other approaches tried:
 - researched a library that detects import issues (didn't find any)
 - tried moving the load_setuptools_entrypoints method call (which throws the error)
    a bit further in the call chain, however tests failed (see previous pipelines in this pull req)

why: some user data jobs have been failing with ImportErrors (where a
user import clashes with internal imports, however this crashed the
vdk runtime without writing a termination message.

testing: added unit test.

Signed-off-by: mrMoZ1 <mzhivkov@vmware.com>


murphp15 (approved):

projects/vdk-core/src/vdk/internal/util/utils.py
```diff
@@ -40,3 +50,27 @@ def parse_config_sequence(
     if sequence:
         sequence = [i.strip() for i in sequence.split(sep)]
     return sequence if sequence else []
+
+
+def exit_with_error(user_error: bool, log: Logger, exception: Exception):
```
tozka, line 55: https://wiki.c2.com/?UseEnumsNotBooleans

we have an enum for that: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/core/errors.py#L55

projects/vdk-core/src/vdk/internal/plugin/plugin.py
```diff
@@ -50,6 +52,8 @@ def load_plugins_from_setuptools_entrypoints(self) -> None:
 
         try:
             self.__plugin_manager.load_setuptools_entrypoints(self.__group_name)
+        except ImportError as e:
+            exit_with_error(True, log, e)
```
tozka, line 56: We exit directly on import error But   raise PluginException(message) on other type of errors. 

I don't think this kind of divergent logic is good.

projects/vdk-core/src/vdk/internal/plugin/plugin.py
```diff
@@ -50,6 +52,8 @@ def load_plugins_from_setuptools_entrypoints(self) -> None:
 
         try:
             self.__plugin_manager.load_setuptools_entrypoints(self.__group_name)
+        except ImportError as e:
```
tozka, line 55: At first glance assuming that each import error must be a user error seems strange. There could be a bug in the plugin itself that results in an Import error. 

Perhaps we discussed that (I have vague memories) but if we had I can't recall the reasoning for why we want this to be user error. What is the reason ?

projects/vdk-core/src/vdk/internal/plugin/plugin.py
mrMoZ1: I have refactored both methods. The behaviour was the same, since when we raised the error, the pod would terminate - since no exception handling yet.

projects/vdk-core/tests/functional/run/test_run_load_setuptools_errors.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+    def test_import_error(
+        self,
+        file_util: MagicMock,
+        patched_exit: MagicMock,
+        load_setuptools_entrypoints: MagicMock,
+    ):
+        """
+        This test simulates a scenario where a user imports a library that
+        has a clash with vdk internal imports. The desired behaviour is to
```
ivakoleva, line 27: > This test simulates a scenario where a user imports a library that
> has a clash with vdk internal imports.

By reading the description, my impression about the use-case is - the user code performs a non-valid import that clashes with vdk internal imports. Then, why patching pluggy's PluginManager? Could you provide a real-world stack trace example?
Adding a functional test will make the use-case more clear

mrMoZ1: Here is a sample stack trace of the problem:

```
  File "/vdk/site-packages/vdk/internal/plugin/plugin.py", line 52, in load_plugins_from_setuptools_entrypoints
    self.__plugin_manager.load_setuptools_entrypoints(self.__group_name)
  File "/vdk/site-packages/pluggy/manager.py", line 299, in load_setuptools_entrypoints
    plugin = ep.load()
  File "/vdk/site-packages/importlib_metadata/__init__.py", line 203, in load
    module = import_module(match.group('module'))
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/vdk/site-packages/supercollider/vdk/supercollider_vdk_plugin.py", line 10, in <module>
    from supercollider.vdk.api.load_plugin import LoadPlugin
  File "/vdk/site-packages/supercollider/vdk/api/load_plugin.py", line 6, in <module>
    from supercollider.vdk.api.load import ImpalaLoader
  File "/vdk/site-packages/supercollider/vdk/api/load.py", line 15, in <module>
    import pandas
  File "/vdk/site-packages/pandas/__init__.py", line 22, in <module>
    from pandas.compat import (
  File "/vdk/site-packages/pandas/compat/__init__.py", line 15, in <module>
    from pandas.compat.numpy import (
  File "/vdk/site-packages/pandas/compat/numpy/__init__.py", line 27, in <module>
    f"this version of pandas is incompatible with numpy < {_min_numpy_ver}\n"
ImportError: this version of pandas is incompatible with numpy < 1.17.3
your numpy version is 1.16.5.
Please upgrade numpy to >= 1.17.3 to use this pandas versionThe above exception was the direct cause of the following exception:Traceback (most recent call last):
  File "/vdk/vdk", line 8, in <module>
    sys.exit(main())
  File "/vdk/site-packages/vdk/internal/cli_entry.py", line 165, in main
    plugin_registry.load_plugins_from_setuptools_entrypoints()
  File "/vdk/site-packages/vdk/internal/plugin/plugin.py", line 67, in load_plugins_from_setuptools_entrypoints
    raise PluginException(message) from e
vdk.api.plugin.plugin_registry.PluginException: Plugin load failed
  WHAT HAPPENED : Cannot load plugin from setuptools entrypoint for group vdk.plugin.run
WHY IT HAPPENED : See exception for possible reason
   CONSEQUENCES : The CLI tool will likely abort.
COUNTERMEASURES : Re-try again. Check exception mesage and possibly uninstall a bad plugin (pip uninstall) Or see what plugins are installed (use `pip list` command) and if there are not issues. Or try to reinstall the app in a new clean environment.Try to revert to previous version of hte CLI tool.If nothing works open a SRE ticket 
```
The import itself is not invalid, it is incompatible with internal imports. The problem here is that the error occurs in the cli_entry.py - main() method which is unfortunately not covered by the CliEntryBasedTestRunner() used in our functional tests another problem is that we need that specific version of numpy and pandas to reproduce it, meaning that we have to install those dependencies in the test environment, that's why I believe it's more convinient to mock the specified method to throw the same type of exception instead.

ivakoleva: Stacktrace shows the issue occurs during `plugin_registry.load_plugins_from_setuptools_entrypoints()` when loading plugins, before any data job user code had been handled yet. 

At the same time:

> why: some user data jobs have been failing with ImportErrors (where a
> user import clashes with internal imports, however this crashed the
> vdk runtime without writing a termination message.

How come the plugin load lifecycle phase - clashes with user code?

projects/vdk-core/src/vdk/internal/cli_entry.py
```diff
@@ -110,7 +111,6 @@ def vdk_main(
         """
         Main method of the CLI. It call all configuration and initialization hooks and start CLI
         """
-
         plugin_registry.add_hook_specs(CoreHookSpecs)
         plugin_registry.load_plugin_with_hooks_impl(builtin_hook_impl, "core-plugin")
```
murphp15, line 116: Could you not exit on this line if root_command is non instead of doing the check twice later on?

mrMoZ1: We need to run all the default hooks as per - https://github.com/vmware/versatile-data-kit/pull/1241#discussion_r1004526029 , initialize configuration and logging.

projects/vdk-core/tests/functional/run/test_run_load_setuptools_errors.py
mrMoZ1: I don't know the intricacies of it all, but it crashes when the user has imported a library that is incompatible with one of our internal ones. So I suppose it fails when the interpreter goes over an import statement in a plugin that clashes with a user import statement.

ivakoleva: I see, https://github.com/vmware/versatile-data-kit/blob/c2444dfe82c4ffeff378988d2e7dc26a850fbdd9/projects/control-service/projects/job-builder-rootless/Dockerfile.python.vdk#L26 -> data job dependencies come preinstalled on image build, that is triggered on `vdk deploy`.  

If you want to deep-dive, then what are the pros and cons of dependency compatibility validation during image build vs run phase?
On another note, what job builder are you using, standard/rootless/secure?

projects/vdk-core/src/vdk/internal/cli_entry.py
```diff
@@ -158,16 +161,20 @@ def main() -> None:
...
+    try:
+        plugin_registry.load_plugins_from_setuptools_entrypoints()
+    except Exception:
+        plugin_load_success = False
     plugin_registry.load_plugin_with_hooks_impl(CliEntry(), "cli-entry")
 
     exit_code = cast(InternalHookSpecs, plugin_registry.hook()).vdk_main(
         plugin_registry=plugin_registry,
-        root_command=cli,
+        root_command=cli if plugin_load_success else None,
```
ivakoleva, line 177: in case of plugin load failure, why evaluating the vdk_main at all? It wasn't evaluated before this change (due to the prior exception raised that is now catched)

projects/vdk-core/src/vdk/internal/cli_entry.py
```diff
@@ -128,12 +128,15 @@ def vdk_main(
...
+            if root_command is not None:
+                exit_code = cast(
+                    InternalHookSpecs, plugin_registry.hook()
+                ).vdk_cli_execute(
+                    root_command=root_command,
+                    command_line_args=command_line_args,
+                    core_context=core_context,
+                    program_name=program_name,
+                )
             return exit_code
```
ivakoleva, line 140: In case of failure, `vdk` returns != 0 by design. That's a major change, modifying the application exit status.

projects/vdk-core/src/vdk/internal/cli_entry.py
mrMoZ1: Because we would only log the exception (without logging functionality initialized) and re-raise it, cutting the execution of all vdk-hooks including vdk_exit which writes the termination message and job execution summary, meaning that users wouldn't see any logs in loginsight and the pod would be deleted from k8s as terminated with Error and the data job would be recorded as failed with platform error in the db. The proposed solution would run all other generic hooks, except the vdk command itself in case of such errors as per Toni's suggestion above.

ivakoleva: In case the vdk-hooks flow is now changed, you should test cover it. Didn't see Toni's suggestion, maybe you could link it?

mrMoZ1: https://github.com/vmware/versatile-data-kit/pull/1241#discussion_r1004526029 . The added unit tests cover new functionality, hit all hooks, including vdk_exit. The only caveat is that I had to use mocks (for the sys.exit and termination message file writer) and introduce the Exceptions as side effects, since it's difficult to introduce this issue with a unit test. The only other test I can think of is to create a specific env with the precise version of clashing dependencies and test exit status codes via bash but I don't like this approach either. N.b the existing functional tests which use CliEntryBasedTestRunner - won't hit the main method in cli_entry.py module.

projects/vdk-core/tests/functional/run/test_run_load_setuptools_errors.py
mrMoZ1: I think we discussed this with Toni, and concluded that the fix should be in vdk-core, since a job builder is decoupled from vdk and this won't entirely resolve the issue for local runs or already deployed jobs with this issue.

ivakoleva: In that case the customer would eventually understand the dependency issues not early on deployment, but late when the first time data job I scheduled and runs. Context switching UX is a downside to me, a user friendly workflow includes both checks. However I think this may be out of scope of this PR.

projects/vdk-core/src/vdk/internal/cli_entry.py
ivakoleva: Could you please link the verification of hooks invocation, because I missed it?

ivakoleva: Also, for `cli_entry.py` coverage maybe can unit test it

mrMoZ1: https://github.com/vmware/versatile-data-kit/blob/62069ddf03322d3a84f2d76b4c1c44758acb74f3/projects/vdk-core/tests/vdk/internal/test_run_load_setuptools_errors.py#L21
This a mocked method in the writer module:
https://github.com/vmware/versatile-data-kit/blob/187853b5a2a812d7cca128ab751f5674da88a13a/projects/vdk-core/src/vdk/internal/builtin_plugins/termination_message/writer.py#L30-L49
which is invoked from the vdk_exit hook. Both unit tests check that it was invoked and that the correct termination message was written e.g Platform or User Error which is the main aim of this change, as previously the termination message wasn't written.

ivakoleva: I don't see test coverage of hooks being innovated now, oppositely to before the MR change.

projects/vdk-core/src/vdk/internal/cli_entry.py
ivakoleva: I saw the change; maybe initialise with 1, to save workflow branches (else)

projects/vdk-core/src/vdk/internal/cli_entry.py
```diff
@@ -125,15 +125,18 @@ def vdk_main(
         configuration = build_configuration(plugin_registry)
         core_context = build_core_context_and_initialize(configuration, plugin_registry)
 
-        exit_code = 0
+        exit_code = 1
```
ivakoleva, line 128: nit: maybe have the comment you had previously provided, explaining why 1 in case no root_command

projects/vdk-core/src/vdk/internal/cli_entry.py
```diff
@@ -125,15 +125,18 @@ def vdk_main(
...
         try:
             log.info(f"Start CLI {program_name} with args {command_line_args}")
-            exit_code = cast(InternalHookSpecs, plugin_registry.hook()).vdk_cli_execute(
-                root_command=root_command,
-                command_line_args=command_line_args,
-                core_context=core_context,
-                program_name=program_name,
-            )
+            if root_command is not None:
```
ivakoleva, line 131: nit: why not shorter `if root_command:`

projects/vdk-core/src/vdk/internal/cli_entry.py
```diff
@@ -158,16 +161,20 @@ def main() -> None:
...
+    plugin_load_success = True
 
     log.debug("Setup plugin registry and call vdk_start hooks ...")
     plugin_registry = PluginRegistry()
     plugin_registry.add_hook_specs(InternalHookSpecs)
-    plugin_registry.load_plugins_from_setuptools_entrypoints()
+    try:
+        plugin_registry.load_plugins_from_setuptools_entrypoints()
+    except Exception:
+        plugin_load_success = False
```
ivakoleva, line 172: Why does not `plugin_load_success` exist into `plugin_registry`, seems contextual to me? the way currently, that value is actually lost when going out of method scope...

projects/vdk-core/tests/functional/run/test_run_load_setuptools_errors.py
```diff
@@ -0,0 +1,87 @@
...
+class TestImportErrors(unittest.TestCase):
+    @patch(
+        "pluggy._manager.PluginManager.load_setuptools_entrypoints",
+        side_effect=ImportError("Test import error."),
+    )
+    @patch("os._exit", side_effect=Exception("VDK run termination exception."))
+    @patch(
+        "vdk.internal.builtin_plugins.termination_message.file_util.WriteToFileAction.write_to_file"
+    )
+    def test_import_error(
```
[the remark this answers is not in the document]
ivakoleva: @mrMoZ1 Is that comment applicable still, or could we resolve?

ivakoleva (approved):
I am okay to merge, my main concern about design still is described in a comment about why doesn't `plugin_load_success exist into plugin_registry`. Lets elaborate before merging

projects/vdk-core/tests/vdk/internal/test_run_load_setuptools_errors.py
```diff
@@ -0,0 +1,109 @@
...
+        "pluggy._manager.PluginManager.load_setuptools_entrypoints",
+        side_effect=Exception("Test general error."),
+    )
+    @patch(f"vdk.internal.cli_entry.build_configuration")
+    @patch(
+        "vdk.internal.builtin_plugins.termination_message.file_util.WriteToFile"
+        "Action.write_to_file"
+    )
+    @patch(f"vdk.internal.cli_entry.build_core_context_and_initialize")
+    def test_hooks_called(
```
ivakoleva, line 85: please pay attention to the static code analysis results, to improve code quality?

tozka (approved):
Looks good to me.

Issue #1243, opened 2022-10-17
Meta Jobs (DAG of jobs)


tozka:
**What is the feature request? What problem does it solve?**
One crucial feature which has been missing from VDK is the ability to express workload (data job) interdependencies, meaning that a certain Data Job wil only trigger upon the successful completion (or other criteria) of one or more other Data Jobs.

An integration with Apache Airflow was developed to address this, however this comes with its downsides: namely that teams would have to manage their own instance of Airflow, or pay for an externally managed service.

**Suggested solution**

**Additional context**

PR #1249, opened 2022-10-19
vdk-meta-jobs: Initial implementation

tozka:
This is a pre-Alpha implementation of Meta Jobs functionality.

One crucial feature which has been missing from VDK is the ability to
express workload (data job) interdependencies, meaning that a certain
Data Job wil only trigger upon the successful completion (or other
criteria) of one or more other Data Jobs.

Integration with Apache Airflow was developed to address this,
however, this comes with its downsides: namely that teams would have to
manage their own instance of Airflow, or pay for an externally managed
service.

The aim of this submission is to build a plugin for Versatile Data Kit
which extends its API with an additional feature that allows users to
trigger so-called Meta Jobs - jobs that trigger one or more other jobs,
wait for their completion, and then trigger another set of jobs until
either the entire job pipeline has succeeded, or at least one of the
jobs has failed.

https://github.com/vmware/versatile-data-kit/issues/1243

Testing Done: functional tests


projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,146 @@
+# Meta Jobs
+
+Express dependecies between data jobs.
+
+A plugin for Versatile Data Kit which extends its Job API with an additional feature which allows users to trigger so called Meta Jobs.
+Meta Jobs are data jobs which trigger one or more other jobs, wait for their completion, and then trigger another set of jobs until either the entire job pipeline has succeeded.
+
+
+## Usage
+
```
ivakoleva, line 10: maybe add information about any existing alternative approach using the current control-service API, and comparison?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_job_runner.py
```diff
@@ -0,0 +1,26 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def get_json(obj):
+    return json.loads(json.dumps(obj, default=lambda o: o.__dict__))
+
+
+class MetaJobInput(IMetaJobInput):
+    def run_meta_job(self, jobs: List[Dict]):
```
murphp15, line 22: Why does `IMetaJobInput` have a different definition for this method? 
The one one on `IMetaJobInput` is much better.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -0,0 +1,214 @@
...
+    SUBMITTED = "submitted"
+    RUNNING = "running"
+    SUCCEEDED = "succeeded"
+    CANCELLED = "cancelled"
+    SKIPPED = "skipped"
+    USER_ERROR = "user_error"
+    PLATFORM_ERROR = "platform_error"
+
+
+class RemoteDataJob:
```
murphp15, line 35: I see that this class has a lot of overlap with the class `VDKhook`. 
Do you plan on deduping the overlap?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job_executor.py
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: Apache-2.0
+from abc import ABC
+from typing import Dict
+
+from vdk.internal.control.configuration.vdk_config import VDKConfig
+from vdk.plugin.meta_jobs.meta import IDataJobExecutor
+from vdk.plugin.meta_jobs.remote_data_job import RemoteDataJob
+
+
+class RemoteDataJobExecutor(IDataJobExecutor):
```
murphp15, line 11: The way that you create a new job each time seems a bit odd to me. 

I would expect that you would call all methods on the same instance.
Maybe they should just be static methods if you aren't keeping the object around. 

```
job = RemoteDataJob(job_name, team_name, vdk_cfg.control_service_rest_api_url)
job.start_job_execution()
job.get_job_execution_status(execution_id)
```

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
```diff
@@ -0,0 +1,66 @@
...
+from typing import Any
+
+
+@dataclass
+class _QueueEntry:
+    element: Any
+    ready_time: float
+
+
+class TimeBasedQueue:
```
murphp15, line 16: I think 99% of this code can be deleted if you use a priorityQueue from the standard python library https://docs.python.org/3/library/queue.html#queue.PriorityQueue?

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,146 @@
+# Meta Jobs
+
+Express dependecies between data jobs.
+
+A plugin for Versatile Data Kit which extends its Job API with an additional feature which allows users to trigger so called Meta Jobs.
+Meta Jobs are data jobs which trigger one or more other jobs, wait for their completion, and then trigger another set of jobs until either the entire job pipeline has succeeded.
+
+
```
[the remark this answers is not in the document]
tozka: The point of hte plugin is that it does NOT require Airflow

projects/vdk-plugins/vdk-meta-jobs/README.md
tozka: README is not a feature spec. Definitely not here if anywhere since I have not investigated alternative approaches.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_job_runner.py
tozka: Because I didn't have time to implement it. That would be future work. I need to do serialization/deserialization because I want to allow meta jobs to be defined as a configuration.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
tozka: I don't like it as well. But I have not hit yet the https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
tozka: That is NOT a priority queue. The point of this queue is to delay dequeu operations until they are ready.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job_executor.py
tozka: Don't focus on the implementation but on the interface

I plan (in the far future!) to refactor the way it's passed configuration. 

VDKConfig is the way vdk-control-cli keeps and passes configuration while vdk-core uses ConfigurationBuilder and vdk_configure plugins. 

I want to streamline it . vdk-control-cli was created before vdk-core and plugins come into place that's why it's different. Iti won't be soon and not in this change.

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,146 @@
+# Meta Jobs
+
+Express dependecies between data jobs.
+
+A plugin for Versatile Data Kit which extends its Job API with an additional feature which allows users to trigger so called Meta Jobs.
+Meta Jobs are data jobs which trigger one or more other jobs, wait for their completion, and then trigger another set of jobs until either the entire job pipeline has succeeded.
```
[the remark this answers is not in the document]
tozka: The descriptoin is "Meta Jobs are data jobs which trigger one or more other jobs, wait for their completion, and then trigger another set of jobs until either the entire job pipeline has succeeded.
"

Can you suggest how to re-word it so it's more clear. 

* Meta Job is normal data job that calls other jobs. Nothing is different really . the same way "ingestion job" is a normal data job that does an ingestion or processing job is a norma data job that does processing.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Use it to enqueue elements until element is ready to be dequeued.
+    """
+
+    def __init__(
+        self,
+        min_ready_time_seconds=10,
+        randomize_delay_seconds=10,
```
murphp15, line 24: why is there some randomized delay added?

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,146 @@
+# Meta Jobs
+
+Express dependecies between data jobs.
+
+A plugin for Versatile Data Kit which extends its Job API with an additional feature which allows users to trigger so called Meta Jobs.
```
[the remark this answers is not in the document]
tozka: Can be triggered in parallel.

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,146 @@
...
+When defining a job to be run following attributes are supported:
+* **job_name**: required, the name of the data job
+* **team_name:**: optional, the team of the data job. If omitted , it will use the meta job's team
+* **fail_meta_job_on_error**: optional, default is true. if true, the meta job will abort and fail if the orchestrated job fails, if false, meta job won't fail and continue.
+* **depends_on**: required (can be empty), list of other jobs that the orchestrated job depends on. The job will not be started until depends_on job have finished.
+
+
+### Example
+
+The following example dependency graph can be implemented with below code.
```
dakodakov, line 40: I believe a clarification about concurrency should be added. My assumption for this specific example is the following:
1. Job 1 will execute
2. After Job 1 is completed, jobs 2,3,4 will start executing in parallel
3. Jobs 5 and 6 will start executing after job 3 completes, but will not wait for the completion of jobs 2 and 4.

Is this assumption correct?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import json
+import logging
+import os
+import time
+from typing import Dict
+
+from taurus_datajob_api import ApiException
```
dakodakov, line 9: taurus_datajob_api? what is Taurus?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
```diff
@@ -0,0 +1,130 @@
...
+        job.details = details
+        log.info(
+            f"Finished data job {job_name}:\n"
+            f"start_time: {details['start_time']}\n"
+            f"end_time: {details['end_time']}\n"
+            f"status: {details['status']}\n"
+            f"message: {details['message']}"
+        )
+        if job.status != JobStatus.SUCCEEDED.value and job.fail_meta_job_on_error:
+            raise UserCodeError(
```
dakodakov, line 68: If the job failed with a platform error, should this raise a PlatformError exception instead?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -0,0 +1,109 @@
...
+            else:
+                result = pprint.pformat(data, indent=2, depth=5, compact=False)
+        return result
+
+    def _start_job(self, node):
+        try:
+            self._job_executor.start_job(node)
+        except ApiException as e:
+            if e.status == 409:
+                log.info(
```
dakodakov, line 96: should this be a warning?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -0,0 +1,109 @@
...
+                    f"Detected conflict with another running job: {e}. Will be re-tried later"
+                )
+                self._delayed_starting_jobs.enqueue(node)
+            elif e.status >= 500:
+                log.info(
```
dakodakov, line 101: should this be an error?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -0,0 +1,109 @@
...
+                job["job_name"],
+                job.get("team_name", self._team_name),
+                job.get("fail_meta_job_on_error", True),
+            )
+            self._job_executor.register_job(trackable_job)
+            self._topological_sorter.add(trackable_job.job_name, *job["depends_on"])
+
+    def execute_dag(self):
+        self._topological_sorter.prepare()
+        while self._topological_sorter.is_active():
```
dakodakov, line 51: Isn't this going to spin infinitely? Do we need some sleep() in this loop?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.__execution_api = self._get_execution_api()
+
+    def start_job_execution(self, **request_kwargs) -> str:
+        """
+        Triggers a manual Datajob execution.
+
+        :param: request_kwargs: Request arguments to be included with the HTTP request
+        """
+        execution_request = DataJobExecutionRequest(
```
dakodakov, line 78: Do we need a new job execution type, e.g. Meta-job-triggered, apart from the existing Scheduled and Manual? I believe we do.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
tozka: Taurus is the second astrological sign in the modern zodiac. 

It appears also that the python API client library generated by the control service uses this as a package name - https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/model/apidefs/datajob-api/config-python.json#L2

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
tozka: Yes! I need to fix that.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
tozka: No, the job will be re-tried. We do not log errors on re-tries.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
tozka: No, the job will be re-tried. We do not log errors on re-tries. On top of that it's not an errorenous condition - it's expected (possible at least) that there might be concurrent job started

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
tozka: We discussed this during the hackaton with our users (who also participated) and we had 2 reasons to do it this way: 

The thought is that if we raise a platform error, this will trigger automatic re-try of the meta job. 
If the orechestated job has 4 re-tries then meta job has 4 re-tries this is 16 re-tries of orchestrated job. 

I cannot recall the second reason now. That's why one should write comments ...

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
tozka: TimeBasedQueue is used to enable delayed re-tries. And we want the delay to be randomized so that not all re-tries trigger in the same time.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
tozka: Not sure. started_by indicates it was started by meta job.
It looks like this "meta-job/aivanov" 

 I will create a separate issue for us to ponder this. But we do not have separate execute type if job is tarted from UI vs CLI currently. Meta Job is just another agent

projects/vdk-plugins/vdk-meta-jobs/.plugin-ci.yml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2021 VMware, Inc.
```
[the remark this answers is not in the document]
tozka: Done

projects/vdk-plugins/vdk-meta-jobs/README.md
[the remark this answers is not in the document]
tozka: Addressed: Re-wrote the summary to this 
```

Express dependencies between data jobs.

A plugin for Versatile Data Kit extends its Job API with an additional feature that allows users to trigger so-called Meta Jobs.

A meta job is a regular Data Job that invokes other Data Jobs using Control Service Execution API. 
In this way, there's nothing different from other data jobs except for its purpose. See [Data Job types](https://github.com/vmware/versatile-data-kit/wiki/User-Guide#data-job-types) for more info.

It's meant to be a much more lightweight alternative to complex and comprehensive workflows solution (like Airflow) 
as it doesn't require to provision any new infrastructure or to need to learn new tool. 
You install a new python library (this plugin itself) and you are ready to go.

Using this plugin you can specify dependencies between data jobs as a direct acyclic graph (DAG). See usage for more information.
```

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,146 @@
+# Meta Jobs
+
+Express dependecies between data jobs.
```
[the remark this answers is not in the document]
tozka: Hard and soft are explained in usage. Re-wrote the summary see comment above.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job_executor.py
tozka: Does that make sense?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
tozka: Created https://github.com/vmware/versatile-data-kit/issues/1272 for this

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
dakodakov: I'm sorry but did you fix this? I see you've marked the conversation as resolved...

projects/vdk-plugins/vdk-meta-jobs/README.md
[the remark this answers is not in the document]
ivakoleva: what I understand from that cited description, is that a meta job is any job triggered from within another data job execution, is that correcT?
also, is a data job triggered via execution api, also a meta job? for example, the link provided is very useful, however it's missing the orchestration capabilities (current, and upcoming ones)

projects/vdk-plugins/vdk-meta-jobs/README.md
ivakoleva: our own execution api is one

ivakoleva: in fact, it is not mentioned, probably purposefully or not

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+def run(job_input: IJobInput) - > None:
+    MetaJobInput().run_meta_job(JOBS_RUN_ORDER)
+```
+
+### FAQ
+
+
+**Q: Will the metajob retry on Platform Error?**<br>
+A: Yes, up to N (configurable by the Control Service) attempts for each job it is orchestrating.
```
ivakoleva, line 113: is it going to be retried immediately, or is there any backoff

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, team_name: str):
+        self._team_name = team_name
+        self._topological_sorter = TopologicalSorter()
+        self._delayed_starting_jobs = TimeBasedQueue(
+            min_ready_time_seconds=int(
+                os.environ.get("VDK_META_JOBS_DELAYED_JOBS_MIN_DELAY_SECONDS", 30)
+            ),
+            randomize_delay_seconds=int(
+                os.environ.get(
+                    "VDK_META_JOBS_DELAYED_JOBS_RANDOMIZED_ADDED_DELAY_SECONDS", 600
```
ivakoleva, line 33: randomized seems rather peculiar strategy to use during orchestration

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import abc
+from dataclasses import dataclass
+
+from vdk.plugin.meta_jobs.api import meta_job
+
+
+class IDataJobExecutor(abc.ABC):
```
ivakoleva, line 9: Maybe a part of the extensible API

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -0,0 +1,214 @@
...
+from taurus_datajob_api import DataJobsExecutionApi
+from urllib3 import Retry
+from vdk.plugin.control_api_auth.authentication import Authentication
+
+log = logging.getLogger(__name__)
+
+
+class JobStatus(str, Enum):
+    """
+    Enum for the possible statuses a job execution can have.
```
ivakoleva, line 23: is that the very first enum we have? maybe have it in core, since it's going to be easier to keep consistent and up-to-date, for plugin usages?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -0,0 +1,214 @@
...
+    SUCCEEDED = "succeeded"
+    CANCELLED = "cancelled"
+    SKIPPED = "skipped"
+    USER_ERROR = "user_error"
+    PLATFORM_ERROR = "platform_error"
+
+
+class RemoteDataJob:
+    """
+    Interact with Verstile Data Kit (VDK) Control Service
```
ivakoleva, line 37: that class should either implement/extend a core concept, or become a part of it itself...
this plugin seems to describe the domain model for the first time, that should be available to reuse in general - and to keep consistent; 
maybe have this as an additional effort

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_job_runner.py
ivakoleva, line 22: it's unusual to have a list of `jobs` being a `List[Dict]`, it should be a list of jobs.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
ivakoleva: in fact, I am looking at core's IStandaloneDataJob that has some common properties used by this plugin domain model, for example name, args (and team that is matching the outer job I guess);

projects/vdk-plugins/vdk-meta-jobs/README.md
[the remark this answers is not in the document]
tozka: > however it's missing the orchestration capabilities (current, and upcoming ones)

See Usage section. They are described there.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
tozka: Hmm, not properly. Do unresolve a comment please when needed. 

I am not sure what is the appropriate etiquet in Github but I do use the resolved comment field to sort of tick that I've addressed a comment (be it with a comment response or a code change) but sometimes I am wrong.

projects/vdk-plugins/vdk-meta-jobs/README.md
tozka: That's not upto this plugin. This is managed by the Control Service (we are talking here about the normal re-try of a data job, nothing fancy).

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
```diff
@@ -0,0 +1,39 @@
...
+    fail_meta_job_on_error: bool = True
+    depends_on: List[str] = field(default_factory=list)
+
+
+@dataclass
+class MetaJob(SingleJob):
+    """
+    TODO
+    """
+
```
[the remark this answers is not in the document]
tozka: This is all TODOs are to be handled in https://github.com/vmware/versatile-data-kit/issues/1271
I added that comment there.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
tozka: It's meant to be randomized backoff delays so that the delay is more randomly spread in time and the API Server is not hammered with requests simultaneously. 

It's hardly a novel thing - the spring re-try framework has similar options - https://docs.spring.io/spring-retry/docs/api/current/org/springframework/retry/annotation/Backoff.html#random--

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
tozka: It's an implementation internal interface. It's not meant to be an API (it's in 'vdk.api' package)

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
tozka: Probably not I added TODO to handle the duplication. I am going to go with rule of 3 on this and wait for now (https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
tozka: Yes, I agree. Created https://github.com/vmware/versatile-data-kit/issues/1280 for this

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_job_runner.py
tozka: That would be handled in https://github.com/vmware/versatile-data-kit/issues/1271

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
tozka: I agree. But This is just a prototype (pre-alpha) plugin. For a prototype, I don't think we should be changing things in core parts.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
tozka: It's there actually - line 64 in meta_dag.py  

I added unit tests that proves that it's not going to spin infinitely (test_execute_dag_busyloop)

dakodakov (approved):

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -0,0 +1,160 @@
...
+* After Job 1 is completed, jobs 2,3,4 will start executing in parallel.
+* Jobs 5 and 6 will start executing after job 3 completes, but will not wait for the completion of jobs 2 and 4.
+
+
+```python
+
+from vdk.api.job_input import IJobInput
+from vdk.plugin.meta_jobs.meta_job_runner import MetaJobInput
+
+JOBS_RUN_ORDER = [
```
doks5, line 63: Out of curiosity, from what I see, all jobs set `fail_meta_job_on_error` to `True`. At the same time jobs 5 and 6 may start executing before jobs 2 and 4 have finished. If, say, job 4 fails while jobs 5 and 6 are still running, this should terminate the meta job. Could that leave the database in an inconsistent state?

tozka: If Job 4 finishes, that would terminate the meta job with an exception, job 5 and 6 would complete as usual (won't be cancelled). 

That also means it's possible for meta job to finish (failure) but orchestrated jobs to finish afterwards.

murphp15 (approved):

PR #1260, opened 2022-10-24
vdk-pipelines-control-service: don't include needless service token.

murphp15:
# Why?
We don't need the token so we shouldn't include it by default

# What?
Set the property so the token is not included

# How has this been tested?
Integration and unit tests are passing
 


tozka:
That is also very useful and it's a security improvement at no cost. (see https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html - "Do not mount the service account credentials in a container if it does not need to access the Kubernetes API.")

So I urge you to finish it and push it. We should do this (preferably separate PR) also for data job started pods.

tozka:
I have recommended some talks on a conference I was [here](https://medium.com/@antoniivanov/scale-19x-recommended-talks-to-watch-ec5e67d52df2) - relevant one for this PR is https://youtu.be/jRrCi1IS7gk?t=14313 (Hardening against Kubernetes Hacks) You can see how this vulnerability (among others) is being exploited . It's very fun talk and very educative about how Kubernetes works

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -1234,6 +1234,7 @@ public void createJob(
     if (StringUtils.isNotEmpty(serviceAccountName)) {
       podSpecBuilder.withServiceAccountName(serviceAccountName);
     }
+    podSpecBuilder.withAutomountServiceAccountToken(false);
```
mivanov1988, line 1237: Consider adding this property through the k8s-data-job-template rather than the source code.

mivanov1988 (approved):

mivanov1988:
This property has been added to one of our internal deployments without any issues.

Issue #1263, opened 2022-10-25
Job Source Upload API security hardening


tozka:
**What is the feature request? What problem does it solve?**

As part of Control Service API we expose [Data Jobs Sources API
](https://iaclqhm5xk.execute-api.us-west-1.amazonaws.com/data-jobs/swagger-ui.html#/Data_Jobs_Sources) which enable users 
to upload their job and we'd version it and store it (in git) 

This channel though opens the opportunity for an attack 
* At the control service we if we do not put proper input validation 
* At the data job runtime if we allow malicious or bad files to be uploaded and potentially executed. 

OWASP has recommendation for file upload validation which we can implement https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html 

**Suggested solution** 

My proposal is 
 

* re-implement [FileUtils.unzipDataJob](https://github.com/vmware/versatile-data-kit/blob/3507c2dfdcfe5dba3187df639df246fad5820de9/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/FileUtils.java#L94) with the implementation in https://wiki.sei.cmu.edu/confluence/display/java/IDS04-J.+Safely+extract+files+from+ZipInputStream
* Impelment JobUploadValidator (see below) which can be used in JopUpload likely after it's unzipped https://github.com/vmware/versatile-data-kit/blob/3507c2dfdcfe5dba3187df639df246fad5820de9/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUpload.java#L132

The OWASP recommends the use of allowlist over blocklist but there may be cases where you are not sure what type of file formats
you need to allow - in our case, in some cases one may want to enable users to upload some helper files - in some cases they may be allowed to even upload .so (pre-compiled dynamic library) files.
 
So it would look like this

```
class JobUploadValidator {  
  /**
    List of file types  that  are blocked from being uploaded.
    Full list of file types are documented in https://tika.apache.org/1.10/formats.html#Full_list_of_Supported_Formats
    If set to empty, then no file is blocked using the blocklist
  */
  @Value{"upload.fileTypes.blocklist:"}
  List<String> blockList;  

  /**
    List of files that are allowed to be uploaded. 
    Full list of file types are documented in https://tika.apache.org/1.10/formats.html#Full_list_of_Supported_Formats
    If set to empty, then all files are allowed using the allowlist. 
  */
[...]

PR #1266, opened 2022-10-26
control-service: add tests for some zip exploits

tozka:
The library we use to unzip (lingala.zip4j) has a protection against zip slip .

But it doesn't have for zip bomb
https://github.com/srikanth-lingala/zip4j/issues/378

In https://github.com/vmware/versatile-data-kit/issues/1263 I proposed we implement proposal from https://wiki.sei.cmu.edu/confluence/x/LjZGBQ

But I've reconsidered. The more dangerous exploit (zip slip) is covered. While DOS is not nice it's unlikely as VDK is expected to be installed inside organization with generally trusted users.

I  dislike the idea of expanding unzip method which is 6 lines into about a 60 in the proposed solution which we need to maintain and test completely.

Testing Done: unit tests


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/FileUtils.java
```diff
@@ -85,6 +85,10 @@ public static void copy(String jobFolder, String jobRepoFolder, String jobName)
    * <p>in case the job_folder name is different than the job name the method will rename it to be
    * the same as the job name.
    *
+   * <p>The function has a protection against zip slip exploit but it doesn't have against zip bomb
+   * (which is type of DOS attack). If someone tries to upload zipbomb then the pod may restart so
+   * it's not that big of an issue.
+   *
```
ivakoleva, line 91: since we already use several apache commons bundles, maybe use Apache POI https://poi.apache.org/apidocs/dev/org/apache/poi/openxml4j/util/ZipSecureFile.html, see `ZipArchiveThresholdInputStream`

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/FileUtils.java
tozka: It's not part of apache commons. It's part of separate module apache poi. It's basically Java API for Microsoft Documents which we do not need. I prefer to convice the net.lingala.zip4j people to accept adding support for this.

dakodakov (approved):

PR #1267, opened 2022-10-27
vdk-core: Allow for modification of dynamic params

doks5:
The ingestion functionality in vdk-core allows for the creation of pre-processor plugins that modify the passed payload. This works for most use-cases, but not when an operator of Versatile Data Kit needs to modify or otherwise ammend the dynamically set parameters like target, destination_table, etc., to comply with specific rules.

This change adds functionality to allow for the modification of dynamically set parameters like target by pre-processor plugins. Once modified, the parameters are to be placed in a dictionary, which is then to be added to the metadata using the `updated_dynamic_params` key.

Testing Done: Unit test

### DIFF
--- a/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
+++ b/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
@@ -184,7 +184,9 @@ class IIngesterPlugin:
 
     pre_ingest_process() - plugins implementing this method do not aim at
                         ingesting the payload, rather to do some processing
-                        of the payload before it is ingested.
+                        of the payload or the dynamic parameters like target
+                        or destination_table before the actual ingestion
+                        happens.
 
     post_ingest_process() - plugins implementing this method aim at doing
                             some metadata processing after the payload is
@@ -234,6 +236,11 @@ class IIngesterPlugin:
     """
 
     IngestionMetadata = NewType("IngestionMetadata", Dict)
+
+    UPDATED_DYNAMIC_PARAMS = "updated_dynamic_params"
+    TARGET_KEY = "target"
+    DESTINATION_TABLE_KEY = "destination_table"
+    COLLECTION_ID_KEY = "collection_id"
 
     def ingest_payload(
         self,
@@ -293,8 +300,10 @@ class IIngesterPlugin:
             ) -> Optional[IngestionMetadata]:
                 result = None
                 metadata: IngestionMetadata = dict()
-                metadata["destination_table"] = destination_table
-                metadata["target"] = target
+                metadata[IIngesterPlugin.DESTINATION_TABLE_KEY] = (
+                            destination_table
+                )
+                metadata[IIngesterPlugin.TARGET_KEY] = target
 
                 try:
                     result = requests.post('example.com', data=payload)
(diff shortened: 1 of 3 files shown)


projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
```diff
@@ -234,6 +238,44 @@ def test_pre_ingestion_operation():
         collection_id=shared_test_values.get("collection_id"),
         metadata=None,
     )
+    ingester_base._ingester.ingest_payload.assert_called_once()
+
+
+def test_pre_ingestion_updated_dynamic_params():
```
murphp15, line 244: I really like this test is it very clean.

murphp15 (approved):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -376,6 +378,28 @@ def pre_ingest_process(self,
...
+                .. code-block:: python
+                    def pre_ingest_process(self,
+                        payload: List[dict],
+                        destination_table: Optional[str],
+                        target: Optional[str],
+                        collection_id: Optional[str],
+                        metadata: Optional[IngestionMetadata],
+                    ) -> Tuple[List[Dict], Optional[IngestionMetadata]]:
+                        if metadata:
+                            metadata["updated_dynamic_params"] = {
```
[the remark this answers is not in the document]
tozka: I am not sure how much we want to change IngestionMetadata. But instead of string , they normal attributes we may have intellisense and such 
It is always better to design interfaces that prevent errors from happening.
 
Have you considered returning them in pre_ingest ? 

return payload, metadata, destination_table, target 


Or perhaps having class PreIngestResult { payload, metadata, destination, target } 

??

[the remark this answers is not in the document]
doks5: I agree that changing IngestionMetadata is not optimal, and at first tried to extend the API with returning a dictionary containing all the updated params. However, due to the possibility to have multiple pre-processors, we would need to also include the data structure/type to the call arguments list. As I was making the changes, realised that this would be quite a breaking change, and it would probably break stuff used by custom pre-ingest plugins used in production environments. The also would make the API heavier than it already is. That's why I went for the most minimal change possible.

That being said, I agree that having a constant is much better choice than expecting users to remember how to type the key correctly. Code is updated for this.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -439,6 +439,21 @@ def _payload_poster_thread(self):
...
+                            )
+                            target = updated_dynamic_params.get("target") or target
+                            collection_id = (
+                                updated_dynamic_params.get("collection_id")
+                                or collection_id
+                            )
+
                     try:
                         ingestion_metadata = self._ingester.ingest_payload(
                             payload=payload_obj,
```
ivakoleva, line 459: why not simplify by reusing `ingestion_metadata` directly, instead of subkey-ing? in case any real parameter name is found in ingestion_metadata, just use it as a parameter for ingest_payload?

doks5: I believe sub-keying `ingestion_metadata` actually makes it cleaner, as we essentially create a sub-section which if exists, then at least one parameter has been updated updated, and we don't need to have a separate check for each and every parameter.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -439,6 +439,21 @@ def _payload_poster_thread(self):
                     for payload_dict in payload_obj:
                         self.__verify_payload_format(payload_dict=payload_dict)
 
+                    if ingestion_metadata:
+                        updated_dynamic_params: Optional[dict] = ingestion_metadata.pop(
+                            IIngesterPlugin.UPDATED_DYNAMIC_PARAMS, None
+                        )
+                        if updated_dynamic_params:
+                            destination_table = (
+                                updated_dynamic_params.get("destination_table")
```
tozka, line 448: Both "destination_table" and "target" should be constants as well. They are used in more than one place for certain and must be the same.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
tozka: What do you mean by "make the API heavier" ?  Where is the term coming from? 

That being said the reasoning sounds good to me. Let's make target and destination constant (or enum) - they are pre-defined set of values that must be set correctly - candidate for constants.

tozka (approved):

projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_base.py
doks5: Thanks

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
doks5: Fixed, thanks

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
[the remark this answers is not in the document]
doks5: I meant it as in "developers need to account for an increasing number of call arguments, etc.". It is not an official term or anything like that.

PR #1268, opened 2022-10-28
control-service: upload validation

tozka:
As part of PR 1263 we want to implement some security hardening of our
upload sources endpoint . OWASP has recommendation for file upload
validation which we try to follow :
https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html

This implements the JobUploadValidator proposal in
https://github.com/vmware/versatile-data-kit/issues/1263

Testing Done: unit tests


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadValidator.java
```diff
@@ -0,0 +1,92 @@
...
+  private void validateFile(String jobName, Path filePath, Path pathInsideJob)
+      throws InvalidJobUpload {
+    if (filePath.toFile().isDirectory()) {
+      return;
+    }
+    try {
+      var fileType = this.formatDetector.detectFileType(filePath);
+      log.debug("Job {}'s file {} is of type {}", jobName, filePath, pathInsideJob);
+      if (Arrays.stream(allowlist)
+          .noneMatch(allowed -> formatDetector.matchTypes(fileType, allowed))) {
```
mivanov1988, line 74:
```suggestion
      if (!Arrays.stream(allowlist)
          .anyMatch(allowed -> formatDetector.matchTypes(fileType, allowed))) {
```

Slightly more efficient in some cases.

mivanov1988:
We should use Ephemeral Storage (emptyDir) for the uploaded files to decouple them from the root file system.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadValidator.java
tozka: It shouldn't be ? Why ? If you mean short-circuiting ...

Both noneMatch and anyMatch are short-circuiting stream operations. And they will stop processing their input when the result becomes known - in our case of it detects single match .

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#noneMatch(java.util.function.Predicate)

tozka:
> We should use Ephemeral Storage (emptyDir) for the uploaded files to decouple them from the root file system.

Yes, I was thinking the same. But not as part of this change. Created https://github.com/vmware/versatile-data-kit/issues/1279 for this.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/FileFormatDetector.java
```diff
@@ -0,0 +1,63 @@
...
+   * returns True match("text/sql", "text/plain") returns False match("text/plain", "application")
+   * returns False
+   *
+   * @param detectedType the type that was detected and need to be verified (must in format
+   *     "type/subtype")
+   * @param targetType the target type against which it is checked (can be either "type" or
+   *     "type/subtype")
+   * @return true or false
+   */
+  public boolean matchTypes(String detectedType, String targetType) {
```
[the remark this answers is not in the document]
tozka: Fixed.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/FileFormatDetector.java
```diff
@@ -0,0 +1,63 @@
...
+    } else { // we compare only by type (text)
+      return detectedMediaType.getType().equals(targetType);
+    }
+  }
+
+  /**
+   * Return the detected file type Full list of file types are documented in
+   * https://tika.apache.org/1.10/formats.html#Full_list_of_Supported_Formats
+   */
+  public String detectFileType(Path filePath) throws IOException {
```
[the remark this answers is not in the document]
tozka: I want to keep it as an instance instead of static so that it's more flexible to changes in the future. I did not add an interface but we can have another type of file type detector in the future. It's not really just an util class.

dakodakov (changes requested):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadValidator.java
```diff
@@ -0,0 +1,92 @@
...
+
+  public JobUploadValidator(
+      @Value("${upload.validation.fileTypes.allowlist:}") String[] allowList) {
+    this.allowlist = allowList;
+    this.formatDetector = new FileFormatDetector();
+    log.debug("Job sources upload validator allowlist: {}", Arrays.toString(allowList));
+  }
+
+  /**
+   * Validate a file in a data job based on its file type : If allowed list is NOT empty and it's
```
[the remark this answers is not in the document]
tozka: fixed

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadValidator.java
mivanov1988: You are right, sorry.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -529,7 +529,7 @@ public List<JobDeploymentStatus> readV1CronJobDeploymentStatuses() {
               .listNamespacedCronJob(
                   namespace, null, null, null, null, null, null, null, null, null, null);
     } catch (ApiException e) {
-      log.warn("Failed to read k8s cron jobs: ", e);
+      log.warn("Failed to read k8s cron jobs: ", new KubernetesException("", e));
```
murphp15, line 532: why do you wrap it in a new exception?

tozka: Because KuberneresException prints more details (it prints the actual error from the K8S API Server which the default exception does not :( )

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadValidator.java
```diff
@@ -0,0 +1,103 @@
...
+   * forbidden. The validation done is by detecting the file type and checking if that file type is
+   * allowed against preconfigurd list controlled by "upload.validation.fileTypes.allowlist". If the
+   * allowList is empty then all files are allowed.
+   *
+   * @param jobName the name of the data job whose files are validated
+   * @param jobDirectory path to the data job directory where unarchived content of the data job
+   *     being uploaded can be seen.
+   * @throws InvalidJobUpload
+   */
+  public void validateJob(String jobName, Path jobDirectory) throws InvalidJobUpload {
```
murphp15, line 50: Is something else going to happen in this method eventually? 
At the moment all it does is call another method?

murphp15 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadValidator.java
tozka: I suppose not. It remained alone after some refactoring. I like to keep the name because it's clear what the method does. 

I like to keep validateJob as the name of the method since we can have multiple validators and in the future, we can create an interface , and that would be one implementation.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/FileFormatDetector.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.tika.mime.MediaType;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+
+class FileFormatDetector {
+
+  private static final Map<String, MediaType> CACHE_MEDIA_TYPES = new HashMap<>();
```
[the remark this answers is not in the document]
tozka: Yes, that makes sense. Spring bean framework would create a single instance anyhow (or rather as few as necessary) so it's not much of an optimization.

dakodakov (approved):

murphp15:
> We should use Ephemeral Storage (emptyDir) for the uploaded files to decouple them from the root file system.

What is the benefit of this?

tozka:
> > We should use Ephemeral Storage (emptyDir) for the uploaded files to decouple them from the root file system.
> 
> What is the benefit of this?

That seems better discussed in the ticket - https://github.com/vmware/versatile-data-kit/issues/1279

PR #1269, opened 2022-10-28
control-service: refactor db query to mitigate race condition

mrMoZ1:
what: Refactored how the watch task updates missing kubernetes job executions. It should now update the statuses directly against the database, instead of loading entities in memory, updating them (in memory) and saving to the database.

why: Users reported that when cancelling executions, the cancelled execution's status is visible as Succeeded, even tough the API returned success and the status was set to Cancelled. This is caused by the watch task which updates executions in bulk and in memory before saving them to the database.

testing: Added unit tests that cover the new repository query. 

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
@@ -373,25 +373,25 @@ public class JobExecutionService {
     if (runningJobExecutionIds == null) {
       return;
     }
-
+    var runningJobStatus = List.of(ExecutionStatus.SUBMITTED, ExecutionStatus.RUNNING);
     List<com.vmware.taurus.service.model.DataJobExecution> dataJobExecutionsToBeUpdated =
         jobExecutionRepository
             .findDataJobExecutionsByStatusInAndStartTimeBefore(
-                List.of(ExecutionStatus.SUBMITTED, ExecutionStatus.RUNNING),
+                runningJobStatus,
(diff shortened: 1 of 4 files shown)


murphp15:
Code change looks good! But can you add a test case for this?

mivanov1988 (approved):

doks5 (approved):
LGTM

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
```diff
@@ -373,25 +373,24 @@ public void syncJobExecutionStatuses(List<String> runningJobExecutionIds) {
     if (runningJobExecutionIds == null) {
       return;
     }
-
+    var runningJobStatus = List.of(ExecutionStatus.SUBMITTED, ExecutionStatus.RUNNING);
     List<com.vmware.taurus.service.model.DataJobExecution> dataJobExecutionsToBeUpdated =
         jobExecutionRepository
             .findDataJobExecutionsByStatusInAndStartTimeBefore(
```
ivakoleva, line 379: Why priorly selecting the entire entity, to then only filter by its id for the subsequent update query?

ivakoleva: @mrMoZ1

mrMoZ1: Created a new pull request addressing your comment: https://github.com/vmware/versatile-data-kit/pull/1278

PR #1277, opened 2022-11-01
vdk-control-service: fix ingress template

murphp15:
# Why?
The current check always returns false

# What?
Change to the property we should actually be checking for 

# How has this been tested?
I created a helm chart locally which looked like this

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  apiVersions.txt: |
    {{ .Capabilities.APIVersions.Has "networking.k8s.io/v1"  }}
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/ingress" }}
  blah: hasC
{{- else }}
  blah: noC
{{- end }}
{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
  blah2: hasC
{{- else }}
  blah2: noC
{{- end }}
```

I rendered it with helm by running `helm template myhelmchart `. The underlying kubectl version was 1.23.x.

the rendered template was: 
```
# Source: apiversiontest/templates/serviceaccount.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  apiVersions.txt: |
    true
  blah: noC
  blah2: hasC

```
This shows that it finds networking v1 but not the ingress version we were using before. 

Bug fix

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/ingress.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/ingress.yaml
@@ -4,7 +4,7 @@ metadata:
   name: {{ include "common.names.fullname" . }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
   annotations: {{- toYaml .Values.ingress.annotations | nindent 4 }}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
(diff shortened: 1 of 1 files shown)


tozka:
How did you do "I checked the property locally." What commands did you use. 

Avoid saying I tested manually or locally but explain what was done locally or manually (e.g what commands, what was verified).

gageorgiev (approved):

murphp15:
@tozka added more detailed testing description.

tozka:
> @tozka added more detailed testing description.

Thanks.

tozka (approved):

PR #1278, opened 2022-11-01
control-service: minor db query optimisation refactor

mrMoZ1:
what: Refactored a query in the JobExecutionRepository to only return the execution id, instead of the whole entity.

why: The query is used by the watch task which after recent changes only needs the id of executions. 

testing: added unit tests that cover the changed repository method.


mivanov1988:
List<com.vmware.taurus.service.model.DataJobExecution> dataJobExecutionsToBeUpdated is used for logging purposes as well, have you considered that?

mrMoZ1:
> 

No I missed that. Is it strictly necessary that we log the full data job entity or can we log just a subset of that information? Essentially is this optimization worth it, or should I discard it?

mrMoZ1:
Closing as we discussed that the whole entity is useful for debugging purposes and that the impact would be minimal since retrieved entities' numbers are quite low.

PR #1283, opened 2022-11-02
vdk-control-service: add support for allowlist in helm chart.

murphp15:
# Why?
We want to be able to configure the file scanning from the helm chart as described here: [TAUR-5633](https://jira.eng.vmware.com/browse/TAUR-5633)

# What?
Update the helm chart to support the property

# How has this been tested?
I created a helm chart locally called `pipelines-control-service-prerelease` and pushed it to the remote helm repo.
I then installed in the vdk aas env.
After I changed the values.yaml files in the vdk aas repo to make sure the property propagated as expected. 
this is the PR where I test: it https://gitlab.eng.vmware.com/taurus/vdk/vdkaas-cicd/-/merge_requests/17 


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -151,7 +151,7 @@ deploymentGitPassword: ""
 # Credentials with read and write access to the Git repository.
 uploadGitReadWriteUsername: ""
 uploadGitReadWritePassword: ""
-
+uploadValidationFileTypesAllowList: ""
```
tozka, line 154: We use values.yaml as an interface so all properties need be documented. 

You can steal the doc I did in application.properties in https://github.com/vmware/versatile-data-kit/pull/1268

tozka:
Testing Done: I think `helm template --debug` or `helm install --dry-run` and verify the pasting here the resulting deployment would have been enough. Though of course your test is much more thorough. 

> pipelines-control-service-prerelease and pushed it to the remote helm repo.
Which helm repo? If it's one we push production releases charts. Please don't. 
Any repo (be it pip, helm, docker) that we use for release should be updated only by our CICD. 

If it's another, then ignore my comment.

tozka (approved):

murphp15:
@tozka 
Its the same repo. 
But I pushed it under the name pipelines-control-service-prerelease so there is no concern of someone picking it up by accident

mivanov1988 (approved):

tozka:
> @tozka Its the same repo. But I pushed it under the name pipelines-control-service-prerelease so there is no concern of someone picking it up by accident

Ok. It's still safer to use your own   repo (e.g the easiest one for you github account there's packages).

PR #1291, opened 2022-11-03
control-service: Container read-only file system

gageorgiev:
To reduce the attack surface of jobs containers, jobs
should run with the root file system set to read-only.
However, this means that job pods cannot write to the
file system, so we need to create an ephemeral file
system that jobs can use instead.

The test job contains one step with the following code:
```py
import logging

import time

from vdk.api.job_input  import IJobInput

def run(job_input: IJobInput):
    log.info("Testing we can still write to the ephemeral filesystem.")

    with open("test-file.txt", 'w') as f:
        f.write("Testing sentence.")

    with open("test-file.txt", 'r') as f:
        if str(f.read()) != 'Testing sentence.':
            raise Exception("File write was unsuccessful.")

    log.info("Testing we cannot write to the root filesystem.")

    try:
        with open("/test-file.txt", 'w') as f:
            f.write("Testing sentence.")
    except OSError:
        log.info("Root filesystem is read-only, as it should.")
    else:
        raise Exception("Root filesystem is not read-only")


```

Testing done: integration test


projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
```diff
@@ -37,4 +37,11 @@ spec:
             runAsUser: 1000
             runAsGroup: 1000
             fsGroup: 1000
+            readOnlyRootFilesystem: true
+          volumeMounts:
```
tozka, line 41: This change requires thought behind UX if we are default to read only. 

Should users be able to create files in the local file system. 

Would this work
```
def run(job_input): 
   with open("my-file", "rw"): 
         # write something   smething ? 
```

And how would it work ?

gageorgiev: Would this not create the file in the ephemeral volume?

tozka: I don't know :) I am asking you that actually and hoping you will test/verify and tell us. 

I think if we set those as default configuration which I think make sense we need to make sure some common use-cases are working or if not there's clear (and easy!) way for users to migrate

ivakoleva: please provide functional test of the local ephemeral storage? Are there any patterns/restrictions on the file paths for end users, or is it completely transparent?

tozka:
This is setting root file system for the builder jobs. And it  looks ok. 


What happens for the data jobs themselves ? Are you planning to do it? It make sense to me to be a separate PR but I just wondered because initially it seems we are changing the data job . 
You need to fix the title if this PR is to be for builder jobs

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -176,6 +179,7 @@ public boolean buildImage(
         builderJobName,
         dockerRegistryService.builderImage(),
         false,
+        readOnlyRootFilesystem,
```
mivanov1988, line 182: JobImageBuilder is responsible for building docker images for data jobs. The builder job shouldn't be affected by this PR.

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mivanov1988: Just defaulting to false?

gageorgiev: It does default to false. What should it default to?

mivanov1988: Yes, but if dataJob.readOnlyRootFileSystem is passed (value: true), it will affect the builder job, which is incorrect. The dataJob.readOnlyRootFileSystem property should affect only the data job.

gageorgiev: gotcha

gageorgiev: amended to have readOnlyRootFilesystem set to false always for builder jobs

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -89,6 +89,9 @@ public class JobImageBuilder {
   @Value("${datajobs.deployment.builder.serviceAccountName}")
   private String builderServiceAccountName;
 
+  @Value("${dataJob.readOnlyRootFileSystem:false}")
+  private boolean readOnlyRootFilesystem;
```
mivanov1988, line 93: not needed anymore.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mivanov1988: Thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
gageorgiev: dropped it, I assumed the Google Java format github action catches unused variables but I guess not.

tozka (changes requested):

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobEphemeralStorageIT.java
```diff
@@ -0,0 +1,380 @@
...
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.header;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@Slf4j
+@Import({DataJobDeploymentCrudIT.TaskExecutorConfig.class})
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+    classes = ControlplaneApplication.class)
+public class DataJobEphemeralStorageIT extends BaseIT {
```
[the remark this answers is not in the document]
gageorgiev: It does not yet, hence the PR still being labelled draft.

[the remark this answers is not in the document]
tozka: Oh, I didn't see. Sorry about that. I saw I was requested for review. I will wait until it's ready.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobEphemeralStorageIT.java
```diff
@@ -0,0 +1,380 @@
...
+                    lambdaMatcher(
+                        s ->
+                            s.endsWith(
+                                String.format(
+                                    "/data-jobs/for-team/%s/jobs/%s",
+                                    TEST_TEAM_NAME, TEST_JOB_NAME)))));
+  }
+
+  @Test
+  public void testEphemeralStorageJob() throws Exception {
```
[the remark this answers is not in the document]
gageorgiev: refactored and moved duplicated code to BaseIT

mivanov1988:
Probably you are going to expose this option through the Helm chart in a separate PR, right?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobEphemeralStorageIT.java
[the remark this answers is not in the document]
tozka: I don't see the refactoring .... 

Btw I am fine if you decide to do this refactoring in separate PR as long as we have an agreement that it will be done with priority.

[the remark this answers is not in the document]
gageorgiev: I cut it as per Miro's suggestion since I was having issues with getting it to work and getting this merged is higher on the priority list.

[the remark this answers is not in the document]
tozka: As I said I think it's ok to merge it as long as we have an agreement on whether it will be refactored or not. 

In any case yes, it's better that purely refactoring changes are done separated from functional changes as it's recommended here - https://github.com/vmware/versatile-data-kit/wiki/How-to-prepare-a-new-PR#tackle-non-functional-or-refactoring-changes-first (the recommendation is that refactoring change is first but the spirit of it is the same).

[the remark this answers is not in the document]
tozka: Just in case it's not clear. I am looking from you if you agree or not. if not why not and we can discuss (I might be wrong). 
 If yes, let's create an issue for that and plan to fix it.

mivanov1988:
> > Probably you are going to expose this option through the Helm chart in a separate PR, right?
> 
> I don't follow. My understanding was that this will be configured through values.yml.

Yes, but now it is not possible, since this property is not exposed yet, right?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobEphemeralStorageIT.java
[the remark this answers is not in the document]
gageorgiev: I will make sure to refactor it, I've made a note of it.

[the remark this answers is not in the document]
tozka: Story: https://github.com/vmware/versatile-data-kit/issues/1330

tozka (approved):

PR #1292, opened 2022-11-04
versatile-data-kit: Adding support to read SQM statement as DataFrame

victor-partico:
#Why
execute_query does not return column name, DataFrames are move useful

#What
Adding new method to job_input

#How has this been tested?
None, could not make build.sh run

#Type

Signed-off-by: Victor Mishovski <vmishovski@vmware.com>

### DIFF
--- a/projects/vdk-core/src/vdk/api/job_input.py
+++ b/projects/vdk-core/src/vdk/api/job_input.py
@@ -7,6 +7,8 @@ from typing import Any
 from typing import List
 from typing import Optional
 
+from pandas import DataFrame
+
 
 class IProperties:
     """
@@ -61,6 +63,29 @@ class IManagedConnection:
         """
         Executes the provided query and returns results from PEP 249 Cursor.fetchall() method, see:
             https://www.python.org/dev/peps/pep-0249/#fetchall
+
+        Query can contain parameters in format -> {query_parameter}. Parameters in the query will
+        be automatically substituted if they exist in Data Job properties and arguments as keys.
+        If parameter with same key is used both as arguments and properties,
+        arguments (get_arguments()) will take precedence over properties (get_property()).
+
+        Note:
+            Parameters are case sensitive.
+            If a query parameter is not present in Data Job properties or arguments it will not be replaced in the query
+            and most likely result in query failure.
+
+        Example usage:
+            job_input.set_all_properties({'target_table': 'history.people','etl_run_date_column': 'pa__arrival_ts'})
+            query_result = job_input.execute_query("SELECT * FROM {target_table} WHERE {etl_run_date_column} > now() - interval 2 hours")
+
+        """
+        pass
+
+    @abstractmethod
+    def read_sql_as_dataframe(self, query_as_utf8_string) -> DataFrame:
+        """
+        Executes the provided query and returns results from PEP 249 pandas.read_sql() method, see:
+            https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html
 
         Query can contain parameters in format -> {query_parameter}. Parameters in the query will
         be automatically substituted if they exist in Data Job properties and arguments as keys.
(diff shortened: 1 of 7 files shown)


tozka:
Thanks for the proposal. I do think it's an interesting and important feature to be able to work with Pandas  easily. 

There are a few things though: 

* Маке sure to follow [vmwclabot](https://github.com/vmwclabot) Instructions 

* I recommend having an associated Github issue that explain what and why do we need this . The very least the "Why" section needs to be much more than 1 sentence . See https://github.com/vmware/versatile-data-kit/blob/main/CONTRIBUTING.md#how-to-prepare-your-change

* I am not sure the change should be part of "vdk-core" itself. It introduces a new dependency in vdk-core - "pandas" - and as vdk-core is used in every single plugin and everywhere (vdk related) it needs to have as little dependencies as possible to avoid dependencies conflicts or some variant of [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell). 

You can consider making it a [plugin](https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-plugins) (vdk-pandas). Changes in vdk-core are much harder to be accepted as they have much bigger impact.
While we welcome all kinds of plugins. 

* Changes in vdk-core and in plugins must be in separate PRs - first for the change in vdk-core and then for the plugins. Though the above point may make this one moot.

tozka:
Btw the reason build.sh and tests fail (at least what I can see from https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3274269747) is because "ModuleNotFoundError: No module named 'pandas'" 

New dependencies are added in setup.cfg (or setup.py) file in[ install_requires section](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#id5)

projects/vdk-core/src/vdk/api/job_input.py
```diff
@@ -79,6 +81,29 @@ def execute_query(self, query_as_utf8_string) -> List[List]:
...
+        arguments (get_arguments()) will take precedence over properties (get_property()).
+
+        Note:
+            Parameters are case sensitive.
+            If a query parameter is not present in Data Job properties or arguments it will not be replaced in the query
+            and most likely result in query failure.
+
+        Example usage:
+            job_input.set_all_properties({'target_table': 'history.people','etl_run_date_column': 'pa__arrival_ts'})
+            query_result = job_input.execute_query("SELECT * FROM {target_table} WHERE {etl_run_date_column} > now() - interval 2 hours")
```
ivakoleva, line 102: include `read_sql_as_dataframe` to the Example usage;
also include into plugin's README.md added

ivakoleva (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_connection_base.py
```diff
@@ -157,6 +158,31 @@ def execute_query(self, query: str) -> List[List[Any]]:
                 List[List[Any]], res
             )  # we return None in case of DML. This is not PEP249 compliant, but is more convenient
 
+    def read_sql_as_dataframe(self, query: str) -> pd.DataFrame:
+        """Read SQL statement and returns the result as pandas DataFrame"""
+        with self.connect() as db:
+            try:
+                res = pd.read_sql(query, con=db)
```
murphp15, line 165: I would probably just do `return pd.read_sql(query, con=db)` here and then `return None` at the end of the catch statement. But might just be personal preference.

projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/postgres_connection.py
```diff
@@ -112,3 +113,9 @@ def execute_query(self, query: str) -> List[List[Any]]:
             return super().execute_query(query)
         finally:
             self.commit()
+
+    def read_sql_as_dataframe(self, query: str) -> List[List[Any]]:
```
murphp15, line 117: Should the return type not be dataframe?

tozka:
We talked with Victor and decided we will go with documentation improvement since ultimately this is currently possible with VDK using 
```
pandas.read_sql("select * from x" , job_input.get_managed_connection()) 
```

PR #1293, opened 2022-11-04
vdk-control-service: add empty dir storage

murphp15:
# Why?
Currently JobUpload uploads the file in the same file system as the container process so if too many requests with big data jobs are uploaded it may crash the process/the whole container

Beyond that it's recommended by https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html#file-storage-location to avoid using web root location for staring user supplied files.

# What?
add an empty directory volume mount to the pipelines-control-service pods. 
Use this to store the files in. 

# How has this been tested?
made a release of the helm chart under a different name and deployed to the vdk vdp cluster. 
Deployed a heartbeat job.

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
@@ -185,7 +185,7 @@ spec:
             - name: UPLOAD_VALIDATION_FILETYPES_ALLOWLIST
               value: "{{ .Values.uploadValidationFileTypesAllowList }}"
             - name: DATAJOBS_TEMP_STORAGE_FOLDER
-              value: /DATAJOBS_TEMP_STORAGE_FOLDER
+              value: /datajobs_temp_storage
             {{- if .Values.datajobTemplate.enabled }}
             - name: K8S_DATA_JOB_TEMPLATE_FILE
               value: "/etc/datajobs/k8s_data_job_template.yaml"
@@ -249,7 +249,7 @@ spec:
             - name: secrets
               mountPath: "/etc/secrets"
               readOnly: true
-            - mountPath: "/DATAJOBS_TEMP_STORAGE_FOLDER"
+            - mountPath: "/datajobs_temp_storage"
               name: DATAJOBS_TEMP_STORAGE_FOLDER_VOLUME
             {{- if .Values.datajobTemplate.enabled }}
             - name: datajob-template-volume
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
@@ -30,14 +30,6 @@ spec:
       {{- if .Values.serviceAccount.create }}
       serviceAccountName: {{ template "pipelines-control-service.serviceAccountName" . }}
       {{- end }}
-      volumes:
-        - name: DATAJOBS_TEMP_STORAGE_FOLDER_VOLUME
-          ephemeral:
-            volumeClaimTemplate:
-              spec:
-                resources:
-                  requests:
-                    storage: 1Gi
       containers:
         - name: pipelines-control-service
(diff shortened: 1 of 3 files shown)


projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -242,10 +244,13 @@ spec:
             - name: {{ $key }}
               value: "{{ $value }}"
             {{- end }}
+
           volumeMounts:
             - name: secrets
               mountPath: "/etc/secrets"
               readOnly: true
+            - mountPath: "/DATAJOBS_TEMP_STORAGE_FOLDER"
```
tozka, line 252: is there a reason that the mounthPath is capitalized. I initially thought it's a env variable.

ivakoleva, line 252: > made a release of the helm chart under a different name and deployed to the vdk vdp cluster.
> Deployed a heartbeat job.

could you also paste the test result file structure found?

ivakoleva (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
murphp15: no reason. I thought it made it a little more consistent everywhere. 
But I can change it.

murphp15: I can't actually provide you with this because the files are cleaned up after each operation.
If you really want to see if I could try run the `ls` operations while it is in the middle of an upload. 
what do you want to see it?

murphp15: here is the result. 
```
I have no name!@vdk-control-service-dep-7ffd6f875c-5pnvd:/application$ ls /DATAJOBS_TEMP_STORAGE_FOLDER
job_562d678e-0396-4aa3-a924-418701ebf2c1
```

murphp15: Will i change the lowercase? I have absolutely no problem doing that!

tozka: If it's a consistency we are looking for - the name of the volumes and directories are in lowercase (e.g "datajob-template-volume") everywhere in the file. Only env. variable are upper case. 


I'd name the volume "datajobs-temp-storage-volume" 
 the folder "datajobs_temp_storage" (folder seems redundant in the name of a folder)
and the env variable "DATAJOBS_TEMP_STORAGE_FOLDER" (capitalized).

tozka (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -246,10 +248,13 @@ spec:
             - name: {{ $key }}
               value: "{{ $value }}"
             {{- end }}
+
           volumeMounts:
             - name: secrets
               mountPath: "/etc/secrets"
               readOnly: true
+            - mountPath: "/datajobs_temp_storage"
```
mivanov1988, line 256: Maybe it's a good idea to expose mountPath as variable in values.yaml and then re-use it in both `envs:` and `volumeMounts:`.

murphp15: why would someone want to change the location? 
I don't want to add extra configuration unless there is a good use case for it.

mivanov1988: It's a bit safer in terms of future updates since it is not so obvious that the mountPath is passed as env var as well. But feel free to discard my comment.

mivanov1988 (approved):

PR #1295, opened 2022-11-07
vdk-control-service: fix helm template

murphp15:
### Why
In a previous commit a helm chart but was submitted. 
### What
This makes a small correction as repo type is no longer used 

### How was it tested
I have added in a test step in the PR pipeline to ensure this never happens again. 

### DIFF
--- a/projects/control-service/cicd/test-testing-pipelines-service.sh
+++ b/projects/control-service/cicd/test-testing-pipelines-service.sh
@@ -37,11 +37,11 @@ helm dependency update --kubeconfig=$KUBECONFIG
 #
 # image.tag is fixed during release. It is set here to deploy using latest change in source code.
 # We are using here embedded database, and we need to set the storageclass since in our test k8s no default storage class is not set.
-helm template $RELEASE_NAME . \
+helm template "$RELEASE_NAME" . \
       --set image.tag="$TAG" \
       --set resources.limits.memory=2G \
       --set credentials.repository="EMPTY" \
-      --set-file vdkOptions=$VDK_OPTIONS_SUBSTITUTED \
+      --set-file vdkOptions="$VDK_OPTIONS_SUBSTITUTED" \
       --set deploymentGitUrl="$CICD_GIT_URI" \
       --set deploymentGitUsername="$CICD_GIT_USER" \
       --set deploymentGitPassword="$CICD_GIT_PASSWORD" \
--- a/projects/control-service/cicd/test-testing-pipelines-service.sh
+++ b/projects/control-service/cicd/test-testing-pipelines-service.sh
@@ -28,10 +28,10 @@ export TRINO_HOST=${TRINO_HOST:-"test-trino"}
 
 # Update vdk-options with substituted variables like sensitive configuration (passwords)
 export VDK_OPTIONS_SUBSTITUTED="${VDK_OPTIONS}.temp"
-envsubst < $VDK_OPTIONS > $VDK_OPTIONS_SUBSTITUTED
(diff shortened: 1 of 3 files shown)


projects/control-service/cicd/test-testing-pipelines-service.sh
```diff
@@ -0,0 +1,61 @@
...
+envsubst < "$VDK_OPTIONS" > "$VDK_OPTIONS_SUBSTITUTED"
+
+cd "$TPCS_CHART" || exit
+helm dependency update --kubeconfig="$KUBECONFIG"
+
+# TODO :change container images with official ones when they are being deployed (I've currently uploaded them once in ghcr.io/tozka)
+#
+# image.tag is fixed during release. It is set here to deploy using latest change in source code.
+# We are using here embedded database, and we need to set the storageclass since in our test k8s no default storage class is not set.
+helm template "$RELEASE_NAME" . \
```
tozka, line 40: I like the test but not the code duplication now when we change the helm arguments we have to do it in two places

tozka (changes requested):

gageorgiev:
I'm not sure I understand what the issue previously was.

projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl
```diff
@@ -211,7 +211,7 @@ VDK distribution docker repository secret name
 {{- end }}
 
 {{- define "shouldCreatePipelinesControlServiceDockerRepoSecret" }}
-  {{- if and (eq .Values.image.registryType "generic") .Values.image.registryUsernameReadOnly .Values.image.registryPasswordReadOnly }}
+  {{- if and (.Values.image.registryUsernameReadOnly) (.Values.image.registryPasswordReadOnly) }}
```
murphp15, line 214: @gageorgiev  there was no property called .Values.image.registryType. That is why it was failing.

projects/control-service/cicd/test-testing-pipelines-service.sh
murphp15: the template function is checking the validity of the possible helm charts. 
I have removed all properties for clarity.
This should catch the bug I previously introduced.

mivanov1988:
can't it be used `helm lint` for the verification?

The output of `helm lint`:
```
==> Linting .
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: template: pipelines-control-service/templates/secret_pull_pipelines_control_service_img.yaml:1:8: executing "pipelines-control-service/templates/secret_pull_pipelines_control_service_img.yaml" at <include "shouldCreatePipelinesControlServiceDockerRepoSecret" .>: error calling include: template: pipelines-control-service/templates/_helpers.tpl:214:14: executing "shouldCreatePipelinesControlServiceDockerRepoSecret" at <eq .Values.image.registryType "generic">: error calling eq: incompatible types for comparison
[WARNING] /home/mivanov/Development/Projects/github/versatile-data-kit/projects/control-service/projects/helm_charts/pipelines-control-service: chart directory is missing these dependencies: common,cockroachdb

Error: 1 chart(s) linted, 1 chart(s) failed
```

tozka (approved):
I think we should prioritize fixing the pipeline on main https://gitlab.com/vmware-analytics/versatile-data-kit/-/commits/main as soon as possible. Usually the easiest way to achieve that is to revert temporarily. 

Otherwise, any other development work is blocked on the control service including any other emergency releases, hotfixes, etc. 

So let's merge this if it fixes it. And you can address any comments in follow-up review

mivanov1988 (approved):

mivanov1988 (approved):

Issue #1275, opened 2022-11-01
Jupyter Notebook Steps


duyguHsnHsn:
**What is the feature request? What problem does it solve?**
For the VDK Jupyter integration we cannot use the sanitary file based step (see [vdk-run](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/file_based_step.py) ) since we will not be using .sql and .py files, instead we will be getting the code for the steps in JSON format (see [nb-json-schema](https://github.com/jupyter/nbformat/blob/main/nbformat/v4/nbformat.v4.schema.json) ) format from Jupyter notebooks.

**Suggested solution**
Add a new plugin for the Jupyter integration and introduce new kind of Job Step and new kinds of step/job run methods to be able to work with Jupyter notebooks.

duyguHsnHsn:
The plugin is added (vdk-notebook), closing this issue.

PR #1297, opened 2022-11-08
vdk-jupyter: Run data jobs with ipynb files

duyguHsnHsn:
what: A solution for [issue-1275]( https://github.com/vmware/versatile-data-kit/issues/1275). With this plugin we can now use ipynb files in our jobs and use CLI run with it.

why: It is a crucial feature for VDK Jupyter integration.

testing: Added unit tests


projects/vdk-plugins/vdk-jupyter/README.md
```diff
@@ -0,0 +1,45 @@
+# jupyter
+
+Plugin for runnig Jupyter notebooks as vdk steps
+
+TODO: what the project is about, what is its purpose
```
tozka, line 5: Prototype or not, that's important.

projects/vdk-plugins/vdk-jupyter/setup.py
```diff
@@ -0,0 +1,34 @@
...
+    long_description=pathlib.Path("README.md").read_text(),
+    long_description_content_type="text/markdown",
+    install_requires=["vdk-core"],
+    package_dir={"": "src"},
+    packages=setuptools.find_namespace_packages(where="src"),
+    # This is the only vdk plugin specifc part
+    # Define entry point called "vdk.plugin.run" with name of plugin and module to act as entry point.
+    entry_points={"vdk.plugin.run": ["jupyter = vdk.plugin.jupyter.jupyter_plugin"]},
+    classifiers=[
+        "Development Status :: 4 - Beta",
```
tozka, line 27: The Development Status should be 2 Pre-Alpha (as defined here https://martin-thoma.com/software-development-stages/#2-pre-alpha) that seems the best fit currently.

projects/vdk-plugins/vdk-jupyter/setup.py
```diff
@@ -0,0 +1,34 @@
...
+        "License :: OSI Approved :: Apache Software License",
```
tozka, line 28: It's not in the template. but it's good idea to add python_requires >=3.7

projects/vdk-plugins/vdk-jupyter/src/vdk/plugin/jupyter/notebook_based_step.py
```diff
@@ -0,0 +1,150 @@
...
+                + " and continuing with other files (if present).",
+                filename,
+                filename,
+            )
+            return success
+        finally:
+            sys.path.remove(str(step.job_dir))
+
+    @staticmethod
+    def invoke_run_function(func: Callable, job_input: IJobInput, step_name: str):
```
tozka, line 105: We discussed face to face to add a TODO to consider removing duplication with vdk-core in the future. Now is not the correct moment to do so - https://refactoring.guru/refactoring/when

tozka (changes requested):
Looks good to me.

projects/vdk-plugins/vdk-jupyter/README.md
```diff
@@ -0,0 +1,40 @@
+# jupyter
+
+Prototype of a plugin for running Jupyter notebooks as vdk steps
+
+TODO: what the project is about, what is its purpose
+
```
ivakoleva, line 6: Are the notebooks compiled to python steps that are evaluated, or are those evaluated directly? I think this is a part of the VEP-994, that links design and implementation stories; 
Also, the relationships are bidirectional, for example the `airflow-provider-vdk` does link the VEP-554 of origin.

duyguHsnHsn: Rhe notebooks are compiled to python and sql steps as the standard ones with one additional field "code" which holds the code we get from cell/cells as string.
Since it is just a prototype, I have not worked on the README that much. When doing the real plugin I will create the links. I will enhance the README and the VEP in the next few days.

PR #1304, opened 2022-11-09
jupyter-vep: Expand details on extensions design

duyguHsnHsn:
What: Changed the high level design diagram
Why:
Since introducing VDK variables directly is not a high level requirement the iPython package module should be removed from the VEP's high level design.  The new diagram for the High level design is much more accurate and gives a better idea of what the changes will be. 
The iPython package with the VDK variables can be introduces later on as an enhancement but should not have a place in the bigger picture of VDK Jupyter Integration.

Signed-off-by:  Duygu Hasan <hduygu@vmware.com>


tozka (approved):

specs/vep-994-jupyter-notebook-integration/README.md
```diff
@@ -76,36 +76,29 @@ Legend (terms and table are based on Pragmatic Marketing recommendations):
 JupyterLab is the next-generation user interface for Project Jupyter offering all the familiar building blocks of the classic Jupyter Notebook (notebook, terminal, text editor, file browser, rich outputs, etc.) in a flexible and powerful user interface.
 Fundamentally, JupyterLab is designed as an extensible environment. JupyterLab extensions can customize or enhance any part of JupyterLab.
 
-The proposed design describes the solution for creating a Jupyter UI for VDK. For this purpose, a JupyterLab extension will be implemented, and it will give us the chance to create graphical elements like buttons and widgets – alternatives of currently used CLI commands. An iPython package will also be implemented to export VDK variables. This package will be imported to JupyterLab, and they will be accessible from there.
+The proposed design describes the solution for creating a Jupyter UI for VDK. For this purpose, a JupyterLab extension will be implemented, and it will give us the chance to create graphical elements like buttons and widgets – alternatives of currently used CLI commands. 
+
```
ivakoleva, line 80: I actually found the iPython package information useful.

specs/vep-994-jupyter-notebook-integration/README.md
```diff
@@ -76,36 +76,29 @@ Legend (terms and table are based on Pragmatic Marketing recommendations):
...
+|:----------------------:|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|:------------------:|:--------:|
+|    Run VDK Jupyter     |                                                                                                                                                                        The user runs a single CLI command which opens the web version of JupyterLab. Examples: vdk jupyter lab                                                                                                                                                                         |         1          |   high   |
```
ivakoleva, line 85: Is it a common practice for jupyter to be started from lets say vdk context, or vice versa -> vdk to be started from jupyter?

specs/vep-994-jupyter-notebook-integration/README.md
```diff
@@ -76,36 +76,29 @@ Legend (terms and table are based on Pragmatic Marketing recommendations):
...
+|          Help          |                                                                                                After the user enters JupyterLab he will see the VDK drop down menu where he can find the "Help" option. There he will be able to check for more information about how to use the VDK extension and where to find specific buttons and other UI elements                                                                                                |         1          |   mid    |
```
ivakoleva, line 86: The users that already have jupyterlab started, could they add vdk?

specs/vep-994-jupyter-notebook-integration/README.md
```diff
@@ -76,36 +76,29 @@ Legend (terms and table are based on Pragmatic Marketing recommendations):
...
+|         Log in         |                                                                                                          After the user enters Jupyter lab he/she will see the VDK drop down menu where he can find the "Log in" option. It is for authentication against Control service. A pop out with the corresponding login options will be introduced.                                                                                                          |         1          |   mid    |
```
ivakoleva, line 87: What's the situation with multiple jupyter instances, is the user session per instance or shared?

ivakoleva (approved):

specs/vep-994-jupyter-notebook-integration/README.md
duyguHsnHsn: The idea is good but as I said in the commit message it is rather a small enhancement(for example not to have to include "import IJobInput"). This is not the main goal of the integration and can be added later on as an enhancement.

specs/vep-994-jupyter-notebook-integration/README.md
duyguHsnHsn: User session per instance, I think they cannot be shared in the default version but there might be some extensions that allow it to be shared.

specs/vep-994-jupyter-notebook-integration/README.md
duyguHsnHsn: Starting Jupyter from VDK context would have some benefits over standard "run Jupyter with installed extension". I don't think starting VDK from Jupyter would have benefits, I amnion even sure whether it is possible.

tozka:
Please avoid generic titles like "changes on README.md" - I mean every single change on readme could have this title. Let's be explicit, what is the chang? E.g Expand details on extensions design or something similar.

specs/vep-994-jupyter-notebook-integration/README.md
```diff
@@ -112,6 +112,24 @@ Explain how does the system handle API violations.
 
 
 ## Detailed design
+### VDK Notebook plugin
+ This VDK plugin will provide the functionality to run Jobs which will retrieve Job steps from notebook files instead of .py and .sql files. This plugin can be used alone without the JupyterLab extension.
+
+### VDK JupyterLab extension
+ This extension will be both front-end and server side extension for JupyterLab.
+ The front-end side will be introducing the graphical elements such as menus, buttons, etc. and will be responsible with sending
```
tozka, line 120: I missed saying that since it can be done at follow-up PR when you have better idea about these things

But we need to be very explicit here. 

What are the exact graphical elements that would be introduced. Having mockup picture would be a huge bonus . 

What are the exact request that would be sent and what commands would be supported? 

Also there's a lot of assumption about what the reader knows and not everyone reading this would know how VDK works and what is file step plugin and etc , so you need either describe or better link to existing documentation that explain concepts you are using.

PR #1309, opened 2022-11-10
Revert "build(deps): bump io.freefair.lombok from 6.1.0-m3 to 6.5.1 in /projects/control-service/projects"

tozka:
Reverts vmware/versatile-data-kit#1306 

The pipeline is consistently failing after merging this - https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/690278941

### DIFF
--- a/projects/control-service/projects/settings.gradle
+++ b/projects/control-service/projects/settings.gradle
@@ -4,7 +4,7 @@ pluginManagement {
       id 'io.spring.dependency-management' version '1.0.11.RELEASE'
       id 'com.palantir.docker' version '0.34.0'
       // https://projectlombok.org/setup/gradle
-      id 'io.freefair.lombok' version '6.5.1'
+      id 'io.freefair.lombok' version '6.1.0-m3'
       id 'com.palantir.git-version' version '0.15.0'
       id 'org.unbroken-dome.test-sets' version '4.0.0'
       // https://gitlab.com/barfuin/gradle-jacoco-log/-/blob/master/README.md


projects/control-service/projects/settings.gradle
```diff
@@ -4,7 +4,7 @@ pluginManagement {
       id 'io.spring.dependency-management' version '1.0.11.RELEASE'
       id 'com.palantir.docker' version '0.34.0'
       // https://projectlombok.org/setup/gradle
-      id 'io.freefair.lombok' version '6.5.1'
+      id 'io.freefair.lombok' version '6.1.0-m3'
       id 'com.palantir.git-version' version '0.15.0'
```
ivakoleva, line 8: Why after instead before / with?

projects/control-service/projects/settings.gradle
```diff
@@ -4,7 +4,7 @@ pluginManagement {
       id 'io.spring.dependency-management' version '1.0.11.RELEASE'
       id 'com.palantir.docker' version '0.34.0'
       // https://projectlombok.org/setup/gradle
-      id 'io.freefair.lombok' version '6.5.1'
+      id 'io.freefair.lombok' version '6.1.0-m3'
```
ivakoleva, line 7: Should we troubleshoot the reason? Should we temp workaround with looking for the last operable version?

ivakoleva (approved):

ivakoleva:
The pipeline seems to have took more time than expected; may be recent networking?

tozka:
> The pipeline seems to have took more time than expected; may be recent networking?

Only the [publish image](https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3302140926) job is slow. And I restarted it a few times. I want to revert to see if this is causing the issue. Other jobs are working well enough.

projects/control-service/projects/settings.gradle
tozka: Yes, we should troubleshoot the reason. But first thing is to bring main CICD to stable.

PR #1311, opened 2022-11-10
vdk-heartbeat: wait for execution to start

murphp15:
# Why?
In the heartbeat test we want to make sure a job is created automatically using the cron. However we often don't give k8s enough time to create it before disabling it again. This causes the test to fail with a cryptic message. 

# What?
Check that the job instance is created before disabling the creation of new ones. 

# How has this been tested?
I ran it many times locally and didn't see it fail again. 

Bug fix

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
@@ -57,7 +57,7 @@ class Heartbeat:
 
             self.clean(run_test, job_controller)
             log.info("Heartbeat has finished successfully.")
-        except Exception as e:
+        except:
             log.info("Heartbeat has failed.")
             job_controller.show_job_details()
             job_controller.show_last_job_execution_logs()
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
@@ -45,6 +45,7 @@ class Heartbeat:
             job_controller.enable_deployment()
             job_controller.wait_job_execution_started()
             job_controller.show_job_details()
+
             run_test.execute_test()
             job_controller.show_last_job_execution_logs()
 
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
@@ -43,8 +43,8 @@ class Heartbeat:
             run_test.setup()
 
             job_controller.enable_deployment()
+            job_controller.wait_job_execution_started()
             job_controller.show_job_details()
-
             run_test.execute_test()
             job_controller.show_last_job_execution_logs()
 
@@ -57,7 +57,7 @@ class Heartbeat:
 
             self.clean(run_test, job_controller)
             log.info("Heartbeat has finished successfully.")
-        except:
+        except Exception as e:
             log.info("Heartbeat has failed.")
             job_controller.show_job_details()
             job_controller.show_last_job_execution_logs()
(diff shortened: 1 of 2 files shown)


projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -336,6 +336,31 @@ def start_job_execution(self):
...
+                    "-o" "json",
+                ]
+                + self.__get_rest_api_url_arg()
+            )
+            execution_list = json.loads(response)
+
+            if execution_list:
+                return
+            time.sleep(10)
+        raise Exception("Job never started")
```
gageorgiev, line 362: Isn't there a better exception class to raise here? Something native to vdk perhaps.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -336,6 +336,31 @@ def start_job_execution(self):
         ), "Failed to start data job execution."
         log.info("Execution started successfully.")
 
+    @LogDecorator(log)
+    def wait_job_execution_started(self):
+        log.info("Checking if data job execution is still running.")
```
gageorgiev, line 341: Maybe move this to debug instead of info.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
gageorgiev: also maybe include some identifying information from the config object in the exception message, it looks a bit bare to me right now

gageorgiev (approved):
looks good overall, I have some small comments which you may want to take under consideration

gageorgiev:
Also correct your PR title.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
murphp15: what exception would you suggest? 
I added job name. 
Let me know if you think it looks good enough?

gageorgiev: Looking through the vdk-heartbeat codebase it seems no vdk-specific exception classes are used. In that case it's probably fine to keep it as Exception.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -336,6 +336,31 @@ def start_job_execution(self):
         ), "Failed to start data job execution."
         log.info("Execution started successfully.")
 
+    @LogDecorator(log)
+    def wait_job_execution_started(self):
+        log.debug("Waiting for job to create")
```
gageorgiev, line 341:
```suggestion
        log.debug("Waiting for job {self.config.job_name} to start.")
```

gageorgiev:
Looks good.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
```diff
@@ -43,6 +43,7 @@ def run(self):
             run_test.setup()
 
             job_controller.enable_deployment()
+            job_controller.wait_job_execution_started()
             job_controller.show_job_details()
 
             run_test.execute_test()
```
ivakoleva, line 49: The reason is, `execute_test` was waiting with configurable max time for the result properties to appear, to conclude the test was successful. There is no need to wait for execution to start in particular -> if results are available, execution had started and completed. 
This change actually introduces RUN_TEST_TIMEOUT_SECONDS * 2

ivakoleva (changes requested):

projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
murphp15: The run test instance I am using is EmptyRunTest and that definitely des not do that.
I also looked at a few other instances and they don't seem to do it either. 

Can you point me to the line of code you think is always running?

PR #1312, opened 2022-11-11
vdk-logging-format: Join JSON and LTSV logging plugins into one

gageorgiev:
This PR joins the vdk-logging-json and vdk-logging-ltsv
plugins into one since they share much of their parts.

Currently, the vdk-logging-json plugin will format logs
to JSON as long as it is installed. This is suboptimal,
as it would be better for this to be configurable in the
case where a user might have multiple log formatting
plugins installed. 

Additionally, one detail which should be mentioned
is that the initial VDK logs are printed before the
config is initialized. This means that to format these
logs to JSON, we check for the `VDK_LOGGING_FORMAT`
env variable instead of accessing the configuration
through the config object. In the case where the
logging format is set through a configuration plugin
instead of an environment variable, the first few
logs will not be formatted correctly. I do not believe
there is a good solution to this problem which wouldn't
require large architectural changes to VDK.

Testing done: tested all cases locally


projects/vdk-plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
```diff
@@ -46,25 +50,38 @@ def format(self, record):
...
-        )
+    if context.core_context.configuration.get_value(LOGGING_FORMAT) == "JSON":
+        attempt_id = context.core_context.state.get(CommonStoreKeys.ATTEMPT_ID)
+        op_id = context.core_context.state.get(CommonStoreKeys.OP_ID)
+        job_name = context.name
+
+        for handler in logging.getLogger().handlers:
+            handler.setFormatter(
+                EcsJsonFormatter(job_name=job_name, attempt_id=attempt_id, op_id=op_id)
+            )
```
tozka, line 87: Please extract it as method https://refactoring.guru/extract-method which can be used above as well.

projects/vdk-plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
```diff
@@ -46,25 +50,38 @@ def format(self, record):
         return json.dumps(result)
 
 
+@hookimpl
+def vdk_configure(config_builder: ConfigurationBuilder):
+    config_builder.add(
+        key=LOGGING_FORMAT,
+        default_value="TEXT",
+        description="The format in which to structure VDK logs.",
```
tozka, line 58: What are the possible values ?  Basically it's a feature flag (disable /enable json formatting). 


Outside of scope of this PR: 

Also, it looks strange for vdk-logging-json plugin to have a format. I wonder if we should have one plugin - logging-format with options ??

tozka (approved):
Looks good to me assuming you address my comments . 
By address I mean 
* Either you agree and complete
* Or you answer that you do not think it's good idea. 

Just do not ingore them :)

projects/vdk-plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
gageorgiev: done

projects/vdk-plugins/vdk-logging-json/src/vdk/plugin/logging_json/logging_json.py
gageorgiev: created one plugin for both formats, deleted other plugins

tozka (commented):
The removal of the plugins needs happens in separate PR. 

At this PR you can update their README with information that they are deprecated and people should use vdk-logging-format with LOGGING_FORMAT configuration set ... 

This would make new release and update the doc at https://pypi.org/project/vdk-logging-json so if people stumble on this , would know.

gageorgiev:
> The removal of the plugins needs happens in separate PR.
> 
> At this PR you can update their README with information that they are deprecated and people should use vdk-logging-format with LOGGING_FORMAT configuration set ...
> 
> This would make new release and update the doc at https://pypi.org/project/vdk-logging-json so if people stumble on this , would know.

gotcha, thank you

projects/vdk-plugins/vdk-logging-format/README.md
```diff
@@ -0,0 +1,68 @@
+This plugins allows for the configuration of the format of VDK logs.
+
+# LTSV
```
tozka, line 3: You should add a few owrlds for "TEXT" . It's a possible value (teh default one) .

gageorgiev: done

murphp15 (approved):
Looks good

PR #1314, opened 2022-11-11
vdk-postgres: add ingest plugin

tozka:
When we were preparing for DSC conference we realized that we are missing simple ingest plugin.

This is really copy pasting the greenplum ingestion. where I changed "greenplum" to "postgres"

I will add this now since we need it . But I think it's time to refactor the db plugins - https://refactoring.guru/refactoring/when

A lot of the code is basically the same with names changed. ... So I will open a PR for that.


victor-partico (approved):

mrMoZ1 (approved):
I'm ok to merge if we address the codacy issues first.

PR #1317, opened 2022-11-11
vdk-jupyter: introduce JupterLab extension

duyguHsnHsn:
What: 
 Generated a Jupyterlab extension package project with frontend (in TypeScript) and backend (in Python) parts using a cookiecutter [template](https://github.com/jupyterlab/extension-cookiecutter-ts).  You can see more in the detailed design section of the [VEP-994](https://github.com/vmware/versatile-data-kit/tree/main/specs/vep-994-jupyter-notebook-integration). Currently the extension is not integrated with VDK. It only loads a log that it is running when the jupyter lab is activated. 
Why: 
 We chose to use JupyterLab and it's extension because it is the next-generation user interface for Project Jupyter offering all the familiar building blocks of the classic Jupyter Notebook  in a flexible and powerful user interface.  JupyterLab extensions can customize or enhance any part of JupyterLab which makes them useful for the VDK Jupyter integration.

PS: you can see the instructions on how to run it locally on README.md. Since it is not a published package yet, you should do "pip install . " in the directory.


projects/vdk-plugins/vdk-jupyter/.github/workflows/binder-on-pr.yml
```diff
@@ -0,0 +1,30 @@
+# Reference https://mybinder.readthedocs.io/en/latest/howto/gh-actions-badges.html
```
tozka, line 1: What are those github actions ?

tozka:
Please make sure the name of the plugin is specific and clear. vdk-jupyterlab-extension seems more clear to me. That's what you've named it also in the VEP.

projects/vdk-plugins/vdk-jupyter/.github/workflows/binder-on-pr.yml
duyguHsnHsn: They were automatically generated by the template and should be removed or changed in the next PR.

murphp15:
I think you need to update the readme to mention that installing npm is a prerequisite.

murphp15:
when I run
```
pip install -e .
```
I get this error
```
ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/Users/paulm2/IdeaProjects/versatile-data-kit/venv/share/jupyter/labextensions/vdk-jupyter/build_log.json'
```

Have you seen this error before?

murphp15:
I can coming to this from the complete outside. 
But how do I open the UI?

PR #1323, opened 2022-11-14
vdk-control-service: set registry name correctly.

murphp15:
### Why
the docker secret is being saved to the wrong path. Previously it was using the path of the entire image 

### What
Now it is correctly pointing to the registry

### How was this tested
printed the chart locally and manually added that secret in the env.


gageorgiev (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl
```diff
@@ -241,5 +241,5 @@ Image Pull Secret in json format
 {{- end }}
 
 {{- define "pipelinesControlServicePullSecretJson" }}
-    {{ include "buildImagePullSecretJson" (list (include "pipelines-control-service.image" .) .Values.image.registryUsernameReadOnly .Values.image.registryPasswordReadOnly) }}
+    {{ include "buildImagePullSecretJson" (list .Values.image.registry .Values.image.registryUsernameReadOnly .Values.image.registryPasswordReadOnly) }}
```
ivakoleva, line 244: isn't that declared in helpers https://github.com/vmware/versatile-data-kit/blob/1dbd84688f07844ee51189e831f6cbd314b2b5ab/projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl#L72

murphp15: Im not using that one anymore.

murphp15: but even if it is why is that relevant?

ivakoleva: I saw you are deleting it. Why are you deleting it? It still seems defined in `_helpers.tpl` after your change? 
Is that change needed, isn't the `list (include "pipelines-control-service.image" .)` actually enlisting all the `pipelines-control-service.image` contained sub values, that already include `.Values.image.registry` as well?
```
{{/*
Return the proper Control Service image name
*/}}
{{- define "pipelines-control-service.image" -}}
{{- $globalRegistryOverrideName := .Values.image.globalRegistryOverride -}}
{{- $registryName := .Values.image.registry -}}
{{- $repositoryName := .Values.image.repository -}}
{{- $tag := .Values.image.tag | toString -}}
{{/*
```

mivanov1988 (approved):

ivakoleva (approved):

PR #1327, opened 2022-11-15
vdk-control-cli: fix cicd

mrMoZ1:
what: Added py dependency in requirements.txt

why: Pipeline was failing with missing module. E.g. https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3323662821

testing: cicd pipeline


dakodakov (approved):

doks5 (approved):

doks5:
Just a note here: According to the project's documentation, `py` is not supposed to be used in other projects, and alternatives are recommended instead, https://pypi.org/project/py/#description

PR #1329, opened 2022-11-15
control-service: Expose LOGGING_FORMAT through helm chart

gageorgiev:
Currently, the logging format of the control service as set by the LOGGING_FORMAT variable cannot be set through the helm chart. This PR corrects that by
exposing it.


tozka:
Why do we need it exposed?

gageorgiev:
> Why do we need it exposed?

To configure VDP CS logs to json, so they can be parsed in Log Insight Cloud.

projects/control-service/projects/pipelines_control_service/src/main/resources/application-prod.properties
```diff
@@ -131,4 +131,4 @@ datajobs.monitoring.sync.initial.delay=${DEFAULT_MONITORING_INITIAL_DELAY_SYNC_I
 #datajobs.deployment.builder.extraArgs=${DATAJOBS_DEPLOYMENT_BUILDER_EXTRAARGS}
 
 # Enables JSON-formatted logging for all logs emitted by the Control Service.
-# LOGGING_FORMAT=JSON
+LOGGING_FORMAT=${CONTROL_SERVICE_LOGGING_FORMAT:""}
```
tozka, line 134: pipelinesControlService is redundant. For other control service options, we do not set a prefix. As it's a pipelines control service chart it's implied. For example, we have "notificationOwnerName" and not "controlServiceOwnerName" . "loggingFormat" or "serviceLoggingFormat" is enough

tozka (approved):

projects/control-service/projects/pipelines_control_service/src/main/resources/application-prod.properties
gageorgiev: gotcha, I'll change it

murphp15 (approved):

PR #1331, opened 2022-11-15
control-service: set registry name correctly.

murphp15:
### Why
The format of the dockerconfigjson is wrong. 
Previously the auth key would look like  `registry/repo` e.g `registry.gitlab.com/project.taurus/data-jobs-stg`.
With all the repositories I tried to authenticate against it should only be the registry name e.g `registry.gitlab.com/project.taurus`.
I reached out the @mivanov1988  privately in slack to check with him that this new way would also work with AWS ECR.

### What
This fixes it

### How has this been tested
Printed out the template and looks good. 


murphp15:
@mivanov1988 Can you please look at this PR.
For me this change allows it to pull from the private jfrog repos. 
However is there some repos you were working with that I might need to test this against to make sure I'm not breaking anything. 

I explain the details on the difference in this comment: https://github.com/vmware/versatile-data-kit/pull/694/files#r1023916365

mivanov1988:
Have you tested it against AWS ECR?

tozka (approved):
LGTM

tozka:
> Have you tested it against AWS ECR?

I think this is fixing a bug - the macro was not working because erroneously it was picking non existing key (repository) from values.yaml (probably we've renamed it in values.yaml and forgot to update the macro). 

@murphp15 Please describe slightly more detailed the problem you are solving in each change/PR. It's not good to make reviewers guess . See https://github.com/vmware/versatile-data-kit/wiki/How-to-prepare-a-new-PR#write-a-good-commit-message

murphp15:
@tozka @mivanov1988  how do I test against aws ECR? 

@tozka  I have added details to the description.

tozka:
> @tozka @mivanov1988 how do I test against aws ECR?
>  
If I understand correctly difference between two is 

the new way the login is against '012285273210.dkr.ecr.us-west-2.amazonaws.com' (the registry)

before it was against '012285273210.dkr.ecr.us-west-2.amazonaws.com/repo-vdk-name' (the repository)

and both work 

```
aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin 012285273210.dkr.ecr.us-west-2.amazonaws.com
```

To test it you need AWS account and ECR registry created. I have those - and I executed above command and it works.

mivanov1988 (approved):

mivanov1988:
LGTM

tozka:
> Given the test Antoni has done do you think it is safe to merge and assume it works with ECR otherwise I need an aws account created.

I don't think we need special tests for AWS really. The way you login and pull or push images from any registry is the same. 

There are some differences in how repository is created across different registries but this PR doesn't really do anything related to repository creation. 

But I did tested it just in case using ECR (I've been wrong before). And that's enough. Let's merge it.

PR #1336, opened 2022-11-22
helm chart for private builder

murphp15:
### Why
We need to be able to pull builders from private repos

### What
add variables to helm chart

### How was this tested
Pipeline linting tests and also printed the charts out an inspected manually


mivanov1988 (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -248,6 +248,11 @@ spec:
             - name: DATAJOBS_READ_ONLY_ROOT_FILESYSTEM
               value: "{{ .Values.dataJob.readOnlyRootFilesystem }}"
             {{- end }}
+            {{- if .Values.deploymentBuilderImage.password }}
```
tozka, line 251: The if condition here and in the secret_pull_builder_img.yaml are different (username vs password). 

That seems like the possibility of a unexpected behaviour e.g somewhere where password is empty .

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -26,6 +26,8 @@ deploymentBuilderImage:
   registry: registry.hub.docker.com/versatiledatakit
   repository: job-builder
   tag: "1.2.3"
+  username:
```
tozka, line 29: nit: being a bit more obvious in the same is not a bad thing. WHy not use registryUsernameReadOnly as above ?

murphp15: I actually think it is cleaner this way. 
They know it is builder registry because of the outer domain context. 
Let me know what you think?

tozka: Ok.

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
murphp15: Resolved

tozka: If I set password but do not set user. then the secret won't be created but this will be created and the deployment would fail ? a

tozka (approved):

PR #1337, opened 2022-11-22
pipelines-control-service: support private base image

murphp15:
### Why
We should be able to pull base images from a private repo
### What
pass the auth into the builder image
### How has this been tested

I tested it manually but this is difficult to test as described in https://github.com/vmware/versatile-data-kit/issues/1339


projects/control-service/projects/pipelines_control_service/src/integration-test/resources/application-private-builder.properties
```diff
@@ -1,3 +1,6 @@
 datajobs.builder.registrySecret=integration-test-docker-pull-secret
 datajobs.builder.registrySecret.content.testOnly=${BUILDER_TEST_REGISTRY_SECRET}
 datajobs.builder.image=${DOCKER_REGISTRY_URL}/versatiledatakit/job-builder-prerelease:1.2.3
+datajobs.deployment.dataJobBaseImage=ghcr.io/murphp15/versatiledatakit/data-job-base-python-3.7:latest
```
tozka, line 4: We have a versatile-data-kit-dev account . Why not use it?

murphp15: because that is used to store the built image. 
I need the base image to come from a different repo to confirm the auth is working correctly. 
The base image and the save image being in the same private registry is actually supported at the moment. 



I actually just submitted this in the last commit to test it. 
I was going to clean this up before merging. 
I will move the PR back into draft .

murphp15: I will address this before merging.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -68,6 +72,37 @@ public class JobImageBuilder {
...
+
+  public String getAuth() throws JsonProcessingException {
+    if (org.springframework.util.StringUtils.hasLength(dataJobBaseImageRegistryPassword)) {
+      if (!deploymentDataJobBaseImage.contains("/")) {
+        throw new IllegalArgumentException(
+            "Base image registry secret is set but the base image is not being pulled from a"
+                + " private registry.");
+      }
+      return ","
+          + objectMapper.writeValueAsString(
```
tozka, line 89: That is the way kaniko accepts authentication, right? It's not standard format ? Is it?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -68,6 +72,37 @@ public class JobImageBuilder {
...
+              Map.of(
+                  deploymentDataJobBaseImage.substring(
+                      0, deploymentDataJobBaseImage.lastIndexOf('/')),
```
tozka, line 92: Using substring to parse out something is a bit of a code smell for me. That means we require certain contract on the value which may not be kept.

projects/control-service/projects/job-builder/build_image.sh
```diff
@@ -14,6 +14,7 @@ git_repository=$7
 registry_type=$8
 registry_username=$9
 registry_password=${10}
+extra_auth=${11}
```
tozka, line 17: We have multiple job-builder images. They all 3 are currently based on kaniko but kaniko is not necessary the only build system. I feel like the contact between what control service passes and the job-builder image accept should not rely on Control Service knowing the implementation of the job-builder image.

PR #1338, opened 2022-11-23
vdk-jupyter: introduce JupterLab extension

duyguHsnHsn:
What:
Generated a Jupyterlab extension package project with frontend (in TypeScript) and backend (in Python) parts using a cookiecutter [template](https://github.com/jupyterlab/extension-cookiecutter-ts). You can see more in the detailed design section of the [VEP-994](https://github.com/vmware/versatile-data-kit/tree/main/specs/vep-994-jupyter-notebook-integration). Currently the extension is not integrated with VDK. It only loads a log that it is running when the JupyterLab is activated.
Why:
We chose to use JupyterLab and it's extension because it is the next-generation user interface for Project Jupyter offering all the familiar building blocks of the classic Jupyter Notebook in a flexible and powerful user interface. JupyterLab extensions can customise or enhance any part of JupyterLab which makes them useful for the VDK Jupyter integration.

PS: Since it is not a published package yet, you should ignore the instruction in README.md and: 
 -install jupyterlab, nodejs, jupyter-packagin in your environment 
 -run
```bash
# Install the package
pip install -ve path-to-vdk-jupyterlab-extension
# Run JupyterLab 
jupyter lab
```


tozka (approved):
I have not looked at it carefully since I assume most of it is auto-generated and only changes are there to make it compile. Any significant changes would be in different PR.

That's a lot of files. If you know for sure we do not need some, remove it. E.g I don't think we'd need RELEASE.md ...

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2022, My Name
```
doks5, line 3: Shouldn't the license file be modified to fit the overall vdk project, given that the extension essentially becomes public the moment it is merged?
cc: @tozka

doks5 (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/LICENSE
tozka: Yes, please remove the license file as well. As I mentioned before. Remove unnecessary files before merging.

PR #1343, opened 2022-11-24
contol-service: test cleanup with the goal of making tests easier to run locally

murphp15:
### Why
1. remove unused method
2. Remove logic for creating dynamic  namespaces (it hasn't worked in a long time and it confuses the code significantly). 
This actually isn't supported at the moment and this code I deleted gives the impression it is. 
It is not supported because DataJobDeploymentExtension expects the namespace to be set already before running its @beforeEach section.  It will run before the beforeEach of test class. 

### How has this been tested
Integration tests.


projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -99,7 +99,7 @@ logging.level.io.swagger.models.parameters.AbstractSerializableParameter=ERROR
 
 datajobs.control.k8s.namespace=
 datajobs.control.k8s.kubeconfig=${HOME}/.kube/config
-datajobs.control.k8s.k8sSupportsV1CronJob=false
+datajobs.control.k8s.k8sSupportsV1CronJob=true
```
gageorgiev, line 102: Is this explicitly set to false in SuperCollider? Otherwise it might cause some issues.

murphp15: Changed, default in tests only now.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -198,6 +198,7 @@ private enum ContainerResourceType {
    */
   protected KubernetesService(
       String namespace, String kubeconfig, boolean k8sSupportsV1CronJob, Logger log) {
+    StringUtils.isNotBlank(namespace);
```
tozka, line 201: That's no-op?

tozka (approved):

tozka:
Thanks for doing this! Cleaning up things around especially unneeded code is pretty important to keep things healthy

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
murphp15: fixed

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -198,6 +199,7 @@ private enum ContainerResourceType {
    */
   protected KubernetesService(
       String namespace, String kubeconfig, boolean k8sSupportsV1CronJob, Logger log) {
+    assertTrue(StringUtils.isNotBlank(namespace));
```
gageorgiev, line 202: Maybe this could print an adequate error message is it is blank?

gageorgiev (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
murphp15: Done.

Issue #1357, opened 2022-11-28
Document how to create custom job builders


murphp15:
**What is the feature request? What problem does it solve?**
At the moment customers can create custom job builders but we don't have doc's on the details of creating these.

**Suggested solution**
Make sure that it describes how default secrets are loaded from the k8s secrets container called `builder-secrets`

sabadzhiev:
Triaged. We will keep it, as being a Documentation enhancement that we would like to address.

PR #1358, opened 2022-11-28
control-service: builder images load secrets from k8s

murphp15:
###  Why
I wrote this PR with the intent of being able to pass extra docker creds to to the image builder so that we would be able to pull a private base image from a docker repo. 
The builder now contains login details for 2 registries. The registry it pushes to and the registry it pulls base image from.

However this PR also introduces the concept of builder secrets being sourced from k8s which leads to much cleaner separation between the control plane and the builder image. I think this will lead to much better support for custom build images in the future.

I added new functionality and didn't document it but have created this ticket to follow up on it. 

https://github.com/vmware/versatile-data-kit/issues/1357

### How was this been tested
tested extensively locally. 
I added a test to check that the base image can be pulled from a private repo but in fact it is pulling from the same repo as it is pushing to and so it is could let bugs through related to the auth not working for the base image repo.

### DIFF
--- a/projects/control-service/projects/job-builder/build_image.sh
+++ b/projects/control-service/projects/job-builder/build_image.sh
@@ -52,6 +52,7 @@ cat > /kaniko/.docker/config.json <<- EOM
       "password":"$registry_password",
       "auth": "$auth"
     }
+  $extra_auth
   }
 }
 EOM
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/resources/application-private-builder.properties
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/resources/application-private-builder.properties
@@ -1,3 +1,4 @@
 datajobs.builder.registrySecret=integration-test-docker-pull-secret
 datajobs.builder.registrySecret.content.testOnly=${BUILDER_TEST_REGISTRY_SECRET}
 datajobs.builder.image=${DOCKER_REGISTRY_URL}/versatiledatakit/job-builder:1.2.3
+datajobs.deployment.dataJobBaseImage=ghcr.io/versatile-data-kit-dev/dp/versatiledatakit/data-job-base-python-3.7:latest
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -2202,6 +2202,7 @@ public abstract class KubernetesService implements InitializingBean {
                     .withRequests(resources(request))
                     .withLimits(resources(limit))
                     .build())
+                .withEnvFrom(new V1EnvFromSource().secretRef(new V1SecretEnvSource().name("builder-secrets").optional(true)))
             .withEnv(
                 envs.entrySet().stream()
                     .map(KubernetesService::envVar)
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -2202,7 +2202,9 @@ public abstract class KubernetesService implements InitializingBean {
                     .withRequests(resources(request))
                     .withLimits(resources(limit))
                     .build())
-                .withEnvFrom(new V1EnvFromSource().secretRef(new V1SecretEnvSource().name("builder-secrets").optional(true)))
+            .withEnvFrom(
+                new V1EnvFromSource()
+                    .secretRef(new V1SecretEnvSource().name("builder-secrets").optional(true)))
             .withEnv(
                 envs.entrySet().stream()
                     .map(KubernetesService::envVar)


projects/control-service/projects/job-builder/build_image.sh
```diff
@@ -52,6 +52,7 @@ cat > /kaniko/.docker/config.json <<- EOM
       "password":"$registry_password",
       "auth": "$auth"
     }
+  $extra_auth
```
tozka, line 55: Is this coming from env variable? 

I think it's good idea to document the env variable at the top of the script. 


It can be 
```
# descritpion
# detailed description 
EXTRA_AUTH=${EXTRA_AUTH:-""}
```
is something I often. This way reader of the script knows what is the expected input.

projects/control-service/projects/pipelines_control_service/src/integration-test/resources/application-private-builder.properties
```diff
@@ -1,3 +1,4 @@
 datajobs.builder.registrySecret=integration-test-docker-pull-secret
 datajobs.builder.registrySecret.content.testOnly=${BUILDER_TEST_REGISTRY_SECRET}
 datajobs.builder.image=${DOCKER_REGISTRY_URL}/versatiledatakit/job-builder:1.2.3
+datajobs.deployment.dataJobBaseImage=ghcr.io/versatile-data-kit-dev/dp/versatiledatakit/data-job-base-python-3.7:latest
```
tozka, line 4: Maybe we can create a second versatile-data-it-dev-second user ? It needs to be documented and  credentials need be in both Lastpass and Gitlab CI.

tozka:
Currently our main user facing documentation of the Control Service configuration is primary though [helm chart values.yaml](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml) so this would need to be documented there as well. 

but we do need more user-friendly one and examples /tutorials - so the ticket that you opened can be about that.

tozka (approved):
Looks goot to me. 

Do we need to port/refactor some of the other configuration options to come from k8s secret file ?

murphp15:
> Do we need to port/refactor some of the other configuration options to come from k8s secret file ?

I am not going to do this now, but in the future we can cleanup those values.

projects/control-service/projects/pipelines_control_service/src/integration-test/resources/application-private-builder.properties
murphp15: I think this is a lot of admin overhead for a single test. 
If we come across a second use case where we need to test multiple private repos I will add it in and test this also. 
Does that sound ok with you?

projects/control-service/projects/job-builder/build_image.sh
```diff
@@ -15,6 +15,10 @@ registry_type=$8
 registry_username=$9
 registry_password=${10}
 
+# Within this property docker config should be included to connect to the registry used to pull the image from.
+# it should be prefixed with a comma
```
ivakoleva, line 19: that is unusual, wouldn't itbe error-prone...

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -2202,6 +2202,9 @@ public static V1Container container(
                     .withRequests(resources(request))
                     .withLimits(resources(limit))
                     .build())
+            .withEnvFrom(
+                new V1EnvFromSource()
+                    .secretRef(new V1SecretEnvSource().name("builder-secrets").optional(true)))
```
ivakoleva, line 2207: losing track of what PR depends on what PR, and how are those linked to a ticket; I hope you are linking them somehow

projects/control-service/projects/base/build.gradle
```diff
@@ -78,6 +78,7 @@ dependencies {
    testImplementation versions.'org.awaitility:awaitility'
    testImplementation versions.'com.github.tomakehurst:wiremock'
    testImplementation versions.'org.mockito:mockito-core'
+   testImplementation versions.'net.bytebuddy:byte-buddy'
```
ivakoleva, line 81: don't see an import change; where do you use the newly introduced library?

ivakoleva (approved):

projects/control-service/projects/job-builder/build_image.sh
ivakoleva: me too; is the comment already resolved ?

projects/control-service/projects/job-builder/build_image.sh
murphp15: Yes this is defo error prone. 
But it is providing way more flexibility to customers that I think it is worth it.

projects/control-service/projects/base/build.gradle
murphp15: In fact this fixes an issue that was introduced into master yesterday where the latest version of mockito causes tests to fail because they need this jar.

projects/control-service/projects/job-builder/build_image.sh
murphp15: yeah

Issue #1333, opened 2022-11-16
The builder job should support pulling the base image from private repositories


murphp15:
**What is the feature request? What problem does it solve?**
At the moment the builder job does not support pulling a base image from a private repo.
This should be configurable

**Suggested solution**
Add configuration to the build_image.sh file.

PR #1359, opened 2022-11-28
control-service: builder base image in helm

murphp15:
### Why
Within this PR we add support in a builder image for using extra info from k8s through secrets in the namespace. 
In this PR we create the actual secret in the namespace if the info required to create it is present.

### How was this tested. 
Basic linting and looking at a printed out chart. 
closes https://github.com/vmware/versatile-data-kit/issues/1333 


tozka (approved):

ivakoleva (approved):
You can check with the `git-commit-template.txt` about describing the change

PR #1363, opened 2022-11-29
Image describing Insert Template Documentation

dimirapetrova:
Based on Radostina's feedback about insert template I would like to upload this image which shows how the INSERT template strategy works.


tozka:
What is the "source" of the image? I usually recommend uploading or linking (if more appropriate) the source of the image so it can be edited later.

dimirapetrova:
The source of the image is done personally by Radostina Kasova and inspired by https://www.datacoral.com/blog/data-connectors-101/

tozka:
I'd prefer we do no put them here. This folder is more for very frequently reusable images (like the vdk logo). 

You can  upload it by dragging and dropping it in the wiki for example. It would upload to images.githubusercontent. 


like this , here is the link: https://user-images.githubusercontent.com/2536458/204569717-e1fbe473-c0cb-45cd-933b-98500a351dd7.png

dimirapetrova:
Thank you. I already added it on the overall page where all templates are described: https://github.com/vmware/versatile-data-kit/wiki/SQL-Data-Processing-templates-examples

I have also created another request for adding it in the detailed INSERT template documentation itself.

tozka:
Ok. I am going to close this PR since we will be using the uploaded to github.usercontent image.

PR #1370, opened 2022-11-29
control-service: use correct secret type 

murphp15:
### Why
Use the opaque data type. The previous data type was copy paste issues. 

### What
Make the correction and also use stringdata instead of data so we don't need to base 64 encode it. 

### How has this been tested
Visual investigation and linting


projects/control-service/projects/helm_charts/pipelines-control-service/templates/secret_pull_base_img.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
   name: builder-secrets
   namespace: {{ .Values.controlK8sNamespace }}
   labels: {{- include "pipelines-control-service.labels" . | nindent 4 }}
-type: kubernetes.io/dockerconfigjson
-data:
+type: Opaque
+stringData:
 {{/* Within this section any set of properties can be set and then used in a job builder image
 at the moment the only one set is extra_auth: Within this property docker config should be included to connect to the registry used to pull the image from. */}}
-  extra_auth: {{- printf ", \"%s\": {\"auth\": \"%s\"}" .Values.deploymentDataJobBaseImage.registry (printf "%s:%s" .Values.deploymentDataJobBaseImage.username .Values.deploymentDataJobBaseImage.password | b64enc) | b64enc }}
```
ivakoleva, line 12: What was the usability problem with data/dockerconfigjson exactly?

ivakoleva (approved):

gageorgiev (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/secret_pull_base_img.yaml
murphp15: when you use type dockerconfigjson it expects certain fields to be present and also it expects certain data in those fields. 
We don't want it to do any of that validation because what we are passing in is not of that format

tozka (approved):

PR #1371, opened 2022-11-30
control-service: run ci on gradle version change 

murphp15:
### Why
At the moment main is broken because we have not been running the build if the versions file is changed. 
Obviously this is a bit of a disaster. 

I also excluded changes to helm charts from running a lot of useless steps/ 

### What
Listen to changes on the version file and if there are changes then run integration tests and unit tests. 

### How has this been tested
Integration and unit tests.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/WebHookServerMockExtension.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/WebHookServerMockExtension.java
@@ -16,7 +16,7 @@ import java.util.concurrent.locks.ReentrantLock;
 import io.netty.handler.codec.http.HttpMethod;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
-import org.mockserver.client.server.MockServerClient;
+import org.mockserver.client.MockServerClient;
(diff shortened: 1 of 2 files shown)


gageorgiev (approved):

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -4,14 +4,30 @@
...
   - projects/control-service/projects/base/**/*
   - projects/control-service/projects/model/**/*
   - projects/control-service/projects/pipelines_control_service/**/*
   - projects/control-service/projects/settings.gradle
+  - projects/control-service/projects/versions-of-external-dependencies.gradle
   - projects/control-service/projects/helm_charts/pipelines-control-service/**/*
 
+# This is a subset of the &control_service_change_locations which listens on changes to all code
+# Changes to this group should trigger runs of runs of unit and integration tests.
+.control_service_code_changes: &control_service_code_change_locations
```
tozka, line 23: Our helm chart deployment uses the current commit as tag for the image of the control service. See [here](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/cicd/deploy-testing-pipelines-service.sh#L16)

murphp15: Yeah that is fine. 
That script is triggered in a section listening on `&control_service_change_locations`.
What do you see as being the issue?

tozka (approved):
I think that would be ok.

Issue #1238, opened 2022-10-13
vdk list -t "team" does not return newly added job


tozka:
**Describe the bug**
I am not sure what is happening . But when querying by team not all jobs are returned.  
It is reproducible by both API and CLI 

**Steps To Reproduce**
1. vdk create -n example-job -t test-team --cloud
2. vdk list -t test-team 
3. Job example-job should is not listed 

 **Expected behavior**
 Job example-job should be listed .... 

**Screenshots**

**Version  (please complete the following information):**
 - Control Service version: 1.5.657062384

 
 
**Additional context**
Add any other context about the problem here. Consider attaching logs if there are any.

gageorgiev:
Integration test relevant to this issue here - https://github.com/vmware/versatile-data-kit/pull/1385 - seems to not be able to reproduce the issue, or the issue is intermittent.

tozka:
Hi @gageorgiev , 

Thanks for looking into this. I tried to reproduce it as well based on the instructions and could not.

If we cannot reproduce it, then let's close it. 

If you have noticed any intermittent failures/issues, please post any findings (including upload details of logs/commands) here so if we re-open it in the future we have some information. Do post what kind of investigation you have done so far.

gageorgiev:
Attempted to reproduce manually by creating a job and listing all jobs, checking that the created job is there. Did this multiple times in a row.
Attempted to reproduce through an integration test which creates 100 jobs are lists them, checking that they all exist.

Either it is a rare issue, or it is no longer a problem. Closing this and closing the IT PR.

tozka:
changed the status to closed as not planned (cannot reproduce)

PR #1385, opened 2022-12-01
control-service: Integration test for listing many jobs

gageorgiev:
This PR introduces an integration test covering the case
where many (in this case 50) jobs are created and then
listed, checking that they all appear in the listing. This is to
check for a bug where occasionally listing jobs won't print
all of them.
Issue https://github.com/vmware/versatile-data-kit/issues/1238

Testing done: introduced integration test


murphp15:
so does this fail sometimes?

gageorgiev:
> so does this fail sometimes?

It should have, but it doesn't seem to have failed this time.

murphp15:
IMO we should keep working on this test till it can reproduce the issue quite often. 
At the moment I think it brings little value.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobListManyJobsIT.java
```diff
@@ -0,0 +1,106 @@
...
+
+  private static final int JOB_COUNT = 50;
+
+  @Test
+  public void testListManyJobs() throws Exception {
+    createDummyJobs();
+
+    mockMvc
+        .perform(
+            get(String.format("/data-jobs/for-team/%s/jobs", TEST_TEAM_NAME))
```
ivakoleva, line 42: could you list the tests that cover this endpoint already?
also, what is the diff between those and these newly added ones, the heavier JOB_COUNT?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobListManyJobsIT.java
```diff
@@ -0,0 +1,106 @@
...
+                        + "\"filter\": ["
+                        + "    {"
+                        + "      \"property\": \"config.team\","
+                        + "      \"pattern\": \""
+                        + TEST_TEAM_NAME
+                        + "\""
+                        + "    }"
+                        + "  ],"
+                        + "\"pageNumber\": 1,"
+                        + "\"pageSize\": "
```
ivakoleva, line 57: does the issue has anything to do with page size? what is the default page size currently, 50? is that why you create 50 jobs, to check the paging on page 2? could you then adjust the page size filter query?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobListManyJobsIT.java
ivakoleva: could you also check the cicd, if those tests already covering the endpoint had failed while listing jobs? if yes, could you apply the pipeline jobs, so we could analyse the common denominator to find the cause?

ivakoleva:
Is there a ticket related to this PR? if yes, you  should link it.

ivakoleva:
> Is there a ticket related to this PR? if yes, you should link it.

Found it https://github.com/vmware/versatile-data-kit/issues/1238
You can ask the issue reporter to show if reproducible, in case you face a problem to do so by yourself. If neither can reproduce it, you should close the ticket/PR

gageorgiev:
Closing this as the target bug cannot be reproduced both manually and through this test. This PR is linked in the closed issue so if this bug appears again in the future, we can reopen this PR.

PR #1387, opened 2022-12-01
control-service: fail fast for tests

murphp15:
### Why
I want to get feedback on this but I think it makes sense. 
The tests take a long time to run and there is normally a queue of jobs waiting to run. 
So I think it would be a lot better working experience if we were getting pass/fails faster. 
I understand the benefit of seeing the results of all tests results but I don't think it is worth the extra wait. 


tozka:
If you look at statistics  for the integration tests
 * How often more than 1 or 2 test fail 
 * How long does it take the test suite when it fails vs with fail-fast ? 
 
 Based on that it can inform us if it's good idea or not.

Issue #1252, opened 2022-10-20
pipeline-control-service is invalid for git commits which is only digits


murphp15:
**Describe the bug**
the pipelines-control-service docker images are tagged with the git commit(https://registry.hub.docker.com/r/versatiledatakit/pipelines-control-service/tags). 

On rare occasions this commit can be entirely digits.
If it is entirely digits helm treats it as a number and formats it in scientific notation before substituting it in a helm chart. 
For example the tag 9505540 would be transformed to 9.50554e+06.

**Steps To Reproduce**
1. within the file .gitlab-ci.yml () replace `git rev-parse --short HEAD` with a number e.g (9505540). 
2. Create a new docker image and helm chart using the pipelines
3. deploy the helm chart to a k8s cluster

**Expected behavior**
The helm chart would deploy without issues.

PR #1393, opened 2022-12-01
control-service: compute the image tag within the script

murphp15:
### Why 
I am seeing an issue where helm charts are being released and they point to a pipelines-control-service image that doesn't exist because they have dropped some numbers from the git hash id.
The bug can be seen here: https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3408315682  
the image it creates doesn’t have the full commit id.  
if you search for image.tag= in the logs you can see that it set the commit to something shorter than the commit id
I also think that adding the quotes when declaring the variable might fix this: [#1252](https://github.com/vmware/versatile-data-kit/issues/1252)
I did test locally but am not able to reproduce either issue locally and I think it is related to the shell version they are using on gitlab.

### What 
Compute the tag within the script

### DIFF
--- a/projects/control-service/cicd/.gitlab-ci.yml
+++ b/projects/control-service/cicd/.gitlab-ci.yml
@@ -301,7 +301,6 @@ control_service_release:
   stage: release
   script:
     - apk --no-cache add bash openssl curl git
-    - export IMAGE_TAG="$(git rev-parse --short HEAD)"
     - export DESIRED_VERSION=v3.6.3 # helm version 3.6.3
     - export CHART_NAME=pipelines-control-service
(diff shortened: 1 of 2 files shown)


tozka:
What makes you think it's a shell issue? 
Shell being so mature and stable seems unlikely to have such a strange behavior even across versions, not for something so basic. Though it happens.
... 

I think  the bug may be in the helm pack plugin 

We use [helm pack plugin](https://github.com/thynquest/helm-pack)  because helm package doesn't support passing values (like image.tag) - https://github.com/helm/helm/issues/3141 

But looking at the helm pack code -https://github.com/thynquest/helm-pack/blob/master/cmd/helmpack/pack.go
It's possible that type inference is lost when passing the arguments.

PR #1403, opened 2022-12-05
vdk-control-cli: allow cli users to explicitly set the user agent tag

murphp15:
### Why
Users are looking to be able to determine where requests originated from when browsing the data.
### How
We are going to let cli users explicitly set the user-agent header. 

### How has this been tested

sent requests to vdk aas and made sure they have the correct user agent in super collider. 
The results of this test can only be validated by VMware employees which isn't ideal for an open source project. https://sql.supercollider.vmware.com/stg/history/index.html?sql=select%20*%20from%20history_staging.taurus_httptrace%20where%20request_headers_user_agent%20like%20%27%25paul%25%27%20limit%2010&ddlOutputType=table&version=1#

Closes: https://github.com/vmware/versatile-data-kit/issues/1404

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
@@ -18,6 +18,8 @@ class VDKConfig:
 
     _op_id = os.environ.get("VDK_OP_ID_OVERRIDE", f"{uuid.uuid4().hex}"[:16])
 
+    _user_agent = os.environ.get("VDK_CONTROL_SERVICE_USER_AGENT", "VDK_CLI")
+
     @property
     def op_id(self) -> str:
         """
@@ -37,6 +39,16 @@ class VDKConfig:
 
         """
         return VDKConfig._op_id
+
+    @property
+    def user_agent(self) -> str:
+        """
+        user agent that is used to make easier troubleshooting.
+
+        This value won't appear in trace logs but will be included in events set to supercollider
+        """
+        return VDKConfig._user_agent
+
 
     @property
     def local_config_folder(self) -> str:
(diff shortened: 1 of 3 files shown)


tozka:
I've created corresponding issue in Github for the work explaining the problem and the solution - #1404

tozka:
For the purposes of this change, you can use a local server with telemetry endpoint like https://webhook.site/ or https://requestbin.com/ 

Thought it is a good point. Our CICD deployment uses vcsa.vmware.com currently for telemetry - https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/cicd/deploy-testing-pipelines-service.sh#L97 
It might be nice to extend vdk-heartbeat to check telemetry as well also

mivanov1988 (approved):

projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
```diff
@@ -18,6 +19,11 @@ class VDKConfig:
 
     _op_id = os.environ.get("VDK_OP_ID_OVERRIDE", f"{uuid.uuid4().hex}"[:16])
 
+    _user_agent = os.environ.get(
+        "VDK_CONTROL_SERVICE_USER_AGENT",
+        f"vdk-control-cli/1.2 ({os.name}; {sys.platform})" + sys.version.split(" ")[0],
```
tozka, line 24: You can get version from projects/vdk-control-cli/src/vdk/internal/control/command_groups/version_group/version.py


But how about this 
```suggestion
        f"vdk-control-cli/{version.__version__} ({sys.platform}; {platform.platform()}; ) Python {platform.python_version()} Node {platform.node()} User {getpass.getuser()}",
```

murphp15: I agree with adding `{version.__version__}`.


I also don't think we want to include the node name.
my node id is paulmxxxxxx.vmware.com'.
Seems like too much info to include in my opinion. 

@tozka Please let me know what you think.

tozka (approved):

PR #1405, opened 2022-12-05
control-service: Reset termination status when job is disabled

doks5:
Currently, when a data job is deleted, we reset its termination status, so as to avoid issues with monitoring services reporting incorrect status. This, however, is not done in case a data job is disabled, even though the job is no longer active and its termination status should not matter.

This change modifies the DeplymentService, so that if a data job is disabled, the termination status is reset.

Testing Done: Added test.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
@@ -207,6 +207,18 @@ public class DataJobMetrics {
   }
 
   /**
+   * Removes the delay notification and termination status gauges
+   * associated with the specific data job. This is done when a data job is
+   * being disabled.
+   *
+   * @param dataJobName The name of the data job for which to clear the gauges.
+   * */
+  public void clearGaugesAtJobDisable(final String dataJobName) {
+    removeNotificationDelayGauge(dataJobName);
+    removeTerminationStatusGauge(dataJobName);
+  }
+
+  /**
    * Removes all gauges associated with data jobs that are not present in the specified iterable.
    *
    * @param dataJobNames The names of the data jobs which will still have gauges.
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
@@ -207,13 +207,12 @@ public class DataJobMetrics {
   }
 
   /**
-   * Removes the delay notification and termination status gauges
-   * associated with the specific data job. This is done when a data job is
-   * being disabled.
+   * Removes the delay notification and termination status gauges associated with the specific data
+   * job. This is done when a data job is being disabled.
    *
    * @param dataJobName The name of the data job for which to clear the gauges.
-   * */
-  public void clearGaugesAtJobDisable(final String dataJobName) {
+   */
+  public void clearTerminationStatusAndDelayNotifGauges(final String dataJobName) {
     removeNotificationDelayGauge(dataJobName);
     removeTerminationStatusGauge(dataJobName);
   }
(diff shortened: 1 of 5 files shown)


projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/monitoring/DataJobMetricsTest.java
```diff
@@ -273,4 +273,40 @@ void testIncrementWatchTaskInvocations() {
         meterRegistry.counter(DataJobMetrics.TAURUS_DATAJOB_WATCH_TASK_INVOCATIONS_COUNTER_NAME);
     Assertions.assertEquals(5.0, counter.count(), 0.001);
   }
+
+  @Test
+  @Order(14)
```
murphp15, line 278: why do you care about the order?

murphp15:
I would prefer that the test is testing the function `patchDeployment` instead of a level below that. 
Because at the moment you aren't really asserting that disabling a job clears the metrics?

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/monitoring/DataJobMetricsTest.java
mivanov1988: Overall the whole test is miswritten and needs to be rewritten. As far as I remember some of the steps have unnecessary dependencies between them.

mivanov1988: So let's keep the order for now.

doks5:
> I would prefer that the test is testing the function `patchDeployment` instead of a level below that. Because at the moment you aren't really asserting that disabling a job clears the metrics?

Hi, @murphp15 
Could you, please, advise where the additional test should go
I searched for a test that checks if the metrics are properly cleared at job deletion, but such does not seem to exist.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
```diff
@@ -206,6 +206,17 @@ public void clearGauges(final String dataJobName) {
     removeTerminationStatusGauge(dataJobName);
   }
 
+  /**
+   * Removes the delay notification and termination status gauges associated with the specific data
+   * job. This is done when a data job is being disabled.
+   *
+   * @param dataJobName The name of the data job for which to clear the gauges.
+   */
+  public void clearGaugesAtJobDisable(final String dataJobName) {
```
tozka, line 215: You should not name a method based on who is using it but based on what it does. We may want to clear the same gauges on deploy Remove as well. 

Which is a follow-up question. Don't we ?

tozka:
Looks in the right direction. I think we need to see what happens when we remove a job deployment as well (I honestly do not know and haven't time to check the code so it's up to you). I am also fine if you decide to handle this case in subsequent PR. 

Don't wait for my approval since I am going on vacation :)

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/monitoring/DataJobMetrics.java
doks5: Agree, the method name didn't make sense. Changed, thanks

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/DeploymentServiceTest.java
```diff
@@ -354,6 +358,32 @@ public void patchDeployment() throws ApiException {
     assertEquals(true, dataJobCaptor.getValue().getEnabled());
   }
 
+  @Test
+  public void patchDeployment_enabled() throws ApiException {
```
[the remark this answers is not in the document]
doks5: Renamed, thanks

murphp15 (approved):

Issue #1397, opened 2022-12-02
Support for Python 3.11 (vdk-snowflake)


tozka:
**What is the feature request? What problem does it solve?**

Python 3.11 is officially released now and latest versions of Operating systems are starting to default to it. 
We should make sure we support it in our python applications

**Suggested solution**
Add tests for python 3.11 for all distributions that test against multiple versions . 
This will happen in multiple PRs and this issue will be closed after all PRs are merged succesfully

ivakoleva:
Should the PR https://github.com/vmware/versatile-data-kit/pull/1395 cover the entire work on this ticket? Because it seems by merging it, the issue will be closed.

tozka:
> Should the PR #1395 cover the entire work on this ticket? Because it seems by merging it, the issue will be closed.

Clarified the description.

tozka:
Consdier updating job-builder as well see https://github.com/vmware/versatile-data-kit/pull/1395#issuecomment-1337241828

ivakoleva:
> Consdier updating job-builder as well see [#1395 (comment)](https://github.com/vmware/versatile-data-kit/pull/1395#issuecomment-1337241828)

I don't quite understand this separation made. If we change the default vdk-core python version, should we change the language level used in the builder  accordingly? If yes, then this should apply to the upper python version supported as well.

ivakoleva:
> What do you mean by changing the default vdk-core python version ? Default where ? In Control Serviec tests ?

That is a funny question, the default Python version should match in all VDK components :) There shouldn't be discrepancies in python versions across components, to minimise risk of discrepancies, unless there is a reason for them to explicitly differ.

> This issue is only about adding official support for 3.11 . And it's done by adding tests everywhere where we already test against multiple versions and changing python_requires where needed.

The issue states (generic) support  for 3.11, or at least not mentioning nor labelling any component in particular. I want to prevent the issue closing with only covering vdk-core and calling it done

tozka:
The only one left is vdk-snowflake. The underlying connector library did not support 3.11 . 
Support seems to have been added a few hours ago now though - https://github.com/snowflakedb/snowflake-connector-python/pull/1349 

So we might be able to ugprade.

PR #1409, opened 2022-12-05
vdk-control-cli and some plugins:  Support for 3.11

tozka:
With the GA release of Python 3.11 a month ago it's time to officially support it.
That means we need to start testing against Python 3.11. See release notes at https://docs.python.org/3/whatsnew/3.11.html

Part of #1397

This PR contains multiple commits - they will be submitted without squash so that each plugin is changed in a separate commit. Each commit updates 2 files - .plugin-ci yaml file and setup.py (or cfg). 
For vdk-heartbeat and airflow-provider-vdk only the setup.py/cfg file is updated because they test with a single version of python. 

Testing Done: this PR pipeline.

### DIFF
--- a/projects/vdk-control-cli/.gitlab-ci.yml
+++ b/projects/vdk-control-cli/.gitlab-ci.yml
@@ -41,6 +41,10 @@ vdk-control-cli-build-with-py310:
   image: "python:3.10"
   extends: .vdk-control-cli-build
 
+vdk-control-cli-build-with-py311:
+    image: "python:3.11"
+    extends: .vdk-control-cli-build
+
 vdk-control-cli-release-acceptance-test:
   stage: pre_release
   before_script:
--- a/projects/vdk-control-cli/setup.cfg
+++ b/projects/vdk-control-cli/setup.cfg
@@ -27,6 +27,7 @@ classifiers =
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.11
 
 [options]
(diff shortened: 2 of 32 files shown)


projects/vdk-control-cli/.gitlab-ci.yml
```diff
@@ -41,6 +41,10 @@ vdk-control-cli-build-with-py310:
   image: "python:3.10"
   extends: .vdk-control-cli-build
 
+vdk-control-cli-build-with-py311:
+    image: "python:3.11"
+    extends: .vdk-control-cli-build
+
 vdk-control-cli-release-acceptance-test:
```
ivakoleva, line 48: some plugin directories are missing .gitlab-ci.yaml file modification.
Is the change actually modifying two files in each plugin dir, `setup.cfg` and `.gitlab-ci.yml`? If yes, you could just say so in the PR description

tozka: Updated

tozka:
> Btw, what do you mean by vdk-heartbeat tests only with one python version? I see two python images (3.9 and 3.10) used in projects/vdk-heartbeat/.gitlab-ci.yml Is there a reason to not test these components with multiple python versions?

[this one](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-heartbeat/.gitlab-ci.yml#L2) is the default image , but it's overridden in the [vdk-heartbeat-test](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-heartbeat/.gitlab-ci.yml#L13) . So it's tested with 3.10 only.

PR #1410, opened 2022-12-05
vdk-core: get_managed_connection should return opened connection

tozka:
Currently job_input.get_managed_connection() return the current managed connection which may not be opened if it's called for a first time. THis may lead to subtle bugs e.g

```
pandas.to_sql(sql, con = job_input.get_managed_connection())
```

will return error if job_input.execute_query has not be run before. But will succeed if it's not . That's clearly error prone and also it's unnecessary to force users to called
job_input.get_managed_connection().connect() explicitly).

Testing Done: functional test

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
@@ -20,6 +20,8 @@ from vdk.internal.builtin_plugins.connection.pep249.interfaces import PEP249Conn
 from vdk.internal.core import errors
 from vdk.internal.core.config import Configuration
 
+log = logging.getLogger(__name__)
+
 
 class ManagedConnectionRouter(IManagedConnectionRegistry):
     """
@@ -97,10 +99,25 @@ class ManagedConnectionRouter(IManagedConnectionRegistry):
             conn.connect()
         return conn
 
+    def __cache_connection(self, dbType: str, conn: ManagedConnectionBase):
+        if dbType in self._connections:
+            # Though not expected, this checks is added to reduce the chance of connection leaks.
+            log.warning(
+                f"There is already cached connection for dbType {dbType}. Replacing it "
+            )
+            prev_conn = self._connections[dbType]
+            try:
+                prev_conn.close()
+            except Exception as e:
+                log.debug(
+                    f"Failed to close cached connection: {e}. Likely it's no longer valid."
+                )
+        self._connections[dbType] = conn
+
     def __create_connection(self, dbtype: str):
         conn = self._connection_builders[dbtype]()
         if isinstance(conn, ManagedConnectionBase):
-            self._connections[dbtype] = conn
+            self.__cache_connection(dbtype, conn)
             if not conn._connection_hook_spec_factory:
                 conn._connection_hook_spec_factory = self._connection_hook_spec_factory
         elif conn is None:
@@ -117,9 +134,11 @@ class ManagedConnectionRouter(IManagedConnectionRegistry):
         else:
             log = logging.getLogger(conn.__class__.__name__)
             conn.close()  # we will let ManagedConnection to open it when needed.
-            self._connections[dbtype] = WrappedConnection(
+            wrapped_conn = WrappedConnection(
                 log,
                 self._connection_builders[dbtype],
                 self._connection_hook_spec_factory,
             )
+            self.__cache_connection(dbtype, wrapped_conn)
+
         return self._connections[dbtype]
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
@@ -24,8 +24,8 @@ from vdk.internal.core.config import Configuration
 class ManagedConnectionRouter(IManagedConnectionRegistry):
     """
     Create ManagedConnection by routing the configured database plugin.
-    Configuration is controlled by DB_DEFAULT_TYPE for defualt connection.
-    Or specfific connection can be specified by open_connection(dbtype)
+    Configuration is controlled by DB_DEFAULT_TYPE for default connection.
+    Or specific connection can be specified by open_connection(dbtype)
     In both cases dbtype must match the string in which the plugin register itself with.
     """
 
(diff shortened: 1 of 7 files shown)


murphp15 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
```diff
@@ -60,8 +61,8 @@ def __init__(
 
     # Connections
 
-    def get_managed_connection(self) -> ManagedConnectionBase:
-        return self.__managed_connection_builder.open_default_connection()
+    def get_managed_connection(self) -> PEP249Connection:
```
ivakoleva, line 64: What is the reason for changing the contract

ivakoleva (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
```diff
@@ -60,8 +61,8 @@ def __init__(
...
+        return self.__managed_connection_builder.open_default_connection().connect()
```
ivakoleva, line 65: Isn't `open_default_connection()` evaluating https://github.com/vmware/versatile-data-kit/blob/5f147364ab8e23a8283964b0500ba526268daff3/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py#L71 that handles the connect?

```
if conn._is_connected():
  return conn
else:
  conn._connect()
  return conn
```

ivakoleva (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
tozka: It was a mistake in the first place. I never intended for get_managed_connection()  to return ManagedConnectionBase.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
tozka: Look at line 86,87 
When we create connection for a first time , we go through those lines. 

But that's a good point. open_connection() works differently depending if it's called once (for same db type) or twice. 
The first time it creates a connection and the second it calls conn._connect() . Which is strange. I feel it's a mistake. I certainly did not intend this on purpose.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
ivakoleva: I may actually remember a discussion

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
ivakoleva: The change doesn't make sense. If a connection is not opened, the place that should be looked into for fixing it, is `open_connection()` method that is intended to do so.

tozka: Yeah , I agree.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input.py
tozka: Possible. But get_managed_connection() is data engineering facing method and ManagedConnectionBase is mostly useful for plugin development to be subclass (though connection hooks should be preferred). Otherwise it's implementation detailed. Both execute_query and get_managed_connection (which ManagedConnectionBase offers) are exposed through IJobInput interface.

ivakoleva: I don't think mixing multiple changes to push a change using the gravity of another change, is optimal.
This should go into a separate PR.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_connection_base.py
```diff
@@ -96,9 +96,12 @@ def method(*args, **kwargs):
     )
     def connect(self) -> PEP249Connection:
         """
+        Checks if we are still properly connected to the database (including by issuing validation query)
+        and re-connect if not.
```
ivakoleva, line 100: Isn't that performed transparently by the driver? The usual `SELECT 1` mechanism to detect timeouts

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_connection_base.py
```diff
@@ -96,9 +96,12 @@ def method(*args, **kwargs):
...
         :return: PEP249 Connection object (managed)
         """
-        if not self._is_db_con_open:
+
+        if not self._is_connected():
```
ivakoleva, line 104: there is no functional change with that modification, since:
```
def _is_connected(self) -> bool:
        if None is self._is_db_con_open:
            return False
        if False is self._is_db_con_open:
            return False
```

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
```diff
@@ -75,25 +75,24 @@ def open_connection(self, dbtype: str) -> ManagedConnectionBase:
...
+        conn = None
         if dbtype in self._connections:
             conn = self._connections[dbtype]
-            if conn._is_connected():
-                return conn
-            else:
-                conn._connect()
-                return conn
         self._log.debug(f"Connection to {dbtype} is missing. Will try to connect")
         if dbtype in self._connection_builders:
```
ivakoleva, line 82: maybe  elseif?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
```diff
@@ -75,25 +75,24 @@ def open_connection(self, dbtype: str) -> ManagedConnectionBase:
...
+            errors.log_and_throw(
+                to_be_fixed_by=errors.ResolvableBy.CONFIG_ERROR,
+                log=self._log,
+                what_happened=f"Provided configuration variable for {DB_DEFAULT_TYPE} has invalid value.",
+                why_it_happened=f"VDK was run with {DB_DEFAULT_TYPE}={dbtype}, however {dbtype} is invalid value for this variable.",
+                consequences=errors.MSG_CONSEQUENCE_DELEGATING_TO_CALLER__LIKELY_EXECUTION_FAILURE,
+                countermeasures=f"Provide either valid value for {DB_DEFAULT_TYPE} or install database plugin that supports this type. "
+                f"Currently possible values are {list(self._connection_builders.keys())}",
+            )
+        conn.connect()
```
ivakoleva, line 94: `conn._connect()` was used priorly,
now by using `connect()` instead then wouldn't you, again, skip the reconnecting part due to:
```
def connect(self) -> PEP249Connection:
        """
        :return: PEP249 Connection object (managed)
        """
        if not self._is_db_con_open:
            db_con = self._connect()
            self._log.debug(f"Established {str(db_con)}")
            self._is_db_con_open = True
            self._db_con = db_con

        return self
```

How is this "reconnect" to return a verified to be opened connection, tested?

ivakoleva (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
tozka: The new functional test covers this.

tozka:
Thanks for the feedback Iva. I will address it next week at some point (after I am back).

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_connection_base.py
tozka: Depends on the driver library. The point here is the remote end (the database server) may have disconnected or the session may have timed out (on the remote oned) but in the client (here in vdk - we are the client) we do not know. So get_managed_connection as you say aims to return verified connection so "select 1" is statement that both is something that works on each db (we think) and it's usually optimized by most db since it's frequently used by validation

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
```diff
@@ -75,25 +75,24 @@ def open_connection(self, dbtype: str) -> ManagedConnectionBase:
         :return: the new connection if succesfull or throws an expception
         """
```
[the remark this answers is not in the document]
ivakoleva: Screenshot 2022-12-12 at 9 50 07
The IDE usually indicates typos, you can leverage this check to provide meaningful documentation

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/impl/router.py
ivakoleva: did checkout the branch to see the reconnect made, looks good to me

projects/vdk-core/tests/functional/run/jobs/pandas-job/10_pandas_step.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import pandas as pd
+from vdk.api.job_input import IJobInput
+
+
+def run(job_input: IJobInput):
+    # The job only uses job_input.get_managed_connection to check it is initialized properly
+    # Should not use job_input.execute_query
```
ivakoleva, line 9: don't see the reconnect-proof test; where do we check for a reconnect made exactly?

ivakoleva (approved):

projects/vdk-core/tests/functional/run/jobs/pandas-job/10_pandas_step.py
tozka: added unit test to cover re-connect scenario.

PR #1415, opened 2022-12-06
vdk-jupyter: notebook-plugin 

duyguHsnHsn:
What:
Adding the hook for initialising a job which works with .ipynb files and the helper classes.

Why: 
It is linked to the [issue](https://github.com/vmware/versatile-data-kit/issues/1275)

Tests: provided unit tests


projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
```diff
@@ -1,33 +1,28 @@
...
-@hookimpl
-def vdk_start(plugin_registry: IPluginRegistry, command_line_args: List):
-    plugin_registry.load_plugin_with_hooks_impl(DummyPlugin(), "DummyPlugin")
+        return
+    file_locator: JobNotebookLocator = JobNotebookLocator()
+    notebook_files = file_locator.get_notebook_files(context.job_directory)
+    if len(notebook_files) >= 1:
+        for file_path in notebook_files:
+            nb: Notebook = Notebook(file_path)
+            NotebookReader.read_notebook_and_save_steps(nb, context)
```
ivakoleva, line 28: Actually, `class Notebook` definition is physically closer to `class NotebookReader`,
so why importing both to `notebook_plugin.py` to just to pass an instance of the one, to the method of the other?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_reader.py
```diff
@@ -0,0 +1,89 @@
...
+                            ]
+                        ),
+                        type=TYPE_PYTHON,
+                        runner_func=NotebookStepFuncFactory.run_python_step,
+                        file_path=notebook.file_path,
+                        job_dir=context.job_directory,
+                        code=python_cells,
+                    )
+                    step_index += 1
+                    context.step_builder.add_step(step)
```
ivakoleva, line 89: nit: 88 and 89 seem to duplicate 70 and 71

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_step.py
```diff
@@ -0,0 +1,29 @@
...
+log = logging.getLogger(__name__)
+
+# The function accept NotebookStep (below class) and IJobInput and
+# return true if the step has been executed and false if it is not (valid) executable step.
+# On error it's expected to raise an exception.
+NotebookStepFunction = Callable[["NotebookStep", IJobInput], bool]
+
+
+@dataclass
+class NotebookStep(Step):
```
ivakoleva, line 21: Why is the NotebookStep in a dedicated file, instead of moving next to `class Notebook` and calling it `notebook.py`? It seems as a compound part of `Notebook` concept, that consists of steps?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -0,0 +1,120 @@
...
+                        code.replace(";", "")
+                        self.sql_and_run_cells.append(
+                            [code.replace("%sql", ""), TYPE_SQL]
+                        )
+                    else:
+                        self.sql_and_run_cells.append(
+                            ["".join(cell["source"]), TYPE_PYTHON]
+                        )
+                else:
+                    self.python_helper_cells.append("".join(cell["source"]))
```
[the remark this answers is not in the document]
ivakoleva: btw, I really like predicates, because they explain the design better due naming them, and could be reused.
For example a named lambda(/function) `code.startswith("%sql")` applied to code: str, what would be....`is_code_sql`
and `"def run(" in code` could be `code_is_python`

Then you may consider moving those top-level, so within method implementation you reuse them, and a `notebook.py` has this functionallity to distinguish between code that is sql or python. And in case you need to change the implementation, you may do within the lambdas(/functions) themselves, or encapsulated.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -0,0 +1,120 @@
...
+            self.sql_and_run_cells = []
+            self.python_helper_cells = []
+            self.file_path = file_path
+
+            content = json.loads(file_path.read_text())
+            cells = [
+                cell
+                for cell in content["cells"]
+                if cell["cell_type"] == "code"
+                and "vdk" in cell["metadata"].get("tags", {})
```
ivakoleva, line 59: nit: this could also be a (potentially top-level) predicate, for example `is_cell_vdk_code` that accepts cell:...

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -0,0 +1,120 @@
...
+                why_it_happened=f"The provided {file_path.name} cannot be loaded into json format and "
+                f"cannot be read as a Jupyter notebook",
+                consequences=errors.MSG_CONSEQUENCE_TERMINATING_APP,
+                countermeasures=f"Check the {file_path.name} format again",
+                exception=e,
+                wrap_in_vdk_error=True,
+            )
+
+    def extract_notebook_steps(self, context: JobContext):
+        for index, cell in enumerate(self.sql_and_run_cells):
```
ivakoleva, line 91: nit: since by design `self.sql_and_run_cells` populated (during init) is a requirement for this method implementation, you may assert on it for the purpose of fail-fast. That makes the future feature extending support less error-prone.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -0,0 +1,120 @@
...
+                            self.file_path.name.replace(".ipynb", "_"),
+                            str(index),
+                        ]
+                    ),
+                    type=TYPE_PYTHON,
+                    runner_func=NotebookStepFuncFactory.run_python_step,
+                    file_path=self.file_path,
+                    job_dir=context.job_directory,
+                    code="\n".join(self.python_helper_cells) + "\n" + cell[0],
+                )
```
ivakoleva, line 119: Screenshot 2022-12-08 at 10 28 03
nit: take a look how similar if-else contents are; maybe deduplicate code?

projects/vdk-plugins/vdk-notebook/tests/test_plugin.py
```diff
@@ -1,30 +1,64 @@
...
-        # and mock large parts of it - e.g passed our own plugins
-        runner = CliEntryBasedTestRunner(notebook_plugin)
+    def test_notebook_plugin(self) -> None:
+        result: Result = self.__runner.invoke(
+            ["run", jobs_path_from_caller_directory("rest-api-job")]
+        )
+        cli_assert_equal(0, result)
+        actual_rs: Result = self.__runner.invoke(
+            ["sqlite-query", "--query", f"SELECT * FROM rest_target_table"]
+        )
```
ivakoleva, line 33: nit: take a look at the codacy notifications to cleanup

projects/vdk-plugins/vdk-notebook/tests/test_plugin.py
```diff
@@ -1,30 +1,64 @@
...
-    with mock.patch.dict(
-        os.environ,
-        {
-            # mock the vdk configuration needed for our test
-        },
-    ):
-        # CliEntryBasedTestRunner (provided by vdk-test-utils) gives a away to simulate vdk command
-        # and mock large parts of it - e.g passed our own plugins
-        runner = CliEntryBasedTestRunner(notebook_plugin)
+    def test_notebook_plugin(self) -> None:
```
ivakoleva, line 26: what's the purpose of this test? is it the happy path or success scenario?

projects/vdk-plugins/vdk-notebook/tests/test_plugin.py
```diff
@@ -1,30 +1,64 @@
...
+        )
+        cli_assert_equal(0, result)
+        actual_rs: Result = self.__runner.invoke(
+            ["sqlite-query", "--query", f"SELECT * FROM rest_target_table_copy"]
+        )
+        assert actual_rs.stdout == (
+            "  userId    id  title                 completed\n"
+            "--------  ----  ------------------  -----------\n"
+            "       1     1  delectus aut autem            0\n"
+        )
```
ivakoleva, line 64: what is the expectation about the mixed file types use-case scenario `(.ipynb +[.sql | .py])`?

projects/vdk-plugins/vdk-notebook/tests/jobs/rest-api-job-code-error/steps.ipynb
```diff
@@ -0,0 +1,90 @@
...
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.9.12"
```
ivakoleva, line 85: what is the notebook python version about, how does it relate to the vdk integration or it does not ?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
duyguHsnHsn: There were some problems with the types, for some reason. When I didn't use if else but directly use variable for the code and the type instead, I got platform error, bc vdk was trying to run the python step as an sql one

projects/vdk-plugins/vdk-notebook/tests/jobs/rest-api-job-code-error/steps.ipynb
duyguHsnHsn: it is the notebook file, it contains all the information about the notebook. We cannot change the format of it, but only use the needed information from it

projects/vdk-plugins/vdk-notebook/tests/test_plugin.py
duyguHsnHsn: currently, we won't have mixed data jobs.

projects/vdk-plugins/vdk-notebook/tests/test_plugin.py
duyguHsnHsn: it is the success  scenario, I might rename the test

projects/vdk-plugins/vdk-notebook/tests/jobs/rest-api-job-code-error/steps.ipynb
ivakoleva: What happens in case of discrepancy with the notebook version property, and the python version we run the vdk + job?

ivakoleva (approved):
LGTM

projects/vdk-plugins/vdk-notebook/tests/test_plugin.py
ivakoleva: Okay, however is it possible for a user to mix those files, and what is expected to happen?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -0,0 +1,106 @@
...
+            log.debug(f"Neither VDK run methods nor SQL statements were detected!")
+        for index, cell in enumerate(self.sql_and_run_cells):
+            cell_type = TYPE_PYTHON if cell.is_vdk_run_cell() else TYPE_SQL
+            runner_func = (
+                NotebookStepFuncFactory.run_python_step
+                if cell.is_vdk_run_cell()
+                else NotebookStepFuncFactory.run_sql_step
+            )
+            if cell.is_vdk_run_cell():
+                cell.add_code(self.python_helper_cells)
```
murphp15, line 92: do all helper cells run before all vdk run cells?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_based_step.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @staticmethod
+    def invoke_run_function(func: Callable, job_input: IJobInput, step: NotebookStep):
+        full_arg_spec = inspect.getfullargspec(func)
+        parameter_names = full_arg_spec[0]
+        possible_arguments = {
+            "job_input": job_input,
+        }
+        # enable plugins to add different types of job input
+        actual_arguments = {
```
murphp15, line 123: Can we not just assume that the only argument is job_input?

murphp15 (approved):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
duyguHsnHsn: In a data job, we only look for the run method.The helper cells code is run, only if  the run method contains calls to other methods, which will be declared in one of the helper cells.

PR #1432, opened 2022-12-13
control-service: disable pet store endpoint from ever displaying

murphp15:


doks5 (approved):
Maybe you can add one line description to the commit, so whoever looks at the git history can have an idea why was the change necessary.

Issue #1441, opened 2022-12-14
control-service: Remove all references to BatchV1beta1Api(kubernetes)


murphp15:
**What is the feature request? What problem does it solve?**
Throughout the code there are references to BatchV1beta1Api. This is deprecated a long time in kubernetes and is removed in the latest version of kubernetes. 
This being present in the code is preventing us from upgrading to the latest k8s versions. 
Here is the PR that is blocked waiting on the fix https://github.com/vmware/versatile-data-kit/pull/1437

PR #1443, opened 2022-12-14
control-service: Remove support for k8s batch/v1Beta. Instead default to batch/v1

murphp15:
# Why?
We can't upgrade to the newer versions k8s because we are on such old versions of their apis. 

In version 1.25 of k8s they removed support for batch/v1Beta (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125).
 
 
 After this PR is merged VDK will work with the latest version of k8s when using the default configuration. 
This will make local development for new comers much easier along with allowing us to use the latest features offered by k8s. 

After this change is merged we won't support batch/v1Beta, which was the previous default. 

# What?
In this PR I remove our dependencies on their very old apis and CRD objects. 
To move to the newest version of k8s we need to remove dependencies on batch/v1beta1. 
I also cleaned up some test code that was using mocks when it didn't need to be using them.

# How has this been tested?
Unit and integration tests. 

Before this is merged we need to test that if a job was originally running as a v1Beta and then changes to v1 that there are no unintended consequences. 

closes: https://github.com/vmware/versatile-data-kit/issues/1441


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -820,7 +820,7 @@ customStartupProbeProbe: {}
 datajobTemplate:
   enabled: false
   template:
-    apiVersion: batch/v1beta1
+    apiVersion: batch/v1
```
ivakoleva, line 823: what are then the supported values after the change?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
```diff
@@ -21,8 +21,7 @@ public class ControlKubernetesService extends KubernetesService {
   // those should be null/empty when Control service is deployed in k8s hence default is empty
   public ControlKubernetesService(
       @Value("${datajobs.control.k8s.namespace:}") String namespace,
-      @Value("${datajobs.control.k8s.kubeconfig:}") String kubeconfig,
-      @Value("${datajobs.control.k8s.k8sSupportsV1CronJob}") boolean k8sSupportsV1CronJob) {
-    super(namespace, kubeconfig, k8sSupportsV1CronJob, log);
+      @Value("${datajobs.control.k8s.kubeconfig:}") String kubeconfig) {
+    super(namespace, kubeconfig, log);
   }
```
ivakoleva, line 26: the boolean toggle was introduced for backwards & forward compatibility that is configurable. Please describe the compatibility proposed

projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
```diff
@@ -1,7 +1,7 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-apiVersion: batch/v1beta1
+apiVersion: batch/v1
 kind: CronJob
```
ivakoleva, line 5: In case we want to switch the VDK defaults to latest, maybe we need to reconsider releasing a major version, and notifying in advance the users so they have time to switch the configurations accordingly. I don't see however why such a change is needed, since it is still configurable, haven't heard a use-case that requires this yet

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
murphp15: No backwards compatibility supported. 
This changes removes backwards compatility support.

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
murphp15: v1 is only supported value.

projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
murphp15: Yeah maybe this needs to be a major release. 
Its not switching the defaults. 
V1 is the only supported value now. v1Beta was deprecated years ago and is being completed removed from the apis now.

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
ivakoleva: nit: since the only value, then why at all configurable?

projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
ivakoleva: I am not familiar with any use case that requires such a change, why is it better t explicitly purposefully remove backwards compatibility?

ivakoleva (changes requested):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
murphp15: Its not configurable, its like the way it says cronJob below. 
Its just describing the values for the template.

projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
murphp15: Because it is really really old and kubernetes are not supporting it anymore. 
It was a beta version of v1 while they tested stuff out. 
But really everyone should have moved to v1 when it became available because that is the stable version.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
ivakoleva: Since the users could be running any Kubernetes, isn't it more valuable for the OSS project to additionally cover those who had not upgraded due to some reason, instead of discarding them from using VDK on older Kubernetes? There is this group of users X that may not be able to upgrade, and still want to use VDK. Why are we now making the k8s version a hard requirement all of a sudden?

projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
ivakoleva: So in fact this is not quite "migrating" to newest k8s version, because currently VDK is able to run on latest version, right? We ar dropping the older versions support, and my suggestion is to rename the PR accordingly.

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
ivakoleva: nit: so when you have a configuration with 1 possible option, then it is not configurable. why not removing the configuration entirely, since it cannot be anything else than this particular value..?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/DataJobsKubernetesService.java
```diff
@@ -24,9 +24,8 @@ public class DataJobsKubernetesService extends KubernetesService {
 
   public DataJobsKubernetesService(
       @Value("${datajobs.deployment.k8s.namespace:}") String namespace,
-      @Value("${datajobs.deployment.k8s.kubeconfig:}") String kubeconfig,
-      @Value("${datajobs.control.k8s.k8sSupportsV1CronJob}") boolean k8sSupportsV1CronJob) {
-    super(namespace, kubeconfig, k8sSupportsV1CronJob, log);
+      @Value("${datajobs.deployment.k8s.kubeconfig:}") String kubeconfig) {
+    super(namespace, kubeconfig, log);
     if (StringUtils.isBlank(kubeconfig) && !new File(kubeconfig).isFile()) {
```
ivakoleva, line 29: please provide to PR description the Kubernetes compatibility expected with the change proposed, before and after

ivakoleva (changes requested):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
ivakoleva: > If we choose to support this beta version VDK we will never be able to upgrade to the latest k8s clients and won't be able to use the latest features k8s has to offer.

Could you please elaborate?

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
murphp15: This isn't configuration. 
It is the template for a cronJob. 

For example you wouldn't suggest removing the line  `kind: CronJob` two lines below it would you? 

We need to specify it here so k8s knows what apiVersion we are using.
Does that answer your q?

ivakoleva: oh yes, that's the default cronjob `apiVersion` value, used by users unless reconfigured.
This is definitely an API change, that requires consideration and documentation accordingly.
Currently, Kubernetes instance is provisioned externally as a prerequisite to installing VDK.

murphp15:
Before this is merged we need to test that if a job was originally running as a v1Beta and then changes to v1 that there are no unintended consequences.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/DataJobsKubernetesService.java
ivakoleva: Could you provide the Kuberenetes versions supported prior/after the change? It is still not clear to me...

ivakoleva: In addition, I understand the defaults need to be switched to v1 in order to use >=1.25. 
So, why not changing the defaults and still keep the backwards compatibility?

tozka:
We cannot support old versions of Kubernetes. But we also should not only support the latest brand new version of Kubernetes (since only the most geeky organizations would use it :) ) 

If I could see k8s API version adoption statistics I'd say go for where 80% are. But I could not find one. 
There is official supported versions from K8s - https://kubernetes.io/releases/ 

Currently, the version supported by the Kubernetes team are 1.23+ so it makes sense for us to support those. 

If we decide to support 1.23+ we will break all our users that use older versions. And I happen to know our biggest one (vmw) are still using 1.20

projects/control-service/projects/pipelines_control_service/src/main/resources/k8s-data-job-template.yaml
ivakoleva: I believe this PR could be split in two. One is the template apiVersion defaults, set to v1 that is required for running on Kubernetes latest.

tozka:
> If we decide to support 1.23+ we will break all our users that use older versions. And I happen to know our biggest one (vmw) are still using 1.20

We *may* break them. I am not sure we will.

murphp15:
@ivakoleva @tozka 

Support for batch/v1Beta1 was added in 1.19 (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/).
Support for batch/v1 was only introduced in 1.21 (https://kubernetes.io/blog/2021/04/09/kubernetes-release-1.21-cronjob-ga/). 
Support for batch/v1Beta1 was removed in 1.25. 


If most of VMWare are using 1.20 then I think @ivakoleva is correct and this PR is too aggressive. 
Let me know what yous think?

ivakoleva (changes requested):
We need to support all our customers along the way, some had invested in using VDK on older Kubernetes already, and may or may not be able to upgrade. 
In other words, is there a  problem with changing the defaults so that we support latest + configurable older k8s APIs, or not?

murphp15:
> We need to support all our customers along the way, some had invested in using VDK on older Kubernetes already, and may or may not be able to upgrade. In other words, is there a problem with changing the defaults so that we support latest + configurable older k8s APIs, or not?

Yeah we can't support the beta version while using the latest k8s client version. It simply won't compile. 
The objects for V1Beta don't exist anymore. 

But int fact I think we shouldn't merge this just yet.

ivakoleva:
> @ivakoleva @tozka
> 
> Support for batch/v1Beta1 was added in 1.19 (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/). Support for batch/v1 was only introduced in 1.21 (https://kubernetes.io/blog/2021/04/09/kubernetes-release-1.21-cronjob-ga/). Support for batch/v1Beta1 was removed in 1.25.
> 
> If most of VMWare are using 1.20 then I think @ivakoleva is correct and this PR is too aggressive.
> 
> At the moment VDK works with k8s 1.19+ however requires config changed to work with 1.25+. If this PR is merged then VDK works with k8s 1.21+.
> 
> Let me know what yous think?

Thanks, please add this info to the PR description

Some customers are actively using v1.20.13, verifiable using `kubect version --short`. However we may presume as well we have customers at each VDK snapshot's supported Kubernetes (since priorly backwards-compatible).

ivakoleva:
> > We need to support all our customers along the way, some had invested in using VDK on older Kubernetes already, and may or may not be able to upgrade. In other words, is there a problem with changing the defaults so that we support latest + configurable older k8s APIs, or not?
> 
> Yeah we can't support the beta version while using the latest k8s client version. It simply won't compile. The objects for V1Beta don't exist anymore.
> 
> But in fact I think we shouldn't merge this just yet.

Maybe conditionally add to class path the compatible k8s api client that corresponds to the configuration?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/DataJobsKubernetesService.java
murphp15: Yeah sorry I see where the confusion is.
We won't be able to use the latest versions of the k8s client SDK. not the latest version of the k8s cluster. 
Does that make sense?

ivakoleva: What's the latest k8s client backwards compatibility, could you please link to the PR?

murphp15:
Will reopen and merge when customers are on newer versions of k8s

ivakoleva:
Please add the information to the project codebase about current Kubernetes versions support, you could link the ticket as well so that it is easier for people to upvote in case needed, so we can prioritize

ivakoleva:
Where users installing the control-service look at, like README

PR #1446, opened 2022-12-14
vdk-jupyer: Jupyter UI monitor environmental variables

duyguHsnHsn:
What: 
Added the VDK Manu with one element in it - an option "Set environmental variables". The idea of this option is: when the user clicks it he gets a pop up where he can see all the variables he had added from Jupyter UI or add a new one. These variables are used for run job. For example the user can add " VDK_INGEST_METHOD_DEFAULT=SQLITE" as an environmental variable. This  pop up will show only the variables added from Jupyter, other externally added ones won't be shown. The user can delete and add new ones. They will be used only when he runs the job.
These variables are saved in the local storage, so when the user opens multiple Jupyter sessions the variables will be set in all of them. They will be deleted when the browser is closed. 

Why:
The users will easily monitor all the vdk variables, since they will be listed in one place. This will give them the chance to chose whether they will set them from the terminal or from the UI. And the project will be more error prone because it is easy to forget about those things when working on UI, but when we have it as an option it will be as a "reminder" to check those things. 

Signed-off-by: Duygu Hasan <hduygu@vmware.com>


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/LICENSE
```diff
@@ -1 +1 @@
-See https://github.com/vmware/versatile-data-kit/blob/main/LICENSE
\ No newline at end of file
+See https://github.com/vmware/versatile-data-kit/blob/main/LICENSE
```
tozka, line 1: Why not delete the file? Do we need it?

duyguHsnHsn: Jupyterlab returns exception for not having a LICENSE  file and the extension cannot be loaded. That's why I added it again

tozka:
Each PR shoudl have linked story where the problem and the proposed solution is described. 

Why is this change necessary? Notebooks have a native way to specify environment variables using `%env` syntax 
And even for credentials as environment variables there are solutions - https://towardsdatascience.com/keeping-credentials-safe-in-jupyter-notebooks-fbd215a8e311

PR #1459, opened 2022-12-19
control-service: graphql wildcard matching filter for team and job names

mrMoZ1:
what: Added wildcard matching e.g *-search-* in graphQl filters for job name and team name and exact matching of search strings.

why: Current implementation uses Java String's .contains() method which doesn't have the ability to return exact matches. Users of the control-service API ran into usability issues when multiple results were returned, without the ability to filter for exact matches.

testing: added tests.


mivanov1988:
LGTM! We should apply this change to the rest of the String properties provided by jobsQuery but in a separate PR.

murphp15 (approved):

gageorgiev (approved):

mivanov1988 (approved):

tozka:
Please update the documentation at api.yaml definition. You may add new example covering the new additions

PR #1463, opened 2022-12-22
vdk-jupyter: python subprocess  security problem

duyguHsnHsn:
What:
This PR introduce a small change that can lower the vulnerability of subprocess call in JupyterLab extension.

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
@@ -36,6 +36,5 @@ class VdkUI:
             process = subprocess.Popen(
                 cmd, stdout=log_file, stderr=log_file, env=os.environ.copy()
             )
-            subprocess.check_output(["/bin/ls", "-l"])
             process.wait()
             return f"{process.returncode}"
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
@@ -34,7 +34,7 @@ class VdkUI:
             )
             cmd = shlex.split(cmd_str)
             process = subprocess.Popen(
-                cmd, stdout=log_file, stderr=log_file, env=os.environ.copy()
+                cmd, stdout=log_file, stderr=log_file, env=os.environ.copy(), shell=False
             )
             process.wait()
             return f"{process.returncode}"


gageorgiev (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
```diff
@@ -34,8 +34,11 @@ def run_job(path, arguments=None):
             )
             cmd = shlex.split(cmd_str)
             process = subprocess.Popen(
-                cmd, stdout=log_file, stderr=log_file, env=os.environ.copy()
+                cmd,
```
tozka, line 37: I thought the [stackoverflow](https://stackoverflow.com/questions/21009416/python-subprocess-security) is clear but this one is clearer ? https://security.openstack.org/guidelines/dg_use-subprocess-securely.html

I think we can ignore the warning. But if you want to remove it. Most likely you need to split the "vdk" from the cmd_str string so that the analyzer realizes that the command (cmd[0]) is not user input . We likely can ignore the warning as the code is safe. But it's better not to have it as not to have doubt.

duyguHsnHsn: Thank you, I understand it better now

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
```diff
@@ -34,8 +34,11 @@ def run_job(path, arguments=None):
             )
             cmd = shlex.split(cmd_str)
```
ivakoleva, line 35: Seems that manually appending to one giant string then splitting, defeats the fix purpose:
```
cmd_str = (
                f"vdk run {path}"
                if not arguments
                else f"vdk run {path} --arguments {arguments}"
             )
cmd = shlex.split(cmd_str)
```
I think, the parameterised cmd items should be added separately to the list `cmd`.

ivakoleva (changes requested):
shell=False LGTM, the request is for a few minor changes

ivakoleva:
Please add links to the subprocess security guidelines you had I'm mind when opening this fix

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk.py
```diff
@@ -26,16 +26,17 @@ def run_job(path, arguments=None):
...
+                cmd.append(f"{arguments}")
             process = subprocess.Popen(
-                cmd, stdout=log_file, stderr=log_file, env=os.environ.copy()
+                cmd,
+                stdout=log_file,
+                stderr=log_file,
+                env=os.environ.copy(),
+                shell=False,
             )
-            subprocess.check_output(["/bin/ls", "-l"])
```
ivakoleva, line 39: why is the line cleared?

duyguHsnHsn: This was there for testing purposes, I have missed to delete it in my previous commit

murphp15 (approved):

PR #1465, opened 2022-12-22
vdk-jupyter:  splitting functionalities of vdk-notebook Cell class

duyguHsnHsn:
What: 
  Splitted the existing Cell class into two different classes - the data class Cell and a class that consists of static methods used for cell operations

Why:
 In this way we describe the data model of a Cell better. 


tozka:
> minor changes on vdk-notebook

Please make sure your title explain what the change is . This way to generic title

gageorgiev (approved):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
```diff
@@ -1,35 +1,42 @@
...
-        return "def run(" in self.source
+    @staticmethod
+    def is_sql_cell(cell: Cell):
+        return cell.source.startswith("%sql")
 
-    def add_code(self, cells):
+    @staticmethod
+    def is_vdk_run_cell(cell: Cell):
+        return "def run(" in cell.source
+
```
tozka, line 29: You should very very very rarely rely on "custom" parsing logic (that is not following some standard (e.g json)) ever. OK? Never do that again, please. 
 It's so easy to break and have corner cases you cannot even think of currently. This will not work long term. Period. This line will have bugs 99.999% sure. 

Instead, you should use the python parser to parse the file and see if there's a method called run there. 

While We need to fix that. I missed that in previous review, and this PR is about other things. So create a story and will fix that later.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
```diff
@@ -1,35 +1,42 @@
...
+class CellFuncFactory:
+    @staticmethod
+    def is_vdk_cell(cell: Cell):
+        return "vdk" in cell.tags
 
-    def is_vdk_run_cell(self):
-        return "def run(" in self.source
+    @staticmethod
+    def is_sql_cell(cell: Cell):
+        return cell.source.startswith("%sql")
```
tozka, line 24: Hmm. Isn't there a better way. See my comment below.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
```diff
@@ -1,35 +1,42 @@
...
     def __init__(self, jupyter_cell):
         self.tags = jupyter_cell["metadata"].get("tags", {})
         self.source = "".join(jupyter_cell["source"])
 
-    def is_vdk_cell(self):
-        return "vdk" in self.tags
 
-    def is_sql_cell(self):
-        return self.source.startswith("%sql")
+class CellFuncFactory:
```
tozka, line 17: Factory seems misnamed. When talking about factory I expect to see the design pattern - https://refactoring.guru/design-patterns/factory-method  or https://refactoring.guru/design-patterns/abstract-factory

This is more like static class with helpful utilities for cells. I'd name mit CellUtils.

tozka (approved):

Issue #1361, opened 2022-11-29
Add data quality validation to processing templates


sbuldeev:
The feature request is to add new functionality to the processing templates in order to allow quality checks to be made before the data is inserted into the target table. 
Currently, the checks done on processing step are not covering if the semantics of the data is correct. Therefore, bad data could went into the target table which could be unwanted behavior.

tozka:
Here is a snippet of conversations about the solution that were done "offline". Posting here for record-keeping purposes so that we know why we've made some deicions 

```
We are considering adding quality checks as part of the templates.
 
Currently execute template work like this:
 
1 - Check arguments and test if source query/view matches target
2- Insert into target from template query (insert overwrite with some logic)
3-Refresh
4- Compute stats
 
We are considering changing them to enable user to run quality check before data is inserted . Something like
 
1. Check arguments and test if source query/view matches target
2. If quality check is set  (more details below on this step but at high level something like this):
2.a Insert into temp table from template query)
2.b Run quality checks on temp table
2.c Insert overwrite into target template from temp table
3. Else if no quality check just insert overwrite as before
4. Refresh
5. Compute stats
 
 
On step 2. There are different approaches.
It’s important that the quality check is run on the same data that will be inserted. So we need a staging table
Considering the impact of Impala can you advice on which approach is acceptable for Impala currently.
For example are DML statements more expensive than DDL ?

Not all templates will need quality checks.
 
The current approaches I see:
 
Approach 1.  Use alter table rename
  Create temp table
  insert into temp table
  run quality check on temp table
  alter table rename target to backup_target
  alter table name temp to target
  drop target
 
This though adds 4 extra DDL statements to the template …
 
Approach 2.  Use 2 inserts and temp
 
  Create temp table
  insert into temp table
  run quality check on temp table
  insert into target from temp table
  drop temp table
 
This adds 2 extra DDL statements and 1 extra DML  (insert)
 
Approach 3.  Use 2 inserts and amortized temp creation
 
Create temp IF NOT EXISTS (OR IF SCHEMA HAS CHANGED) table
  insert into temp table
  run quality check on temp table
  insert into target from temp table
 
This will add 1 extra DDL statement  when template is run for a first time or target schema has changed (hence cost is amortized over time)
And 1 extra DML statement
```

```
Both options will add more IO operations and eventually may saturate network (especially INSERT FROM SELECT)
 
Have you considered to add temporary partition column, verify data were this column is set and just drop the column again.
 
[...]

PR #1472, opened 2023-01-04
vdk-impala: Introduce checks for scd1 template

sbuldeev:
What:
Adding functionality to handle the scd1 template behavior if there are checks provided by the user

Why:
It is linked to https://github.com/vmware/versatile-data-kit/issues/1361

Tests: provided positive and negative regression tests


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,94 @@
...
+
+SQL_FILES_FOLDER = (
+    os.path.dirname(os.path.abspath(__file__)) + "/02-requisite-sql-scripts"
+)
+
+
+def run(job_input: IJobInput):
+    job_arguments = job_input.get_arguments()
+
+    check = job_arguments.get("check")
```
tozka, line 17: I think we should allow users to omit passing check (and assume it's None). 

```suggestion
    check = job_arguments.get("check", None)
```

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,94 @@
...
+    target_schema = job_arguments.get("target_schema")
+    target_table = job_arguments.get("target_table")
+    staging_schema = job_arguments.get("staging_schema")
```
tozka, line 20: Let's have a default as well. If not specified uses same as target_schema 
```suggestion
    staging_schema = job_arguments.get("staging_schema". target_schema)
```

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,94 @@
...
+    insert_query = get_query("02-insert-into-target.sql")
+
+    if check:
+        if not staging_schema:
+            raise Exception(
```
tozka, line 25: It's considered bad practice to use the base Exception class . 

It's better to use specific exception. For Illegal arguments ValueError is a frequently used

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
```diff
@@ -108,6 +109,56 @@ def test_load_dimension_scd1_bad_target_schema(self) -> None:
...
+                "source_schema": test_schema,
+                "source_view": source_view,
+                "target_schema": test_schema,
+                "target_table": target_table,
+                "check": sample_check,
+                "staging_schema": staging_schema,
+            },
+        )
+
+        assert res.exception
```
tozka, line 160: You may want to add an assert that verifies that the target has not been changed.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,94 @@
...
+            f"{target_schema}.{target_table}", staging_table, job_input
+        )
+
+        insert_into_staging = insert_query.format(
+            target_schema=staging_schema, target_table=target_table
+        )
+        job_input.execute_statement(insert_into_staging)
+
+        if check(staging_table):
+            insert_into_target = insert_query.format(
```
tozka, line 40: Copying from staging table to target you probably do not want to reuse the same query. 

It works for this template because it's insert overwrite, but won't for the more complex templates. 

And we want to keep the code as for the next template we'd likely want to abstract teh common functionalities and reuse them.

tozka (commented):
Looks good to me. I made some comments.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,94 @@
...
+            raise Exception(
+                "No staging_schema specified to execute the defined data checks against."
+            )
+
+        staging_table = f"{staging_schema}.{target_table}"
+        align_stg_table_with_target(
+            f"{target_schema}.{target_table}", staging_table, job_input
+        )
+
+        insert_into_staging = insert_query.format(
```
murphp15, line 34: This probably is not an issue with this PR but my understanding of VDK. 
Here the insert statement is created with target table and schema. 
However how does it know what the source table and scheme is?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,94 @@
...
+            target_schema=staging_schema, target_table=target_table
+        )
+        job_input.execute_statement(insert_into_staging)
+
+        if check(staging_table):
+            insert_into_target = insert_query.format(
+                source_schema=staging_schema, source_view=target_table
```
murphp15, line 41: If you are trying to insert into the target table should it not be target_schema and target_view instead of source_schema and source_view?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: Do we need passing a default value as None? The default behavior of dict.get("key") is to return None if the key is not present.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: Wouldn't this cause an error? In the current logic the table name will be the same for both target and staging schemas.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: Changed that.

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
sbuldeev: Added the suggested assert.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: I will have to think of a more generic way to transfer the data in order to fit all templates needs.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: I expect all the arguments that are not passed explicitly to be inherited from the arguments passed to the template call.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: On this step I want to insert from the staging table(where checks are performed) into target table. Therefore I only set the source to the staging table, because I want to avoid using the source view.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
tozka: Oh, I forgot, In this case I guess no.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
tozka: Add a TODO here if you are not ready to solve it now. It's easier to figure out what is "common" after a second instance of something.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
tozka: That's true. I didn't realize. 

But it's better UX to require as few users as possible to get started with. This means as many as possible parameters to have defaults.
Wouldn't be better to prefix the table as well? If one uses a staging db for multiple purposes, this way it'd be easy to see those tables are generated for checks. Maybe we can add prefix "vdk_check_" ? 

What do you think ?

tozka:
I don't know if you've noticed but there are some failing tests: https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3552673519 

Have you be able to run them locally ? In IDE ?

sbuldeev:
> I don't know if you've noticed but there are some failing tests: https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3552673519
> 
> Have you be able to run them locally ? In IDE ?

These tests are failing, because at some point the execute_tempalate behavior is trying to dump all the arguments and since there is a function passed as an argument this is not possible.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
sbuldeev: Okay lets proceed without that for now. 
We've discussed internally a way to introduce such generic functionality, but it will take time to be implemented, so we prefer to stick with what we have and introduce it on the next template modification.

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
```diff
@@ -108,6 +109,56 @@ def test_load_dimension_scd1_bad_target_schema(self) -> None:
...
+        actual_rs = self._run_query(f"SELECT * FROM {test_schema}.{target_table}")
+        expected_rs = self._run_query(f"SELECT * FROM {test_schema}.{source_view}")
+        assert actual_rs.output and expected_rs.output
+        assert actual_rs.output == expected_rs.output
+
+    def test_load_dimension_scd1_checks_negative(self) -> None:
+
+        test_schema = "vdkprototypes"
+        source_view = "vw_dim_org"
+        target_table = "dw_dim_org"
```
tozka, line 141: To avoid situaitons in which another test populates the data in a table. Let's use distinct names

```
source_view="vw_dim_org_check_negative"
target_table ="dw_dim_org_check_negative"

```

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
```diff
@@ -108,6 +109,56 @@ def test_load_dimension_scd1_bad_target_schema(self) -> None:
             template_args=template_args,
         )
 
+    def test_load_dimension_scd1_checks_positive(self) -> None:
+
+        test_schema = "vdkprototypes"
+        source_view = "vw_dim_org"
+        target_table = "dw_dim_org"
```
tozka, line 116: To avoid situaitons in which another test populates the data in a table. Let's use distinct names

```
source_view="vw_dim_org_check_positive"
target_table ="dw_dim_org_check_positive"

```

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/dimension/scd1/02-handle-quality-checks.py
```diff
@@ -0,0 +1,105 @@
...
+import re
+
+from vdk.api.job_input import IJobInput
+
+SQL_FILES_FOLDER = (
+    os.path.dirname(os.path.abspath(__file__)) + "/02-requisite-sql-scripts"
+)
+
+
+def run(job_input: IJobInput):
```
tozka, line 13: Let's add a comment which explains what the step does at high level (e.g what is its purpose). To help future maintainers.

sbuldeev: Added high level description.

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
sbuldeev: Done.

projects/vdk-plugins/vdk-impala/tests/functional/template_regression_test.py
sbuldeev: Done

tozka (approved):
Looks good to me. 
Thanks for the hard work on this change.

tozka:
@sbuldeev, before merging do ping me so I can walk you through the proper process for merging a change. 

A) It's fairly important that there's one commit (in this case particularly) - so you need to use Squash option 
B) It's important that the commit description is the same as the PR description. 
   There also should be Signed-off-by: Humpty Dumpty <humpty.dumpty@example.com> in the commit description.

murphp15 (approved):
Looks good

PR #1483, opened 2023-01-06
vdk-jupyter: create iPython extension

duyguHsnHsn:
What:
Add VDK iPython extension which using the StandaloneDataJob class loads job_input variable to a notebook. The job_input is created for the data job that the user specifies for example with giving the path to the job's parent directory.
You can see the README file to see how the extension is used.

Why: This extension allows users to use Jupyter with VDK in a familiar for them way. Since they will be able to work with job_input almost as same as working with a variable declared in a notebook.

Tests: unit tests are introduced


projects/vdk-plugins/vdk-ipython-ext/README.md
```diff
@@ -0,0 +1,53 @@
+# ipython-ext
```
tozka, line 1: ext is not needed . We do not add "plugin" or "extension" or similar prefix/suffix to any other plugins. 

Just `vdk-ipython` .

duyguHsnHsn:
Please, only look at the vdk-ipython plugin

projects/vdk-plugins/vdk-ipython/.plugin-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    PLUGIN_NAME: vdk-ipython-ext
+  extends: .build-plugin
+
+build-py37-vdk-ipython-ext:
+  extends: .build-vdk-ipython-ext
+  image: "python:3.7"
+
+
+build-py38-vdk-ipython-ext:
+  extends: .build-vdk-ipython-ext
```
tozka, line 17: This should likely be renamed as well `build-vdk-ipython-ext` (above and below)

projects/vdk-plugins/vdk-ipython/setup.py
```diff
@@ -0,0 +1,31 @@
...
+__version__ = "0.1.0"
+
+setuptools.setup(
+    name="vdk-ipython",
+    version=__version__,
+    url="https://github.com/vmware/versatile-data-kit",
+    description="Ipython extension for VDK",
+    long_description=pathlib.Path("README.md").read_text(),
+    long_description_content_type="text/markdown",
+    install_requires=["vdk-core"],
```
tozka, line 17: Seems suspicious that ipython is missing as dependency here. 

It's explained in the requirements.txt comment : 
```
this [requirements.txt] file is used to provide testing requirements
for requirements (dependencies) needed during and after installation of the plugin see (and update) setup.py install_requires section
```

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def load_job(
+    path: str = None, name: str = None, arguments: str = None, template: str = None
+):
+    path = pathlib.Path(path) if path else pathlib.Path(os.getcwd())
+    with StandaloneDataJobFactory.create(
+        data_job_directory=path, name=name, job_args=arguments, template_name=template
+    ) as job_input:
+        get_ipython().push(variables={"job_input": job_input})
```
tozka, line 45: The way it's written, we store "job_input" in get_ipython().push . and then load_job exits and that means that StandaloneDataJobFactory.__exit__ will be called and will finalize the job. 

OK. .. Let's merge it this way but this is not what we want, right?

ivakoleva, line 45: job_input potentially contains connection(s), being a IJobInput instance, and what I saw https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.interactiveshell.html#IPython.core.interactiveshell.InteractiveShell.push is that it supports dict & str & list|tuple of str only

duyguHsnHsn: In this paragraph I think they talk about the way the variables are pushed. For example, I use  dict {"job_input": job_input},  str can also be used or list/tuple
I don't know whether this answers the question
Let me know if you mean something else

duyguHsnHsn: No, maybe we will need to push StandaloneDataJobFactory instead of this, which will give the user more responsibility, tho. I will try to fix this in further changes.

murphp15 (approved):

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
ivakoleva: is with resources closing the job_input after the code block containing push?

ivakoleva (approved):

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
duyguHsnHsn: not quite sure, since the documentation of push method is not much detailed. I am not sure whether it creates a copy of the job_input or not. This will be checked in the later changes, since it will take some time, it would be better to not include t in this PR.

PR #1487, opened 2023-01-06
vdk-jobs-troubleshooting: Improve robustness of the plugin

dakodakov:
Why?
Stop the plugin from crashing a data job run if the configuration is not complete or the target port is alredy used.

What?
Changed default value of TROUBLESHOOT_UTILITIES_TO_USE Added find free port logic to server start.
Added try/except logic to prevent the plugin from crashing the job run.

How has this been tested?
Added another test case in the functional test.

Bug fix

Signed-off-by: Dako Dakov <ddakov@vmware.com>

### DIFF
--- a/projects/vdk-plugins/vdk-jobs-troubleshooting/src/vdk/plugin/jobs_troubleshoot/troubleshoot_utilities/healthcheck_server.py
+++ b/projects/vdk-plugins/vdk-jobs-troubleshooting/src/vdk/plugin/jobs_troubleshoot/troubleshoot_utilities/healthcheck_server.py
@@ -1,6 +1,7 @@
 # Copyright 2021 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import logging
+import socket
 from http.server import HTTPServer
 from threading import Thread
 from typing import Any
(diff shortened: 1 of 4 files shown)


murphp15 (approved):

projects/vdk-plugins/vdk-jobs-troubleshooting/src/vdk/plugin/jobs_troubleshoot/troubleshoot_utilities/healthcheck_server.py
```diff
@@ -21,25 +22,50 @@ def __init__(self, port: int, handler: Any = None):
...
+                port = self.find_open_port(port)
+                self._server = HTTPServer(("", port), handler)
+                self._thread = Thread(target=self._server.serve_forever)
+                log.info(f"Troubleshooting utility server started on port {port}.")
+            else:
+                log.error(
+                    "Troubleshooting utility handler not specified. Will not start the server."
+                )
+        except Exception as e:
+            logging.error(
```
doks5, line 36: Why not `logging.warning()`? This way the full stacktrace would be added to the message, and the exc_info would not be needed?

doks5 (approved):

projects/vdk-plugins/vdk-jobs-troubleshooting/src/vdk/plugin/jobs_troubleshoot/troubleshoot_utilities/healthcheck_server.py
dakodakov: Do you mean logging.exception()? As far as I can tell to get the exception in the when calling warning() I still have to pass the exc_info paramter?
Anyway... I prefer to keep it like that if you don't mind.

doks5: I meant `logging.warning()` (according to the documentation, https://docs.python.org/3/library/logging.html#logging.Logger.warning, it behaves like debug which means that it includes the stacktrace by default), but I am ok to leave like that.

PR #1512, opened 2023-01-13
vdk-control-service: fix null dereferences

dakodakov:
Why?
Fix null dereference uncovered by coverity scaner

What?
Multiple changes mostly related to json conversions.

Bug fix

Signed-off-by: Dako Dakov ddakov@vmware.com

### DIFF
--- a/projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
+++ b/projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
@@ -117,29 +117,27 @@ public class HttpTracer implements org.springframework.boot.actuate.trace.http.H
     ObjectMapper mapper = new ObjectMapper();
     mapper.registerModule(new JavaTimeModule());
     ObjectNode node = mapper.convertValue(trace, ObjectNode.class);
-    if (node != null) {
-      node.put("@type", "taurus_httptrace");
-      node.put("deployment_mode", String.join(",", environment.getActiveProfiles()));
-      String opId = operationContext.getOpId();
-      if (null != opId) {
-        node.put("@id", opId);
-      }
-      HttpServletRequest request = operationContext.getRequest();
-      if (null != request) { // methods from the actuator don't have the request set
-        String requestMapping =
-            "" + request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
-        node.put("request_mapping", requestMapping);
-      }
-      String user = operationContext.getUser();
-      if (null != user) {
-        node.put("user", user);
-      }
-      String team = operationContext.getTeam();
-      if (null != team) {
-        node.put("team", team);
-      }
-      String json = node.toPrettyString();
-      this.telemetryClient.sendAsync(json);
+    node.put("@type", "taurus_httptrace");
+    node.put("deployment_mode", String.join(",", environment.getActiveProfiles()));
+    String opId = operationContext.getOpId();
+    if (null != opId) {
+      node.put("@id", opId);
     }
+    HttpServletRequest request = operationContext.getRequest();
+    if (null != request) { // methods from the actuator don't have the request set
+      String requestMapping =
+          "" + request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
+      node.put("request_mapping", requestMapping);
+    }
+    String user = operationContext.getUser();
+    if (null != user) {
+      node.put("user", user);
+    }
+    String team = operationContext.getTeam();
+    if (null != team) {
+      node.put("team", team);
+    }
+    String json = node.toPrettyString();
+    this.telemetryClient.sendAsync(json);
   }
 }
--- a/projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
+++ b/projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
@@ -117,7 +117,7 @@ public class HttpTracer implements org.springframework.boot.actuate.trace.http.H
     ObjectMapper mapper = new ObjectMapper();
     mapper.registerModule(new JavaTimeModule());
     ObjectNode node = mapper.convertValue(trace, ObjectNode.class);
-    if (node != null){
+    if (node != null) {
       node.put("@type", "taurus_httptrace");
       node.put("deployment_mode", String.join(",", environment.getActiveProfiles()));
       String opId = operationContext.getOpId();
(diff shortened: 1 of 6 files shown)


murphp15:
How do you run coverity? 
where do you see the output from it?

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
```diff
@@ -117,27 +117,29 @@ void sendTelemetry(HttpTrace trace) {
...
-      node.put("request_mapping", requestMapping);
-    }
-    String user = operationContext.getUser();
-    if (null != user) {
-      node.put("user", user);
-    }
-    String team = operationContext.getTeam();
-    if (null != team) {
-      node.put("team", team);
+    if (node != null) {
```
murphp15, line 120: This only returns null if httpTrace is null? 
Maybe we could add a @NotNull annotation on the parameter and maybe an assertion instead of doing this?

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/methodintercept/DiagnosticsPublisher.java
```diff
@@ -64,12 +64,16 @@ public void accept(DiagnosticsContext context) {
 
       ObjectMapper mapper = new ObjectMapper();
       ObjectNode node = mapper.convertValue(metrics, ObjectNode.class);
-      node.put("@table", "taurus_api_call");
-      node.put("@id", opId);
-      String json = node.toPrettyString();
-      log.trace("Sending telemetry: {}", json);
-      this.telemetryClient.sendAsync(json);
+      if (node != null) {
```
murphp15, line 67: How could this be nulll?
Metrics can't ever be null at this point.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobsService.java
```diff
@@ -140,7 +140,9 @@ private boolean isInvocationSuccessful(Optional<WebHookResult> resultHolder) {
   public boolean updateJob(DataJob jobInfo) {
     var dataJob =
         jobsRepository.existsById(jobInfo.getName()) ? jobsRepository.save(jobInfo) : null;
-    dataJobMetrics.updateInfoGauges(dataJob);
+    if (dataJob != null) {
```
murphp15, line 143: Nice find. 

I think I would prefer if there was only a single if statement. 

like 

```
if(jobsRepository.existsById(jobInfo.getName())){
      dataJobMetrics.updateInfoGauges(jobsRepository.save(jobInfo));
      return true;
    }
    return false;
 ```
 
 
 But happy to merge this also.

File: projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java [hunk 4]
```diff
@@ -57,7 +57,7 @@ private void checkDependencies() {
       ProcessExecutor executor = new ProcessExecutor();
       executor.readOutput(true);
       ProcessResult result = executor.command("/bin/sh", "-c", "which kadmin").execute();
-      if (result.getExitValue() != 0) {
+      if (result != null && result.getExitValue() != 0) {
```
murphp15, line 60: If result is null it would exit successfully. Are you sure that is what we want?

murphp15 (approved):

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
dakodakov: The node object is a result of the conversion of the trace object by the object mapper and according to the coverity scan and the object mapper java doc, this method could still return a null value even if the argument is not null, hence this check.

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/methodintercept/DiagnosticsPublisher.java
dakodakov: According to the object mapper java doc, the convertValue method could return a null value even if the argument is not null, hence this check.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/JobsService.java
dakodakov: Agreed.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
dakodakov: good catch!

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/methodintercept/DiagnosticsPublisher.java
murphp15: I don't see that in the docs ? https://fasterxml.github.io/jackson-databind/javadoc/2.8/com/fasterxml/jackson/databind/ObjectMapper.html#convertValue(java.lang.Object,%20java.lang.Class)

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/HttpTracer.java
murphp15: Where in the docs does it say that it can return null for a non null object ? these are the docs I'm looking at https://fasterxml.github.io/jackson-databind/javadoc/2.8/com/fasterxml/jackson/databind/ObjectMapper.html#convertValue(java.lang.Object,%20java.lang.Class)

dakodakov: I agree.. sorry... I looked at some examples of edge cases where either the object was null or was of primitive type, which cannot happen at this part of the code. 

btw, we are using a newer version of the library and the correct javadoc link is: https://fasterxml.github.io/jackson-databind/javadoc/2.13/com/fasterxml/jackson/databind/ObjectMapper.html#convertValue(java.lang.Object,%20java.lang.Class)

projects/control-service/projects/base/src/main/java/com/vmware/taurus/service/diag/methodintercept/DiagnosticsPublisher.java
dakodakov: I agree.. sorry... I looked at some examples of edge cases where either the object was null or was of primitive type, which cannot happen at this part of the code. 

I've reverted this change. 

btw, we are using a newer version of the library and the correct javadoc link is: https://fasterxml.github.io/jackson-databind/javadoc/2.13/com/fasterxml/jackson/databind/ObjectMapper.html#convertValue(java.lang.Object,%20java.lang.Class)

mivanov1988 (approved):

tozka (approved):

PR #1520, opened 2023-01-18
vdk-jupyter: modify the way we read notebooks in notebook plugin

duyguHsnHsn:
What:
Changes on the way we read the JSON schema of a notebook. 

Why:
Since we introduced job_input as Jupyter variable with an iPython extension see https://github.com/vmware/versatile-data-kit/pull/1483 , our users now have direct access to the job_input variable and no longer need to use def run function to access it. 

Testing: Unit tests


specs/vep-994-jupyter-notebook-integration/README.md
```diff
@@ -98,59 +98,32 @@ You can see the full video in the parent directory of this file.
...
-| Ease of learning: how fast a user who has never seen the user interface before can accomplish basic tasks | 1 <br /> it is harder since the magic commands and writing a function to multiple cells is Jupyter specific                          | 2 <br />  calling functions from one function is a thing we do generally in coding                                                                  |
-
-    Solution 2 total score: 18       Solution 1 total score: 23
-
-Both solutions are not Mutually Exclusive. At some point they can be implemented both. But for the first iteration/version based on above analysis we will go with Solution 1.
-Final decision: Solution 1
+|                  Operation                   |                                                                                                                                                                                                                       Flow                                                                                                                                                                                                                        | Covered  use cases | Priority |
```
ivakoleva, line 101: Have you decided to drop the multiple solutions considered pros & cons, and have the final decision documented?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_based_step.py
```diff
@@ -31,10 +30,21 @@ class NotebookStep(Step):
...
+        self,
+        name,
+        type,
+        runner_func,
+        file_path,
+        job_dir,
+        code,
+        module=None,
+        parent=None,
+    ):
```
ivakoleva, line 43: l'm looking for parameters documentation, it would be useful I think for code readability

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -46,28 +47,56 @@ class Notebook:
...
+                            notebook_steps.append(step)
+                            context.step_builder.add_step(step)
                         else:
-                            self.python_helper_cells.append(cell)
-            log.debug(
-                f"{len(self.sql_and_run_cells) + len(self.python_helper_cells)} "
-                f"cells with vdk tag were detected!"
-            )
+                            step = NotebookStep(
+                                name="".join(
```
ivakoleva, line 82: nit: step name seems the same in both if-else blocks; maybe calculate it in advance, then reuse it (deduplicate common code)

ivakoleva (approved):
cells.jpeg and working-with_notebook-steps.png LGTM, the overall direction of simplifying the way user interacts with VDK, and the model.
Except png and jpeg, GitHub review does not visualise the gifs

specs/vep-994-jupyter-notebook-integration/README.md
duyguHsnHsn: Yes, they seem to be outdated, since many things change. But I will try to document the final decision (this one) better documented

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -46,28 +47,47 @@ class Notebook:
...
 
-    def __init__(self, file_path: Path):
+    @staticmethod
+    def register_notebook_steps(file_path: Path, context: JobContext):
         try:
-            self.sql_and_run_cells = []
-            self.python_helper_cells = []
-            self.file_path = file_path
+            python_module = imp.new_module("notebook")
+            exec("job_input = 1", python_module.__dict__)
```
tozka, line 54: What is this?

tozka (approved):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
duyguHsnHsn: I do it in order to pass our  job_input variable in the module. I declare it like that then access it via module.job_input = job_input (the second one is ours). I couldn't find another way to pass the "real" job_input, so I use it as that. Open to recommendation on using something else to pass job_input into the module

tozka: Ok. Currently I am not sure. I feel there should be. 

Do add a comment above this snippet of code that explains the reason . 
It's a good idea any place that the reason for doing something is not obvious from the code itself to comment it.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -46,28 +47,47 @@ class Notebook:
...
     """
 
-    def __init__(self, file_path: Path):
+    @staticmethod
+    def register_notebook_steps(file_path: Path, context: JobContext):
         try:
-            self.sql_and_run_cells = []
-            self.python_helper_cells = []
-            self.file_path = file_path
+            python_module = imp.new_module("notebook")
```
murphp15, line 53: I see this warning?

`the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses `

Can you use importlib instead? 
I know nothing about those libraries.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
murphp15, line 54: Can you explain? 
what does job_input = 1 mean?

duyguHsnHsn: see this https://github.com/vmware/versatile-data-kit/pull/1520#discussion_r1081285633 
I will put a comment there

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
duyguHsnHsn: Yes, I know currently in vdk core we use imp, too. If I cannot find any good alternatives I will leave it like that for now, if I find something good, will change it

duyguHsnHsn: I changed it with importlib

PR #1521, opened 2023-01-18
vdk-spec: summary, glossary, motivation

murphp15:
# Why
I want to merge summary glossary and motivation first and tackle the next sections later. 

### DIFF
--- a/specs/vep-1507-vdk-operations-ui/README.md
+++ b/specs/vep-1507-vdk-operations-ui/README.md
@@ -15,36 +15,25 @@
 - [Alternatives](#alternatives)
 
 ## Summary
-
-<!--
-Short summary of the proposal. It will be used as user-focused
-documentation such as release notes or a (customer facing) development roadmap.
-The tone and content of the `Summary` section should be
-useful for a wide audience.
--->
+At the moment customers can deploy and manage jobs through the VDK CLI tool or through the API's.
+We want to extend on this and create a web frontend for the VDK. 
+The frontend will let customers, create,delete deploy and generally managed nad browse their jobs.
+Frontend is written in an extensible format to easily allow other teams to build extensions.
+To achieve this extensibility the frontend will be built in angular.
 
 ## Glossary
-<!--
-Optional section which defines terms and abbreviations used in the rest of the document.
--->
+data-pipelines: This is the npm project of the bootstrapping codebase it contains all the vdk specific frontend code. for more details please see [README.md](/projects/frontend/data-pipelines/README.md)
+shared-components, @vdk/shared: This is a set of components that are used within the data-pipelines project but are generic and could be re-used by other projects [README.md](/projects/frontend/data-pipelines/README.md)
 
 ## Motivation
-<!--
-It tells **why** do we need X?
-Describe why the change is important and the benefits to users.
-Explain the user problem that need to be solved.
--->
+We want to improve the usability of VDK. With a UI customers will be able to use VDK much quicker and much more proficiently because UIs are easier to understand and become familiar with than CLI tools or APIs.
+
 
 ## Requirements and goals
-<!--
-It tells **what** is it trying to achieve?
-List the specific goals (functional and nonfunctional requirements)? How will we
-know that this has succeeded?
-
-Specify non-goals. Clearly, the list of non-goals can't be exhaustive.
-Non-goals are only features, which a contributor can reasonably assume were a goal.
-One example is features that were cut during scoping.
--->
+1. It should be easy to setup and run locally or in a k8s cluster
+2. It should be easy to extend with new features
+3. It should be well tested
+4. It should be well documented
 
 ## High-level design
 
@@ -134,8 +123,3 @@ Consider at least the below topics but you do not need to cover those that are n
 Optionally, describe what are the implementation stories (eventually we'd create github issues out of them).
 -->
 
-## Alternatives
-<!--
-Optionally, describe what alternatives has been considered.
-Keep it short - if needed link to more detailed research document.
--->
--- a/specs/vep-1507-vdk-operations-ui/README.md
+++ b/specs/vep-1507-vdk-operations-ui/README.md
@@ -55,11 +55,18 @@ It tells **why** do we need X?
 Describe why the change is important and the benefits to users.
 Explain the user problem that need to be solved.
 -->
+
+### Goals:
 1. It should be easy to setup and run locally or in a k8s cluster.
 2. It should be easy to extend with new features
    1. Adding a new section should be well documented
    2. We expect programmers who didn't work on the initial build of the project to be able to build a section for their needs in isolated manner
    3. After they have created a section they can easily add it to a deployed instance
+
+
+
+### Non Goals:
+1. Screens for rendering data lineage graphs.
 
 ## High-level design
 
--- a/specs/vep-1507-vdk-operations-ui/README.md
+++ b/specs/vep-1507-vdk-operations-ui/README.md
@@ -38,8 +38,16 @@ data-pipelines: This is the npm project of the bootstrapping codebase it contain
 shared-components, @vdk/shared: This is a set of components that are used within the data-pipelines project but are generic and could be re-used by other projects [README.md](/projects/frontend/data-pipelines/README.md)
 
 ## Motivation
-We want to improve the usability of VDK. With a GUI customers will be able to use VDK much quicker and much more proficiently because GUIs are quicker to understand and become familiar with compared to CLI tools or APIs.
-Adding a GUI now will allow us to build features in the future that would be close to impossible to build without a GUI. An example of this would be rendering DAGs from lineage data.
+One of the core goals of the VDK is to provide a common and efficient framework for implementing, executing, monitoring, and troubleshooting data jobs (and data pipelines).
+
+In a sufficiently complex system, a single data team would need to manage many different data jobs - upgrade them, monitor them, fix incidents and issues, etc. Using CLI or API directly for the purpose of troubleshooting or operating multiple jobs doesn't scale. The CLI is very limited in terms of filtering, searching and visualizing complex information. And both require a higher level of technical skills to use.
+
+In an UI we want user to be able to
+
+For Developers - To be able to manage data jobs that allows for more flexibility - update configuration, check latest version, etc.
+For Operator/Support - ability to detect status of data jobs and easily access troubleshooting tools, as well as the ability to track and see log data jobs
+For team leads to be able to have aggregated view of the status of all jobs how often they fail, why do they fail.
+
 
 ## Requirements and goals
 <!--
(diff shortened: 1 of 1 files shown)


specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
 ## Glossary
-<!--
-Optional section which defines terms and abbreviations used in the rest of the document.
--->
+data-pipelines: This is the npm project of the bootstrapping codebase it contains all the vdk specific frontend code. for more details please see [README.md](/projects/frontend/data-pipelines/README.md)
+shared-components, @vdk/shared: This is a set of components that are used within the data-pipelines project but are generic and could be re-used by other projects [README.md](/projects/frontend/data-pipelines/README.md)
 
 ## Motivation
-<!--
-It tells **why** do we need X?
```
ivakoleva, line 33: I think it's good to keep the comments, so that we are guided by those during populating the VEP, especially in the very early stages.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
-Describe why the change is important and the benefits to users.
-Explain the user problem that need to be solved.
--->
+We want to improve the usability of VDK. With a UI customers will be able to use VDK much quicker and much more proficiently because UIs are easier to understand and become familiar with than CLI tools or APIs.
```
ivakoleva, line 30: May add the GUI is user friendly.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
```
ivakoleva, line 29: UI seems ambiguous. UI is "user interface", and user interface could be console user interface -a narrowed type of user interface. And we already provide console user interface.

ivakoleva, line 29: I think "much quicker" or "much more proficient" depends on the user audience. For example the GUI could be easier for a novice user. Also, GUI is usually utilised by more people than CLI. Yet some people may find it rather quicker to configure and use a CLI. 
So my suggestion is to rather elaborate on the broader audience and user friendliness, instead of generalising the speed of usage.

ivakoleva, line 29: Is there anything the CLI is missing, and the GUI is providing it? 
For example, a graph/chart visualisation is rather very specific for GUI.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
 
-## Requirements and goals
-<!--
-It tells **what** is it trying to achieve?
-List the specific goals (functional and nonfunctional requirements)? How will we
-know that this has succeeded?
```
ivakoleva, line 42: My recommendation here is to provide the use cases. Those seem missing.
We need to write down the Use Cases, those describe the requirements and workflows expected.
The focus may be on happy paths.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
 
-Specify non-goals. Clearly, the list of non-goals can't be exhaustive.
```
ivakoleva, line 44: Is deploying the data job a GUI capability we look for?
We need to be able to complete the initiative, and have a new one - for features that are out of scope.
The VEP could be then further extended, rather slowing down the small atomic steps with endless requirements.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
-Non-goals are only features, which a contributor can reasonably assume were a goal.
-One example is features that were cut during scoping.
--->
+## Requirements and goals
+1. It should be easy to setup and run locally or in a k8s cluster
```
ivakoleva, line 33: not sure this is a requirement. how would you measure if easy or not?? Easy for whom?
We should have the operational setup as a part of the CI/CD, and it should be functioning properly (deploying and verifying the GUI component).

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
+2. It should be easy to extend with new features
```
ivakoleva, line 34: what is this extensibility? Sounds very generic. How do you measure if extensible or not, what's the criteria.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
+3. It should be well tested
```
ivakoleva, line 35: what is well tested? does any other component specify explicitly anything about testing? Or is it a general VDK repository requirement, that applies to any component implicitly?
We may as well drop this requirement, I would rather expect for every VDK component to be in fact well tested.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
+4. It should be well documented
```
ivakoleva, line 36: don't understand why mentioning this explicitly. Do we mention any documentation requirements for other VDK components explicitly? We may presume it is well documented, matching the project documentation guidelines.
CONTRIBUTING.md and root documents would apply to any component.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
 
 ## High-level design
```
ivakoleva, line 39: we need to provide information on what the component is design-wise. I believe we already have the information, we need to summarise it.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -133,9 +122,3 @@ Consider at least the below topics but you do not need to cover those that are n
 <!--
 Optionally, describe what are the implementation stories (eventually we'd create github issues out of them).
 -->
-
-## Alternatives
-<!--
-Optionally, describe what alternatives has been considered.
-Keep it short - if needed link to more detailed research document.
--->
```
ivakoleva, line 141: nit: here is a good place to elaborate on code quality and security scanners, and their adoption.

ivakoleva:
> I want to merge summary glossary and motivation first and tackle the next sections later.

In that case, please split the other section changes in separate PRs?
Also, ignore the comment on missing design for example, it is meant to remind about next steps. Shouldn't be a blocker for this PR.

specs/vep-1507-vdk-operations-ui/README.md
murphp15: I would rather do it in its own PR. no point in trying to include everything in one PR.

specs/vep-1507-vdk-operations-ui/README.md
murphp15: Alternative security scanners? 

I would think that is too low level. 
I would think alternatives here would be auto generated GUI from swagger docs etc... 

e.g alternatives to having a frontend GUI as opposed to alternatives for low level implementation details.

specs/vep-1507-vdk-operations-ui/README.md
murphp15: I don't think extensibility is generic for frontends.
Most of the time it wouldn't be trivial to add a new screen to an existing GUI. 

How I would measure.
1. Is adding a new screen documented?
2. Do we expect programmers who didn't work on the initial build of the project to be able to build a section for their needs in  isolated manner
3. After they have created a section can they easily add it to a deployed instance?

specs/vep-1507-vdk-operations-ui/README.md
ivakoleva: I find the comments useful. I am easily reminded what the section is meant for. For example, VEP 994 on Jupyter notebook integration keeps its comments and focus. I like this approach

specs/vep-1507-vdk-operations-ui/README.md
ivakoleva: My suggestion here is to postpone the extensibility statements from the VEP - since it's still an open question, and it would require a dedicated focus to resolve it in my opinion.
I think the Use Cases should rather be a priority at this early point.

specs/vep-1507-vdk-operations-ui/README.md
ivakoleva: So, what you say is that the Frontend alternatives are Control Service API and VDK CLI (API client), is that so?

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -15,36 +15,25 @@
...
-
-<!--
-Short summary of the proposal. It will be used as user-focused
-documentation such as release notes or a (customer facing) development roadmap.
-The tone and content of the `Summary` section should be
-useful for a wide audience.
--->
+At the moment customers can deploy and manage jobs through the VDK CLI tool or through the API's.
+We want to extend on this and create a web frontend for the VDK.
+The frontend will let customers, create,delete deploy and generally managed nad browse their jobs.
```
ivakoleva, line 20: not sure about 20, could you please provide the Frontend Use Cases we want to provide for the users?

specs/vep-1507-vdk-operations-ui/README.md
tozka: One of the core goals of the VDK is to provide a common and efficient framework for implementing, executing, monitoring, and troubleshooting data jobs (and data pipelines). 

In a sufficiently complex system, a single data team would need to manage many different data jobs - upgrade them, monitor them, fix incidents and issues, etc. Using CLI or API directly for the purpose of troubleshooting or operating multiple jobs doesn't scale. The CLI is very limited in terms of filtering, searching and visualizing complex information. And both require a higher level of technical skills to use. 

In an UI we want user to be able to 
* For Developers - To be able to manage data jobs that allows for more flexibility  - update configuration, check latest version, etc. 
* For Operator/Support - ability to detect status of data jobs and easily access troubleshooting tools, as well as the ability to track and see log data jobs
* For team leads to be able to have aggregated view of the status of all jobs how often they fail, why do they fail.

tozka (approved):
I think this looks good to me. 
I suggested how to extend the motivation section and should address Iva's comments.

ivakoleva (approved):

specs/vep-1507-vdk-operations-ui/README.md
murphp15: Added back in.

specs/vep-1507-vdk-operations-ui/README.md
murphp15: @ivakoleva  I added details below. 
Can you read and let me know what you think!

specs/vep-1507-vdk-operations-ui/README.md
murphp15: There are now use cases in the motivation section.

PR #1522, opened 2023-01-18
vdk-control-service: fix possible NPE

dakodakov:
What?
Fix a possible NPE which was introduced recently.

Signed-off-by: Dako Dakov ddakov@vmware.com

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
@@ -62,7 +62,7 @@ public class KerberosCredentialsRepository implements CredentialsRepository {
             "Missing required dependencies. "
                 + " kadmin are missing or are not found on PATH."
                 + "Possible reason: "
-                + result.getOutput().getUTF8()
+                + (result != null ? result.getOutput().getUTF8() : "")
                 + "Without those credential management will not work. ");
       }
       var krbConf = System.getenv().getOrDefault("KRB5_CONFIG", "/etc/krb5.conf");


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
```diff
@@ -62,7 +62,7 @@ private void checkDependencies() {
             "Missing required dependencies. "
                 + " kadmin are missing or are not found on PATH."
                 + "Possible reason: "
-                + result.getOutput().getUTF8()
+                + (result != null ? result.getOutput().getUTF8() : "")
```
murphp15, line 65: can getOutput return null?

murphp15 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
dakodakov: No... It can throw an IllegalStateException, but it cannot return a null.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
```diff
@@ -62,7 +62,7 @@ private void checkDependencies() {
             "Missing required dependencies. "
                 + " kadmin are missing or are not found on PATH."
                 + "Possible reason: "
```
ivakoleva, line 64:
```
if (result == null || result.getExitValue() != 0) {
        log.error(
             "Missing required dependencies. "
                 + " kadmin are missing or are not found on PATH."
                 + "Possible reason: "
                 + result.getOutput().getUTF8()
                 + (result != null ? result.getOutput().getUTF8() : "")
```
So in case of result == null we print an empty-reasoned error message?

ivakoleva (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/KerberosCredentialsRepository.java
dakodakov: yes, as we don't have a valid reason.

PR #1527, opened 2023-01-20
[vdk-plugins] vdk-meta-jobs: Fix TimeoutErrors in start_job

doks5:
Currently, if the network is unstable while executions of data jobs are started by a meta-job, timeout exceptions may occur, which are not properly handled.

This change improves the robustness of the vdk-meta-jobs plugin by introducing logic to handle urllib3.exceptions.TimeoutError cases, which may sporadically occur when the network or the Control Service is unstable.

The new logic adds a retry mechanism to `start_job()` in the TrackingDataJobExecutor class through new method called `start_new_job_execution()`. The idea is that the network instability mey be temporary or even an one-off event, and the meta-job should not fail. In addition, a new api call is introduced to get the list of all recent executions and filter out everything but the latest execution. This is needed as, even though a timeout error has been raised, that does not mean the operation was unsuccessful. In fact, in most cases, a data job execution has been started, and the timeout has been caused by the response from the Control Service not arriving in time, thus starting yet another execution might actually be undesirable.

Testing done: Unit and Integration tests


projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
```diff
@@ -39,6 +42,18 @@ def details_job(self, job_name: str, team_name: str, execution_id: str) -> dict:
         """
         pass
 
+    @abc.abstractmethod
```
murphp15, line 45: I would probably say that they are sorted such that the last element in the list is the latest execution.
That seems counter intuitive to me. I would expect for the 0th element to be the last. execution.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
```diff
@@ -127,3 +133,78 @@ def get_all_jobs(self):
...
+                 executions are available.
+        """
+        latest_execution_id = None
+        executions_list = self._executor.job_executions_list(
+            job_name=job_name, team_name=team
+        )
+
+        if executions_list:
+            # We need only the latest execution which is the last element of the list
+            latest_execution = executions_list[-1]
```
murphp15, line 208: Do you really need to create the `latest_execution` variable? can you not just do `executions_list[-1].id`

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
```diff
@@ -127,3 +133,78 @@ def get_all_jobs(self):
...
+            try:
+                execution_id = self._executor.start_job(job_name, team_name)
+                return execution_id
+            except TimeoutError as e:
+                log.info(
+                    f"A timeout exception occurred while starting the {job_name} data job. "
+                    f"Exception was: {e}"
+                )
+
+            if not execution_id:
```
murphp15, line 173: I don't see how this is not always true? 
Because if it is set then it is retuned above on line 166?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
```diff
@@ -127,3 +133,78 @@ def get_all_jobs(self):
...
+        """
+        current_retries = 0
+        execution_id = None
+
+        latest_available_execution_id = self.get_latest_available_execution_id(
+            job_name=job_name, team=team_name
+        )
+
+        while current_retries < ALLOWED_RETRIES:
+            try:
```
murphp15, line 164: `

            try:
                execution_id = self._executor.start_job(job_name, team_name)
                return execution_id
            except TimeoutError as e:
                log.info(
                    f"A timeout exception occurred while starting the {job_name} data job. "
                    f"Exception was: {e}"
                )
                execution_id = self.get_latest_available_execution_id(
                    job_name=job_name, team=team_name
                )
                if execution_id and execution_id != latest_available_execution_id:
                     return execution_id
                else:
                     current_retries += 1
                time.sleep(SLEEP_TIME)  # Sleep for 10 seconds before re-try.
`

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -135,6 +136,17 @@ def get_job_execution_status(self, execution_id: str) -> DataJobExecution:
         )
         return job_execution
 
+    def get_job_executions(self) -> Optional[List[DataJobExecution]]:
```
murphp15, line 139: again I think the docs need to declare the sorting order of the returned executions.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
doks5: Updated the doc. Tested the Control Service API with different jobs and it constantly returns the most recent execution as the last element of the list.
Didn't want to do reverse list operation as it would have been yet another operation which seemed unnecessary.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
doks5: Oopsie, good catch! It was not necessary. Fixed, thanks

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
doks5: Good point. Fixed, thank you for the review

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
doks5: Updated doc. For more information, see https://github.com/vmware/versatile-data-kit/pull/1527#discussion_r1083932427

dakodakov (approved):

PR #1528, opened 2023-01-20
vdk-jupyter: improve error handling on the UI

duyguHsnHsn:
What: 
This PR introduces a new way to handle errors in the Jupyter UI, using a  function called showErrorMessage from jupyterlab/apputils ( you can see https://jupyterlab.readthedocs.io/en/stable/api/modules/apputils.html )

Why: Up to now we used to console log the errors we et on the UI (mainly to check the behaviour of the extension).

Testing: Tested the dialogs manually 

PS: test for the whole UI are u to be introduced soon


murphp15:
Can you include a screenshot of what it looks like?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx
```diff
@@ -52,7 +55,7 @@ export function updateVDKMenu(commands: CommandRegistry) {
                 body: <DeleteJobDialog jobName={defaultJobName} jobTeam={sessionStorage.getItem("delete-job-team")!}></DeleteJobDialog>,
                 buttons: [Dialog.okButton(), Dialog.cancelButton()],
             }).then(async result => {
-                if (!result.value) {
+                if (result.button.accept) {
                     let bodyMessage = 'Do you really want to delete the job with name ' + sessionStorage.getItem("delete-job-name") + " from " + sessionStorage.getItem("delete-job-rest-api-url") + "?"
                     showDialog({
```
murphp15, line 60: I find it every odd that you mix async/awaits with then/catch. 
I think you should try use async/await everywhere.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -1,94 +1,147 @@
...
- /**
-    * Sent a GET request to the server to get current working directory
-    */
-export function getCurrentPathRequest(){
-    requestAPI<any>('run', {
-        method: 'GET',
+/**
+ * Sent a GET request to the server to get current working directory
+ */
+export function getCurrentPathRequest() {
```
murphp15, line 11: That looks like a race condition to me. 
I think you need to mark this as async and await the result

duyguHsnHsn:
This is an example error for downloading a job
Screenshot 2023-01-23 at 12 30 34

murphp15 (approved):
Looks good, really good work!

tozka (approved):

Issue #1526, opened 2023-01-19
Upgrade to newer kerberos libraries 


murphp15:
**What is the feature request? What problem does it solve?**
At the moment we use extremely old libraries for kerberous auth. 
https://mvnrepository.com/artifact/org.springframework.security.kerberos/spring-security-kerberos-web/1.0.1.RELEASE

It hasn't been update in 8 years. 
It depends on classes in the JDK that aren't easily accessible in JDK 17 onwards.

People online say that we should migrate to apache kirby: https://stackoverflow.com/a/69479644/6401281

**Suggested solution**
Migrate to apache Kirby or a similar framework (https://github.com/bedrin/kerb4j)

tozka:
It does seem we have 2 optoins 
A) Either adopt the latest version of spring security kerberos web by compiling and releasing it ourselves. 
B) Adopt https://github.com/bedrin/kerb4j . It does have spring integration and although there are not updates in last year . It's still more recently updated than the spring web and also the maintainer has stated that there are not major issues. 

So I'd recommend going with kerb4j. I haven't tried it out yet to see how seamless and easy the integration is (hopefully just renaming a few packages and upgrade gradle manifest). But at worst case we. should be able to do it in 2 or 3 days even if we hit some issues.

sabadzhiev:
Triaged. Closed as being already resolved.

PR #1539, opened 2023-01-23
control-service: remove old kerberous test dependency

murphp15:
# Why
This dependency was preventing us from migrating to the newest version of java. 
It helps in eventually tackling this ticket also: https://github.com/vmware/versatile-data-kit/issues/1526 

# What 
Instead of using MiniKdc provided by spring I use simpleKdcServer which is provided by one of the underlying libraries. 
In this PR I also clean up the object lifecycle a lot too. 
Previously we were explicitly wiring some beans together using setters. 
I removed this as it is an anti pattern. 

# How was this tested?
Its only changes to tests and so I just run integration and unit tests.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/JobCredentialsService.java
```diff
@@ -22,6 +22,8 @@
 import java.util.Collections;
 import java.util.Optional;
 
+import static graphql.Assert.assertTrue;
```
tozka, line 25: That looks worng. Why JobCredentialService depends on anything in graphql package ?

murphp15, line 25: Not sure if ppl want me to avoid using dependencies like this?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/JobCredentialsService.java
```diff
@@ -54,11 +56,13 @@ public void createJobCredentials(String jobName) {
     String principal = getJobPrincipalName(jobName);
     File keytabFile;
     try {
+      // create a temp file name by creating a temp file and then deleteing the actual file.
       keytabFile = File.createTempFile(principal, ".keytab");
+      assertTrue(keytabFile.delete());
```
tozka, line 61: I assume this raises an AssertionError which is not good practice in production code (though they are arguments for both sides). In this project we avoid them. 

Logging a warning message may be more suitable here ?

tozka (changes requested):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/JobCredentialsService.java
murphp15: Change made. 
I think a log line would be excessive as it should never happen.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/JobCredentialsService.java
tozka: Yes. Definitely. I am not sure how successful But I am trying to keep each "service" - e.g Execution Service , Data Jobs Service, Deployment Servie decoupled from each other and independent. 

That also means dependencies dedicated for "Data Jobs Service" (like graphql) should not be used outside. Now without proper compile time checks, this is not easy to do so we need to rely on reviews for that. 

But even as a general rule. The less things any module (be it class, package, namespace) depends , the better.

murphp15: Yeah I agree.

mivanov1988 (approved):

PR #1540, opened 2023-01-23
vdk-test-utils: add cli_assert_output_contains

tozka:
We may need to verify that something exists in output or some other verification. In the same time we'd like to make use of cli_assert more deetailed output in case of an error (prints output logs, thread dump of exception).

This is extending `cli_assert` method to support passing custom message. And adding new method `cli_assert_output_contains` which would test if output string contains substring. As this is frequently used verification method in a lot of plugins.

Testing Done: new unit tests


murphp15 (approved):
Looks good

projects/vdk-plugins/vdk-test-utils/src/vdk/plugin/test_utils/util_funcs.py
```diff
@@ -43,18 +43,41 @@
...
+
+def cli_assert_output_contains(expected_substring: str, result: Result) -> None:
+    """
+    Check if the output of a CLI run (logs or stdout) contains the passed expected substring.
+    If not error is raised with debug details about the CLI run.
+    """
+    cli_assert(
+        expected_substring in result.output,
+        result,
+        f"result output does not contains {expected_substring}",
```
doks5, line 80:
```suggestion
        f"result output does not contain {expected_substring}",
```

doks5 (approved):

PR #1545, opened 2023-01-24
control-service: only delete file in test path

murphp15:
# Why
In my previous PR I merged code which deletes an empty keytab file, so that we keep only the filename. 
This is the desired behaviour for the test kdc server as it expects no file to be at the file location. 
However the production path expects there to be an empty file there. 

# What 
Push the delete logic into the test impl so that we aren't modifying the production code path. 

# How has this been tested
entire deployment pipeline

# Open questions
Should more of the pipeline run on branches as it would have caught this issue?


tozka:
How did you catch the regression ? Did vdk-heartbeat post deployment test fail? It seems so. That's good. That's the purpose of the test to catch issues in real deployment. Or was some other way ?

tozka (approved):

tozka:
The integration test is now failing btw.

murphp15:
> How did you catch the regression ? Did vdk-heartbeat post deployment test fail? It seems so. That's good. That's the purpose of the test to catch issues in real deployment. Or was some other way ?

Yeah was caught by the heartbeart test

mivanov1988 (approved):

PR #1548, opened 2023-01-24
Control-service: Remove dependency on old docker image which is not needed

murphp15:
# Why
The old docker image with the aws-iam-authenticator is giving us a lot of trouble. we don't need it. 
Removing it, everything works well. 
# What
Change the image we are using

# How has this been tested?
At the commit [c483219](https://github.com/vmware/versatile-data-kit/pull/1548/commits/c4832191a3de4f883b5ed4760e17c03e2005e69b) I pushed extra changes to run the actual deployment to make sure it worked.  Then reverted them when everything looked good.


projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -225,7 +225,7 @@ control_service_publish_api_client:
 # Go to the namespace -> StorageClass
 control_service_deploy_testing_data_pipelines:
   stage: pre_release
-  image: proum/aws-iam-authenticator
+  image: intellibank/aws-iam-authenticator:0.5.12
```
tozka, line 228: Yes, I think you need some testing . You can run it with docker run and see if the script would work at least teh AWS authentication part - https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/cicd/deploy-testing-pipelines-service.sh ( which really is KUBECONFIG and kuberenetes commands really). 

Just now running it locally , itseems there's no shell installed - `docker run --entrypoint sh --network="host" -v /Users/aivanov/dev/workspace/github:/repo -it intellibank/aws-iam-authenticator:0.5.12` 

While the previous did have "sh" installed. Without shell the below script seems unlikely to be able to run.

tozka: I don't think it's big deal to install awi-iam-authenticator ourself . It's one line curl command - https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html

mivanov1988 (approved):

mivanov1988:
LGTM!

tozka (approved):
Yes, The authenticator is no longer needed: 

https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html

> If you're running the AWS CLI version 1.16.156 or later, then you don't need to install the authenticator. Instead, you can use the aws eks get-token command. For more information, see Create kubeconfig file manually.

tozka (approved):

PR #1550, opened 2023-01-25
control-service: Fix release step in pipeline

murphp15:
# Why
The release step fails with the newer docker image version. But works with the previous setup.We need the newer version for java 17 support for other steps. 
# What
Revert the changes for that step only. 
# How has this been tested?
Tested on the previous commit and changes reverted to merge.


tozka:
Why does it fail ?  Can you link and copy paste the failure? Can we fix it? 


But first let's merge this as I understand the main pipeline is red and fixing the main branch CICD pipeline this is highest priority

tozka (approved):

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -287,7 +287,13 @@ control_service_release:
 
 # TODO: automatically rebuild image on vdk-heartbeat change
 control_service_vdk_heartbeat_release:
-  extends: .images:dind
+  image: docker:19.03.15
```
tozka, line 290: nit: write a comment that are are using older version because ...

murphp15:
An example failure https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3658293978

projects/control-service/cicd/.gitlab-ci.yml
murphp15: Done

Issue #1254, opened 2022-10-20
Have nightly builds for vdk plugins, integration tests etc.


murphp15:
**What is the feature request? What problem does it solve?**
Nightly builds of the vdk plugins and integration tests would help ensure the stability of the project.

**Suggested solution**
We should have nightly builds of the vdk plugins and integration tests.

tozka:
I think we can schedule a [Gitlab Scheduled Pipeline](https://docs.gitlab.com/ee/ci/pipelines/schedules.html). 
We can mark certain Gitlab Jobs using https://docs.gitlab.com/ee/ci/jobs/job_control.html#run-jobs-for-scheduled-pipelines

I do not think that we should make nightly releases. It would make it harder for us to determine what is new in last month (e.g what was released) and it's a bit misleading that there are changes to a component when there are not.

PR #1551, opened 2023-01-26
vdk-cicd: during a scheduled run publish_artifacts and release shouldn't run

murphp15:
# Why
We want to be able to run scheduled jobs. 
But as part of the scheduled jobs we want to skip `publish_artifact` jobs and also `release` jobs.

This change should not have any impact on the pipeline and I will start running the scheduled jobs once we see the pipeline is still green for merge triggered builds. 

In the future I might extract the rule into a shared abstract job. But that will be a later PR
# What
Add a guard on all `publish_artifact` jobs and also `release` jobs so that they won't run if it is triggered by a scheduled operation. 
I also upgraded helm if we are testing everything we should try upgrade that too. The hope here is that it fixes https://github.com/vmware/versatile-data-kit/issues/1252. 

# How has this been tested
Not much, mostly visual inspection. 

closes https://github.com/vmware/versatile-data-kit/issues/1254

### DIFF
--- a/projects/control-service/cicd/.gitlab-ci.yml
+++ b/projects/control-service/cicd/.gitlab-ci.yml
@@ -151,6 +151,8 @@ control_service_publish_job_base_image-secure:
     - bash -ex ./publish-job-base.sh
   retry: !reference [.control_service_retry, retry_options]
   rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule"'
+      when: never
     - if: '$CI_COMMIT_BRANCH == "main"'
       changes:
         - projects/control-service/projects/job-base-image-secure/**/*
@@ -165,6 +167,8 @@ control_service_publish_job_builder_image:
     - bash -ex ./publish-vdk-job-builder.sh
   retry: !reference [.control_service_retry, retry_options]
   rules:
+    - if: '$CI_PIPELINE_SOURCE == "schedule"'
+      when: never
     - if: '$CI_COMMIT_BRANCH == "main"'
       changes:
         - projects/control-service/projects/job-builder/version.txt
--- a/projects/control-service/cicd/.gitlab-ci.yml
+++ b/projects/control-service/cicd/.gitlab-ci.yml
@@ -212,13 +212,13 @@ control_service_publish_image:
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule"'
       when: never
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: [ projects/control-service/projects/helm_charts/pipelines-control-service/version.txt ]
-#      when: never
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: *control_service_code_change_locations
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: *control_service_helm_change_locations
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: [ projects/control-service/projects/helm_charts/pipelines-control-service/version.txt ]
+      when: never
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: *control_service_code_change_locations
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: *control_service_helm_change_locations
 
 control_service_publish_api_client:
   image: docker:20.10.22
@@ -303,10 +303,10 @@ control_service_release:
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule"'
       when: never
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: *control_service_code_change_locations
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: *control_service_helm_change_locations
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: *control_service_code_change_locations
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: *control_service_helm_change_locations
 
 # TODO: automatically rebuild image on vdk-heartbeat change
 # We are using an old version of docker here because we get error on the newest version which at the time was tag=20.10.22
@@ -332,7 +332,7 @@ control_service_vdk_heartbeat_release:
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule"'
       when: never
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: *control_service_code_change_locations
-#    - if: '$CI_COMMIT_BRANCH == "main"'
-#      changes: *control_service_helm_change_locations
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: *control_service_code_change_locations
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      changes: *control_service_helm_change_locations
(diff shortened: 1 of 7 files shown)


tozka:
Please assign the https://github.com/vmware/versatile-data-kit/issues/1254 to yourself. And linked it in teh ticket.

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -150,15 +150,10 @@ control_service_publish_job_base_image-secure:
     - export VERSION_TAG="1.$CI_PIPELINE_ID"
     - bash -ex ./publish-job-base.sh
   retry: !reference [.control_service_retry, retry_options]
-  only:
```
murphp15, line 153: @tozka  I don't think we need the except clause here because it is not in the folder where the changes must happen. 
Do you agree?

tozka: Do you mean this except clause 
```
except:
    changes:
```
Then yes, we do not need it.

Yes, we do not need it. We had it at times when Control Service releases were manual. Since we did not want to trigger everything just to make new release . We had "release" pipelines that only ran the release job. 

As releases of Control Service are automated , this except is no needed at all.

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -107,7 +107,7 @@ control_service_integration_test:
 control_service_test_helm_chart:
   stage: build
   script:
-    - export DESIRED_VERSION=v3.6.3 # helm version 3.6.3
+    - export DESIRED_VERSION=v3.11.0 # helm version 3.11.0
```
ivakoleva, line 110: seems the PR merges multiple changes
(however cannot tell, since there is no exact PR description of what had changed exactly, and why)

ivakoleva (commented):
could explain in the PR description what are the changes expected? 
for example, I am wondering why the vdk-heartbeat publish/release stages are handled, and the vdk-control-cli are not?

projects/control-service/cicd/.gitlab-ci.yml
ivakoleva: haven't we split that in two already in PR https://github.com/vmware/versatile-data-kit/pull/1547? why not merging it first in case a prerequisite?

murphp15:
> could explain in the PR description what are the changes expected? for example, I am wondering why the vdk-heartbeat publish/release stages are handled, and the vdk-control-cli are not?

Fixed

ivakoleva (commented):
Still don't understand why some gitlab ci yaml files are changed, and some are not.....for example https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/.gitlab-ci.yml#L86 ?

projects/vdk-control-cli/.gitlab-ci.yml
```diff
@@ -84,7 +84,8 @@ vdk-control-cli-release:
     - pip install -U pip setuptools wheel twine
     - python setup.py sdist --formats=gztar
     - twine upload --repository-url "$PIP_REPO_UPLOAD_URL" -u "$PIP_REPO_UPLOAD_USER_NAME" -p "$PIP_REPO_UPLOAD_USER_PASSWORD" dist/*tar.gz --verbose
-  only:
-    refs:
-      - main
-    changes: *vdk_control_cli_changes_locations
+  rules:
```
ivakoleva, line 87: have we covered all the release/publish stages of all artifacts at that point?

ivakoleva (approved):

ivakoleva:
for convenience, you can also link the gitlab schedule created, to save time (I went to find it on my own)

projects/vdk-control-cli/.gitlab-ci.yml
murphp15: there is no stage called publish, there is `publish_artifacts`, `release` and  `release_image`. and yeah they are all covered.

murphp15:
> for convenience, you can also link the gitlab schedule created, to save time (I went to find it on my own)

the schedule is not actually created yet. 
that schedule you are looking at it for a branch and it does a pretty bad job of testing it. 
I will create the schedule when I know I have not introduced regressions in the current builds.

murphp15:
> Still don't understand why some gitlab ci yaml files are changed, and some are not.....for example https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-control-cli/.gitlab-ci.yml#L86 ?

It is changed, but I only changed it after you pointed it out. 
Do you see other examples where I am not doing it?

projects/control-service/cicd/.gitlab-ci.yml
murphp15: I am going to delete that PR. 
I never finished that PR out because it was hard to figure out how to effectively test it.

tozka:
Is it possible to configure the rule to be "opt-in" - run on schedule only the jobs marked as such. 
Instead of opt-out as it's now (excluding jobs) ?

With opt-in , it's easier to start with smaller set of jobs and add more incrementally until we are happy

murphp15:
> Is it possible to configure the rule to be "opt-in" - run on schedule only the jobs marked as such. Instead of opt-out as it's now (excluding jobs) ?
> 
> With opt-in , it's easier to start with smaller set of jobs and add more incrementally until we are happy

IMO opt out is better so we are testing more by default. 

I think the problem with current setup with both my solution and your suggested solution is that it touches soo many lines of code. 
I think we should try make an abstract job for each stage. all jobs in that stage should extend it. 
That way it is super easy to skip an entire stage. 

However I only want to try that after I have identified all the edge cases etc..  


Let me know what you think?

projects/vdk-control-cli/.gitlab-ci.yml
ivakoleva: Yes, I went looking for it, and found vdk-core gitlab ci yaml file for example is also not patched.

I managed to do it by fetching all the latest changes in the project, then click find in projects/ dir while narrowing the file types to .y*ml, and looking for each occurrence of "publish" and "release" words.

Attaching a screenshot of how I did the check:
Screenshot 2023-01-26 at 13 57 36

murphp15: Oh sorry! thanks!
Should be all fixed now

tozka:
> > Is it possible to configure the rule to be "opt-in" - run on schedule only the jobs marked as such. Instead of opt-out as it's now (excluding jobs) ?
> > With opt-in , it's easier to start with smaller set of jobs and add more incrementally until we are happy
> 
> IMO opt out is better so we are testing more by default.
> 
> I think the problem with current setup with both my solution and your suggested solution is that it touches soo many lines of code. I think we should try make an abstract job for each stage. all jobs in that stage should extend it. That way it is super easy to skip an entire stage.
> 
> However I only want to try that after I have identified all the edge cases etc..
> 
> Let me know what you think?


With opt-out if we add a new component/plugin build job, it would be automatically included in the scheduled nightly build. Which is a good thing. 

 But if we add a new release job then it would be automatically included which is not what we want. 

But if we can extend a stage, as you suggest, then we can opt out "release" stage. Unless somebody releases in the wrong stages, there's no problem with wrong jobs being included.

 So I like this.. 

OK. Let's go with the proposed plan. First let's identify the  what jobs need to be included and what not. 
Then we can refactor to support inheriting abstract job for each stage. Which would likely be useful for other purposes as well.

projects/control-service/cicd/.gitlab-ci.yml
tozka: But I'd prefer if you remove them in separate PR if possible/feasible.

tozka (approved):
Looks good to me. 

Since there's some risk. But I'd suggest that You do not merge it soon before clocking out and monitor the pipeline afterward. It's very unlikely that the new rules would break something but you never know

projects/control-service/cicd/.gitlab-ci.yml
murphp15: DOne

PR #1555, opened 2023-01-26
job-builder: fix ci/cd steps

mivanov1988:
# Why
The job builder steps failed due to the docker version - https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3666206273 .
# What
Revert the docker version for the job builder steps.
# Testing done
CI/CD


dakodakov (approved):

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -121,7 +121,12 @@ control_service_test_helm_chart:
       - projects/control-service/projects/helm_charts/pipelines-control-service/version.txt
 
 control_service_publish_job_base_image:
-  extends: .images:dind
+  image: docker:19.03.15
```
ivakoleva, line 124: 20.10.23 (in favour of [20.10.22](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/cicd/.gitlab-ci.yml#L8)) does not solve the issue? what was the issue?

projects/control-service/projects/job-builder-secure/version.txt
```diff
@@ -1 +1 @@
-1.0.0
+1.1.0
```
ivakoleva, line 1: nit: do we still use the version.txt mechanism or is it redundant?

ivakoleva (approved):

tozka:
Why do we need older version? What is the difference between the new and the old?

projects/control-service/projects/job-builder-secure/version.txt
mivanov1988: we use it for job-builder

mivanov1988:
> Why do we need older version? What is the difference between the new and the old?

Just reverted the version in order to unblock the pipeline - more details here - https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3666206273.

projects/control-service/cicd/.gitlab-ci.yml
mivanov1988: Just reverted the version in order to unblock the pipeline - more details here - https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3666206273.

mivanov1988:
> Why do we need older version? What is the difference between the new and the old?

I don't know what is the difference between the new and the old - @murphp15 do you know?

PR #1558, opened 2023-01-26
control-service: Expose v1CronJob flag as env var

doks5:
Currently, if a user decides to use the `batch/v1` version of the kubernetes cronjob API, they need to set the `datajobs.control.k8s.k8sSupportsV1CronJob` application property to true. This, however, is a code change and applies to all deployment environments, which means that if the user has separate test and production setups, they cannot switch first one of the environments and then the second, they need to switch them both at the same time.

This change introduces an environment variable `DATAJOBS_SUPPORT_V1CRONJOB`, and adds support for setting it through the helm chart. The change, therefore, allows for the cronjob API to be switched per environment.

Testing Done: Tested locally against a local kind cluster.

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
@@ -252,6 +252,10 @@ spec:
             - name: DATAJOBS_BUILDER_REGISTRY_SECRET
               value: {{ template "pipelines-control-service.fullname" . }}-dtjb-buldr-rgstry-scrt
             {{- end }}
+            {{- if .Values.controlK8sK8sSupportsV1CronJob }}
+            - name: DATAJOBS_SUPPORT_V1CRONJOB
+            value: "{{ .Values.controlK8sK8sSupportsV1CronJob }}"
+            {{- end }}
 
 
 {{- if .Values.extraVars }}
--- a/projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
+++ b/projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
@@ -100,7 +100,7 @@ logging.level.io.swagger.models.parameters.AbstractSerializableParameter=ERROR
 
 
 datajobs.control.k8s.kubeconfig=${HOME}/.kube/config
-datajobs.control.k8s.k8sSupportsV1CronJob=false
+datajobs.control.k8s.k8sSupportsV1CronJob=${DATAJOBS_SUPPORT_V1CRONJOB:false}
 datajobs.control.k8s.jobTTLAfterFinishedSeconds=3600
 # Location to a K8s cronjob yaml file which will be used as a template
 # for all data jobs. If the location is missing, the default internal


projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -252,6 +252,10 @@ spec:
             - name: DATAJOBS_BUILDER_REGISTRY_SECRET
               value: {{ template "pipelines-control-service.fullname" . }}-dtjb-buldr-rgstry-scrt
             {{- end }}
+            {{- if .Values.controlK8sK8sSupportsV1CronJob }}
+            - name: DATAJOBS_SUPPORT_V1CRONJOB
+            value: "{{ .Values.controlK8sK8sSupportsV1CronJob }}"
```
tozka, line 257: If you are adding new values.yaml field `controlK8sK8sSupportsV1CronJob` you need to document it in values.yaml 

The double k8s seem redundant `controlK8sSupportsV1CronJob` is enough.

tozka:
> This, however, is a code change and applies to all deployment environments

Not necessarily . You can pass it as environment variable using [extraEnvVars](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml#L758):



```
extraEnvVars:
   DATAJOBS_CONTROL_K8S_K8SSUPPORTSV1CRONJOB: true
```

doks5:
> > This, however, is a code change and applies to all deployment environments
> 
> Not necessarily . You can pass it as environment variable using [extraEnvVars](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml#L758):
> 
> Spring properties are automatically inferred from env variables (with dots replaces with underscores):
> 
> ```
> extraEnvVars:
>    DATAJOBS_CONTROL_K8S_K8SSUPPORTSV1CRONJOB: true
> ```

Correct. I missed that option, which basically makes the change unnecessary. Thank you

mivanov1988:
> > This, however, is a code change and applies to all deployment environments
> 
> Not necessarily . You can pass it as environment variable using [extraEnvVars](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml#L758):
> 
> Spring properties are automatically inferred from env variables (with dots replaces with underscores):
> 
> ```
> extraEnvVars:
>    DATAJOBS_CONTROL_K8S_K8SSUPPORTSV1CRONJOB: true
> ```

I prefer this property to be exposed and documented in values.yaml so our clients can easily find and use it.

tozka:
> I prefer this property to be exposed and documented in values.yaml so our clients can easily find and use it. What do you think?

Wouldn't it be better to make it default to True? 

Then only users that use some old version of Kubernetes would need it and as time progresses older and older, which should be fewer and fewer as time progresses.

WDYT (What do you think) ?

mivanov1988:
> > I prefer this property to be exposed and documented in values.yaml so our clients can easily find and use it. What do you think?
> 
> Wouldn't it be better to make it default to True?
> 
> Then only users that use some old version of Kubernetes would need it, which should be fewer and fewer as time progresses (as the version would be older and older as time goes)
> 
> WDYT (What do you think) , @mivanov1988 ?

Makes sense.

PR #1562, opened 2023-01-30
control-service: integration tests refactoring

mivanov1988:
# Why
Currently, a lot of integration tests duplicate steps for data job creation, deployment, and execution. Although the DataJobDeploymentExtension exists, it does not provide capabilities for custom data job deployment.

# What
Extended the DataJobDeploymentExtension to support custom data job deployment. Refactored all integration tests that use real data jobs.

# Testing done
CI/CD integration tests

Signed-off-by: Miroslav Ivanov [miroslavi@vmware.com](mailto:miroslavi@vmware.com)


tozka:
Is there Github issue for that? While not always required. If it is planned I strongly encourage having a Github issue where the issue/solution could be discussed earlier.

tozka:
> > Is there Github issue for that? While not always required. If it is planned I strongly encourage having a Github issue where the issue/solution could be discussed earlier.
> 
> There is no GitHub issue for that since it is not planned.

Yes, This is perfect. I encourage people to fix things as they find them and not let them fester when it's feasible/possible. 
And thanks for doing that!

But I want to keep reminding since it's important for planned work to have been discussed and to have clear visibility into the issue and the solutions and the problem which tend to be well provided in a Github issue.

murphp15 (approved):
Looks good! So much duplication removed!

PR #1564, opened 2023-01-30
vdk-jupyter: modify the way we work with notebooks in notebook plugin 

duyguHsnHsn:
Why:
Currently we offer this way of working with SQL and Python cells: 
working-with_notebook-steps

There are the reasons why we need to change it:
* Even though it looks pretty and easy to use, this parsing method is error-prone. Since we read the whole notebook file as a JSON and get the code as string, it becomes really hard to tests the cells which include magic (for example when the cell includes both magic and python code). And vdl-core does not work with iPython.
* Since the users does not use the job_input in the %sql cells this requires them to create new connections to the database by themselves - they cannot use the connections created by VDK
* Since these %sql cells does not include job_input, to make them run with vdk we need to do changes on the cell execution of Jupyter, which will take a lot of time and will be hard to be achieved. Also, changing the cell execution cannot be done via ipython extension, we will need to do modifications using jupyterlab-extension. Having one solution separated into two totally different packages does not seem to be a good solution.

What:
Changes on the way we read the JSON schema of a notebook.
Screenshot 2023-01-30 at 10 44 14

As you can see, in this notebook we use job_input variable to execute a SQL query instead of %sql magic.
* It does exactly the same thing as the solution above but it is a safer option to chose because:
* Since the SQL code directly turn into python code which calls for SQL execute, this solution is more error-proof (since everything for VDK will be python code, we will not need to parse SQL code by hand, if there is a problem with the SQL statement we will get an exception and throw it) 
* The users use the job_input variable directly, so they can use the VDK connections directly, too
All the solution remains in one place instead of being separated into different packages

Testing: Unit tests


tozka (approved):
I agree that the current approach for handling %sql only cells was hacky and we are better for now supporting just Python cells. 

Eventually in the future we can add support for SQL cells to be executed by VDK. There are non-hacky solutions to do that. But it's not that great priority. 


I suggest you start wiki page with all kinds of things that are cut now but are future improvements. This will build us a roadmap for this feature.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -65,29 +63,21 @@ def register_notebook_steps(file_path: Path, context: JobContext):
             for jupyter_cell in content["cells"]:
                 if jupyter_cell["cell_type"] == "code":
                     cell = Cell(jupyter_cell)
-                    if CellUtils.is_vdk_cell(cell):
-                        is_sql = CellUtils.is_sql_cell(cell)
-                        cell_type = TYPE_SQL if is_sql else TYPE_PYTHON
-                        runner_func = (
-                            NotebookStepFuncFactory.run_sql_step
-                            if is_sql
-                            else NotebookStepFuncFactory.run_python_step
```
ivakoleva, line 74: nit: so the future implementation could be 
if sql then transform/decorate to execute in python job_input.execute_query(sql)

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
```diff
@@ -12,21 +12,3 @@ class Cell:
...
-        return "vdk" in cell.tags
-
-    @staticmethod
-    def is_sql_cell(cell: Cell):
-        return cell.source.startswith("%sql")
-
-    @staticmethod
-    def get_cell_code(cell: Cell):
-        if cell.source.startswith("%sql"):
-            code = cell.source.replace("%sql", "")
```
ivakoleva, line 30: nit: if we still plan to have %sql in foreseeable future, we could keep CellUtils and link to VEP's explanation on it; otherwise we need to go back in git log and cherry-pick to bring it back, that's also a way to do it

ivakoleva (approved):
I agree with the proposal where the sql query is executed in the standard way via `job_input.execute_query()`. I think we need to keep this logic, with any kind visualization we decide on in the future.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
tozka: I don't know. I kinda feel it's better to delete it now. Code that is not used is better off deleted in most cases, right?

In the future it may be an entirely different solution. I am thinking of potential integration with SQLAlchemy for example 
IPythonSQL supports providing connection string using [SQLAlchemy URL](http://docs.sqlalchemy.org/en/latest/core/engines.html#database-urls) standard. 

Alternatively, we might just be able to simply set teh db connection directly 
(https://github.com/catherinedevlin/ipython-sql/blob/master/src/sql/connection.py#L60)

ivakoleva: well will the %sql convenient interface change, or should the backing implementation rather change?
In case we want to keep the %sql user interface, then the code could match the VEP description.

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
duyguHsnHsn: Yes, indeed. Tho the problem is not really with executing the query, but with the fact that VDK works with python not ipython and we get the whole cell content as a string, so we need to recognise all the ipython content by hand and leave only the sql query for vdk which is really error-prone, but ofc it can be done in the future. But it seems like leaving it like that (only with python steps) for now is the best option.

PR #1580, opened 2023-02-01
control-service: Cronjob API backwards compatibility

doks5:
Currently there is no backwards compatibility support when a user switches to the V1 Kubernetes Cronjob API (from the currently default V1beta1). This results in situations, where if there are deployed data jobs which use the v1beta1 API, and a switch is made to the V1 API, these jobs are suddenly shown as `NOT DEPLOYED`, and cannot be properly managed.

This change adds backwards compatibility support in the Kubernetes Service, to allow for a switch to V1 Cronjob API in clusters, where V1beta1 API cronjobs are deployed.

Testing Done: Unit and Integration tests (new and existing).

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -492,10 +492,11 @@ public abstract class KubernetesService implements InitializingBean {
    */
   public List<JobDeploymentStatus> readJobDeploymentStatuses() {
     if (getK8sSupportsV1CronJob()) {
-      return Stream.concat(readV1CronJobDeploymentStatuses().stream(), readV1beta1CronJobDeploymentStatuses().stream())
-              .collect(Collectors.toList());
-    }
-    else {
+      return Stream.concat(
+              readV1CronJobDeploymentStatuses().stream(),
+              readV1beta1CronJobDeploymentStatuses().stream())
+          .collect(Collectors.toList());
+    } else {
       return readV1beta1CronJobDeploymentStatuses();
     }
   }
@@ -768,15 +769,15 @@ public abstract class KubernetesService implements InitializingBean {
     log.debug("Listing k8s cron jobs");
     Set<String> v1Set = Collections.emptySet();
 
-    if(getK8sSupportsV1CronJob()) {
+    if (getK8sSupportsV1CronJob()) {
       var v1CronJobs =
-              new BatchV1Api(client)
-                      .listNamespacedCronJob(
-                              namespace, null, null, null, null, null, null, null, null, null, null);
+          new BatchV1Api(client)
+              .listNamespacedCronJob(
+                  namespace, null, null, null, null, null, null, null, null, null, null);
       v1Set =
-              v1CronJobs.getItems().stream()
-                      .map(j -> j.getMetadata().getName())
-                      .collect(Collectors.toSet());
+          v1CronJobs.getItems().stream()
+              .map(j -> j.getMetadata().getName())
+              .collect(Collectors.toSet());
       log.debug("K8s V1 cron jobs: {}", v1Set);
     }
 
@@ -1194,13 +1195,11 @@ public abstract class KubernetesService implements InitializingBean {
     if (getK8sSupportsV1CronJob()) {
       try {
         new BatchV1Api(client)
-                .deleteNamespacedCronJob(name, namespace, null, null, null, null, null, null);
+            .deleteNamespacedCronJob(name, namespace, null, null, null, null, null, null);
         log.debug("Deleted k8s cron job: {}", name);
         return;
       } catch (Exception e) {
-        log.debug(
-                "An exception occurred while trying to delete cron job. Message was: ",
-                e);
+        log.debug("An exception occurred while trying to delete cron job. Message was: ", e);
       }
     }
 
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -61,6 +61,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.util.function.Predicate.not;
 
@@ -490,9 +491,13 @@ public abstract class KubernetesService implements InitializingBean {
    *     data
    */
   public List<JobDeploymentStatus> readJobDeploymentStatuses() {
-    return getK8sSupportsV1CronJob()
-        ? readV1CronJobDeploymentStatuses()
-        : readV1beta1CronJobDeploymentStatuses();
+    if (getK8sSupportsV1CronJob()) {
+      return Stream.concat(readV1CronJobDeploymentStatuses().stream(), readV1beta1CronJobDeploymentStatuses().stream())
+              .collect(Collectors.toList());
+    }
+    else {
+      return readV1beta1CronJobDeploymentStatuses();
+    }
   }
 
   public List<JobDeploymentStatus> readV1beta1CronJobDeploymentStatuses() {
@@ -761,16 +766,30 @@ public abstract class KubernetesService implements InitializingBean {
 
   public Set<String> listCronJobs() throws ApiException {
     log.debug("Listing k8s cron jobs");
-    var cronJobs =
+    Set<String> v1Set = Collections.emptySet();
+
+    if(getK8sSupportsV1CronJob()) {
+      var v1CronJobs =
+              new BatchV1Api(client)
+                      .listNamespacedCronJob(
+                              namespace, null, null, null, null, null, null, null, null, null, null);
+      v1Set =
+              v1CronJobs.getItems().stream()
+                      .map(j -> j.getMetadata().getName())
+                      .collect(Collectors.toSet());
+      log.debug("K8s V1 cron jobs: {}", v1Set);
+    }
+
+    var v1BetaCronJobs =
         new BatchV1beta1Api(client)
             .listNamespacedCronJob(
                 namespace, null, null, null, null, null, null, null, null, null, null);
     var set =
-        cronJobs.getItems().stream()
+        v1BetaCronJobs.getItems().stream()
             .map(j -> j.getMetadata().getName())
             .collect(Collectors.toSet());
-    log.debug("K8s cron jobs: {}", set);
-    return set;
+    log.debug("K8s V1Beta cron jobs: {}", set);
+    return Stream.concat(v1Set.stream(), set.stream()).collect(Collectors.toSet());
   }
 
   public void createCronJob(
@@ -1168,6 +1187,23 @@ public abstract class KubernetesService implements InitializingBean {
 
   public void deleteCronJob(String name) throws ApiException {
     log.debug("Deleting k8s cron job: {}", name);
+
+    // If the V1 Cronjob API is enabled, we try to delete the cronjob with it and exit the method.
+    // If, however, the cronjob cannot be deleted, this means that it might have been created
+    // with the V1Beta1 API, so we need to try again with the beta API.
+    if (getK8sSupportsV1CronJob()) {
+      try {
+        new BatchV1Api(client)
+                .deleteNamespacedCronJob(name, namespace, null, null, null, null, null, null);
+        log.debug("Deleted k8s cron job: {}", name);
+        return;
+      } catch (Exception e) {
+        log.debug(
+                "An exception occurred while trying to delete cron job. Message was: ",
+                e);
+      }
+    }
+
     try {
       new BatchV1beta1Api(client)
           .deleteNamespacedCronJob(name, namespace, null, null, null, null, null, null);
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -85,6 +85,7 @@ public abstract class KubernetesService implements InitializingBean {
   public static final String LABEL_PREFIX = "com.vmware.taurus";
   private static final int WATCH_JOBS_TIMEOUT_SECONDS = 300;
   private static final String K8S_DATA_JOB_TEMPLATE_RESOURCE = "k8s-data-job-template.yaml";
+  private static final String V1_K8S_DATA_JOB_TEMPLATE_RESOURCE = "v1-k8s-data-job-template.yaml";
 
   private static int fromInteger(Integer value) {
     return Optional.ofNullable(value).orElse(0);
@@ -241,7 +242,7 @@ public abstract class KubernetesService implements InitializingBean {
     try {
       if (getK8sSupportsV1CronJob()) {
         loadV1CronjobTemplate(
-            new ClassPathResource("v1-" + K8S_DATA_JOB_TEMPLATE_RESOURCE).getFile());
+            new ClassPathResource(V1_K8S_DATA_JOB_TEMPLATE_RESOURCE).getFile());
       } else {
         loadV1beta1CronjobTemplate(new ClassPathResource(K8S_DATA_JOB_TEMPLATE_RESOURCE).getFile());
       }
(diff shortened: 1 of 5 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -487,9 +488,14 @@ public Optional<JobDeploymentStatus> readV1CronJob(String cronJobName) {
    *     data
    */
   public List<JobDeploymentStatus> readJobDeploymentStatuses() {
-    return getK8sSupportsV1CronJob()
-        ? readV1CronJobDeploymentStatuses()
-        : readV1beta1CronJobDeploymentStatuses();
+    if (getK8sSupportsV1CronJob()) {
```
murphp15, line 491: Why not always just return both?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -758,16 +764,30 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
 
   public Set<String> listCronJobs() throws ApiException {
```
murphp15, line 765: So this function was never implemented properly? 
It always only supported v1Beta?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -1165,6 +1185,21 @@ public void updateV1CronJob(
 
   public void deleteCronJob(String name) throws ApiException {
```
murphp15, line 1186: Again this function wasn't implemented properly to begin with?

murphp15:
Nice spot! I think it looks really good. However I think it would be soo much better if you are able to test it with an integration test using actual jobs.

murphp15:
@doks5  I actually think even if beta1 is enabled we should check for v1 jobs. 
Because if we deploy a few jobs on v1 and need to roll back to betav1 for some unexpected reason then the issue will introduce its self again. 
Plus it should make the code a bit simpler?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: Correct.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: Correct.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -758,16 +764,30 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
 
   public Set<String> listCronJobs() throws ApiException {
     log.debug("Listing k8s cron jobs");
-    var cronJobs =
+    Set<String> v1Set = Collections.emptySet();
```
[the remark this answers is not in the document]
doks5: Renamed, thanks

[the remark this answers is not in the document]
doks5: Doc added, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -1165,6 +1185,21 @@ public void updateV1CronJob(
...
+      try {
+        new BatchV1Api(client)
+            .deleteNamespacedCronJob(name, namespace, null, null, null, null, null, null);
+        log.debug("Deleted k8s cron job: {}", name);
+        return;
+      } catch (Exception e) {
+        log.debug("An exception occurred while trying to delete cron job. Message was: ", e);
+      }
+    }
+
```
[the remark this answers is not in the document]
doks5: Good point. Fixed, thank you

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -758,16 +764,30 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
 
   public Set<String> listCronJobs() throws ApiException {
     log.debug("Listing k8s cron jobs");
-    var cronJobs =
+    Set<String> v1Set = Collections.emptySet();
+
+    if (getK8sSupportsV1CronJob()) {
+      var v1CronJobs =
```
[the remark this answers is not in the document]
doks5: Refactoring is indeed important to make logic easier to understand, and the chatGPT suggestion looks pretty good.

That being said, it assumes that `BatchV1Api` and `BatchV1betaApi` both extend `BatchApi`, which is not the case.
If reduction in the size of the `listCronJobs()` is really wanted, I guess we can use the above suggestion and overload the `getCronJobNames()` method, but at the end we will still have two additional methods.

```java
public Set<String> listCronJobs() throws ApiException {
  log.debug("Listing k8s cron jobs");
  Set<String> v1Set = Collections.emptySet();
  Set<String> v1BetaSet = getCronJobNames(new BatchV1beta1Api(client));
  log.debug("K8s V1Beta cron jobs: {}", v1BetaSet);

  if (getK8sSupportsV1CronJob()) {
    v1Set = getCronJobNames(new BatchV1Api(client));
    log.debug("K8s V1 cron jobs: {}", v1Set);
  }

  return Stream.concat(v1Set.stream(), v1BetaSet.stream()).collect(Collectors.toSet());
}

private Set<String> getCronJobNames(BatchV1beta1Api batchApi) throws ApiException {
  var cronJobs = batchApi.listNamespacedCronJob(
      namespace, null, null, null, null, null, null, null, null, null, null);
  return cronJobs.getItems().stream()
      .map(j -> j.getMetadata().getName())
      .collect(Collectors.toSet());
}

private Set<String> getCronJobNames(BatchV1Api batchApi) throws ApiException {
  var cronJobs = batchApi.listNamespacedCronJob(
      namespace, null, null, null, null, null, null, null, null, null, null);
  return cronJobs.getItems().stream()
      .map(j -> j.getMetadata().getName())
      .collect(Collectors.toSet());
}
```

@tozka Do you think it is worth the two new methods?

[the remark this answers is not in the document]
tozka: You decide as you have the code in the IDE. It looks easier to read but the difference doesn't seem big and it's likely just a opinion. 

It doesn't seem to matter much. And it doesn't remove the duplication really. It's small enough though.

doks5:
Associated Github Issue: https://github.com/vmware/versatile-data-kit/issues/1595

murphp15:
@doks5  
What I would do is create a new class called LegacyDataJobKubernetesService.
It would extend DataJobKubernetesService and it would always set k8sSupportsV1CronJob to false.
in the test setup I would have  k8sSupportsV1CronJob set to true. 


Now you have one bean that you can use for legacy inserts/update/etc and one for new inserts updates etc.. 

Please let me know if that isn't clear as I am happy to help.

doks5:
This change is blocked by https://github.com/vmware/versatile-data-kit/issues/1612

doks5:
@tozka, Can you take a look? I believe to have fixed all outstanding issues.

tozka (approved):
If you have done any manual testing please include details on what you've exactly done. 

Include comment in the Git description on why we decided not to make IT tests (you can include a link to the ticket where we discussed this (if it was in a ticket). 

Looks good to me. Thanks for the change.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -1040,7 +1078,7 @@ public void updateCronJob(
       Map<String, String> jobLabels,
       List<String> imagePullSecrets)
       throws ApiException {
-    if (getK8sSupportsV1CronJob()) {
+    if (readV1CronJob(name).isPresent()) {
```
mivanov1988, line 1081:
```suggestion
    if (getK8sSupportsV1CronJob() && readV1CronJob(name).isPresent()) {
```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -983,7 +1021,7 @@ public void updateCronJob(
       List<V1Volume> volumes,
       Map<String, String> jobDeploymentAnnotations)
       throws ApiException {
-    if (getK8sSupportsV1CronJob()) {
+    if (readV1CronJob(name).isPresent()) {
```
mivanov1988, line 1024:
```suggestion
    if (getK8sSupportsV1CronJob() && readV1CronJob(name).isPresent()) {
```

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
```diff
@@ -420,6 +425,69 @@ public void testCreateV1CronJobFromInternalResource() {
...
+    Mockito.when(mock.readV1CronJobDeploymentStatuses()).thenReturn(v1TestList);
+    Mockito.when(mock.readV1beta1CronJobDeploymentStatuses()).thenReturn(v1BetaTestList);
+    Mockito.when(mock.readJobDeploymentStatuses()).thenCallRealMethod();
+    List<JobDeploymentStatus> resultStatuses = mock.readJobDeploymentStatuses();
+
+    Assertions.assertEquals(mergedTestLists, resultStatuses);
+  }
+
+  @Test
+  public void testReadCronJob() {
```
mivanov1988, line 459: Split it into two methods, one per scenario.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -312,7 +311,8 @@ private V1beta1CronJob loadInternalV1beta1CronjobTemplate() {
 
   private V1CronJob loadInternalV1CronjobTemplate() {
     try {
-      return loadV1CronjobTemplate(new ClassPathResource(K8S_DATA_JOB_TEMPLATE_RESOURCE).getFile());
+      return loadV1CronjobTemplate(
+          new ClassPathResource("v1-" + K8S_DATA_JOB_TEMPLATE_RESOURCE).getFile());
```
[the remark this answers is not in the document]
doks5: Done. Thanks for the review

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -447,14 +447,25 @@ public Set<String> listJobs() throws ApiException {
   }
 
   public Optional<JobDeploymentStatus> readCronJob(String cronJobName) {
-    return getK8sSupportsV1CronJob() ? readV1CronJob(cronJobName) : readV1beta1CronJob(cronJobName);
+    /**
```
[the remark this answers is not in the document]
doks5: Moved.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -757,17 +773,37 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
   }
 
   public Set<String> listCronJobs() throws ApiException {
+    /**
```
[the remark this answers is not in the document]
doks5: Ditto.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -757,17 +773,37 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
...
+     * Returns a set of cron job names for a given namespace in a Kubernetes cluster. The cron jobs
+     * can be of version V1 or V1Beta.
+     *
+     * @return a set of cron job names
+     * @throws ApiException if there is a problem accessing the Kubernetes API
+     */
     log.debug("Listing k8s cron jobs");
-    var cronJobs =
-        new BatchV1beta1Api(client)
+    Set<String> V1CronJobNames = Collections.emptySet();
```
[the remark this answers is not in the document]
doks5: Good catch. Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -757,17 +773,37 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
...
             .collect(Collectors.toSet());
-    log.debug("K8s cron jobs: {}", set);
-    return set;
+    log.debug("K8s V1 cron jobs: {}", V1CronJobNames);
+
+    var v1BetaCronJobs =
+        initBatchV1beta1Api()
+            .listNamespacedCronJob(
+                namespace, null, null, null, null, null, null, null, null, null, null);
+    var V1BetaCronJobNames =
```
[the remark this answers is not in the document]
doks5: Ditto.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -757,17 +773,37 @@ public void cancelRunningCronJob(String teamName, String jobName, String executi
...
-        new BatchV1beta1Api(client)
+    Set<String> V1CronJobNames = Collections.emptySet();
+
+    var v1CronJobs =
+        initBatchV1Api()
             .listNamespacedCronJob(
                 namespace, null, null, null, null, null, null, null, null, null, null);
-    var set =
-        cronJobs.getItems().stream()
+    V1CronJobNames =
```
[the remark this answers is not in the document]
doks5: Variable is defined on line 784 above.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: If for some reason there are cron jobs using v1 and v1beta1 in the same namespace, wouldn't checking the flag cause unexpected behaviour?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: Same as above.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
mivanov1988: The flag indicates if the v1 is supported or not. In my opinion, it will be strange to see the v1 job if the flag is set to false.

mivanov1988: @tozka What’s your view on that?

doks5: This was done in the context of https://github.com/vmware/versatile-data-kit/pull/1580#issuecomment-1413626260

tozka: Paul's comment makes sense. In case of a revert, if a job is created with the new API, the control service will 'forget' it. However, this would require an additional Kubernetes API call, which is not optimal. Therefore, the decision to make it forward-compatible or not should be based on the expected performance overhead it would add.

It is worth noting that any new users would be using Kubernetes 1.21, where the V1 API is supported, and we plan to switch k8sSupportsV1CronJob to true by default soon. As a result, I would be okay with an approach that is not forward-compatible (i.e., without the extra API call).

On a separate note, I wish we had automated testing to measure how many API calls (and other resources) we use.

mivanov1988 (approved):

PR #1589, opened 2023-02-02
Control service: remove deprecated dependency 

murphp15:
# Why
At the moment we are blocked on spring upgrades because we are using a number of legacy classes. https://github.com/vmware/versatile-data-kit/pull/1530
This remove one of those dependencies. 
Even after this is merged I don't expect that PR to be fixed as there will be a number of other issues but this is a good first step
# What
I followed the guidelines here on how to remove it https://spring.io/blog/2022/02/21/spring-security-without-the-websecurityconfigureradapter

Closes: https://github.com/vmware/versatile-data-kit/issues/1590


DeltaMichael:
This seems to follow the migration guide closely, so no problem with the code itself. I'm just curious if and how the security config is tested.

ivakoleva (approved):
mind that merging may then auto-close the issue;

mivanov1988 (approved):

mivanov1988:
LGTM

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -111,22 +112,20 @@ public SecurityConfiguration(
...
-      enableSecurity(http);
-    } else {
-      log.info("Security is disabled.");
-      http.csrf().disable().authorizeRequests().anyRequest().anonymous();
-    }
+  @Bean
+  @Primary
+  protected SecurityFilterChain configure(
+      HttpSecurity http, AuthenticationManager authenticationManager) throws Exception {
+    return enableSecurity(http, authenticationManager);
```
murphp15, line 119: I'm looking into why I removed the else statement here before merging.

PR #1592, opened 2023-02-03
vdk-meta-jobs: Better error message for misspelled job name 

gageorgiev:
If a user misspells a job name in the depends_on section, currently they will receive an error informing them that the job is not registered. Job registration is an
implementation detail and is not a concept users would be familiar with, so this PR amends the error message to tell the user that they might have mistyped the job name.

Testing done: trivial


projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
```diff
@@ -90,7 +90,7 @@ def __get_job(self, job_name) -> TrackableJob:
         job: TrackableJob = self._jobs_cache.get(job_name)
         if job is None:
             raise IndexError(
-                f"The job {job_name} has not been registered. Use register_job first. "
+                f"The job {job_name} cannot be found. Verify that the job name is spelled correctly."
```
ivakoleva, line 93: maybe both, 
```suggestion
                f"The job {job_name} cannot be found. Verify the data job is registered, and the job name is spelled correctly."
```

ivakoleva (approved):

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/cached_data_job_executor.py
tozka: The fix should not be here IMO.
 The error message is not bad if the user is the developer. It says to developer using this class the job has not been registered . If this error is raised in production,  it indicates a bug .


2nd, it's too late to make such a check. If depends_on job is misspelled, it's likely the misspelled name to create issues in other places of code. 

Validation should be done as close as possible to the source/input of the data that need to be validated.

ivakoleva: I cannot seem to find the "register" term used with the vdk core api:
Screenshot 2023-02-08 at 10 59 13

tozka: The error specifies: `has not been registered. Use register_job first` . `register_job` is a method of CachedDataJobExecutor and should be used first before other methods. 

It's not supposed to be exception visible by end users and should be handled by the code up the stack. 
We should revise all end user facing errors and make sure that they follow https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#error-handling 

@gageorgiev maybe we should have a ticket for that as part of our effort to improve vdk-meta-jobs.

Issue #1600, opened 2023-02-06
Control-service: KubernetesService class is very bloated and needs to be cleaned up 


murphp15:
**What is the feature request? What problem does it solve?**
the KubernetesService class is an abstract class which is extended by two classes, the ControlKubernetesService and the DataJobsKubernetesService.

The DataJobsKubernetesService is responsible for deploying and reporting status updates data jobs. 
All of the methods that are specific to the DataJobsKubernetesService actually exist in the KubernetesService. 

This is very confusing, the fact that most of these methods live in the abstract class gives the impression that these would be used by both extending classes but this is not the case.
It also leads to extra confusion when creating an instance of ControlKubernetesService and the property `supportsK8sV1CronJobs` needs to be set even though it will have absolutely no impact.

**Suggested solution**
Move all code specific to the DataJobsControlService into the DataJobsControlService.
Leaving the class KubernetesService to be able to focus on generic abstract functionality. 
In general I also think we should avoid inheritance where possible in favour of composition https://en.wikipedia.org/wiki/Composition_over_inheritance#:~:text=Composition%20over%20inheritance%20(or%20composite,than%20inheritance%20from%20a%20base.

tozka:
I think the Control Servie as collection of services - at least that was the initial idea- the public API tends to reflect that (https://iaclqhm5xk.execute-api.us-west-1.amazonaws.com/data-jobs/swagger-ui/index.html) : 
- Data Jobs Service
- Properties Service
- Execution Service 
- Deployment Service 
- Sources Service

They live in single monolith for development and resource efficiency reasons. But they should be independent. So I'd have liked to have separate Kubernetes Service for each of those services .

There is also other issue discussing refactoring of KubernetesService - https://github.com/vmware/versatile-data-kit/issues/586

murphp15:
@tozka is that to say that you don't want to move forward with this PR https://github.com/vmware/versatile-data-kit/pull/1601 ? 

I think the best path forward is iterative changes like in that PR.

tozka:
@murphp15 

I did not see #1601 when I wrote that. 

I am a very firm believer in doing things iteratively and in small increments (and I've tried to explain why [here](https://github.com/vmware/versatile-data-kit/wiki/How-to-prepare-a-new-PR#break-your-code-commits-into--small-self-contained-units)). I was only sharing how I see the end state which might be in the far future.

sabadzhiev:
Triaged. We will keep it and we are working on this iteratively.

PR #1601, opened 2023-02-06
control-service: cleanup kubernetesService class

murphp15:
# Why
as described in this ticket https://github.com/vmware/versatile-data-kit/issues/1600 KubernetesService class is very bloated. 
Within this PR I move DataJobs specific code to its own class. 

# What
Copy all methods from the KubernetesService to the DataJobsKubernetesService if they only make sense in the context of DataJobs. 

# How has this been tested? 
Unit and integration tests


tozka:
I think this looks good to me. Moving the methods related to Cron job (which is our Kubernetes implementation of a data job deployment) is an excellent set to start with. 

The fact that Kubernetes servicea is halved in size is a big win for me with this change.

duyguHsnHsn (approved):

tozka:
Closing this PR one for now as it has been stale for more than a month. 
When Paul is back (or  if someone else decides to take it) we can re-open it as the issue is still important and relevant.

PR #1604, opened 2023-02-06
control-service: cleanup tests to ease testing on control service

murphp15:
# Why
Alot of tests which are written on the KubernetesService should actually be written on the DataJobsControlService as the tests themselves are focused on Cron job functionality which should only happens on the data jobs. 
For more info regarding this please look here: https://github.com/vmware/versatile-data-kit/issues/1600

# What
Within this PR I have modified a lot of tests to use declare the controlService with var instead of ControlService, so we can change it for DataJobsControlService very easily in an upcoming PR. 
This is a stand alone PR to make reviewing easier.


mivanov1988 (approved):

doks5:
> Within this PR I have modified a lot of tests to use declare the controlService with var instead of ControlService

For a split second I got a bit confused here. You mean Kubernetes Service, not Control Service, correct?

murphp15:
> > Within this PR I have modified a lot of tests to use declare the controlService with var instead of ControlService
> 
> For a split second I got a bit confused here. You mean Kubernetes Service, not Control Service, correct?

@doks5 I do! updated comment.

gageorgiev (approved):

duyguHsnHsn (approved):

PR #1621, opened 2023-02-13
vdk-jupyter: vdk-jupyterLab-extension changes on contribution part of the read me 

duyguHsnHsn:
What:
Added some more information on the contribution part of the README file of the extension.

Why:
To make it easier for other contributors to understand how the extension works.

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
@@ -97,15 +97,15 @@ In development mode, you will also need to remove the symlink created by `jupyte
 command. To find its location, you can run `jupyter labextension list` to figure out where the `labextensions`
 folder is located. Then you can remove the symlink named `vdk-jupyterlab-extension` within that folder.
 
-### Front-end extension 
+### Front-end extension
 This extension uses [JSX](https://reactjs.org/docs/introducing-jsx.html).
 
-The components of the front-end extension are located in the /src directory. All the new UI elements are added there. 
+The components of the front-end extension are located in the /src directory. All the new UI elements are added there.
 
-The main script for the extension is index.ts - this is where the front-end extension is loaded. 
+The main script for the extension is index.ts - this is where the front-end extension is loaded.
 In handlers.ts the connection with the server is done, while in serverRequests.ts all the requests are sent.
 
-In the subdirectory /components are located the JSX components that represent VDK menu elements. 
+In the subdirectory /components are located the JSX components that represent VDK menu elements.
 All the components use the sessionStorage to save data that is submitted from the user.
 Here is how all components use it:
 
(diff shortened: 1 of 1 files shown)


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
```diff
@@ -97,6 +97,43 @@ In development mode, you will also need to remove the symlink created by `jupyte
...
+The main script for the extension is index.ts - this is where the front-end extension is loaded.
+In handlers.ts the connection with the server is done, while in serverRequests.ts all the requests are sent.
+
+In the subdirectory /components are located the JSX components that represent VDK menu elements.
+All the components use the sessionStorage to save data that is submitted from the user.
+Here is how all components use it:
+
+|  Component  |                                                Specified SessionStorage item set by the component                                                 |
+|:-----------:|:-------------------------------------------------------------------------------------------------------------------------------------------------:|
+|  CreateJob  |                                              "local": set if the job is going to be  created locally                                              |
```
ivakoleva, line 114: Working with a code abstraction is easier than matching a human-readable nomenclature. My suggestion is, to rather have an abstraction introduced in the code

duyguHsnHsn: I will try to implement your idea, if it works good I will delete this PR.
Will not delete the PR for now, if we decide to keep the sessionStorage afterwards we can merge this.

duyguHsnHsn: @ivakoleva u can check this https://github.com/vmware/versatile-data-kit/pull/1624

ivakoleva: A nomenclature could be simply an enum of values.
The convenience with this approach is, that all of those are found at one place altogether. And updating a property (like `isCloud` becoming `cloud`), is easier and less error-prone.

duyguHsnHsn: Instead of an enum, I introduced simple classes with one method - setting everything to default( empty string). You can see the PR. I don't think changing it to enum would change anything much.

ivakoleva: I'll try to explain it better,
The problem is, we need to document the session memory model of VDK in Jupyter. So it's clear what is stored where. 

We can approach solving this problem in two ways:
1. have a purely text explanation of the memory model (property names, and what are they purposed for), or
2. have a data structure to describe that instead.

Having a data model contract, in (1.) a text format, makes it prone to discrepancies. All developers need to be aware of the existence and location of the freeform contract, and continuously sync the code with it. The downside is, there is no tools to automatically verify the code reflects the freeform text contract at all.

(2.) Having a data model contact in the code, could be easily linked in the VEP. Also, the compiler checks for discrepancies, so aany mistaken enum value typo would be fast-failing therefore fixed early on during the development life-cycle. 
Contract is usually implemented using interfaces, enums (immutable).

ivakoleva (approved):

ivakoleva:
@duyguHsnHsn shall we then close this PR in favour of https://github.com/vmware/versatile-data-kit/pull/1624 ?

Issue #1484, opened 2023-01-06
Change or document what data is session stored in JupyterLab front-end extension


duyguHsnHsn:
**What is the feature request? What problem does it solve?**
We should keep track on what data is session-stored by design in the JupiterLab front-end extension

**Suggested solution**
Introduce a .md file which explains the use cases of the data and why it is stored. 
Or implement a data structure(object) that describes the properties stored (self-explanatory for the code reader)

**Additional context**
You can see use of session storage here: https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx

PR #1624, opened 2023-02-13
vdk-jupyter: Jupyter UI change the strategy on storing the data from user input 

duyguHsnHsn:
What:
Changed the strategy on storing the data from user input for each vdk operation.

Why:
It is linked to the https://github.com/vmware/versatile-data-kit/issues/1484
It is also linked to the problem - saving SessionStorage data is really error-prone. SessionStorage keys are pure strings and even a single typo can create many unwanted behaviour of the project.
With this solution we remove the sessionStorage saved data and save them in global variables which are dataclasses that describe the input data from a user for specific vdk operation.

Note: If this solution is reviewed better than the current solution with the SessionStorage, the https://github.com/vmware/versatile-data-kit/pull/1621 PR will be removed


ivakoleva:
I got it, this change aims to shift the session-scoped data to conversation-scoped one. 
If logically possible to narrow the scope of visibility, then this would be an improvement.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/CreateJob.tsx
```diff
@@ -103,7 +112,7 @@ export default class CreateJobDialog extends Component<ICreateJobDialogProps> {
         if (!value) {
             value = this.props.jobPath;
         }
-        sessionStorage.setItem("create-job-name", value);
+        createJobData.jobName = value;
```
ivakoleva, line 115: here, there is `jobName` being a property of an `createJobData`,
then further there is `jobName` being a string key ```input_data["jobName"]```;

So, "jobName" is a part of the data contract with the user. Each user will provide a custom "jobName" value, that lives in context for a particular amount of time (in this case, during a unique `vdk create job` conversation is done).

Maybe we could have the "jobName", and the rest of the data contract, in one place altogether? So the data contract with the user is visible with one glance at the code file :)

ivakoleva (commented):
LGTM

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/CreateJob.tsx
duyguHsnHsn: _**input_data**_ lives in the server extension not in the front-end part, while createJobData, runJobData, etc. live in the front-end
The data entered by the user lives in the front-end until a request is sent to the server where according to the data a specific operation is done. When the operation is done the server sends a specific response to the front-end, after the from-end gets the response all the current saved data is deleted (all the properties are set to "").

About this: **_"Maybe we could have the "jobName", and the rest of the data contract, in one place altogether? So the data contract with the user is visible with one glance at the code file"_** 

- Are you talking about having the data separated into runJobDat, createJobDat, etc.? Maybe it is not good.
- Another solution would be to implement a class: for example UserInputData. And to create a global variable that would "live" from starting the Jupyter Kernel until it stops. The class would have different properties as jobName, jobTeam, etc. and we would only use the needed ones for the specific operation.  In this way it would be easy to see all the information that we would retrieve from the user in one place.

ivakoleva: I want to simplify it. 
Should be able to link in the VEP -> a file, pointing to a data model stored by VDK. For example, a property we store is `jobName`, and this is described "The name of the data job. Unique workload name given by the a team member.". 

Therefore, in all scopes (session, conversations, requests, etc), `jobName` means that same thing, and I can click in the VEP and navigate to a file, that contains all of those mappings property<->description.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/CreateJob.tsx
```diff
@@ -64,35 +65,40 @@ export default class CreateJobDialog extends Component<ICreateJobDialogProps> {
    */
     private _onLocalClick() {
         return (event: React.MouseEvent) => {
-            this._handleCheckboxEvent("Local", "jobPath", "local")
+            let checkbox = document.getElementById("Local");
```
DeltaMichael, line 68: This and _onCloudClick seem to do the same thing with different input. Might be worth exporting this logic to its own method and calling it in the callbacks.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/dataClasses/jobData.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2023-2023 VMware, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import data from "./jobDataModel.json";
+
+export var jobData = JSON.parse(JSON.stringify(data));
```
DeltaMichael, line 8: Please make sure there's no multithreading involved here somehow. Otherwise, awesome job, very elegant.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
```diff
@@ -54,15 +55,15 @@ def post(self):
...
-            local = True if input_data["isLocal"] else False
-            cloud = True if input_data["isCloud"] else False
+            local = True if job_data.local else False
+            cloud = True if job_data.cloud else False
             status = VdkUI.create_job(
-                input_data["jobName"],
-                input_data["jobTeam"],
-                input_data["restApiUrl"],
-                input_data["jobPath"],
+                job_data.jobName,
```
DeltaMichael, line 63: Consider refactoring the VdkUI methods to take just the new job object as an argument. You can access the separate fields inside the methods, instead of passing them here.
```
VdkUI.create_job(job_data)
```

DeltaMichael: I'm not too familiar with Jupyter internals, so take this suggestion as optional.

In fact, if you add the job type to the data model, you could do something like below in every handler.

```
VdkUI.run(job_data)
```

and then inside the run method, you can decide if you want to call create_job, download_job, etc. based on the job type.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/CreateJob.tsx
duyguHsnHsn: Primarily, I did it like that but I removed it while removing the sessionStorage. Will implement it again

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
duyguHsnHsn: > Consider refactoring the VdkUI methods to take just the new job object as an argument. You can access the separate fields inside the methods, instead of passing them here.
> 
> ```
> VdkUI.create_job(job_data)
> ```

I really like this idea. It would definitely improve the quality of the handlers.
Tho for the second comment I think having the operations separated is better for now.

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk_ui.py
```diff
@@ -45,53 +47,58 @@ def run_job(path, arguments=None):
             return f"{process.returncode}"
 
     @staticmethod
-    def delete_job(name: str, team: str, rest_api_url: str):
+    def delete_job(job_data: DictObj):
```
ivakoleva, line 50: job_data of type `DictObj` seems an odd convention for naming

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk_ui.py
```diff
@@ -45,53 +47,58 @@ def run_job(path, arguments=None):
...
-        :param rest_api_url: The base REST API URL.
+        :param job_data: the job data
+        following the projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/dataClasses/jobDataModel.json
         :return: message that the job is deleted
         """
-        cmd = JobDelete(rest_api_url)
-        cmd.delete_job(name, team)
-        return f"Deleted the job with name {name} from {team} team. "
+        cmd = JobDelete(job_data.restApiUrl)
+        cmd.delete_job(job_data.jobName, job_data.jobTeam)
```
ivakoleva, line 58: Is the contract on job_data ensuring a `jobName` property presence?

duyguHsnHsn: yes, the job_data object follows the projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/dataClasses/jobDataModel.json

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
```diff
@@ -17,19 +18,17 @@ def get(self):
 
     @tornado.web.authenticated
     def post(self):
-        input_data = self.get_json_body()
-        status_code = VdkUI.run_job(input_data["jobPath"], input_data["jobArguments"])
+        job_data = DictObj(self.get_json_body())
```
ivakoleva, line 21: The `DictObj` could be eventually initialised using any json-
maybe have a singleton instance with a stronger connection/mapping (within the class itself) to always hydrate from the exact json file, then have it hydrated optimally once (since those model properties are rather static)?

ivakoleva: Also, it is rather strange to populate job_data from DictObject type, maybe consider JobData or similar?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/dict_object.py
```diff
@@ -0,0 +1,14 @@
...
+class DictObj:
+    def __init__(self, in_dict: dict):
+        assert isinstance(in_dict, dict)
+        for key, val in in_dict.items():
+            if isinstance(val, (list, tuple)):
+                setattr(
+                    self, key, [DictObj(x) if isinstance(x, dict) else x for x in val]
+                )
+            else:
+                setattr(self, key, DictObj(val) if isinstance(val, dict) else val)
```
ivakoleva, line 14: What is the benefit of having a dynamic type based on json properties, in favour of having a class containing an enum of properties that are accompanied by their description? 

I think enum is better for enumerating nomenclatures that are static during runtime. It is a subject to compiler checks. For example, a duplication could be prevented.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
```diff
@@ -38,14 +37,9 @@ def post(self):
...
-        input_data = self.get_json_body()
+        job_data = DictObj(self.get_json_body())
         try:
-            status = VdkUI.download_job(
-                input_data["jobName"],
-                input_data["jobTeam"],
-                input_data["restApiUrl"],
-                input_data["parentPath"],
-            )
+            status = VdkUI.download_job(job_data)
```
ivakoleva, line 42: I fancy the (first) approach of having the needed-only exact parameters, rather a blob of a generic object;

ivakoleva: Thinking about input_data[JOB_NAME], where "jobName" is not a free-text value -such is prone to errors

DeltaMichael: If we're enforcing a specific object model, we shouldn't be worried about passing the correct data to methods like this one. This is a lot better imo, instead of having a method signature with 5 parameters, some of which are optional and have default values.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
DeltaMichael: @duyguHsnHsn Maybe you can do something like what's suggested here
https://stackoverflow.com/questions/6578986/how-to-convert-json-data-into-a-python-object

Not sure if this is good practice, haven't really had to do this myself. I don't like copying code just because someone claims it works. @ivakoleva what do you think about it?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
ivakoleva: Three possibilities to design interface surface for those methods `VdkUI.*` (for example `VdkUI.download_job()`): 
1. use simple data typed params (like before, jobName: str, ...)
2. use a class type, per method; for example, DownloadJobDto (that itself is composite of the params needed)
3. use a generic object that could/could not contain the params needed -> and assert on the params requires do exist during method implementation (the way latest change attempts to do, which I am not in favour of, due poor data contract)

I am looking to have the most simplistic, per key access of `input_data[KEY]`, where KEY is a constant (could be easily reused, and is subjected to compiler changes, and could be accompanied with a description).

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
ivakoleva: Well I personally think the prior (the change) approach was most simplistic, by accessing directly the input_data properties by key. No unnecessary unique DTOs per function/method to pile up. 
I would only suggest to have a constant of each key, for example usage of the constant:
```
input_data[JOB_NAME]
```
..and the constant is mapped to name, and description (both string).
The description could be used to populate any explanation of what job name is, or job path is.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/ui_job_properties.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from enum import Enum
+
+
+class JobProperties(Enum):
```
ivakoleva, line 6: maybe VdkOptions, due restApiUrl is not a job property, and the file name also
```suggestion
class VdkOption(Enum):
```

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/ui_job_properties.py
```diff
@@ -0,0 +1,13 @@
...
+    name = "jobName"
```
ivakoleva, line 7: nit: usually, the code style naming convention of enum entries is capital letters with underscore (matching constant values)

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright 2023-2023 VMware, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * This enum is automatically generated from the enum from  ../vdk-jupyterlab-extension/ui_job_properties.py
+ * Using https://github.com/Syndallic/py-to-ts-interfaces#example
+ */
+export enum JobProperties {
+  name = 'jobName',
```
ivakoleva, line 11: shouldn't the auto-generated enum be in a separate file then? are the rest of the file contents also generated or merged in a way?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
```diff
@@ -0,0 +1,62 @@
...
+  local = 'local',
+  arguments = 'jobArguments'
+}
+
+/*
+ * In the front-end extension a global variable is introduces (jobData) which holds the information about the current job.
+ * The values of its properties are meant to be changed during a VDK operation and after the operation ends they need to be set to default.
+ */
+export var jobData = new Map<string, string>([
+  [JobProperties.name, ''],
```
ivakoleva, line 25: nit: could save ourselves some work (adding manually new entries) -> if the map is automatically populated by auto-iterating over the enum values, then set those automatically, with the default value accordingly.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
```diff
@@ -0,0 +1,62 @@
...
+  [JobProperties.cloud, ''],
+  [JobProperties.local, ''],
+  [JobProperties.arguments, '']
+]);
+
+/*
+ * Function responsible for reverting all property values of jobData to default
+ */
+export function revertJobDataToDefault() {
+    jobData.set(JobProperties.name, '');
```
ivakoleva, line 38: couldn't we reuse this function to actually populate the map the first time?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
```diff
@@ -0,0 +1,62 @@
...
+    jobData.set(JobProperties.arguments, '');
+}
+
+
+/*
+ * Function taht return the JSON object of jobData
+ */
+export function getJobDataJsonObject() {
+    const jsObj = {
+        'jobName': jobData.get(JobProperties.name),
```
ivakoleva, line 53: I don't understand the purpose of this, in case you need to use the enum structure as json,
maybe have   enum to json serialization?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
```diff
@@ -97,6 +97,42 @@ In development mode, you will also need to remove the symlink created by `jupyte
...
+
+### Server extension
+This extension uses [Tordnado](https://www.tornadoweb.org/en/stable/).
+
+All the requests handlers are located in handlers.py file where the communication with the front-end is created.
+
+The connection with VDK is done in the vdk_ui.py file - all the vdk operations are handled there.
+
+### Job data model
+Follows the enum JobProperties from [ui-job-properties.py](vdk-jupyterlab-extension/ui_job_properties.py).
```
ivakoleva, line 120: how does it follow it?
maybe add information about python enum primary artifact, and auto-generation mechanism during build;

are the build instructions up-to-date with the extra step? don't see any py-to-ts-interfaces dependencies, except a mention in the comments.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
```diff
@@ -97,6 +97,42 @@ In development mode, you will also need to remove the symlink created by `jupyte
...
+
+
+In the front-end extension a global variable is introduces (jobData) which holds the information about the current job.
```
ivakoleva, line 123: so we now use a broader than session-scope (and that is global) variable, for the purposes to solve the UX consistency in multi-tab scenario? Please add a few words about this

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
```diff
@@ -97,6 +97,42 @@ In development mode, you will also need to remove the symlink created by `jupyte
...
+All the information about the data job can be accessed using the enum JobProperties from jobData/ts(src/jobData.ts),
+which is generated automatically from  [ui-job-properties.py](vdk-jupyterlab-extension/ui_job_properties.py) enum and
+shall not be changed directly in the ts file.
```
ivakoleva, line 126: The disclaimer about class modification, maybe add it to the class code docs?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
```diff
@@ -49,7 +51,7 @@ export default class RunJobDialog extends Component<IRunJobDialogProps> {
         if(!value){
             value = this.props.jobPath;
         }
-        sessionStorage.setItem("job-path", value)
+        jobData.set(JobProperties.path, value);
```
ivakoleva, line 54: is the stateful struct a vdkGlobalStorage, a custom alternative to Jupyter's sessionStorage? could you describe why we have it, in the struct code documentation?

ivakoleva (approved):
LVGTM

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
ivakoleva: nit: this could be in a separate effort

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
duyguHsnHsn: There is a problem in ES6 and we cannot directly convert the map into json, this is the only solution that works for now but off can be changed later

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
duyguHsnHsn: You are right. It will be better to separate it. Going to change it now

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
duyguHsnHsn: At the moment it is not added to the build, but will be added as separate effort

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
duyguHsnHsn: sadly no, it places undefined instead of empty string

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/ui_job_properties.py
duyguHsnHsn: Done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/ui_job_properties.py
duyguHsnHsn: Done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
duyguHsnHsn: Done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
duyguHsnHsn: TODO added

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/jobData.ts
duyguHsnHsn: done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
duyguHsnHsn: done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk_ui.py
```diff
@@ -93,5 +93,5 @@ def create_job(
         if local:
             cmd.validate_job_path(path, name)
 
-        cmd.create_job(name, team, path, cloud, local)
+        cmd.create_job(name, team, path, bool(cloud), bool(local))
```
ivakoleva, line 96: I like the good contract, using the data types accordingly. The way `create_job()` was accepting boolean in favour of string (anything could be represented as string).

My suggestion is, to parse boolean from string before invoking `create_job()`, for example:
```bool(input_data[VdkOption.CLOUD.value])```

ivakoleva (approved):
LGTM

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/vdk_ui.py
duyguHsnHsn: fixed

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
duyguHsnHsn: added

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
duyguHsnHsn: done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
duyguHsnHsn: added the auto-generation mechanism during build;

PR #1625, opened 2023-02-13
versatile-data-kit: git pre-commit hooks config

ivakoleva:
Frontend config files were failing git pre-coming hooks due:
* too large 
* JSON format inspection
* PR title prefix

License insertion now covers TypeScript, JavaScript, (S)CSS, 
HTML, Y(A)ML file types in addition.
PR titles can be now associated with "frontend" component.
Modified the pre-commit and gitlint configurations accordingly.

Testing done: installed the hook locally, and verified all
frontend/data-pipelines changes from multiple stacked PRs 
were committed and auto-modified correctly
with the code header expected.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,6 +82,7 @@ repos:
       - id: insert-license
         files: \.ts$
         args:
+          - --use-current-year
           - --license-filepath
           - NOTICE.txt
           - --comment-style
@@ -89,6 +90,7 @@ repos:
       - id: insert-license
         files: \.js$
         args:
+          - --use-current-year
           - --license-filepath
           - NOTICE.txt
           - --comment-style
@@ -96,6 +98,7 @@ repos:
       - id: insert-license
         files: \.html
         args:
+          - --use-current-year
           - --license-filepath
           - NOTICE.txt
           - --comment-style
@@ -103,10 +106,17 @@ repos:
       - id: insert-license
         files: (\.scss$|\.css$)
         args:
+          - --use-current-year
           - --license-filepath
           - NOTICE.txt
           - --comment-style
           - /*| *| */
+      - id: insert-license
+        files: (\.yaml$|\.yml$)
+        args:
+          - --use-current-year
+          - --license-filepath
+          - NOTICE.txt
       - id: insert-license
         files: \.py$
         args:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,11 +10,11 @@ repos:
         # exclude helm chart templates
         exclude: ^projects/control-service/projects/helm_charts
       - id: check-json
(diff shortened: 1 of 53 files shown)


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/.eslintrc.js
```diff
@@ -1,3 +1,8 @@
+/*
+ * Copyright 2021 VMware, Inc.
```
sabadzhiev, line 2: Isn't 2023 applicable here?

ivakoleva: Well noticed; did a quick check and my impression is, we need to update the NOTICE.txt year accordingly.

ivakoleva: It seems we need 2021-2023. What are your thoughts on this?

gageorgiev (approved):

.pre-commit-config.yaml
```diff
@@ -10,9 +10,11 @@ repos:
         # exclude helm chart templates
         exclude: ^projects/control-service/projects/helm_charts
       - id: check-json
+        exclude: (^projects/frontend/data-pipelines/gui/tsconfig.json|^projects/frontend/data-pipelines/gui/package-lock.json|^projects/frontend/data-pipelines/gui/projects/data-pipelines/.eslintrc.json|^projects/frontend/data-pipelines/gui/projects/data-pipelines/tsconfig.lib.prod.json|^projects/frontend/data-pipelines/gui/projects/data-pipelines/tsconfig.spec.json|^projects/frontend/data-pipelines/gui/projects/data-pipelines/tsconfig.lib.json|^projects/frontend/data-pipelines/gui/projects/ui/.eslintrc.json|^projects/frontend/data-pipelines/gui/projects/ui/tsconfig.app.json|^projects/frontend/data-pipelines/gui/projects/ui/tsconfig.spec.json)
```
DeltaMichael, line 13: Do we want to exclude all jsons under `data-pipelines`? If this is the case, it might be worth doing something like below (not sure about the exact regex pattern)

```
^projects/frontend/data-pipelines/.*json$
```

ivakoleva: not sure wewant to generalise that much, would prefer to pin the concrete files for exclusion at this point;
That would prevent adding a file that is excluded from the pre-commit hook - without the author realising it.

What are your thoughts on this?

DeltaMichael: I'm fine with leaving it like this. Was just wondering if we can't use a shorter regex, but it seems like you want to make it more granular.

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/.eslintrc.js
sabadzhiev: Agree.

Issue #1271, opened 2022-10-28
Implement Object Oriented Design of invoking meta jobs


tozka:
Currently meta jobs are specifed as dictionary of some fields. 

There's is a design to have API Meta Job and and IMetaJobInput so that users instead of calling

run_meta_job(custom_dict) , then can call run_meta_job (new MetaJob())

The approach with custom dict is easier for non software developers to use though as it's really configuration - no need to create pesky classes and related them to each other. 

But for maintenance purposes (of the code) and if we want to express more complex cases we should be OOP model as well. 

_Originally posted by @tozka in https://github.com/vmware/versatile-data-kit/pull/1249#discussion_r1005574353_

See also https://github.com/vmware/versatile-data-kit/pull/1640#discussion_r1108634403

sabadzhiev:
Triaged. Closed as stale.

PR #1640, opened 2023-02-14
versatile-data-kit: Meta Job example

gageorgiev:
This PR introduces a wiki example demonstrating how a user could implement a Meta Job and component jobs.


examples/metajob-example/README.md
```diff
@@ -0,0 +1,458 @@
+Overview
+--------
+Meta Jobs allow VDK users to schedule jobs in a directed acyclic graph.
+This means that jobs can be configured to run only when a set of previous jobs have finished successfully.
+
+In this example we will use the Versatile Data Kit to develop four Data jobs - two of these jobs will read data
+from separate json files, and will subsequently insert the data into Trino tables. The third job will read the data
+inserted by the previous two jobs, and will print the data to the terminal. The fourth Data Job will be a Meta Job
+which will manage the other three and ensure that the third job run only when the previous two finish successfully.
```
dvalkova, line 9: small typo -> ...the third job runS only when...

dvalkova:
I went through the README and the sample data jobs. I think the example is great, easy to follow, with no unnecessary text or explanations. The only thing I would add to the README is a short explanation about the MetaJobInput interface - just 1-2 sentences to stress on the fact that this is a different interface to the IJobInput we're used to using (or maybe add just a comment in the meta job script about that). This is not a requirement, you can decide whether to implement it.

dvalkova (approved):
Ship it :)

examples/metajob-example/example-meta-job/example_meta_job.py
```diff
@@ -0,0 +1,29 @@
...
+        "job_name": "read-data-job",
+        "team_name": "my-team",
+        "fail_meta_job_on_error": True,
+        "depends_on": ["ingest-job1", "ingest-job2"],
+    },
+]
+
+
+def run(job_input):
+    MetaJobInput().run_meta_job(JOBS_RUN_ORDER)
```
victor-partico, line 29: Just an idea. I think it would be easier for user if the paramethers are implemented is such way that can be used like:
```
some_object.run_meta_job(
    job_name = "ingest-job1",
    team_name = "my-team",
    fail_meta_job_on_error = True,
    depends_on = []
)

some_object.run_meta_job(
  **{
      "job_name": "read-data-job",
      "team_name": "my-team",
      "fail_meta_job_on_error": True,
      "depends_on": ["ingest-job1", "ingest-job2"]
  }
)
```
I like what I currently see. This is just a suggestion.

victor-partico (approved):

examples/metajob-example/example-meta-job/example_meta_job.py
gageorgiev: will consider this, but it is not in the scope of this PR

examples/metajob-example/README.md
gageorgiev: noted

examples/metajob-example/example-meta-job/example_meta_job.py
tozka: There is a ticket about that idea here - https://github.com/vmware/versatile-data-kit/issues/1271 
Feel free to follow it /upvote it @victor-partico  :)

PR #1663, opened 2023-02-21
frontend: test and publish vdk/shared

DeltaMichael:
### What

frontend-pr workflow:
- build and test vdk/shared on every PR that introduces changes to vdk/shared or vdk/data-pipelines

frontend-relase workflow:
- buld and test vdk/shared on every merge to main that introduces changes to vdk/shared or vdk/data-pipelines
- publish vdk/shared to public npm registry
- bump vdk/shared patch version by one

### How was this tested

Running actions on my VDK fork
https://github.com/DeltaMichael/versatile-data-kit/actions

### What kind of change is this

feature, non-breaking


.github/workflows/frontend-release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - '!projects/frontend/shared-components/gui/projects/shared/package.json'
+
+jobs:
+  build-test-publish-vdk-shared:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
DeltaMichael, line 20: Not sure if the GITHUB_TOKEN has write permissions to the repo. I don't have the correct access level to change it if it doesn't.

.github/workflows/frontend-release.yml
```diff
@@ -0,0 +1,55 @@
...
+
+    - name: Test
+      working-directory: ./projects/frontend/shared-components/gui
+      run: npm run test
+    
+    - name: Publish Shared Components
+      working-directory: ./projects/frontend/shared-components/gui/dist/shared
+      run: npm publish --ignore-scripts
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
DeltaMichael, line 43: Also not sure where to get this.

tozka:
We use Gitlab CI for CICD for all projects. 

Github  Actions only is used for integration with 3 party tools and services like Codacy, Pre-commit and not really anything created by us in the CI CD pipeline.

As there are some potential drawbacks to using multiple CI/CD  tools in the same project:
* Complexity to Dev process: Using multiple CI/CD tools can add complexity to the development process, as developers need to be familiar with the syntax and configuration of both tools.
* Maintenance:  may increase the amount of maintenance required to keep the build and deployment process running smoothly. Each tool may require separate updates, bug fixes, and configuration changes, which can be more time-consuming to manage.
* Reduced visibility: Using multiple CI/CD tools may reduce visibility into the build and deployment process, as each tool may have its own dashboard or reporting system

It might be more efficient and cost-effective to choose one tool and use it consistently throughout the development process. 

That being said, our repo is also structured as a mono repo. With each project being separate and independent. If we decide it's better to use Github Actions for the front end , that's ok if the benefits outway the costs.. But I feel we should not mix them in the in a single project (frontend) and the decision require analysis of cons and pros.

cc @ivakoleva

PR #1668, opened 2023-02-22
versatile-data-kit: simplify release process

tozka:
Having to enumerate every single package is time-consuming and is never used by anyone.  Those versions are rather arbitrary and point-to-time snapshots at the time of release announcement which could be found if needed from the GItlab CI and does not represent anything special (like a more stable or battle-tested version). 

So instead we can simplify the process and only include install instructions.


ivakoleva (approved):
I'm approving because I want to first unblock this change. 
Please make sure in addition, the packages repositories are described in the docs, and it is clear where the packages come from (or go to).

CONTRIBUTING.md
```diff
@@ -145,10 +145,7 @@ To make a new public release, follow these steps:
   - Major features include
   - Backwards incompatible changes
   - Removal of features
-- Review which components have received changes since the previous release - this should be apparent from the changelog;
-- Record the latest versions of the components included in the release (from the auto generated notes) and list them in the Release description;
-    - You can see the current version (as of time of release) of a Python package in https://pypi.org/search/?q=
-    - You can see the current version (as of time of release) of the Control Service in [the helm repo here](https://gitlab.com/vmware-analytics/versatile-data-kit-helm-registry/-/packages)
+- Include Install instructions (link to main install page).
```
tozka, line 148:
```suggestion
- Include Install instructions (link to https://github.com/vmware/versatile-data-kit/wiki/Installation).
```

doks5 (approved):

PR #1669, opened 2023-02-23
control-service: Remove license headers from helm charts

DeltaMichael:
### What

Remove existing license headers from helm charts
Disable auto-insert of pre-commit hooks for helm charts Clean up pre-commit config

### Why

License headers break parsing for helm charts
This breaks control-service deployments

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,11 +44,6 @@ repos:
     hooks:
       - id: pylint
         args: [ --exit-zero ]
-  #-   repo: https://github.com/pre-commit/mirrors-mypy
-  #    rev: v0.812
-  #    hooks:
-  #        -   id: mypy
-  #            files: ^(src/|tests/|plugins/)
   - repo: https://github.com/asottile/reorder_python_imports
     rev: v3.9.0
     hooks:
@@ -60,14 +55,6 @@ repos:
     hooks:
       - id: pyupgrade
         args: [ --py37-plus ]
-  # security check
-  #- repo: https://github.com/PyCQA/bandit
-  #  rev: '1.7.0'
-  #  hooks:
-  #      - id: bandit
-  #        exclude: tests
-  #        args: [ -s, B101 ]
-  # add copyright notice
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.4.2
     hooks:
@@ -113,6 +100,8 @@ repos:
           - /*| *| */
       - id: insert-license
         files: (\.yaml$|\.yml$)
+        # exclude helm chart templates
+        exclude: ^projects/control-service/projects/helm_charts
         args:
           - --use-current-year
           - --license-filepath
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -113,6 +100,8 @@ repos:
           - /*| *| */
       - id: insert-license
         files: (\.yaml$|\.yml$)
+        # exclude helm chart templates
+        exclude: ^projects/control-service/projects/helm_charts
```
ivakoleva, line 104: why excluding the copyright notices from yaml files? 
having those, is in accordance with company legal

DeltaMichael: Because helm can't parse the yaml files for some reason if they have comments on the first line.

ivakoleva: I did helm lint, and the chart seems well formatted. Also, helm documentation states that yaml templates do support comments

DeltaMichael: Actual error we're getting

```
Error: UPGRADE FAILED: error validating "": error validating data: apiVersion not set
```

https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3813924321

ivakoleva: should we remove the copyright notice, or is there any other way to keep it? removing it seems harshly and against legal

DeltaMichael: I can check if we can add it to the bottom of the file instead.

ivakoleva: > Put the two copyright notices ("Copyright Notice" and "Copying Permission Statement") together, near the top of each source file.

DeltaMichael: > Put the two copyright notices ("Copyright Notice" and "Copying Permission Statement") together, near the top of each source file.
Bummer 😕

I'll dig a bit deeper, then. Maybe repro locally. But this is most likely what's causing the problem. Maybe we need to update helm in the runner image if it's been fixed upstream.

tozka: You can remove the copyright notice. There were not there before as well (until very recently). And that was on purpose, not an omission. When we open-sourced the project a year ago I consulted and it was ok to remove it if it would break the build and there was no alternative.

tozka: To explain a bit more:  Under the Apache License (2.0) it is recommended that all source files include a copyright notice to protect intellectual property and provide clarity around ownership of the code. However, the Apache License does not require a copyright notice to be included in the source files.

The Apache License has specific requirements for attribution and notices that must be included in the LICENSE and NOTICE files distributed with the project (which we have), but it does not require that copyright notices be included in individual source files. That being said, having a copyright notice in each source file can help prevent confusion about the origins of the code and can help protect against unauthorized use or distribution. Hence we put it everywhere

DeltaMichael: I actually found a way to keep the licenses, although it might result in some annoying debug output. Also turned out that I didn't remove the correct license headers. I'll update the PR shortly.

PR #1670, opened 2023-02-23
control-service: Update helm template license headers

DeltaMichael:
### What
Update helm template license headers to use template-style comments Ignore helm templates in pre-commit yaml license hook Create new pre-commit license hook to add template-style comments to helm templates

### Why
Regular yaml comments at the top of files break parsing for helm charts This breaks control-service deployments

### How was this tested

Tested installing the charts on a local k8s cluster using helm

Ran `helm upgrade --install --debug --wait --timeout 10m0s cicd-control-service .` on the chart, which gave me the same output as the failing pipeline. Edited the comments until the apiVersion error disappeared.

```
Error: UPGRADE FAILED: error validating "": error validating data: apiVersion not set
```

https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3813924321

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ tests.xml
 temp
 gradlew
 gradlew.bat
+
+# Ignore downloaded helm charts
+projects/control-service/projects/helm_charts/pipelines-control-service/charts/*
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -113,10 +113,20 @@ repos:
           - /*| *| */
       - id: insert-license
         files: (\.yaml$|\.yml$)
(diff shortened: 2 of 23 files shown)


tozka (changes requested):
Please do not include the downloaded charts in source control (the binary files). Those are generated automatically.

tozka:
> Tested installing the charts on a local k8s cluster using helm

It is good idea when doing any manual testing to include the actual commands being executed in the 'manual' test and some relevant for verification output of the command. That makes it clearer what exactly was tested and easier to review if something could be tested better.

tozka (approved):

ivakoleva (approved):

Issue #1533, opened 2023-01-20
Documentation improvement: improve VDK images


zverulacis:
**What is the feature request? What problem does it solve?**
From the problem definition interviews:

_-The image is confusing (VDK before/after)._
_-If I'm an engineer, I don't think in terms of teams - Data Team / Infra team._ 
_-If I have no data background, it will not support me in understanding the framework._ 
_-specifically about the images: images are not explaining anything but are pretty, to a Data Engineer it's not clear what does it mean silos, or actually, none of the text in the first image is clear_
_-images seem to be very marketing-oriented / salesy software product type: "so the images look nice, but I'm reading this, but how am I going to apply this?"_
_-they compared the images and the text in it to software tools, as to say that it seems to solve all their issues and problems in life and work and so on, but it's not clear what it actually does and how._
_-One person said that after living ten years in the US they hadn't heard the word silo and didn't know what it meant without googling._
_-DEs don't get the language used there: "what is a silo? what are the tensions? why is it focused on teams? what is fragmented infrastructure?"_
_-but the assumption of DEs is that images are useful for the infrastructure decision-makers most likely; maybe they know the terms._
_-images that are used in the repository are helpful only to these people, not data engineers (+ almost everyone I've had a conversation about this are saying the same)_

**Suggested solution**
Use simple language and make the benefits of the Versatile Data Kit clear for everyone in the images 

**Additional context**
The problem is a good place to start with solving the issue. It needs to be understood first before implementing any changes; simpler/clearer language might support any type of user/reader.

Addition to this ticket:
Please check existing images if any of them are irrelevant and remove them if they are.

Radsotina: just remove things not relevant to VDK OSS, some images that I added were copied - like download keytab

PR #1676, opened 2023-02-25
Documentation improvement: improve main README VDK images - gif

zverulacis:
This PR is part of the issue https://github.com/vmware/versatile-data-kit/issues/1533 

The goal is to create a gif to preview vdk CLI usage.


DeltaMichael (commented):
What's the point of having a separate gif for the help command?

README.md
```diff
@@ -57,6 +57,9 @@ VDK creates data processing workflows to:
     - A plugin framework that allows users to extend the framework according to their specific requirements.
 - Control Service: The Control Service allows users to create, deploy, manage, and execute data jobs in a Kubernetes runtime environment.
 
+A preview of the VDK CLI commands - `vdk --help`, `vdk create` and `vdk run`:
```
DeltaMichael, line 60: What's the point of previewing the help command? Create and run are super valuable to have as a gif preview. Having help is a standard part of any CLI, so it doesn't seem special in that respect. Plus, whoever is reading this will not have time to read the help output and it's a bit annoying, imo.

zverulacis: My idea was to show all the possibilities of VDK, but I don't want to put a separate image of the help command.
Also, gif is a catchy way to do it. The idea is not to show the help command, of course, but all the other commands that I don't show. But I can leave it out if you think it doesn't make sense.

gary-tai:
Consider a bullet list the commands...

A preview of the VDK CLI commands:
- `vdk create`
- `vdk run`
-  `vdk deploy`

DeltaMichael (approved):

zverulacis:
I still want to make the black bar nicer, maybe gray too, or idk yet
But I'd like to hear from you @tozka if it's the right direction and if you like it there. Because the commands are actually pretty self-explanatory and with the list of what I'm doing it might not be necessary.

tozka:
I think this GIF would be great to include in the tutorial on the VDK Wiki, especially since it demonstrates the VDK commands clearly. For example, we could replace "my-job" with "Hello World" as shown in the tutorial https://github.com/vmware/versatile-data-kit/wiki/getting-started.


However, for a first quick overview of the CLI capabilities, it may not be sufficient. To make it more effective, it should be tailored towards a specific problem or use case. And if data or analytics engineers have similar use case, this would give users the impression that the VDK CLI can help them solve their challenges.

For instance, Let's say we have a scenario where a data engineering team needs to process and transform large amounts of customer data for a retail company. The team needs to extract data from multiple sources, transform it into a unified format, and load it into a data warehouse for further analysis. That's a common scenario in the data engineering world 

The first VDK CLI command that appears is vdk create --name retail-data --team data-engineering.

Next, the GIF shows an animation of the VDK CLI command `vdk run retail-data`, which starts the data job. The GIF shows the status of each step in the job as it progresses, such as extracting data from each source, performing transformations, and loading the data into the warehouse. So the steps should be named appropriately.

After the job completes successfully, the GIF shows an animation of the VDK CLI command vdk deploy --name retail-data-pipeline --team data-engineering, which deploys the pipeline to the cloud. The GIF shows the pipeline being versioned and uploaded to the cloud, the job being deployed and ready for use.

tozka (approved):
Can we decrease the size by 30% ? I feel it's taking almost teh full screen and with smaller screens it might be hard to see. 

I am going to approve it. Because I think even if we merge it now it would be an improvement over the previous state. But I think we can iterate and improve even further.

gary-tai (approved):
lgtm

Issue #1559, opened 2023-01-27
Stabilise nightly build pipelines


murphp15:
**Describe the bug**
At the moment the nightly build pipeline is very unstable. 
The result of this ticket is that nightly builds should always be green when no regressions have been introduced. 

**Steps To Reproduce**
Look at previous nightly runs to get a list of unstable jobs.

PR #1677, opened 2023-02-27
vdk-trino: stabilize vdk-trino tests

tozka:
This change is fixing vdk-trino tests which are failing regularly in the nightly builds . See
https://github.com/vmware/versatile-data-kit/issues/1559

The tests were failing intermittently for two main reasons 
* Docker failures which are being fixed by adopting testcontainers  (in favour of pytest-docker)
* Hidden dependencies between  test cases (because different test use same schema and tables on shared db instance)

Until now we used pytest-docker to start service container, this is switching it to testcontainers-python.

There are multiple advantages of using it.

* Supposedly Testcontainers is built with performance in mind and uses a variety of techniques to minimize container startup time and resource usage like image reuse, container reuse, asynchronous operations.
* It is far more popular lately - https://hugovk.github.io/top-pypi-packages/  is a site I frequently use to decide which library is likely to be more stable/used . It is showing testcontainers is 2000 places ahead of pytest-docker
* Testcontainers provides a wider range of features for managing containers, such as starting and stopping them, waiting for them to become available, and exposing their ports to the host system
* Testcontainers-Python allows you to define and configure containers using Python code instead of yaml which pytest-docker required.

Testing Done: ran the test suite of vdk-trino a few times and it was more stable and quick. I have not run actual benchmarks and recorded times but it seems about twice faster possibly. We will verify this as part of the nightly builds.

### DIFF
--- a/projects/vdk-plugins/vdk-trino/tests/conftest.py
+++ b/projects/vdk-plugins/vdk-trino/tests/conftest.py
@@ -5,29 +5,15 @@ import time
 from unittest import mock
 
 import pytest
-from vdk.plugin.test_utils.util_funcs import CliEntryBasedTestRunner
-from vdk.plugin.trino import trino_plugin
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.waiting_utils import wait_for_logs
+
+TRINO_IMAGE = "trinodb/trino:latest"
 
 VDK_TRINO_HOST = "VDK_TRINO_HOST"
 VDK_DB_DEFAULT_TYPE = "VDK_DB_DEFAULT_TYPE"
 VDK_TRINO_PORT = "VDK_TRINO_PORT"
 VDK_TRINO_USE_SSL = "VDK_TRINO_USE_SSL"
-
-
-def is_responsive(runner):
-    try:
-        result = runner.invoke(["trino-query", "--query", "SELECT 1"])
-        if result.exit_code == 0:
-            return True
-    except:
-        return False
-
-
-@pytest.fixture(scope="session")
-def docker_compose_file(pytestconfig):
-    return os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "docker-compose.yml"
-    )
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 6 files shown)


projects/vdk-plugins/vdk-trino/tests/conftest.py
```diff
@@ -40,19 +24,80 @@ def docker_compose_file(pytestconfig):
...
-    time.sleep(3)
-
-    docker_services.wait_until_responsive(
-        timeout=30.0, pause=0.3, check=lambda: is_responsive(runner)
-    )
+    port = int(os.environ[VDK_TRINO_PORT])
+    container = DockerContainer("trinodb/trino:408").with_bind_ports(port, port)
+    try:
+        container.start()
+        wait_for_logs_with_print(container, "SERVER STARTED", timeout=120)
```
ivakoleva, line 35: is there a configurable retry with backoff, likewise the pytest-docker approach?

projects/vdk-plugins/vdk-trino/tests/conftest.py
```diff
@@ -40,19 +24,80 @@ def docker_compose_file(pytestconfig):
...
+        os.environ[VDK_TRINO_HOST] = container.get_container_host_ip()
+        os.environ[VDK_TRINO_PORT] = str(container.get_exposed_port(port))
```
ivakoleva, line 37: nit: reinitializing the env variables...immutable configurations is actually simple and less error prone

projects/vdk-plugins/vdk-trino/tests/conftest.py
```diff
@@ -40,19 +24,80 @@ def docker_compose_file(pytestconfig):
...
+    except Exception as e:
+        print(f"Failed to start Trino service: {e}")
+        print(f"Container logs: {container.get_logs()}")
+        print(
+            "var logs: "
+            + container.exec(command=["cat", "/var/log/trino/server.log"])
+            + "\n"
+        )
+        print("java version: " + container.exec(command=["java", "-version"]) + "\n")
+        time.sleep(120)
```
ivakoleva, line 50: May have this value adjustable/tunable in global scope, seems it has to do with graceful start wait

projects/vdk-plugins/vdk-trino/tests/conftest.py
```diff
@@ -40,19 +24,80 @@ def docker_compose_file(pytestconfig):
...
+        Number of seconds until the predicate was satisfied.
+    """
+    if isinstance(predicate, str):
+        import re
+
+        predicate = re.compile(predicate, re.MULTILINE).search
+    import time
+
+    start = time.time()
+    while True:
```
ivakoleva, line 90: nit: `while true` could be questionable in terms of design (due the implementation may change, and it's a potential endless loop trap); I would rather suggest `while condition` (in that case, may be duration), so completing the loop in favour of returning success, would result in timeouterror

ivakoleva (commented):
LGTM
(mind the codacy report)

projects/vdk-plugins/vdk-trino/tests/conftest.py
tozka: I should have marked the change as draft. I added this peace of code (it's copy paste from testcontainers library) to verify something. it's removed now in favour of wait_for_logs method in testcontainers.

projects/vdk-plugins/vdk-trino/tests/conftest.py
tozka: this was added temporarily. Removed now.

projects/vdk-plugins/vdk-trino/tests/conftest.py
tozka: I should have marked the change as draft. I added this peace of code (it's copy paste from testcontainers library) to verify something. it's removed now in favour of wait_for_logs method in testcontainers.

projects/vdk-plugins/vdk-trino/tests/test_vdk_templates.py
```diff
@@ -27,42 +28,54 @@
...
-    if from_table_name == "tmp_dw_scmdb_people" and to_table_name == "dw_scmdb_people":
+    if from_table_name.startswith("tmp_dw_people") and to_table_name.startswith(
+        "dw_people"
+    ):
         obj.drop_table(from_db, from_table_name)
-    if (
-        from_table_name == "backup_dw_scmdb_people"
-        and to_table_name == "dw_scmdb_people"
+
+    if from_table_name.startswith("backup_dw_people") and to_table_name.startswith(
```
ivakoleva, line 46: Why moving tests from an exact table name to a non-exact one?
I now see that some of the tests source & target tables are changed to dynamic ones (and some are kept static), why?

projects/vdk-plugins/vdk-trino/tests/conftest.py
```diff
@@ -40,19 +26,28 @@ def docker_compose_file(pytestconfig):
...
-
-    docker_services.wait_until_responsive(
-        timeout=30.0, pause=0.3, check=lambda: is_responsive(runner)
-    )
+    port = int(os.environ[VDK_TRINO_PORT])
+    container = DockerContainer(TRINO_IMAGE).with_bind_ports(port, port)
+    try:
+        container.start()
+        wait_for_logs(container, "SERVER STARTED", timeout=120)
+        time.sleep(2)
```
ivakoleva, line 38: nit: may add a comment explaining the purpose

ivakoleva (approved):

projects/vdk-plugins/vdk-trino/tests/test_vdk_templates.py
tozka: The schema name is dynamic and different for each test case because some tests were failing because they were written in the same table which was one of the reasons for the intermittent failures of the tests (explained it also in the commit description). 

I renamed some tables to make it easier to troubleshoot - having scmdb_people everywhere in the logs is not as useful as knowing it scmdb_people_scd1 which indicates the test case.

duyguHsnHsn (approved):

PR #1678, opened 2023-02-27
 vdk-ipython: change the way job_input is introduced in a notebook

duyguHsnHsn:
What:
Change the way job_input is introduced in a notebook by creating a class that wraps IStandaloneDataJob instance. Instead of job_input a JobControl object is pushed to Jupyter when the %reload_data_job magic is used. Using that object the user can get initialised job_input by calling get_initialized_job_input() method of the JobControl object. 

Why: The previous version introduced a job_input that was in finalised state.

Tests: unit tests are introduced


tozka:
The title should be prefixed by the name of the plugin being changed 

`vdk-ipython: change ...`

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
```diff
@@ -7,28 +7,55 @@
...
-    StandaloneDataJobFactory,
-)
+from vdk.internal.builtin_plugins.run import standalone_data_job
+
+
+class JobControl:
+    """
+    The main idea of thi class is to make it easier to  work with IStandaloneDataJob instances
+    Since the user will be using this class
+    """
```
ivakoleva, line 17: Trying to figure out why it would be easier to work with a wrapper than the VDK class?
Didn't get that info from the PR description, nor the docs

ivakoleva, line 17: I don't understand what `The previous version introduced a job_input that was in finalised state.` in "why" section of PR description is,
may you explain what the problem is?

duyguHsnHsn: I thing, it will looks strange to use data_job.__enter__ (if I was a standard user I would not think it return an initialised job_input). Instead having a class that is more user friendly is a better option in my opinion. I am not sure whether my class is that good, but I think I will be nice to have something like a "wrapper" for the users. I am open to any other ideas

tozka:
Please include Github issue that this change is part of.

tozka:
Some general comments: 

While the code is generally well-structured, it could benefit from having docstrings added to the JobControl class and the load_job() and magic_load_job() functions. This would make it easier for other developers to understand the purpose of each component and how to use them.

Adding type hints to the function arguments and return values would make the code more readable and help prevent errors

tozka (changes requested):

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
```diff
@@ -7,28 +7,55 @@
...
+            name=name,
+            job_args=arguments,
+            template_name=template,
+            data_job_directory=path,
+        )
+
+    def get_initialized_job_input(self):
+        return self.job.__enter__()
+
+    def finalise(self):
```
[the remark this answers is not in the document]
duyguHsnHsn: I was thinking to introduce it in a different PR, as well. I will add the TODO

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
```diff
@@ -7,28 +7,55 @@
...
+        path = pathlib.Path(path) if path else pathlib.Path(os.getcwd())
+        self.job = standalone_data_job.StandaloneDataJob(
+            name=name,
+            job_args=arguments,
+            template_name=template,
+            data_job_directory=path,
+        )
+
+    def get_initialized_job_input(self):
+        return self.job.__enter__()
```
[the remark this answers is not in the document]
duyguHsnHsn: I will check it. But using Singleton seems to be better, I will change and try it.

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
ivakoleva: Isn't the wrapper (re)implementing the StandaloneDataJobFactory `create()`, see
https://github.com/vmware/versatile-data-kit/blob/768ec74b9dfc05b8a0207b4a4eac5955bf2e42cd/projects/vdk-core/src/vdk/internal/builtin_plugins/run/standalone_data_job.py#L196 ?

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
ivakoleva, line 37: may consider improving the `StandaloneDataJob` class, since you need to use it, and it seems that you have found a better interface surface to do so? For example, you think `finalize()` is a better way have an `__exit__` alias?

duyguHsnHsn: Well, this interface surface fits vdk_ipython use of StandaloneDataJob but I am not sure whether it is so for all use case scenarios of the class. If I remember correctly the main idea of using __enter__ and __exit__ in this class is to use it with the StadaloneDataJobFactory (which I do not use in this solution)
I don't know whether it will be ok to change the whole class for this specific interface surface
We might get more opinion, @tozka what do yo. think about it?

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
duyguHsnHsn: Yes, it was but I changed the solution a little bit. U can check it again, I added one more attribute to handle calling initialise multiple times by the users

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
```diff
@@ -1,34 +1,84 @@
...
+        """
+        Get initialised IJobInput object for the current job if present
+            :return:  an IJobInput object
+        """
+        if not self.job_input:
+            self.job_input = self.job.__enter__()
+        return self.job_input
+
+    # TODO: should automate calling finalize (if it is not called explicitly)
+    def finalise(self):
```
tozka, line 56:
```suggestion
    def finalize(self):
```

I think we stick mostly to American English spelling of words (e.g Initialize not Initialise)

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
tozka: It's an adapter. It infers things like job directory , name etc from the notebook location. 

I am tno sure about the names. JobControl is internal implementation name and won't be exposed to end user so it's less important what is it. 

The end user would use `job. get_initialized_job_input()` and `job.finalize()`

We probably would need to iterate on the names and looks for user feedback .

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
```diff
@@ -38,8 +88,5 @@ def magic_load_job(line: str):
 def load_job(
     path: str = None, name: str = None, arguments: str = None, template: str = None
 ):
-    path = pathlib.Path(path) if path else pathlib.Path(os.getcwd())
-    with StandaloneDataJobFactory.create(
-        data_job_directory=path, name=name, job_args=arguments, template_name=template
-    ) as job_input:
-        get_ipython().push(variables={"job_input": job_input})
+    job = JobControl(path, name, arguments, template)
+    get_ipython().push(variables={"data_job": job})
```
tozka, line 92: I was thinking. Wouldn't be better to be "VDK" ? 


Like 

```
data_job.get_initialized_job_input()
```

or 
```
VDK.get_initialized_job_input()
```

We can add more stuff related to VDK in the future. VDK is the more familiar and user facing name . That's why. 

Similarly to how click project exposes a lot of interfaces using `click.xxx` interface (click is the name of the project and main entry point)

duyguHsnHsn: I like this naming more, as well

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
duyguHsnHsn: done

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -18,21 +21,151 @@ def ip(session_ip):
...
+    )
+    ip.get_ipython().run_cell("payload = response.json()")
+
+    # send data for ingestion
+    ip.get_ipython().run_cell(
+        'job_input.send_object_for_ingestion(payload=payload,destination_table="rest_target_table")'
+    )
+
+    # wait until the table is populated
+    await asyncio.sleep(1)
```
tozka, line 118: There is `time.sleep` to sleep . There's not need to add 3th party library just for that.

How sure you are that 1 second is enough ?  It's always better not to rely on timing but to verify explicitly (e.g wait till it's available)

tozka (approved):

projects/vdk-plugins/vdk-ipython/README.md
```diff
@@ -14,19 +14,21 @@ Then to load the extension in Jupyter the user should use:
 ```
 %reload_ext vdk_ipython
 ```
-And to load the job_input:
+And to load the data_job (Job Control object):
```
ivakoleva, line 17: Having a "data job" from the perspective of:
a) sources the users create (by template, or from scratch)
b) at the same time, a notebook runtime-loaded data job context
, could be confusing. For example, a user will be referring to "data job" and it would not be clear what he or she is referring to exactly;

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
ivakoleva: The vdk-sqlite ingestion tests seem to directly lookup the data upon ingestion, see https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-sqlite/tests/test_sqlite_plugin.py#L71

projects/vdk-plugins/vdk-ipython/README.md
```diff
@@ -14,19 +14,21 @@ Then to load the extension in Jupyter the user should use:
...
 
 ### Example
 The output of this example is "myjob"
 ```
 %reload_ext vdk_ipython
 
-%reload_job_input --name=myjob
+%reload_data_job --name=myjob
+
+job_input = data_job.get_initialized_job_input()
```
ivakoleva, line 31: Does it make a difference for the end user to be utilising `get_initialized_job_input` or simply `get_job_input`? Is "initialized" meaningful to keep in mind from user perspective?

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
ivakoleva: nit: Is this a concern valid for other StandaloneDataJob usages? Seems to me, likewise jupyter notebook users, other users could also be finding themselves in a situation with initializing multiple times a StandaloneDataJob...

ivakoleva (approved):

projects/vdk-plugins/vdk-ipython/README.md
duyguHsnHsn: I left it like because of the finalise method, because I thought that users might think "What did I initialise to have a finalization" or sth like that. When we are ready with the finalise method, we can change the naming

projects/vdk-plugins/vdk-ipython/README.md
duyguHsnHsn: I changed it to VDK, as mentioned in https://github.com/vmware/versatile-data-kit/pull/1678#discussion_r1120069027 
I forgot to update the README. I will update it now

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
duyguHsnHsn: > The vdk-sqlite ingestion tests seem to directly lookup the data upon ingestion, see https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-sqlite/tests/test_sqlite_plugin.py#L71

I used it because of the get_ipython.get_output function

projects/vdk-plugins/vdk-ipython/src/vdk_ipython.py
duyguHsnHsn: yes, it is. We might raise an issue about it

PR #1681, opened 2023-02-28
[vdk-plugins] vdk-meta-jobs: implement a limit on starting jobs at once

yonitoo:
What:
Create a validation function that checks if in a Meta Jobs DAG there exists a state with more than N (set to 15 right now) jobs are ready to start at once. The validation function validate_job_limit() is used when the DAG is built before anything else.

Why: We want to limit the number of simultaneously starting Meta jobs in order not to overload the SC

Tests: unit test is introduced

Signed-off-by: Yoan Salambashev [ysalambashev@vmware.com](mailto:ysalambashev@vmware.com)


projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -12,12 +12,56 @@
...
+    sorted_jobs = []
+    for job_name in topo_sorter.static_order():
+        job = next((j for j in jobs if j["job_name"] == job_name), None)
+        if job:
+            sorted_jobs.append(job)
+
+    out_degree = {job["job_name"]: 0 for job in jobs}
+    for job in sorted_jobs:
+        for pred in job["depends_on"]:
+            print(pred)
```
tozka, line 39: Don't use print. Always use logging.

tozka:
Please link the github issue that you are addressing in this PR

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -12,12 +12,56 @@
...
+
+
+def validate_job_limit(jobs: List[Dict]):
+    topo_sorter = TopologicalSorter()
+    for job in jobs:
+        topo_sorter.add(job["job_name"], *job["depends_on"])
+
+    sorted_jobs = []
+    for job_name in topo_sorter.static_order():
+        job = next((j for j in jobs if j["job_name"] == job_name), None)
```
ivakoleva, line 32: Is there a possibility that "depends_on" job name is not in `jobs` list (param) keys? 
For example:
```
A -(depends on)-> B
C - ...
```
How so (asking about the logic here), and should we at least log a message

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -12,12 +12,56 @@
...
 from vdk.plugin.meta_jobs.cached_data_job_executor import TrackingDataJobExecutor
 from vdk.plugin.meta_jobs.meta import TrackableJob
 from vdk.plugin.meta_jobs.remote_data_job_executor import RemoteDataJobExecutor
 from vdk.plugin.meta_jobs.time_based_queue import TimeBasedQueue
 
 log = logging.getLogger(__name__)
+max_starting_jobs = 15
+
+
+def validate_job_limit(jobs: List[Dict]):
```
ivakoleva, line 25: Could you provide code docs about the purpose of `validate_job_limit`, for example explaining expected behaviour towards depends_on graph elements

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -12,12 +12,56 @@
...
+        for pred in job["depends_on"]:
+            print(pred)
+            out_degree[pred] += 1
+
+    ready_jobs = set()
+    for job in sorted_jobs:
+        if out_degree[job["job_name"]] == 0:
+            ready_jobs.add(job["job_name"])
+
+    for job in sorted_jobs:
```
ivakoleva, line 47: nit: a lot of loops, it's a good code style to comment on each code block (why necessary, what is the purpose), to make it easily readable.
then, you can think if optimisations (doing same stuff by reusing same loop) are possible

tozka:
Btw the correct prefix is just the name of the plugin `vdk-meta-jobs: ...` 

We use the titles to auto-generate the release notes: https://github.com/vmware/versatile-data-kit/releases

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_dag.py
```diff
@@ -67,3 +70,46 @@ def mock_get_finished_job_names(*args, **kwargs):
 
     # check for busyloop (this would have been called hundreds of times if there is busyloop bug)
     assert calls[0] <= 4
+
+
+def test_build_dag_over_max_starting_jobs_sad_case():
+    job1 = dict(job_name="job1", depends_on=[])
```
DeltaMichael, line 76: You can init these using a loop or list comprehension.

DeltaMichael (commented):
It's a bit hard to understand what the code does. Just by looking at it and without having tried to run it, it should do the following:

1. Sort the DAG topologically
2. Get each level of the sorted DAG
3. If a level exceeds the max limit, raise an error

If this is correct, there should be a good library to use without having to implement it yourself. In case you still decide to implement it yourself, I suggest you break the code down into simple pieces, so it's more understandable, e.g. have one function that sorts, one that gets the levels and one that checks the length. This is a good approach regardless of how you handle the levels that are over the limit (@tozka's suggestion)

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Done.

tozka:
We met face-to-face (zoom to zoom) with @yonitoo  and agreed on an approach where the MetaDag class ensures that there are at most N (configurable) concurrent jobs running by delaying the start of a new job if it exceeds the limit. We will reuse the already-built delay job functionality.

We also realized that the current validation approach would not work because it doesn't account for the fact that job duration could be different, and job start can be delayed for different reasons. As a result, it's possible that parts of the DAG may be ahead of other parts.

yonitoo:
Thanks a lot for the feedback! @tozka @ivakoleva @DeltaMichael 
After a discussion with Antoni today, we decided that from UX perspective, it would be best to just check before the start of each job if there are more than a configurable amount of jobs (15 by default) starting concurrently. If so, the rest of the jobs would be delayed. Therefore, the previous validation function is replaced by the aforementioned check as well as we would no longer throw an error. Also, the test is now remade and some long list initializations are replaced by list comprehension.

yonitoo:
Actually, I just saw that @tozka already wrote the update, sorry for the spam.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -109,7 +117,16 @@ def default_serialization(o: Any) -> Any:
 
     def _start_job(self, node):
         try:
-            self._job_executor.start_job(node)
+            if (
+                len(self._job_executor.get_currently_running_jobs())
+                > self.max_concurrent_starting_jobs
```
tozka, line 122: Probably should be `>=`
```suggestion
                len(self._job_executor.get_currently_running_jobs())
                >= self.max_concurrent_starting_jobs
```

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -34,12 +35,19 @@ def __init__(self, team_name: str):
                 )
             ),
         )
+        self._max_concurrent_starting_jobs = int(
```
tozka, line 38:
```suggestion
        self._max_concurrent_running_jobs = int(
```

Isn't ita better name ?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -109,7 +117,16 @@ def default_serialization(o: Any) -> Any:
 
     def _start_job(self, node):
         try:
-            self._job_executor.start_job(node)
+            if (
+                len(self._job_executor.get_currently_running_jobs())
+                > self.max_concurrent_starting_jobs
+            ):
+                log.info(
+                    "Starting job fail - too many concurrently running jobs. Will be re-tried later"
```
tozka, line 125: It would be helpful to log the number of currently running jobs and the maximum allowed number of concurrent jobs in the log message when   delaying a job. This will help in debugging efforts

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_dag.py
```diff
@@ -67,3 +70,42 @@ def mock_get_finished_job_names(*args, **kwargs):
...
+            "job16",
+            "job17",
+        ],
+    ]
+
+    dag.execute_dag()
+
+    assert [
+        call(job["job_name"]) for job in jobs[:17]
+    ] == dag._job_executor.start_job.call_args_list
```
tozka, line 111: I am not sure how do we verify that at max VDK_META_JOBS_MAX_CONCURRENT_STARTING_JOBS number of jobs are started and the last few are pending. 

Also to simplify the test (less jobs) and also not break the test if the default is changed, It's good idea to explicitly set the max concurrent jobs setting:  

You can do that with 
```
@mock.patch.dict(
            os.environ,  { xxx: "val" }   )
```

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_dag.py
yonitoo: Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Not relevant anymore.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_dag.py
```diff
@@ -4,9 +4,12 @@
 from unittest.mock import call
 from unittest.mock import MagicMock
 
+from pytest import raises
```
tozka, line 7: Also please fix the codacy warnings. There are unused imports.

yonitoo: Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: You are right, thank you. Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Yep, changed it.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_dag.py
yonitoo: I added the mock patch with the default value of 15 and still figuring out the way to verify the last few pending jobs.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -109,7 +116,18 @@ def default_serialization(o: Any) -> Any:
 
     def _start_job(self, node):
         try:
-            self._job_executor.start_job(node)
+            if (
+                len(self._job_executor.get_currently_running_jobs())
+                >= self.max_concurrent_running_jobs
```
tozka, line 121: Looks like a bug 
```suggestion
                >= self._max_concurrent_running_jobs
```

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_dag.py
tozka: To make the test easier to read/consume. lower the value to 5. 

You can checkout https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py for ideas.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
```diff
@@ -303,3 +303,36 @@ def test_meta_job_circular_dependency(httpserver: PluginHTTPServer):
...
+
+
+def test_meta_job_exceed_limit(httpserver: PluginHTTPServer):
+    jobs = [("job" + str(i), [200], "succeeded") for i in range(1, 8)]
+    api_url = _prepare(httpserver, jobs)
+
+    with mock.patch.dict(
+        os.environ,
+        {
+            "VDK_CONTROL_SERVICE_REST_API_URL": api_url,
```
tozka, line 315: Because we have delayed jobs, you may want to speed up the test by lowering VDK_META_JOBS_DELAYED_JOBS_MIN_DELAY_SECONDS and VDK_META_JOBS_DELAYED_JOBS_RANDOMIZED_ADDED_DELAY_SECONDS to 0 or 1 so we do not wait and re-try almost immediately.

yonitoo: Done. Thank you.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
```diff
@@ -303,3 +303,36 @@ def test_meta_job_circular_dependency(httpserver: PluginHTTPServer):
...
+        # Check if the limit has been reached (we add 1 for the request for job1)
+        if (
+            len(post_requests)
+            == int(os.getenv("VDK_META_JOBS_MAX_CONCURRENT_RUNNING_JOBS", "5")) + 1
+        ):
+            print("change state")
+
+        cli_assert_equal(0, result)
+
+        assert len(post_requests) == 17
```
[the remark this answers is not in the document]
yonitoo: Thanks for this idea, Toni! Appreciate it. I tried implementing this but when the limit is reached (after job6 in the test example), the test is stuck - we don’t receive any new requests and never reach job7 due to the set limit. The delayed job7 is not started.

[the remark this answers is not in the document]
tozka: Well , that's good to know. There's a bug in the implementation (most likely). Debug it. The test should not be stuck.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -63,6 +63,10 @@ def execute_dag(self):
         while self._topological_sorter.is_active():
             for node in self._topological_sorter.get_ready():
                 self._start_job(node)
+                if len(self._job_executor.get_currently_running_jobs()) >= self._max_concurrent_running_jobs:
```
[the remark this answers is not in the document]
yonitoo: The root cause was that as soon as the limit is hit, the method "_start_delayed_jobs()" is trying to be executed but there is no space left for the delayed jobs and that's why I decided that this is a good moment to finalise the current running jobs as soon as we hit the limit and be able to start the delayed ones.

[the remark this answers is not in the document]
yonitoo: Actually just realized what I did there... Now I moved the check out of the loop.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
```diff
@@ -303,3 +303,55 @@ def test_meta_job_circular_dependency(httpserver: PluginHTTPServer):
...
+                assert (
+                    running_jobs <= expected_max_running_jobs
+                )  # assert that max concurrent running jobs is not
+                # exceeded
+            if (
+                request.method == latest_req
+                and "executions/job" in request.path
+                and response.status < "300"
+            ):
+                running_jobs = 0  # back-to-back GET reqs means the jobs have been finalized successfully
```
tozka, line 350: When testing, it's important to focus on testing what the software is supposed to do, instead of how it's built.  
* If the implementation changes, the tests will need to be updated. This is especially bad for refactoring changes. Refactoring is the process of making changes to the code to improve it  without changing its behavior.
* Tests that depend on specific implementation details are limited in scope and are likely to catch less potential bugs

tozka: So I think it would be better instead to keep a set of jobs (the key in the set would be the job name) when a job is started we add to a set, and if we get status finished response we remove the job from the set . Wouldn't this be more robust ?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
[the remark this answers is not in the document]
yonitoo: Yep, it makes sense that this would not fix the bug. It would indeed loop forever due to the number of delayed jobs exceeding the concurrent limit. I will change my approach to fix that.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
yonitoo: Yep, it would be. I was way too excited to have a working test and got carried away... Will change according to your feedback.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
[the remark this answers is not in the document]
tozka: > loop forever due to the number of delayed jobs exceeding the concurrent limit.

I am not sure I see it. That does not seem to be the reason I see or I do not understand you correctly. 
Please, it's really important to read the code correctly and understand what the code does when making changes.

tozka:
Improving your code reading abilities is essential for becoming a better programmer. So let me offer you some tips which I've found useful for me and I hope you find useful as well: 

* Take your time: When you read code, don't rush through it. Take your time to understand each line and how it fits into the bigger picture.
* Identify the purpose of the code.  Ask yourself what problem the code is trying to solve and how it fits into the larger project or system.
* Take some time to familiarize yourself with the data structures used in the code.
* For each method or class (or data structure) used: Read the method/class signature, Understand the parameters, understand the return value, Read the description/documentation, and when necessary read the implementation 
* If you can do all that entirely in your mind, that's very good but it takes time to get used to it. You can also use a pen and paper or notebook editor - it may be helpful to take notes on what each line of code does.
* Use a debugger, just for reading purposes, to step through the code can help you understand how it works and what each line does. 
* Practice, practice, practice - for example, reading code from other open-source projects is an excellent way to learn from other programmers.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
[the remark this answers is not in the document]
yonitoo: By this I meant that if there are more than N delayed jobs (let's say N = limit), then when we go into "_start_delayed_jobs()" we would be forever stuck in the loop there, delaying the already delayed jobs because the running jobs hit the limit and there is no room for these delayed ones.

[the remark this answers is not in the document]
yonitoo: It no longer hangs with the latest changes.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -38,6 +41,10 @@ def __init__(self, team_name: str, meta_config: MetaPluginConfiguration):
             time_between_status_check_seconds=meta_config.meta_jobs_time_between_status_check_seconds(),
         )
 
+    @property
+    def max_concurrent_running_jobs(self):
```
tozka, line 45: why exposing it as a public property? It's not used outside of this class.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -54,8 +61,14 @@ def execute_dag(self):
         while self._topological_sorter.is_active():
             for node in self._topological_sorter.get_ready():
                 self._start_job(node)
-            self._start_delayed_jobs()
-
+            if self._delayed_starting_jobs.size() > 0:
```
tozka, line 64: is this if needed?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -54,8 +61,14 @@ def execute_dag(self):
...
+                self._start_delayed_jobs()
+            while (
```
tozka, line 66: Is this while cycle needed ?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -78,7 +91,10 @@ def _finalize_jobs(self, finalized_jobs):
             self._finished_jobs.append(node)
 
     def _start_delayed_jobs(self):
-        while True:
+        while (
+            len(self._job_executor.get_currently_running_jobs())
```
tozka, line 95: Shouldn't this check be enough to solve the issue?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Same as above.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: The if statement goes along with the while loop. Actually, I decided to use the if+while because, at least from what I saw, it performs better in terms of time - the number of operations are nearly 1/2 of the ones using the other approach. However, I agree that removing the if+while, the code would be cleaner. If optimality is not the high priority, should I remove it still?

tozka: Performance is important here because we are sending a lot of requests. to the API: 

```
it performs better in terms of time - the number of operations are nearly 1/2 of the ones using the other approach
```

How do you measure it? What operations do you mean (API calls? -those are fairly expensive and reducing them is a bonus).

tozka (approved):
Looks good to me. Assuming the tests pass let's merge it.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Removed it.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: Removed it.

gageorgiev (approved):

PR #1682, opened 2023-02-28
versatile-data-kit: Meta Jobs blogpost

gageorgiev:


examples/metajob-blogpost.md
```diff
@@ -0,0 +1,25 @@
+# Managing VDK Data Jobs in a directed acyclic graph using Meta Jobs
+
+### Intro
+
+Modern data pipelines are comprised of many workloads of different types which often are interdependent on each other. Versatile Data Kit has provided greater atomization to individual data workloads through its Data Jobs API, and now with the release of the vdk-meta-jobs package, VDK offers the ability to express Data Job dependencies in an idiomatic fashion through managing workloads called Meta Jobs.
+
+### Meta Jobs Overview
```
tozka, line 7: Overall the structure looks pretty good. 
I think uou are going into technical details too much which is more tailored for architecture document. 

Here are suggestions for small improvements in the structure, for example sections can be: 

1. The Challenges of Managing Complex Data Pipelines (take this out of our discussion with users we have had)
2. Introducing VDK Meta Jobs 
   In this section, introduce the VDK Meta Jobs package and explain how it addresses the challenges discussed in the previous section. This could include topics such as:
- An overview of the Meta Jobs and its capabilities
- How Meta Jobs allow for the creation and management of directed acyclic graphs (DAGs) for data pipelines
- The benefits of using Meta Jobs for managing complex data pipelines, such as improved efficiency, scalability, and reliability

I'd lighten the focus on the implementation (e.g answer what does it do  and not how )
3. Demo use case
Reuse simplified version of a real use case. Consider co-writing the blog post with some of VDK users who already are using meta jobs and let them help with that part
4. Getting Started with VDK Meta Jobs is better title than "Installation/Setup"
  In this section, provide guidance and resources for readers who want to start using vdk-meta-jobs in their own projects. 
5. Conclusion: Sum up the benefits of using Meta Jobs with VDK and encourage readers to try it out for themselves.

Considering how much vdk meta jobs look like Luigi / Airflow, it would be good to call out somewhere the comparison. Some short section. VDK Meta Jobs is a lightweight and simple solution for managing data jobs in a DAG that is tightly integrated with VDK, while Airflow offers a more powerful and flexible platform with a larger feature set and community.

tozka: At this stage using markdown is fine. 
But Please write the blog post in medium directly. Medium has its own way of providing review comments. 

For the full blog post review, it would be important to review the format, pictures chosen, and layout, which we cannot do in the markdown file properly.

tozka: > But Please write the blog post in medium directly. Medium has its own way of providing review comments.

That being said, if you want to write the blog post iteratively (which is always good idea).

In other words write first the introduction and the conclusion as [this guy advices](https://blog.hubspot.com/marketing/how-to-write-in-depth-blog-posts) , it's fine to keep it here. Though if layout and formating is important even then it might be better directly in medium. 

I like those advies about blogs as well https://www.grammarly.com/blog/how-to-write-a-blog

PR #1696, opened 2023-03-02
Improve documentation: before and after image 

zverulacis:
From the problem definition interviews:​

_- The image is confusing (VDK before/after).​_
_- If I'm an engineer, I don't think in terms of teams - Data Team / Infra team.​_
_- If I have no data background, it will not support me in understanding the framework. ​_

Instead of bringing clarity, the images are confusing potential users.​

Solution: Use simple language and make the benefits of the Versatile Data Kit clear for everyone in the images​

Images created/edited:​
- Image of Without VDK vs. With VDK (problem/solution)​
- Image of Without VDK vs. With VDK (code)​


ivakoleva (approved):
LGTM
I like the colour and visualization, the idea seems compelling to me

gageorgiev (commented):
Love the idea! Two notes I have:
1. The k8s logo on the target databases is confusing to me - target databases don't need to be k8s-based;
2. The SQL step and Python step blocks seem somewhat out of place to me.

duyguHsnHsn (commented):
I love the photos but I think it is hard to read "data sources" in the second one. And why are they placed in one place? 
Isn't the idea that they can be in different places but still will be pointing all to VDK and not as the first one pointing to different places

duyguHsnHsn:
> Love the idea! Two notes I have:
> 
> 1. The k8s logo on the target databases is confusing to me - target databases don't need to be k8s-based;
> 2. The SQL step and Python step blocks seem somewhat out of place to me.

2. seems off to me as well, bc in general there wouldn't be just 1 sql step and 1 python step, and VDK is way more than just steps. I think leaving it just Versatile data kit is better

duyguHsnHsn:
Is the .svg only for dark mode? I think some of the text won't be visible in light mode.

duyguHsnHsn (approved):
LGTM! I really like the second image for VDK

tozka:
Please explain in the PR description, why are doing this change, what is the goal of the before and after image. 

If this explanation is available in a github issue, you can leave it in the github issue (provide link). 

please link the github issue from which this PR comes in any case.

PR #1701, opened 2023-03-02
vdk-plugins: bump docker version and resuse common dind

tozka:
Last change upgraded the shared plugin-ci to use latest version of docker. 3 plugins do not use the shared build-plugin-dind which means they are still using an old version.
Switch the 3 plugins to use build-plugin-dind

Testing Done: this pr ci


duyguHsnHsn (approved):

ivakoleva (commented):
vdk-trino seems to use .build-plugin still, https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-trino/.plugin-ci.yml#L6

Maybe reuse and override the tls, is needed?

Basically, any docker-compose.yml would have to be migrated? Should we add a ticket

tozka:
> vdk-trino seems to use .build-plugin still, https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-trino/.plugin-ci.yml#L6
> 
> Maybe reuse and override the tls, is needed?
> 
> Basically, any docker-compose.yml would have to be migrated? Should we add a ticket

disable tls is something I am not sure is needed now that I look at it. I added while trying to get tests running but it did not solve the issue. 

Otherwise added trino to use build-plugin-dind

PR #1705, opened 2023-03-06
frontend: data-pipelines e2e tests

ivakoleva:
A prerequisite is a control-service instance running.
Auth configured with CSP STG provisioned OAuth app, redirect URLs and
TTL set common to the prior app configs.
e2e tests sync to freezepoint - 
Had to fetch the latest changes from the internal repo contributed,
to unify the diverged with the public repo.

Added test_e2e_data_pipelines.sh script for cicd/local dev purposes.
Added Gitlab job to run. 
Provisioned the clientId with CSP STG.
Synced the `frontend/data-pipelines/gui/e2e/*`
This PR is dependent on syncing the functional changes in the scope of:
https://github.com/vmware/versatile-data-kit/pull/1746  

Testing done: ci/cd; run locally `npm start` and `node --max_old_space_size=4096 ./node_modules/cypress/bin/cypress open --env OAUTH2_API_TOKEN=$VDK_API_TOKEN,CYPRESS_TERMINAL_LOGS='./e2e/logs'` where auth is successful
Testing the e2e tests functional changes will be additionally done.


projects/frontend/cicd/test_e2e_data_pipelines.sh
```diff
@@ -0,0 +1,31 @@
...
+
+cd "$(dirname $0)" || exit 1
+cd "../data-pipelines/gui"
+
+if [ $# -eq 0 ]
+  then
+    echo "ERROR: No OAuth2 token provided."
+    exit 3
+fi
+export OAUTH2_API_TOKEN=$1
```
tozka, line 28: Instead of passing the API token as an argument to the script, I think it's better to use environment variables to store sensitive information. The environment variables can be set in the GitLab CI/CD settings and then accessed within the script. 

A few reasons : 
 * Environment variables are more secure than arguments because they are not visible in the process list or command history. 
 * Avoiding clutter in the future - As a script grows in complexity, it can become difficult to manage and maintain multiple command line arguments.

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -34,8 +34,24 @@ frontend-data-pipelines-build:
...
 
+frontend-data-pipelines-e2e-tests:
+  image: "versatiledatakit/vdk-cicd-base-gui:2.0.0"
+  stage: pre_release_test
+  before_script:
+    - cd projects/frontend/
+  script:
+    - ./cicd/test_e2e_data_pipelines.sh "$VDK_API_TOKEN"
+  retry: !reference [.retry, retry_options]
+  rules:
```
tozka, line 49: Consider generating and publishing npm test repo: 


```
artifacts:
    paths:
      - reports/ # Artifacts for the report
```
 
 to make it easier to debug / see issues in Gitlab website .

projects/frontend/cicd/test_e2e_data_pipelines.sh
ivakoleva: The proposal made, makes quite less likely for the (dev local setup) secret to be copied and shared by accident. That applies for a usage, where a user pasted the parameter directly. 

Not sure an environment-scoped secret (broader visibility) than a parameter (local scope) is more secure, due longer life-cycle and global accessibility. In case of parameter, the user may decide to source it in different ways (including env var). 
Some examples and analysis made https://smallstep.com/blog/command-line-secrets/ 

Do you think we need to transition?

tozka: I think env. variables are a bit more secure than command-line arguments. But we could do better. 
I do also think that they are more maintainable (in the same way that keyword arguments are easier to maintain than positional arguments) 

But we've used command line arguments before (though I've never recommended it) and we can continue using it and re-evaluate in separate change (the focus of this change is different).

gorankokin (commented):
General question is: against which deployment are this e2e tests executed?
They are written (changed) to be executed in internal VMware network against testing environment where Teams exists, etc...
What about github environment?

projects/frontend/cicd/.gitlab-ci.yml
ivakoleva: I've separated the changes. Wiring the e2e tests goes in https://github.com/vmware/versatile-data-kit/pull/1754, where I've added e2e reports.

ivakoleva:
This PR remains for transferring diverged e2e tests code, and will most probably be closed - due the diverged e2e tests code had been meanwhile coupled with another non-VDK service.

ivakoleva:
Agreed with @gorankokin to take a look and resolve Team service (internal) dependencies with current PR

PR #1706, opened 2023-03-06
frontend: bump data-pipelines dependency versions

DeltaMichael:
#### Why

Fixes following vulnerabilities

https://github.com/vmware/versatile-data-kit/security/dependabot/6 https://github.com/vmware/versatile-data-kit/security/dependabot/7

#### What

Bump patch versions for moment.js and ua-parser-js

#### How has this been tested?

Manual build data-pipelines and run tests

#### What type of change are you making?

Bugfix

### DIFF
--- a/projects/frontend/data-pipelines/gui/projects/data-pipelines/package.json
+++ b/projects/frontend/data-pipelines/gui/projects/data-pipelines/package.json
@@ -53,7 +53,7 @@
     "lodash": "~4.17.21",
     "lottie-web": "~5.9.4",
     "lz-string": "~1.4.4",
-    "moment": "~2.29.3",
+    "moment": "~2.29.4",
     "ng2-date-picker": "~12.0.4",
     "ngx-clipboard": "~15.1.0",
     "ngx-cookie-service": "~13.2.1",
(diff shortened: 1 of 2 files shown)


tozka (approved):

ivakoleva (approved):
May use a bound in favour of pinning?

projects/frontend/data-pipelines/gui/projects/data-pipelines/package.json
```diff
@@ -53,7 +53,7 @@
     "lodash": "~4.17.21",
     "lottie-web": "~5.9.4",
     "lz-string": "~1.4.4",
-    "moment": "~2.29.3",
+    "moment": "~2.29.4",
```
ivakoleva, line 56: Why does moment gets pinned to 2.29.3? Is there any other moment references?

DeltaMichael: Just the ones that are listed here. We should have a broader discussion about how we handle dependancies anyway. It's something we inherited from Taurus and I'm not sure what the best practice is.

PR #1725, opened 2023-03-08
versatile-data-kit: automate merging of dependabot PRs

tozka:
Dependabot is creating PR. This is helping to automate them by
1. Automatically approving them
2. Automatically trying to merge them (only if all tests have passed)
3. Automatically updating them with latest changes

Testing Done: created a fork and tested the flow there.
using custom labels (as a proxy for dependabot)

### DIFF
--- a/.github/workflows/dependabot.autoapprove.yml
+++ b/.github/workflows/dependabot.autoapprove.yml
@@ -1,3 +1,6 @@
+# Copyright 2023-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
 name: Auto approve
 on: pull_request_target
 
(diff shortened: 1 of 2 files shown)


.github/workflows/dependabot.autoapprove.yml
```diff
@@ -0,0 +1,14 @@
...
+
+name: Auto approve
+on: pull_request_target
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    if: github.actor == 'dependabot[bot]'
```
DeltaMichael, line 12: Should we also not filter the automerge action like this?

ivakoleva (approved):

.github/workflows/dependabot.autoapprove.yml
ivakoleva: May you elaborate?

DeltaMichael (approved):

PR #1730, opened 2023-03-09
vdk-plugins: remove global "variables" in CI

tozka:
As part of effort to stabilize builds -
https://github.com/vmware/versatile-data-kit/issues/1559

I am removing global variables and names since they can be overwriten by accident across differnet projects.

### DIFF
--- a/projects/vdk-plugins/.plugin-common.yml
+++ b/projects/vdk-plugins/.plugin-common.yml
@@ -27,7 +27,7 @@
   artifacts:
     when: always
     reports:
-      junit: tests.xml
+      junit: "projects/vdk-plugins/$PLUGIN_NAME/tests.xml"
 
 .build-plugin-on-vdk-core-release:
   stage: pre_release_test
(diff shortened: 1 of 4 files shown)


projects/vdk-plugins/.plugin-common.yml
```diff
@@ -27,7 +27,7 @@
   artifacts:
     when: always
     reports:
-      junit: tests.xml
+      junit: "projects/vdk-plugins/$PLUGIN_NAME/tests.xml"
```
ivakoleva, line 30: may you link a pipeline before/after?

nit: the PR description does not seem to match the draft changes 👀

projects/vdk-plugins/vdk-impala/.plugin-ci.yml
```diff
@@ -1,7 +1,6 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
```
ivakoleva, line 4: nit: those are redundant, since overwritten? what's the effective python version definition?

projects/vdk-plugins/.plugin-common.yml
tozka: It's a draft . I am still working on the change. And I needed to test using the Gitlab CI since I cannot locally as easily for this. 

Anyhow I opened a PR for the change since it should be separate from this one : https://github.com/vmware/versatile-data-kit/pull/1732

PR #1743, opened 2023-03-14
frontend: Fix shebang in build_shared.sh

DeltaMichael:
Change shebang in build_shared.sh to !/bin/bash -e

This points to the correct execution shell for build scripts and passes the -e flag, which makes the script fail on the first error exit code

### DIFF
--- a/projects/frontend/cicd/install_shared.sh
+++ b/projects/frontend/cicd/install_shared.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash -e
 # Copyright 2023-2023 VMware, Inc.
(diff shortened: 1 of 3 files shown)


ivakoleva (approved):
LGTM; Are you also considering adding some additional messaging to enable us troubleshooting https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/3927259840? Is that one-time only, or reoccurring event? 

(nit: may look in k8s cicd namespace to find any traces of runner pods execution)

duyguHsnHsn (approved):

DeltaMichael:
Also added the `verbose` flag to `.npmrc` to enable verbose logging for troubleshooting.

PR #1744, opened 2023-03-14
vdk-ipython: add finalize job 

duyguHsnHsn:
What:
Added finalise_job for JobControl object which finalises the job instance in JobControl.
The job can be finalised manually or when exiting the iPython shell.

Why: A way to finalise the initialized job in ipython should be provided.

Tests: pytest test


projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +169,19 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
     assert "2021-01-01  GOOG          123" in (
         ip.get_ipython().getoutput("! " "vdk " "sqlite-query -q 'SELECT * FROM stocks'")
     )
+
+
+def test_finalise(ip):
```
tozka, line 174: We are missing some scenarios : 

Calling get_initialized_job_input multiple times with no issues
``
job_input = VDK.get_initialized_job_input()
job_input = VDK.get_initialized_job_input() # should be the same instance 
VDK.finalize()
```

Another scenario calling get_initialized_job_input after previously calling finalize
```
job_input = VDK.get_initialized_job_input()
VDK.finalize()
job_input_second = VDK.get_initialized_job_input()
```
 

Calling the VDK.finalize() method multiple times and verifying that it does not cause any issues.
```
job_input = VDK.get_initialized_job_input()
VDK.finalize()
VDK.finalize()
```

Scenario in if the kernel is restarted before shutdown. I don't actually know what would happen ?

Scenario in which the kernel is reset - https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-reset That removes all the variables but does not perform again the imports.

Comment (duyguHsnHsn) [reply projects/vdk-plugins/vdk-ipython/tests/test_plugin.py:174]: Thanks a lot, added a  few new tests and did a few changes on the code since some of them were failing with current implementation.

File: projects/vdk-plugins/vdk-ipython/tests/test_plugin.py [hunk 2]
```diff
@@ -169,3 +169,50 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+    )
+    ip.get_ipython().run_cell("VDK.finalize()")
+    assert (
+        "finalize_job"
+        in ip.get_ipython().run_cell("VDK.job._plugin_registry.hook().__dict__").result
+    )
+
+
+def test_get_initialized_job_input_multiple_times(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
```
tozka, line 191: Try to style each test method using AAA pattern : http://wiki.c2.com/?ArrangeActAssert

```
ip.run_line_magic(magic_name="reload_VDK", line="")

# act 
ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
first_result = ip.get_ipython().run_cell("job_input").result
ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")

# assert 
assert first_result == ip.get_ipython().run_cell("job_input").result
```

In practice I do not actually write comments `#act` `#assert` but leave an empty line between each phase. 

```
ip.run_line_magic(magic_name="reload_VDK", line="")

ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
first_result = ip.get_ipython().run_cell("job_input").result
ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")

assert first_result == ip.get_ipython().run_cell("job_input").result
```

But I recommend thinking in this way.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +169,50 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    assert first_result == ip.get_ipython().run_cell("job_input").result
+
+
+def test_get_initialized_job_input_after_finalize(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
```
tozka, line 203: you are missing verify/assert phase. WHat is supposed to happen (or not happen) if we call VDK.finalize multiple times.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +169,50 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+
+
+def test_call_finalize_multiple_times(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
```
tozka, line 211: same comment here. If the idea is that the method won't fail. Just add a comment at the end that it is what we are verifying.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +169,50 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+
+
+def test_call_finalize_before_get_initialized_job_input(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
```
tozka, line 218: same comment as above. what are we verifying/asserting ?

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
duyguHsnHsn: Yes, this was the idea. I will add comments to each tests that is for that.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
duyguHsnHsn: asserts added

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
duyguHsnHsn: comment about the verification is added

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +180,66 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+    assert isinstance(ip.user_global_ns["job_input"], IJobInput)
+
+
+def test_call_finalize_multiple_times(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
+
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    # verifying that calling finalize multiple times won't produce any errors (the method will not fail)
```
tozka, line 236: I think that's not enough verification. 

We want to 

A) Assert that the new job_input is different than the now finalized , right ? 
B) Assert that finalize has been called for both new and old job_input.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +180,66 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+
+
+def test_call_finalize_before_get_initialized_job_input(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
+
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    # verifying that calling finalize before get_initialized_job_input won't produce errors(the method will not fail)
```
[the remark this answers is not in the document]
tozka: I think that's not enough verification. 

We want to 

A) Assert that the new job_input is different than the now finalized one, right ? 
B) Assert that finalize has been called for both new and old job_input.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
tozka: It might be good idea to verify that the second VDK.finalize does nothing. We do not want to finalize the job twice.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -23,15 +23,17 @@ def ip(session_ip):
 
 def test_load_vdk_with_no_arguments(ip):
     ip.run_line_magic(magic_name="reload_VDK", line="")
+
     assert ip.user_global_ns["VDK"] is not None
     assert isinstance(ip.user_global_ns["VDK"], JobControl)
 
 
 def test_load_vdk_with_valid_argument(ip):
     ip.run_line_magic(magic_name="reload_VDK", line="--name=test")
```
tozka, line 32: You are calling reload_VDK every time. Why not just move it toe the fixture ?

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -169,3 +180,66 @@ def test_extension_with_pure_sql_job(ip, tmpdir):
...
+def test_get_initialized_job_input_after_finalize(ip):
+    ip.run_line_magic(magic_name="reload_VDK", line="")
+
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+    ip.get_ipython().run_cell("VDK.finalize()")
+    ip.get_ipython().run_cell("job_input = VDK.get_initialized_job_input()")
+
+    assert ip.user_global_ns["job_input"] is not None
```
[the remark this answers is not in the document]
duyguHsnHsn: the test was supposed to verify that no errors occur if we want to get job_input for the second time. It does not test whether it is different or not. I might add a test for checking the difference , too.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
duyguHsnHsn: the test is about testing the first finalise(), not the second. The first one does nothing (even though there is no initialized job calling finalise will not end with an error).

tozka: I mean this method test_call_finalize_multiple_times and the next method test_call_finalize_before_get_initialized_job_input are the same ?

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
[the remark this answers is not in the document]
tozka: There's no need to add multiple tests. We can use multiple asserts verifications for different things in the same test if the setup(arrange) and the code under test (act) are the same.

Otherwise we have a lot of code duplication in the test which is not good.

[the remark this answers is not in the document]
tozka: alternatively you can extract the common code in a method and call it from both tests, but if the assert part is not complex simple overkill

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
duyguHsnHsn: I changed them

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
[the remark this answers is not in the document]
duyguHsnHsn: Changes were done on the test

tozka (approved):

gageorgiev (approved):

PR #1756, opened 2023-03-20
VEP-1493: Vault Integration For Secrets Storage

dakodakov:
This VEP proposes integration with Hashicorp Vault for secrets storage.


specs/vep-1493-vault-integration/README.md
```diff
@@ -0,0 +1,223 @@
...
+          $ref: '#/components/responses/Secrets'
+      operationId: dataJobSecretsRead
+      summary: Get Data Job secrets.
+      description: <i>(Introduced in v1.0)</i>
+    put:
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DataJobSecrets'
```
mivanov1988, line 101: Add the definition of DataJobSecrets

specs/vep-1493-vault-integration/README.md
```diff
@@ -0,0 +1,223 @@
...
+        description: Data job deployment id.
+        schema:
+          $ref: '#/components/schemas/DataJobDeploymentId'
+        in: path
+        required: true
+```
+
+When storing secrets, the request body is expected to provide a JSON object, containing all secrets:
+
+```json
```
mivanov1988, line 136: Have you considered the following format?

```json
{
  "secrets": [
    {"name": "host", "value": "host1.com"},
    {"name": "user", "value": "user1"},
    {"name": "pass", "value": "1234"}
  ]
}
```

specs/vep-1493-vault-integration/README.md
```diff
@@ -0,0 +1,223 @@
...
+{
+  "secret1": {},
```
mivanov1988, line 138: Why the secret value is an object?

specs/vep-1493-vault-integration/diagrams/high_level_design.puml
```diff
@@ -0,0 +1,31 @@
...
+User(admin, "Administrator", " ")
+User(engineer, "User", " ")
+
+rectangle "K8s Cluster" {
+   component "Control Service" as cs #teal {
+    rectangle "Properties " as properties_service #teal
+   }
+}
+
+rectangle "  <&cog*4>\nVDK SDK" as sdk #teal
```
tozka, line 15: SDK or UI ? ,   the flow using UI and CLI would be the same

specs/vep-1493-vault-integration/diagrams/high_level_design.puml
```diff
@@ -0,0 +1,31 @@
...
+database "\nCockroachDB\n" as DB
+rectangle "<$vault>\n  Vault " as vault #yellow
+
+engineer -- sdk : get/set property\nfor a data job
+sdk -- properties_service : API call
```
tozka, line 20: You may indicate auth. in some ways. Maybe `API Call (User Oauth2 token)`

specs/vep-1493-vault-integration/README.md
```diff
@@ -0,0 +1,223 @@
...
+- [High-level design](#high-level-design)
+- [API Design](#api-design)
+- [Detailed design](#detailed-design)
+- [Implementation stories](#implementation-stories)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Provide integration with Hashicorp Vault for secure storage of secrets. Control Service operators will be able to
+optionally configure a vault instance to be used for storing data job properties which are marked as secret.
```
ivakoleva, line 35: a. It seems the proposal is dual properties storage, where entries marked as secret go to another data store. Is that correct?

b. Is the feature meant to be reused by other plugins, swapping the secret storage to another one than Vault? Asking about properties conditional multi-storage routing, and reusability

c. Is there any reason for excluding the more basic use case for storing all properties to Vault?

specs/vep-1493-vault-integration/README.md
dakodakov: Yes, however it is different than the current properties format which means that either we'll have two different interfaces for the users, or we'll have to convert them in the code...

tozka: I agree with Dako. If we want to change the format, we should do it for both.

[Vault supports JSON format](https://developer.hashicorp.com/vault/docs/commands/kv/put). I suggest we use it And store in in the same format as we do in Properties (JSON).  I don't believe we should limit what structure the user choose to store their secrets in (only string). 


This reminds me that we should consider size limits (I read that a common limitation for many vault storage backends is around 512 KB per key-value pair in Vault). That would go in Detail design section. But it came to mind now.

mivanov1988: As you said the interface will remain the same, we will convert them in the code. I am not saying that this is better or not, but if you think it is better I don't see any problem to proceed with it.

specs/vep-1493-vault-integration/diagrams/high_level_design.puml
dakodakov: fixed.

specs/vep-1493-vault-integration/diagrams/high_level_design.puml
dakodakov: done.

specs/vep-1493-vault-integration/diagrams/high_level_design.puml
```diff
@@ -0,0 +1,31 @@
...
+admin -- cs : Configure \nVault \ninstance
+properties_service -- vault : stores\nsecrets
```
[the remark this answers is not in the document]
dakodakov: done.

specs/vep-1493-vault-integration/README.md
mivanov1988: In my opinion, we are a bit limited with this approach since we don't have the ability to update only one of the properties/secrets (although it is possible to update a particular field in the JSON column in the database, it is hard to be implemented). In the case of a data job with a lot of properties/secrets that updates only one property/secret, we need to load all properties/secrets and then send them back to the CS, which will be slightly inefficient.

specs/vep-1493-vault-integration/README.md
ivakoleva: The a. and b. are referring the current design suggested.

I guess the separate PR is about c. swapping the entire backing storage that is simpler use case. I feel like connecting to Vault storage is an implementation step, preceding the conditional routing step.

specs/vep-1493-vault-integration/README.md
radostkassova: I think that we need to support CRUD operations in the API for a single secret.
Why because when updating the secrets in bulk, you need to retrieve the other values and they offer the user in  VDK/UI to edit all values.

This was an issue with the UI implementation when it was started in the summer.

radostkassova: I believe this also need to be added for properties, as the UI to be consistent or very close.

radostkassova: in the VEP, it is suggested to support a command "vdk secrets --set "my-password" - which I suggest - single secret update, how is this going to be supported if the API supports bulk edit?
{
  "secrets": [
    {"name": "host", "value": "host1.com"},
    {"name": "user", "value": "user1"},
    {"name": "pass", "value": "1234"}
  ]
}

tozka: > In my opinion, we are a bit limited with this approach since we don't have the ability to update only one of the properties/secrets

I am not sure we are that limited 
By Contract ([API defintion](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/model/apidefs/datajob-api/api.yaml#L1222) and [Job Input SDK definition](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/job_input.py#L32) ) the properties are dictionary where the key must be string, the value can be primtiive type, list or dict. 


In other words Properties type is Dict[str, PropertyValue]) where PropertyValue = Union[int, float, str, list, dict, None] ([here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py#LL70C46-L70C71))
and [here](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/properties/controller/DataJobsPropertiesController.java#LL36C61-L36C71)


So it will not be a breaking change to introduce:  '/data-jobs/for-team/{team_name}/jobs/{job_name}/deployments/{deployment_id}/properties/top_level_key/ = PropertyValue

should we decide to Impl wise it would still need to fetch all properties and update them. But it's db - to service communication now behind interface. 
As Properties DB model is implemntation detail we can always change it if we are not happy.

dakodakov: I've added a couple of points in the Alternatives section for clarity.

specs/vep-1493-vault-integration/README.md
dakodakov: I've added a couple of points in the Alternatives section for clarity.

tozka (approved):
Looks good to me. I made some very small (formatting) comments.

My main 3 comments (and those can be addressed in separate PR): 
 
1. ) Would vdk secrets CLI be the same as vdk properties (with only difference not allowing values on command line) ? Right ?

2) I'd like us to think (and hence describe) a bit about performance, availability and security. 
  -  Availability. If Vault is down or slow ,  
     As vault is external service I really recommend implementing read-through cache with fallback (read from storage and on error from cache) . Because during a job run we only will read secrets. And writes would be outside of data job run. That would help keep the SLA of job runs , so in case Vault is unavailable we will rely on the server cache to make sure we still get return the values. As values would tend to be changed fairly rarely - having even hours exipry time (before failing) would be enough . (Spring CacheManager should made this easy)
      
   - Performance - the secrets will be loaded and cached at teh start of the SDK run job. (the same is currently for properties ), right? 
   
  3) The SDK API changes are missing . We will add new interface ISecrets in job_input (https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/job_input.py) , and then we can re-use basically the whole properties builtin plugin - https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties with difference possibly having SecretsRouter. That's teh direction my thoughts go but other options are ok.

specs/vep-1493-vault-integration/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+def run(job_input):
+   secrets = job_input.get_all_secrets()
+   secrets['MY_TOKEN'] = get_new_token()
+   job_input.set_all_secrets()
+```
+
+#### Storing/retrieving secrets via the CLI
+
+We are going to introduce a new set of commands for secrets, similar to "vdk properties":
```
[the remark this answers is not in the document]
dakodakov: yes. extended explanation for clarification.

dakodakov:
> 

1. Yes
2. 
- Availability: let's not solve a problem we currently don't have
- Performance: yes, that is correct
3. Fixed - updated the readme.

mivanov1988:
The final version looks great and I am very satisfied with the job you did.

specs/vep-1493-vault-integration/README.md
mivanov1988: Although I think it would be even better to improve the API format, it sounds outside the scope of the VEP. Let's be consistent with the Properties API.

PR #1760, opened 2023-03-20
vdk-plugins: local differential privacy plugin 

murphp15:
# Why?
In the last few years we have seen the emergence of Big Data within the data space. From making and validating business decisions to training machine learning models, data has shown itself to be vital to the operation of many businesses.

However, this trend has coincided with a greater interest in data privacy. Users are interested more than ever in protecting their personal data.

This raises the question, how can we collect as much data as possible, while respecting our users’ privacy? Different answers have been given to this query.

Historically, data teams would simply obfuscate personally identifiable information, such as names for example, in their datasets; however this is not sufficient for protecting a user’s privacy, as a malicious actor can easily pick out quasi-identifying elements within the dataset.

Differential privacy has been proposed as a solution to this problem. Simply put, the idea is that we introduce some controlled amount of statistical noise within the data that we collect; this means that the data cannot be used reliably to identify users, while at the same time preserving its usefulness for decision making or training ML models.

However, a main obstacle to differential privacy’s popularisation is the lack of tooling to support it: while some libraries for data processing or plugins for databases exist, there are few integrations with data engineering infrastructure frameworks that can be picked up and easily used.

# What?
This PR contains a plugin that can be used to add statistically noise to enum like data and/or boolean data during the ingestion phase.

# How has this been tested?
new unit tests added

New Feature

# What is missing? 
1. at the moment the amount of noise is not configurable but the code is written in a way to make it easily configurable in the future.


projects/vdk-plugins/vdk-local-differential-privacy/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+At the moment we support noise on two different types of data. 
+1. [Boolean data with Random response](https://programming-dp.com/ch13.html#randomized-response)
+2. [Enum type data with Unary encoding](https://programming-dp.com/ch13.html#unary-encoding)
+
+
+## Usage
+
+```
+pip install local-differential-privacy
```
tozka, line 20: Reserve the name in pypi.org `local-differential-privacy` If you are providing instructions on people to install something. You need to make sure that we cannot suffer dependency confusion attack (https://fossa.com/blog/dependency-confusion-understanding-preventing-attacks/) 

I had to modify a few examples recently for that.

tozka (approved):
Looks good to me. Looking forward to seeing it in the presentation.

You should think of some cool use case to based it around I think. 

I asked chat GPT (based on your readme) and it offered me those options 

> 
> A hospital has collected patient data that includes sensitive medical information such as diagnoses, treatments, and medications. The hospital wants to share this data with a research organization to help advance medical research. However, the hospital is concerned about patient privacy and wants to ensure that the data is anonymized before sharing it. The hospital can use this plugin to apply differential privacy techniques to the data before sending it to the research organization.
> 
> Financial Industry: A bank has collected customer data that includes personal and financial information such as names, addresses, account numbers, and transaction histories. The bank wants to share this data with a credit bureau to help improve credit scores for its customers. However, the bank is concerned about customer privacy and wants to ensure that the data is anonymized before sharing it. The bank can use this plugin to apply differential privacy techniques to the data before sending it to the credit bureau.
>

projects/vdk-plugins/vdk-local-differential-privacy/tests/test_random_response_ingestion.py
```diff
@@ -0,0 +1,59 @@
...
+
+    payload = destination_plugin.payloads[0].payload
+    count_of_true = np.count_nonzero([a["sensitive_key"] for a in payload])
+    count_of_false = len(payload) - count_of_true
+
+    # assert that the plugins have no unintended impact on other columns.
+    assert list(payload[0].keys()) == ["str_key", "sensitive_key"]
+    assert payload[0]["str_key"] == "str"
+
+    # Within the test data the actual of true vs false is 0 to 60.
```
ivakoleva, line 34: nit: seems that a data observer is still able to distinguish the correct response, by simply aggregating the data?

projects/vdk-plugins/vdk-local-differential-privacy/tests/test_random_response_ingestion.py
```diff
@@ -0,0 +1,59 @@
...
+    # within the noisy data we want to test to ensure that false is still significantly higher count.
```
ivakoleva, line 35: nit: what about noice shift balance to the left or right? For example, a person fetches a data subset, or a person does select a partial previous data batch then appends this batch?

ivakoleva (approved):

PR #1761, opened 2023-03-21
Control Service: Add supported python version configuration

doks5:
As part of the effort to support different python versions for data job deployments in Versatile Data Kit, we need to be able to configure what python versions are supported by the Control Service.

This change adds new configuration to the Control Service's helm chart, which allows administrators of the service to specify what python versions are allowed for deployments and what vdk and job base images are to be used for the corresponding python versions. The configuration is in the form of a map as described in VEP-1739 and is injected in the Control Service deployment as a `DATAJOBS_DEPLOYMENT_SUPPORTED_PYTHON_VERSIONS` environment variable.

The configuration is also added to the service's application properties files.

Additionally, a set of methods is implemented to support a variety of use cases related to accecssing and managing the supported python versions configuration.

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
@@ -265,10 +265,6 @@ spec:
               value: "{{ .Values.dataJob.deployment.initContainer.resources.limits.memory }}"
             - name: DATAJOBS_DEPLOYMENT_INITCONTAINER_RESOURCES_LIMITS_CPU
               value: "{{ .Values.dataJob.deployment.initContainer.resources.limits.cpu }}"
-            {{- if .Values.dataJob.deployment.supportedPythonVersions }}
-            - name: DATAJOBS_DEPLOYMENT_SUPPORTED_PYTHON_VERSIONS
-              value: "{{ .Values.dataJob.deployment.supportedPythonVersions }}"
-            {{- end}}
 
 
 {{- if .Values.extraVars }}
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
@@ -265,6 +265,10 @@ spec:
               value: "{{ .Values.dataJob.deployment.initContainer.resources.limits.memory }}"
             - name: DATAJOBS_DEPLOYMENT_INITCONTAINER_RESOURCES_LIMITS_CPU
               value: "{{ .Values.dataJob.deployment.initContainer.resources.limits.cpu }}"
+            {{- if .Values.dataJob.deployment.supportedPythonVersions }}
+            - name: DATAJOBS_DEPLOYMENT_SUPPORTED_PYTHON_VERSIONS
+              value: "{{ .Values.dataJob.deployment.supportedPythonVersions }}"
+            {{- end}}
 
 
 {{- if .Values.extraVars }}
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
@@ -47,7 +47,6 @@ public class JobImageDeployer {
   @Value("${datajobs.deployment.readOnlyRootFilesystem:}")
   private boolean readOnlyRootFilesystem;
 
-
   private static final String VOLUME_NAME = "vdk";
   private static final String VOLUME_MOUNT_PATH = "/vdk";
   private static final String EPHEMERAL_VOLUME_NAME = "tmpfs";
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
@@ -47,6 +47,7 @@ public class JobImageDeployer {
   @Value("${datajobs.deployment.readOnlyRootFilesystem:}")
   private boolean readOnlyRootFilesystem;
 
+
   private static final String VOLUME_NAME = "vdk";
   private static final String VOLUME_MOUNT_PATH = "/vdk";
   private static final String EPHEMERAL_VOLUME_NAME = "tmpfs";
@@ -95,6 +96,7 @@ public class JobImageDeployer {
       String lastDeployedBy) {
     Validate.notNull(jobDeployment, "jobDeployment should not be null");
     Validate.notNull(jobDeployment.getImageName(), "Image name is expected in jobDeployment");
+
     try {
       return updateCronJobWithNotification(
           dataJob, jobDeployment, sendNotification, lastDeployedBy);
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
@@ -95,7 +95,6 @@ public class JobImageDeployer {
       String lastDeployedBy) {
     Validate.notNull(jobDeployment, "jobDeployment should not be null");
     Validate.notNull(jobDeployment.getImageName(), "Image name is expected in jobDeployment");
-
     try {
       return updateCronJobWithNotification(
           dataJob, jobDeployment, sendNotification, lastDeployedBy);
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -24,7 +24,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class SupportedPythonVersions {
 
-  public record DeploymentDetails(String baseImage, String vdkImage) {}
+  record DeploymentDetails(String baseImage, String vdkImage) {}
 
   @Value("#{${datajobs.deployment.supportedPythonVersions:{}}}")
   private Map<String, DeploymentDetails> supportedPythonVersions;
@@ -50,9 +50,7 @@ public class SupportedPythonVersions {
    * @return A set of all python versions supported by the Control Service.
    */
   public Set<String> getSupportedPythonVersions() {
-    return Optional.ofNullable(supportedPythonVersions)
-        .map(Map::keySet)
-        .orElse(Collections.emptySet());
+    return supportedPythonVersions.keySet();
   }
 
   /**
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -106,13 +106,8 @@ public class SupportedPythonVersions {
    * from the datajobs.deployment.defaultPythonVersion application property.
    *
    * @return a string indicating the default python version supported by the Control Service.
-   * @throws InvalidParameterException
    */
-  public String getDefaultPythonVersion() throws InvalidParameterException {
-    if (defaultPythonVersion != null) {
-      return defaultPythonVersion;
-    } else {
-      throw new InvalidParameterException("Missing value for defaultPythonVersion!");
-    }
+  public String getDefaultPythonVersion() {
+    return defaultPythonVersion;
   }
 }
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -5,7 +5,7 @@
 
 package com.vmware.taurus.service.deploy;
 
-import java.util.HashMap;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -14,14 +14,19 @@ import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Component;
 import lombok.extern.slf4j.Slf4j;
 
-/** Handles operations related to supported python versions for data job deployments. */
+/**
+ * Handles operations related to supported python versions for data job deployments.
+ * The class contains utility methods which provide functionality to read the configuration
+ * related to the python versions supported by the Control Service.
+ * These utility methods are meant to be used in other components, as needed.
+ * */
 @Component
 @RequiredArgsConstructor
 @Slf4j
 public class SupportedPythonVersions {
 
   @Value("#{${datajobs.deployment.supportedPythonVersions}}")
-  private HashMap<String, HashMap<String, String>> supportedPythonVersions;
+  private Map<String, Map<String, String>> supportedPythonVersions;
 
   @Value("${datajobs.vdk.image}")
   private String vdkImage;
@@ -36,12 +41,12 @@ public class SupportedPythonVersions {
   /**
    * Check if the python_version passed by the user is supported by the Control Service.
    *
-   * @param python_version python version passed by the user.
+   * @param pythonVersion python version passed by the user.
    * @return true if the version is supported, and false otherwise.
    */
-  public boolean isPythonVersionSupported(String python_version) {
+  public boolean isPythonVersionSupported(String pythonVersion) {
     if (!supportedPythonVersions.isEmpty()) {
-      return supportedPythonVersions.containsKey(python_version);
+      return supportedPythonVersions.containsKey(pythonVersion);
     } else {
       return false;
     }
@@ -72,12 +77,12 @@ public class SupportedPythonVersions {
    * according to the configuration, the base image corresponding to the python_version is returned.
    * Otherwise, the default base image name as set in deploymentDataJobBaseImage is returned.
    *
-   * @param python_version a string indicating the python version passed by the user
+   * @param pythonVersion a string indicating the python version passed by the user
    * @return a string of the data job base image.
    */
-  public String getJobBaseImage(String python_version) {
-    if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(python_version)) {
-      return supportedPythonVersions.get(python_version).get(BASE_IMAGE);
+  public String getJobBaseImage(String pythonVersion) {
+    if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(pythonVersion)) {
+      return supportedPythonVersions.get(pythonVersion).get(BASE_IMAGE);
     } else {
       return deploymentDataJobBaseImage;
     }
@@ -93,12 +98,12 @@ public class SupportedPythonVersions {
    * configuration, the vdk image corresponding to the python_version is returned. Otherwise, the
    * default vdk image name as set in vdkImage is returned.
    *
-   * @param python_version a string indicating the python version passed by the user
+   * @param pythonVersion a string indicating the python version passed by the user
    * @return a string of the data job base image.
    */
-  public String getVdkImage(String python_version) {
-    if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(python_version)) {
-      return supportedPythonVersions.get(python_version).get(VDK_IMAGE);
+  public String getVdkImage(String pythonVersion) {
+    if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(pythonVersion)) {
+      return supportedPythonVersions.get(pythonVersion).get(VDK_IMAGE);
     } else {
       return vdkImage;
     }
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -65,11 +65,9 @@ public class SupportedPythonVersions {
    */
   public List<String> getSupportedPythonVersions() {
     if (!supportedPythonVersions.isEmpty()) {
-      return supportedPythonVersions
-              .keySet()
-              .stream()
-              .map((obj) -> Objects.toString(obj, null))
-              .collect(Collectors.toList());
+      return supportedPythonVersions.keySet().stream()
+          .map((obj) -> Objects.toString(obj, null))
+          .collect(Collectors.toList());
     } else {
       try {
         return Arrays.asList(getDefaultPythonVersion());
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -7,7 +7,6 @@ package com.vmware.taurus.service.deploy;
 
 import java.security.InvalidParameterException;
 import java.util.*;
-import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.annotation.Value;
 import lombok.RequiredArgsConstructor;
@@ -45,15 +44,15 @@ public class SupportedPythonVersions {
 
   /**
    * Returns a set of the python versions supported by the Control Service, in the format: [3.7,
-   * 3.8, ...]. If the supportedPythonVersions configuration is not set, the method returns an
-   * empty set.
+   * 3.8, ...]. If the supportedPythonVersions configuration is not set, the method returns an empty
+   * set.
    *
    * @return A set of all python versions supported by the Control Service.
    */
   public Set<String> getSupportedPythonVersions() {
     return Optional.ofNullable(supportedPythonVersions)
-            .map(Map::keySet)
-            .orElse(Collections.emptySet());
+        .map(Map::keySet)
+        .orElse(Collections.emptySet());
   }
 
   /**
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -31,11 +31,9 @@ public class SupportedPythonVersions {
   @Value("${datajobs.deployment.defaultPythonVersion}")
   private String defaultPythonVersion;
 
-
   private static final String VDK_IMAGE = "vdkImage";
 
   private static final String BASE_IMAGE = "baseImage";
-
 
   /**
    * Check if the pythonVersion passed by the user is supported by the Control Service.
@@ -70,9 +68,9 @@ public class SupportedPythonVersions {
 
   /**
    * Returns the name of the data job base image as stored in the docker registry. If
-   * supportedPythonVersions is set, and the pythonVersion passed by the user is supported
-   * according to the configuration, the base image corresponding to the pythonVersion is returned.
-   * Otherwise, the default base image is returned.
+   * supportedPythonVersions is set, and the pythonVersion passed by the user is supported according
+   * to the configuration, the base image corresponding to the pythonVersion is returned. Otherwise,
+   * the default base image is returned.
    *
    * @param pythonVersion a string indicating the python version passed by the user
    * @return a string of the data job base image.
@@ -81,7 +79,9 @@ public class SupportedPythonVersions {
     if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(pythonVersion)) {
       return supportedPythonVersions.get(pythonVersion).get(BASE_IMAGE);
     } else {
-      log.warn("An issue with the passed pythonVersion or supportedPythonVersions configuration has occurred. Returning default job base image");
+      log.warn(
+          "An issue with the passed pythonVersion or supportedPythonVersions configuration has"
+              + " occurred. Returning default job base image");
       return getDefaultJobBaseImage();
     }
   }
@@ -92,9 +92,9 @@ public class SupportedPythonVersions {
 
   /**
    * Returns the name of the vdk image as stored in the docker registry. If supportedPythonVersions
-   * is set, and the pythonVersion, passed by the user, is supported according to the
-   * configuration, the vdk image corresponding to the pythonVersion is returned. Otherwise, the
-   * default vdk image is returned.
+   * is set, and the pythonVersion, passed by the user, is supported according to the configuration,
+   * the vdk image corresponding to the pythonVersion is returned. Otherwise, the default vdk image
+   * is returned.
    *
    * @param pythonVersion a string indicating the python version passed by the user
    * @return a string of the vdk image.
@@ -103,7 +103,9 @@ public class SupportedPythonVersions {
     if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(pythonVersion)) {
       return supportedPythonVersions.get(pythonVersion).get(VDK_IMAGE);
     } else {
-      log.warn("An issue with the passed pythonVersion or supportedPythonVersions configuration has occurred. Returning default vdk image");
+      log.warn(
+          "An issue with the passed pythonVersion or supportedPythonVersions configuration has"
+              + " occurred. Returning default vdk image");
       return getDefaultVdkImage();
     }
   }
@@ -113,8 +115,8 @@ public class SupportedPythonVersions {
   }
 
   /**
-   * Returns the default python version supported by the Control Service. The version number is
-   * read from the datajobs.deployment.defaultPythonVersion application property.
+   * Returns the default python version supported by the Control Service. The version number is read
+   * from the datajobs.deployment.defaultPythonVersion application property.
    *
    * @return a string indicating the default python version supported by the Control Service.
    * @throws InvalidParameterException
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
@@ -23,10 +23,12 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class SupportedPythonVersions {
 
-  record DeploymentDetails(String baseImage, String vdkImage) {}
+  private static String BASE_IMAGE = "baseImage";
+
+  private static String VDK_IMAGE = "vdkImage";
 
   @Value("#{${datajobs.deployment.supportedPythonVersions:{}}}")
-  private Map<String, DeploymentDetails> supportedPythonVersions;
+  private Map<String, Map<String, String>> supportedPythonVersions;
 
   @Value("${datajobs.deployment.defaultPythonVersion}")
   private String defaultPythonVersion;
@@ -38,7 +40,7 @@ public class SupportedPythonVersions {
    * @return true if the version is supported, and false otherwise.
    */
   public boolean isPythonVersionSupported(String pythonVersion) {
-    return supportedPythonVersions.containsKey(pythonVersion);
+    return !supportedPythonVersions.isEmpty() && supportedPythonVersions.containsKey(pythonVersion);
   }
 
   /**
@@ -49,7 +51,9 @@ public class SupportedPythonVersions {
    * @return A set of all python versions supported by the Control Service.
    */
   public Set<String> getSupportedPythonVersions() {
-    return supportedPythonVersions.keySet();
+    return Optional.ofNullable(supportedPythonVersions)
+            .map(Map::keySet)
+            .orElse(Collections.emptySet());
   }
 
   /**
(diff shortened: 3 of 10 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/** Handles operations related to supported python versions for data job deployments. */
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class SupportedPythonVersions {
+
+  @Value("#{${datajobs.deployment.supportedPythonVersions}}")
+  private HashMap<String, HashMap<String, String>> supportedPythonVersions;
```
ivakoleva, line 24: I think the requirement to enumerate specific versions supported, is not quite practical. I would imagine rather a condition to narrow those, in case needed. 
May I propose an agile approach towards describing versions compatibility requirements? 
Find https://developer.hashicorp.com/terraform/language/expressions/version-constraints

doks5: Hi, @ivakoleva 

I think I understand what you mean, but am not sure how practical it is from the administrator point of view. Data Job base images are build for specific major python versions like 3.7, 3.9, etc. Same applies to vdk images, and we need to take into account that users would most likely end up installing packages compatible with specific python versions only.

It seems to me that if we add some looser constraints to the versioning, we may get ourselves in some mess, where we need to do even more work to ensure we don't hit corner cases, deploying data jobs with the wrong images.

If you have some specific suggestion/example in mind, I am open to discuss it.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Returns a string of the python versions supported by the Control Service, in the format: [3.7,
+   * 3.8, ...]. If the supportedPythonVersions configuration is not set, the method returns the
+   * default python version set in the deploymentDataJobBaseImage configuration property.
+   *
+   * @return A string of all python versions supported by the Control Service
+   */
+  public String getSupportedPythonVersions() {
```
tozka, line 57: This looks strange. Why return a string and not a list ?

What is the purpose of this method . How is it goingot be used. I don't see it used anywhere

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
```diff
@@ -95,6 +95,7 @@ public boolean scheduleJob(
       String lastDeployedBy) {
     Validate.notNull(jobDeployment, "jobDeployment should not be null");
     Validate.notNull(jobDeployment.getImageName(), "Image name is expected in jobDeployment");
+
```
tozka, line 98: Revert this , seems like accidentally adding a line

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -265,6 +265,10 @@ spec:
               value: "{{ .Values.dataJob.deployment.initContainer.resources.limits.memory }}"
             - name: DATAJOBS_DEPLOYMENT_INITCONTAINER_RESOURCES_LIMITS_CPU
               value: "{{ .Values.dataJob.deployment.initContainer.resources.limits.cpu }}"
+            {{- if .Values.dataJob.deployment.supportedPythonVersions }}
+            - name: DATAJOBS_DEPLOYMENT_SUPPORTED_PYTHON_VERSIONS
+              value: "{{ .Values.dataJob.deployment.supportedPythonVersions }}"
```
tozka, line 270: I think you need toJson . https://helm.sh/docs/chart_template_guide/function_list/#tojson-musttojson 

How was this tested ?

I don't see where Values.dataJob.deployment.supportedPythonVersions is defined (values.yaml should be changed)

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.HashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.springframework.beans.factory.annotation.Value;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+import lombok.extern.slf4j.Slf4j;
+
+/** Handles operations related to supported python versions for data job deployments. */
```
tozka, line 17: I struggle to understand the purpose of this class. 
 
Let's improve the java doc - we should include a clear and concise description of the class, its purpose, and any important information related to its usage or behavior.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Value("#{${datajobs.deployment.supportedPythonVersions}}")
+  private HashMap<String, HashMap<String, String>> supportedPythonVersions;
+
+  @Value("${datajobs.vdk.image}")
+  private String vdkImage;
+
+  @Value("${datajobs.deployment.dataJobBaseImage:python:3.9-slim}")
+  private String deploymentDataJobBaseImage;
+
```
tozka, line 31: Why do we need these ?

tozka (commented):
I asked a few questions because I was confused a bit. Let's see them answered first and I will make a second pass.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+import org.springframework.stereotype.Component;
+import lombok.extern.slf4j.Slf4j;
+
+/** Handles operations related to supported python versions for data job deployments. */
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class SupportedPythonVersions {
+
+  @Value("#{${datajobs.deployment.supportedPythonVersions}}")
```
mivanov1988, line 23: Use Map instead of HashMap.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private static final String BASE_IMAGE = "baseImage";
+
+  /**
+   * Check if the python_version passed by the user is supported by the Control Service.
+   *
+   * @param python_version python version passed by the user.
+   * @return true if the version is supported, and false otherwise.
+   */
+  public boolean isPythonVersionSupported(String python_version) {
```
mivanov1988, line 42: Camel case

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
ivakoleva: @doks5 I see, what you are concerned with, please correct me if I'm missing something.

From one side, we have the optional constraints on possible base images. Admins may set those rules. For example, python >=3.7 may be a requirement for a VDK deployment in env X.

From the other side, we have an optional per-job configurable python version. It is only valid if satisfying admin constraints. If not customized, will default to latest possible. For example, the image will default to 3.11, that honours the >=3.7 admin constraints. 

https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/job-base-image/Dockerfile-data-job-base#L4 the [parent or base scratch image](https://docs.docker.com/build/building/base-images/) is becoming configurable or not?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
mivanov1988: Agreed with @tozka , this method should return a list of Strings.

mivanov1988: Also, I think we should define the use cases and the behavior. 

We always have a default python version so the supportedPythonVersions should not be empty right? Somehow we should connect the default version to the list of supported versions.

How about that (or something similar)?
datajobs.deployment.supportedPythonVersions={3.7: {vdkImage: 'some-test-img', baseImage: 'python:3.7-slim'}}
datajobs.deployment.defaultPythonVersion=3.7

And we definitely need validation during the CS startup.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Returns the default python version supported by the Control Service. The version number is
+   * extracted from the datajobs.deployment.dataJobBaseImage application property. The property is
+   * set as for example, `python:3.9-slim`, and we use a regex to match `3.9` and return it to the
+   * caller.
+   *
+   * @return a string indicating the default python version supported by the Control Service.
+   * @throws IllegalArgumentException
+   */
+  public String getDefaultPythonVersion() throws IllegalArgumentException {
```
mivanov1988, line 120: I don't think it is a good approach, definitely won't work in all cases.

projects/control-service/projects/pipelines_control_service/src/main/resources/application-prod.properties
```diff
@@ -58,6 +58,7 @@ datajobs.kadmin_password=${KADMIN_PASSWORD}
 # This is the full image name of the data job builder
 datajobs.builder.image=registry.hub.docker.com/versatiledatakit/job-builder:1.1
 datajobs.deployment.dataJobBaseImage=registry.hub.docker.com/versatiledatakit/data-job-base-python-3.7:latest
+datajobs.deployment.supportedPythonVersions=${DATAJOBS_DEPLOYMENT_SUPPORTED_PYTHON_VERSIONS:{}}
```
mivanov1988, line 61: Move this to the application.properties since it is the base for all property files.

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
mivanov1988: This configuration should be exposed further on.

tozka: Yes, ok that make sense. Let's add a task (issue) for that.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Can you elaborate? The base images we build contain the python version number, and unless we introduce yet another application property, we don't have any way of knowing what version is the default.

projects/control-service/projects/pipelines_control_service/src/main/resources/application-prod.properties
doks5: Done, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Refactored, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Changed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,130 @@
...
+  private static final String BASE_IMAGE = "baseImage";
+
+  /**
+   * Check if the python_version passed by the user is supported by the Control Service.
+   *
+   * @param python_version python version passed by the user.
+   * @return true if the version is supported, and false otherwise.
+   */
+  public boolean isPythonVersionSupported(String python_version) {
+    if (!supportedPythonVersions.isEmpty()) {
```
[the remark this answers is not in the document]
doks5: In addition to the empty check, or instead of the empty check?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: These are, where we currently set the vdk and job base images to be used, so the idea is for them to act as defaults, in case the supportedPythonVersions configuration is not set.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Expanded the doc.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Updated the method to return a list of strings, instead of one big string.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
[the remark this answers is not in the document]
mivanov1988: Instead

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
mivanov1988: Such logic is error-prone if the administrator passes the image without the python version in it.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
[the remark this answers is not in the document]
doks5: Fixed, thanks

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
doks5: Removed this for the time being and created an issue for it - https://github.com/vmware/versatile-data-kit/issues/1776

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
tozka: Why not provide default supported python versions using quickstart-vdk as the default vdk image are the standard python containers: python:3.10 

quickstart-vdk is the default image used anyhow , we simply would change the . 

I understand if you want to keep it for backwards compatibility reasons. But we should deprecate it. We should strive to remove unnecessary elements by keeping only essential features .

I might be barking the wrong tree here.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
tozka: I am in favour of `datajobs.deployment.defaultPythonVersion` property and deprecating those two. 

Overall we add one and remove 2 so we are one ahead :)

If user do not specify datajobs.deployment.defaultPythonVersion then we could default to latest supported version. 
If we specify default version that is not supported we could default to closest one (and generate error message)

I think this is better because 
A) As [Zen of Python](https://peps.python.org/pep-0020/) says "explicit is better than implicit"
B) this it would be easier to maintain and update, as the Python version can be changed independently of the base image name 
C) Code is clearer and less error-prone (one advice from me - always be wary of introducing parsing logic unless you follow well-defined standards (RFC / IEEE level)  and it guarantees that something you need will be there in the format.)

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Removed the properties, thanks.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Did a complete refactor.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: @tozka @mivanov1988 

Did a complete refactoring of the logic and removed the use of the `datajobs.vdk.image` and `datajobs.deployment.dataJobBaseImage` application properties, and replaced them with `datajobs.deployment.supportedPythonVersions` and `datajobs.deployment.defaultPythonVersion` properties. Could you, please, do another pass of the PR? Thank you

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+   * Returns the name of the data job base image as stored in the docker registry. If
+   * supportedPythonVersions is set, and the pythonVersion passed by the user is supported according
+   * to the configuration, the base image corresponding to the pythonVersion is returned. Otherwise,
+   * the default base image is returned.
+   *
+   * @param pythonVersion a string indicating the python version passed by the user
+   * @return a string of the data job base image.
+   */
+  public String getJobBaseImage(String pythonVersion) {
+    if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(pythonVersion)) {
```
mivanov1988, line 79:
```suggestion
    if (isPythonVersionSupported(pythonVersion)) {
```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+    } else {
+      log.warn(
+          "An issue with the passed pythonVersion or supportedPythonVersions configuration has"
+              + " occurred. Returning default job base image");
+      return getDefaultJobBaseImage();
+    }
+  }
+
+  public String getDefaultJobBaseImage() {
+    return supportedPythonVersions.get(defaultPythonVersion).get(BASE_IMAGE);
```
mivanov1988, line 90: Add null check

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+   * Returns the name of the vdk image as stored in the docker registry. If supportedPythonVersions
+   * is set, and the pythonVersion, passed by the user, is supported according to the configuration,
+   * the vdk image corresponding to the pythonVersion is returned. Otherwise, the default vdk image
+   * is returned.
+   *
+   * @param pythonVersion a string indicating the python version passed by the user
+   * @return a string of the vdk image.
+   */
+  public String getVdkImage(String pythonVersion) {
+    if (!supportedPythonVersions.isEmpty() && isPythonVersionSupported(pythonVersion)) {
```
mivanov1988, line 103:
```suggestion
    if (isPythonVersionSupported(pythonVersion)) {
```

same here

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+    } else {
+      log.warn(
+          "An issue with the passed pythonVersion or supportedPythonVersions configuration has"
+              + " occurred. Returning default vdk image");
+      return getDefaultVdkImage();
+    }
+  }
+
+  public String getDefaultVdkImage() {
+    return supportedPythonVersions.get(defaultPythonVersion).get(VDK_IMAGE);
```
mivanov1988, line 114: null check

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Returns the default python version supported by the Control Service. The version number is read
+   * from the datajobs.deployment.defaultPythonVersion application property.
+   *
+   * @return a string indicating the default python version supported by the Control Service.
+   * @throws InvalidParameterException
+   */
+  public String getDefaultPythonVersion() throws InvalidParameterException {
```
mivanov1988, line 124: not needed any more

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
```diff
@@ -0,0 +1,167 @@
...
+    ReflectionTestUtils.setField(
+        supportedPythonVersions, "supportedPythonVersions", new HashMap<>());
+
+    Assertions.assertFalse(supportedPythonVersions.isPythonVersionSupported("3.7"));
+  }
+
+  @Test
+  public void isPythonVersionSupported_versionSupported() {
+    var supportedVersions = new HashMap<String, HashMap<String, String>>();
+    supportedVersions.put(
```
mivanov1988, line 36: Use Map.of

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
```diff
@@ -0,0 +1,167 @@
...
+    ReflectionTestUtils.setField(
+        supportedPythonVersions, "supportedPythonVersions", supportedVersions);
+
+    Assertions.assertTrue(supportedPythonVersions.isPythonVersionSupported("3.7"));
+  }
+
+  @Test
+  public void isPythonVersionSupported_versionNotInSupported() {
+    var supportedVersions = new HashMap<String, HashMap<String, String>>();
+    supportedVersions.put(
```
mivanov1988, line 54: Map.of

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
```diff
@@ -0,0 +1,167 @@
...
+    Assertions.assertEquals(defaultBaseImage, supportedPythonVersions.getJobBaseImage("3.11"));
+  }
+
+  @Test
+  public void getJobBaseImage_multipleSupportedVersions() {
+    var supportedVersions = generateSupportedPythonVersionsConf();
+
+    final String resultBaseImg = "python:3.8-slim";
+    ReflectionTestUtils.setField(
+        supportedPythonVersions, "supportedPythonVersions", supportedVersions);
```
mivanov1988, line 110: Create a constant for `supportedPythonVersions` and etc.

projects/control-service/projects/pipelines_control_service/src/test/resources/application-unittest.properties
```diff
@@ -25,6 +25,7 @@ datajobs.git.password=
 
 datajobs.control.k8s.namespace=${CONTROL_K8S_NAMESPACE:default}
 datajobs.deployment.k8s.namespace=${DEPLOYMENT_K8S_NAMESPACE:default}
+datajobs.deployment.supportedPythonVersions={3.7: {vdkImage: 'some-test-img', baseImage: 'python:3.7-slim'}}
```
mivanov1988, line 29: Missing default version

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+ * python versions supported by the Control Service. These utility methods are meant to be used in
+ * other components, as needed.
+ */
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class SupportedPythonVersions {
+
+  @Value("#{${datajobs.deployment.supportedPythonVersions}}")
+  private Map<String, Map<String, String>> supportedPythonVersions;
```
murphp15, line 29: I think it would be a bit nicer if you made a record for the second map. 

```
public record DeploymentDetails (String vdkImage, String baseImage) {}

private Map<String, DeploymentDetails> supportedPythonVersions;
```


It would let you remove lines 34-36.
on lines 104,114, 80 and 90 you wouldn't need to worry about the keys being missing. 
Plus I also think it documents the class a bit better. 

But I actually don't know would spring deserialise that from config so natively.

murphp15: If this is too much work you can ignore and just mark the comment as resolved.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Returns a list of the python versions supported by the Control Service, in the format: [3.7,
+   * 3.8, ...]. If the supportedPythonVersions configuration is not set, the method returns the
+   * default python version set in the defaultPythonVersion configuration property.
+   *
+   * @return A list of all python versions supported by the Control Service.
+   */
+  public List<String> getSupportedPythonVersions() {
+    if (supportedPythonVersions != null && !supportedPythonVersions.isEmpty()) {
```
murphp15, line 60: again here I think that the map should only be able to be empty or non empty. never both.

projects/control-service/projects/pipelines_control_service/src/test/resources/application-unittest.properties
murphp15: that is acceptable config because it inherits from application.properties where it is set.

murphp15:
It looks good, very small amount of changes and very easy to follow considering how much more benefit it will bring to the system,

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
mivanov1988: Even better follow Paul's comment :)

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -124,6 +124,13 @@ datajobs.status.watch.initial.delay=10000
 # On top of it the job source and its dependencies are installed for each job
 datajobs.deployment.dataJobBaseImage=python:3.9-slim
 
+# The map of python version and respective data job base and vdk images that would be
+# used for data job deployments
+datajobs.deployment.supportedPythonVersions=${DATAJOBS_DEPLOYMENT_SUPPORTED_PYTHON_VERSIONS:{}}
```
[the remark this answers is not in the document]
doks5: Good point, fixed.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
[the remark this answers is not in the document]
doks5: Refactored, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Returns a list of the python versions supported by the Control Service, in the format: [3.7,
+   * 3.8, ...]. If the supportedPythonVersions configuration is not set, the method returns the
+   * default python version set in the defaultPythonVersion configuration property.
+   *
+   * @return A list of all python versions supported by the Control Service.
+   */
+  public List<String> getSupportedPythonVersions() {
```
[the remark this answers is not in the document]
doks5: Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Did so. Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Refactored, so no needed anymore

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: I am not sure I understand. We need a method to return the default version to the users.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
doks5: Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
doks5: Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+    Assertions.assertFalse(supportedPythonVersions.isPythonVersionSupported("3.7"));
+  }
+
+  @Test
+  public void getSupportedPythonVersions_multipleSupportedVersions() {
+    var supportedVersions = generateSupportedPythonVersionsConf();
+
+    var res = new ArrayList<String>();
```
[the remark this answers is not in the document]
doks5: Fixed.

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/SupportedPythonVersionsTest.java
doks5: Done, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,131 @@
...
+  private static final String BASE_IMAGE = "baseImage";
+
+  /**
+   * Check if the pythonVersion passed by the user is supported by the Control Service.
+   *
+   * @param pythonVersion python version passed by the user.
+   * @return true if the version is supported, and false otherwise.
+   */
+  public boolean isPythonVersionSupported(String pythonVersion) {
+    if (supportedPythonVersions != null) {
```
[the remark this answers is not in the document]
doks5: Good point. Refactored, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Done, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Refactored, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Returns a set of the python versions supported by the Control Service, in the format: [3.7,
+   * 3.8, ...]. If the supportedPythonVersions configuration is not set, the method returns an empty
+   * set.
+   *
+   * @return A set of all python versions supported by the Control Service.
+   */
+  public Set<String> getSupportedPythonVersions() {
+    return Optional.ofNullable(supportedPythonVersions)
```
mivanov1988, line 53: You can just return the keySet since supportedPythonVersions has a default value.
```suggestion
    return supportedPythonVersions.keySet();
```

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -124,6 +124,13 @@ datajobs.status.watch.initial.delay=10000
...
+# The default python version, which is to be used when selecting the vdk and job base images from
+# the supportedPythonVersions for data job deployments
+datajobs.deployment.defaultPythonVersion=${DATAJOBS_DEPLOYMENT_DEFAULT_PYTHON_VERSION:3.7}
```
mivanov1988, line 132: This default configuration is not correct. DefaultPythonVersion is 3.7 but you don't have any image for it. I think the best default version here is the one you used in dev profile.

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: Fixed, thanks

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
doks5: Changed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
doks5: The base images that can be used would depend on what images are built by the administrators.
The idea is for users to be able to specify what python version they want their jobs to be deployed with, and yes, the base image will be made configurable in a separate PR. The base image will be passed based on the value in supportedPythonVersions.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/SupportedPythonVersions.java
```diff
@@ -0,0 +1,120 @@
...
+ * contains utility methods which provide functionality to read the configuration related to the
+ * python versions supported by the Control Service. These utility methods are meant to be used in
+ * other components, as needed.
+ */
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class SupportedPythonVersions {
+
+  public record DeploymentDetails(String baseImage, String vdkImage) {}
```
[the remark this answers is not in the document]
doks5: Fixed.

murphp15 (approved):

PR #1762, opened 2023-03-21
vdk-jupyter: load data before vdk operations

duyguHsnHsn:
What:
Load data to jobData and UI components before rendering them by getting the job data from current working directory and config.ini file if it exists.
 
Example how deploy job component is rendered when we are in /rest-api-job directory with config.ini file: 
Screenshot 2023-03-21 at 10 22 20

Why: Loading data before rendering the UI components reduces the need for manual input, saving time and effort for users who need to interact with job data. The interface can better adapt to the specific needs and requirements of each user. This automatic data population minimizes the risk of input errors and helps users quickly access the information they need.

Tests: jest tests


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/props.tsx
```diff
@@ -0,0 +1,19 @@
+
+
+export interface IJobPathProp {
+    /**
+     * Current Job path or the path to job's parent directory
+     */
+    jobPath: string;
```
ivakoleva, line 7: it just seems strange from OOP perspective, having an interface  field. 
Interface is a stateless abstraction
https://peps.python.org/pep-0245/

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/index.ts
```diff
@@ -11,24 +11,40 @@ import {
...
 import { updateVDKMenu } from './commandsAndMenu';
 
-import { getCurrentPathRequest } from './serverRequests';
+import { FileBrowserModel, IFileBrowserFactory } from '@jupyterlab/filebrowser';
+import { IChangedArgs } from '@jupyterlab/coreutils';
+
+/**
+ * Current working directory in Jupyter
+ */
+export var workingDirectory = '';
```
ivakoleva, line 20: nit: what's the visibility of the var?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/handlers.py
```diff
@@ -7,15 +7,28 @@
 from jupyter_server.base.handlers import APIHandler
 from jupyter_server.utils import url_path_join
 
+from .job_data import JobData
 from .vdk_options.vdk_options import VdkOption
 from .vdk_ui import VdkUI
 
 
-class RunJobHandler(APIHandler):
+class JobDataHandler(APIHandler):
```
ivakoleva, line 15: nit: the name of handler may be more meaningful, suggesting what action or event it is handling?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk-jupyterlab-extension/job_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+try:
+    import hybridcontents
+except ImportError:
+    hybridcontents = None
+
+
+class JobData:
+    """
+    class responsible for retrieving data for the job
```
ivakoleva, line 11: nit: The class name may impress an object like model or DTO. Yet the docs suggest it does something, it performs actions in favour of purely storing data. The suggestion is to have it named in a way, that suggests the action

ivakoleva (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/props.tsx
DeltaMichael: Fields in interfaces are allowed in typescript. If it looks two weird, maybe @duyguHsnHsn can replace it with a type alias instead. They're effectively the same thing, minus extensibility.

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/VdkTextInput.tsx
```diff
@@ -58,7 +58,7 @@ export default class VDKTextInput extends Component<IVdkTextInputProps> {
   private onInputChange = (event: any): void => {
     const nameInput = event.currentTarget as HTMLInputElement;
     let value = nameInput.value;
-    if (!value) value = this.props.value;
+    if (value == "") value = this.props.value;
```
DeltaMichael, line 61: nit: An empty string should be falsy, so you can try

```
if (!value) value = this.props.value;
```

DeltaMichael (approved):

tozka:
Please link github issue

tozka:
> Why: To make the Jupyter UI more user-friendly.

Please, try to be more detailed in the why section. Why it would make it more user-friendly (quick advice - learn to self-employ the 5 whys method - https://en.wikipedia.org/wiki/Five_whys) 

 If this explanation is in another place (like VEP or Github issue, then just link to it.)

In this case, I'd say:
Loading data before rendering the UI components reduces the need for manual input, saving time and effort for users who need to interact with job data. The interface can better adapt to the specific needs and requirements of each user.  This automatic data population minimizes the risk of input errors and helps users quickly access the information they need.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/index.ts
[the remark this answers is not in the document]
tozka: My experience with TypeScript is close to zero but I am learning things about TypeScript now. And it seems you should prefer let over var . 

See https://www.typescriptlang.org/docs/handbook/variable-declarations.html#let-vs-const and I agree with the argumentation. As a best practice (in any programming language) you want to minimize the scope of the declaration to be as narrow as possible)

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -92,3 +83,26 @@ export async function jobRequest(endPoint: string) {
     }
   }
 }
+
+/**
+ * Sent a POST request to the server to get information about the data job of current directory
+ */
+export async function jobdDataRequest() {
```
tozka, line 90: nit : Add a return type for the function, even if it is void.
https://www.typescriptlang.org/docs/handbook/2/functions.html#return-type-void

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/__tests__/deploy-job-component.spec.ts
```diff
@@ -2,14 +2,13 @@
...
-  IDeployJobDialogProps
-} from '../components/DeployJob';
+import DeployJobDialog from '../components/DeployJob';
 import { render, fireEvent } from '@testing-library/react';
 import { jobData } from '../jobData';
 import { VdkOption } from '../vdkOptions/vdk_options';
+import { IJobNameAndTeamProps, IJobPathProp } from '../components/props';
 
-const defaultProps: IDeployJobDialogProps = {
+const defaultProps: IJobNameAndTeamProps & IJobPathProp = {
```
tozka, line 11: nit: is there reason for 2 interfaces , why not 1  - IJobDetails or something. 

Interface called Field1&Field is a code smell a bit. You could not figure out what is the correct abstraction.

duyguHsnHsn: I changed the interface into type and created a third type which represents  Field1&Field2

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/props.tsx
duyguHsnHsn: changed with type.

PR #1763, opened 2023-03-21
control-service: Only CLI executions are "Manual"

gageorgiev:
Currently, all executions started through the Execution API are labelled
as "Manual". This makes it confusing in cases where users use the Execution
API for starting executions not through the CLI, for example through the
Airflow integration or by a Meta Job - users will still see the execution
as "Manual" in the UI.

This change ensures that only executions which were ran through the
VDK CLI are labelled as "Manual". Further work is required for
extracting and globalizing the "vdk-control-cli" string as it
currently appears as a magic string in both the vdk-control-cli
and the control-service.

Testing done: ran data job through meta job and verified its label


yonitoo (approved):
LGTM!

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
```diff
@@ -97,7 +97,13 @@ public String startDataJobExecution(
               ? jobExecutionRequest.getStartedBy() + "/" + operationContext.getUser()
               : operationContext.getUser();
       annotations.put(JobAnnotation.STARTED_BY.getValue(), startedBy);
-      annotations.put(JobAnnotation.EXECUTION_TYPE.getValue(), ExecutionType.MANUAL.getValue());
+      annotations.put(
```
tozka, line 100: I believe that the current approach to address the issue by modifying the code at this specific location may not be the most effective solution in the long run. The "manual" condition, as it stands, is quite specific, being solely based on whether the job was started using the Execution API. By relying on this criterion alone, we may face complications when considering various scenarios.

For instance, if the "vdk-control-cli" is renamed in the future, this code change would no longer function as intended. Additionally, if other workflow solutions such as Airflow DAG or Dagster are employed, our current solution may not accurately categorize the execution types as manual or scheduled.

Given these potential issues, I propose that we rethink our approach and instead focus on making adjustments to the User Interface (UI). For example instead of showing execution type in the UI , we can show started by in the UI which is more specific.

gageorgiev: I'm not in agreement with your perspective. To explain why, let's look at the types of execution from both the dev perspective and the user perspective, in a point of time before airflow integration and metajobs (i.e. the Execution API is only used for running jobs with `vdk execute --start ...`):
 - from the dev perspective, there are two kinds of executions: scheduled, which are managed by a CronJob object, and manual, which are started using the Execution API - let's refer to these as dev-scheduled and dev-manual for simplicity;
 - from the user perspective, once again there are two: scheduled, which are run according to a configured schedule, and manual, which are started with a CLI command - let's refer to these as user-scheduled and user-manual.

Until a point, there was a 1-to-1 correspondence between these two types of runs - dev-scheduled and user-scheduled referred to the same process in different terms, and same for dev-manual and user-manual. However, when we introduced the Airflow integration and subsequently the Meta Jobs package, now there is a discrepancy. Running a job through a Meta Job is dev-manual, since it's done through the Execution API, but it is also user-scheduled, since it happens (roughly) according to the schedule of the Meta Job. 

My point here is that the word "manual" is more closely tied to a user experience than it is to a particular process of execution. Labeling all runs done by the Execution API as "manual" fails on a semantic level there; it also does not correspond to the way we currently use the Execution API (there's nothing manual about Meta Job-managed executions).

To conclude, I think the "scheduled" vs "manual" distinction should only exist on a UX-level. "Manual" runs are ones intently triggered by a user - a user thinks "I want a run of this job *now*" and triggers it.

The issue you've brought up about renaming "vdk-control-cli" is one I've included a todo for - I'm not sure how to configure a global string for projects of different technologies though - Python and Java in this case.

tozka: I agree overall that the "scheduled" vs "manual" distinction should only exist on a UX-level

In hindsight, ExecutionType.MANUAL is not an appropriate name at all since it causes confusion, Clearly, you can automate it using API (like what airflow or meta jobs do). I'd change it to "ExecutionType.API" 

---

Let's also look at another scenario. 

If I start A Meta job manually (e.g click execute meta jobs from CLI or UI) : I would expect all started meta jobs to be manual as well. 

If a meta job is started on schedule, then I'd expect all jobs started by it to be "Scheduled" 
So in a way a parent meta job should be able to propagate its started by to its children so to speak.

gageorgiev: This is an interesting suggestion, however I think it is outside of the scope of this change. Changing the ExecutionType label would require changes to all three major component groups (SDK, control-service and frontend).

I see a potential solution for the label propagation suggestion (amending the Execution API to allow the API call to specify the execution type).

I think we should proceed with this change as it is, and discuss at some point creating a epic dedicated to solidifying the execution types on both levels (so scheduled vs API on a dev-level and scheduled vs manual on a user-level).

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
```diff
@@ -97,7 +97,12 @@ public String startDataJobExecution(
...
+          JobAnnotation.EXECUTION_TYPE.getValue(),
```
tozka, line 101: Considering how much we discuss it. Add a comment explaining why do we do it this way for future mantainers.

gageorgiev: Added the following comment:
```
// 'Manual' executions must have their `startedBy` follow the structure 'manual/*triggering-mechanism*'
// For example, executions started from the CLI would have `startedBy` equal to 'manual/vdk-control-cli'
// Additionally we also check if it contains 'vdk-control-cli' here for the purposes of backwards compatibility
```

tozka (approved):

PR #1775, opened 2023-03-23
vdk-server: upgrade minimum k8s cluster version and upgrade vdk-quickstart k8s version

murphp15:
# Why
When we merged this PR https://github.com/vmware/versatile-data-kit/pull/1580/files we unintentionally changed the minimum required version from 1.19 to 1.21. 

We always check for v1 versions of batch jobs. checking for that version on clusters <1.21 will result in errors.

# What
Change the docs to state that 1.21 is a minimum requirement and vdk server --install now uses a newer version of k8s. 

# How has this been tested
installed locally and tested locally

# Not handled in this PR
How do we prevent issues like this in the future? 
Maybe integration tests run against many versions of k8s?

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/README.md
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/README.md
@@ -2,7 +2,7 @@
 Versatile Data Kit is a platform that enables Data Engineers to implement automated pull ingestion (E in ELT) and batch data transformation into a database (T in ELT).
 
 ## Prerequisites
-- Kubernetes 1.19<=x>1.25 works with no config changes
+- Kubernetes 1.21<=x>1.25 works with no config changes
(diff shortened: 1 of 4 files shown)


doks5:
> When we merged this PR https://github.com/vmware/versatile-data-kit/pull/1580/files we unintentionally changed the minimum required version from 1.19 to 1.21.

If this is indeed the case, then there is a regression in the Control Service that needs to fixed ASAP, as we still have deployments on clusters with k8s <1.21
How was this discovered?

murphp15:
> How was this discovered?

quickstart vdk runs on 1.20 and was failing.

murphp15:
@doks5 
Does the following sound good? 
I am going to kill this PR. 
I am going to create a new PR which supports version of k8s older than 1.21.

doks5:
> @doks5 Does the following sound good? I am going to kill this PR. I am going to create a new PR which supports version of k8s older than 1.21.

Yup, we need to prevent the Control Service from failing on older k8s.

PR #1787, opened 2023-03-28
control-service: enable usage of aws temporary credentials

mrMoZ1:
what: added the ability to use an IAM user with Role permissions
when the data job image registry type is ECR. Configuration through control-service properties.
Previous functionality should remain unchanged if properties are unset.

why: User needs of the control-service require the Service Pattern to be introduced.

testing: Integration test that deploys a data job via the new method.


mivanov1988:
When you are planning to expose the new configuration through the Helm chart?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -124,6 +131,19 @@ public boolean buildImage(
       String imageName, DataJob dataJob, JobDeployment jobDeployment, Boolean sendNotification)
       throws ApiException, IOException, InterruptedException {
 
+    String awsSecretAccessKey = this.awsSecretAccessKey;
```
mivanov1988, line 134: How about encapsulating the AWS credentials management in AWSCredentialsService? Something like awsCredentialsService.getCredentials().

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeployTemporaryCredsIT.java
```diff
@@ -0,0 +1,173 @@
...
+      "datajobs.control.k8s.k8sSupportsV1CronJob=true",
+      "datajobs.aws.assumeIAMRole=true",
+      "datajobs.aws.RoleArn=arn:aws:iam::850879199482:role/svc.supercollider.user",
+      "datajobs.docker.registryType=ecr"
+    })
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+    classes = ControlplaneApplication.class)
+@Disabled("Draft pull request.")
+public class DataJobDeployTemporaryCredsIT extends BaseIT {
```
mivanov1988, line 53: Better follow the DataJobEphemeralStorageIT style.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -250,19 +270,25 @@ public boolean isBuildingJobInProgress(String dataJobName) {
...
+    envMap.put("IMAGE_REGISTRY_PATH", dockerRepositoryUrl);
+    envMap.put("BASE_IMAGE", deploymentDataJobBaseImage);
+    envMap.put("EXTRA_ARGUMENTS", builderJobExtraArgs);
+    envMap.put("GIT_SSL_ENABLED", Boolean.toString(gitDataJobsSslEnabled));
+
+    if (!awsSessionToken.isBlank()) {
+      // Don't include session token in env if blank.
+      envMap.put("AWS_SESSION_TOKEN", awsSessionToken);
+    }
+    return Collections.unmodifiableMap(envMap);
```
ivakoleva, line 291: nit: may hint of read-only return type

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -250,19 +270,25 @@ public boolean isBuildingJobInProgress(String dataJobName) {
...
+    envMap.put("JOB_NAME", jobName);
+    envMap.put("DATA_JOB_NAME", jobName);
+    envMap.put("GIT_COMMIT", jobVersion);
+    envMap.put("JOB_GITHASH", jobVersion);
+    envMap.put("IMAGE_REGISTRY_PATH", dockerRepositoryUrl);
+    envMap.put("BASE_IMAGE", deploymentDataJobBaseImage);
+    envMap.put("EXTRA_ARGUMENTS", builderJobExtraArgs);
+    envMap.put("GIT_SSL_ENABLED", Boolean.toString(gitDataJobsSslEnabled));
+
+    if (!awsSessionToken.isBlank()) {
```
ivakoleva, line 287: nit: is awsSessionToken nullable?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -250,19 +270,25 @@ public boolean isBuildingJobInProgress(String dataJobName) {
...
-    return Map.ofEntries(
-        entry("JOB_NAME", jobName),
-        entry("DATA_JOB_NAME", jobName),
-        entry("GIT_COMMIT", jobVersion),
-        entry("JOB_GITHASH", jobVersion),
-        entry("IMAGE_REGISTRY_PATH", dockerRepositoryUrl),
-        entry("BASE_IMAGE", deploymentDataJobBaseImage),
-        entry("EXTRA_ARGUMENTS", builderJobExtraArgs),
-        entry("GIT_SSL_ENABLED", Boolean.toString(gitDataJobsSslEnabled)));
+    var envMap = new HashMap<String, String>();
```
ivakoleva, line 277: nit: why the style change

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -206,3 +206,10 @@ upload.validation.fileTypes.allowlist=${UPLOAD_VALIDATION_FILETYPES_ALLOWLIST:}
 # If the job builder image is saved in a private docker registry then this
 # property should have the name of the secret in the env.
 datajobs.builder.registrySecret=${DATAJOBS_BUILDER_REGISTRY_SECRET:}
+
+# Variables for using temporary AWS credentials instead of long term.
+datajobs.aws.assumeIAMRole=${DATAJOBS_AWS_ASSUME_IAM_ROLE:false}
+# This variable must be set if assumeIAMRole is true
+datajobs.aws.RoleArn=${DATAJOBS_AWS_ROLE_ARN:}
+# Default value is 30 minutes
```
ivakoleva, line 214: nit: may add a short explanation about vars purpose

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeployTemporaryCredsIT.java
```diff
@@ -0,0 +1,173 @@
...
+
+  @AfterEach
+  public void cleanUp() throws Exception {
+    mockMvc
+        .perform(
+            delete(
+                    String.format(
+                        "/data-jobs/for-team/%s/jobs/%s/sources", TEST_TEAM_NAME, TEST_JOB_NAME))
+                .with(user("user")))
+        .andExpect(status().isOk());
```
ivakoleva, line 171: nit: is the test cleanup a part of the feature's functional verification?

ivakoleva (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mrMoZ1: Refactored the credential service. Turns out its credentials must be different than AWS credentials - datajobs.aws.accessKeyId and secretAccessKey used by the control service, because if the K8S cluster is hosted on a EKS instance, some API calls to the instance made by the control-service could fail.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mrMoZ1: refactored, method is now as it was

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mrMoZ1: refactored, method is now as it was

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeployTemporaryCredsIT.java
mrMoZ1: No, but if the test fails midway we need to make sure any data jobs created are deleted.

mrMoZ1:
> Are you planning to expose the new configuration through the Helm chart?

Yes, in a different PR, if there aren't any objections to the variable names.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mivanov1988: My idea is to get rid of AWS credentials/data/properties from the JobImageBuilder and provide the following interface:

```
AwsCredentials/Properties awsCredentials = awsCredentialsService.getCredentials();
```
getCredentials() will be responsible to return the correct AWS credentials/data/properties based on the certain checks which currently are distributed between JobImageBuilder and AwsCredentialsService. Also, the AwsCredentials will contain all related to the AWS data.

```
AwsCredentials.getRegion()
AwsCredentials.getSecret()
AwsCredentials.getKey()
...
etc.
```

In this way, you will encapsulate the all related to AWS credentials logic.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSServiceAccountCredentialsService.java
```diff
@@ -0,0 +1,64 @@
...
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClientBuilder;
+import com.amazonaws.services.securitytoken.model.AssumeRoleRequest;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AWSServiceAccountCredentialsService {
+
+  @Value("${datajobs.aws.serviceAccountSecretAccessKey:}")
```
tozka, line 22: I really recommend against using @Value on fields. It is better to use constructor injection with@Value. 

Using constructor injection allows for mandatory dependencies to be passed in at the time of object creation, making it clear which dependencies are required for the object to function properly.
 Very importantly it also makes it easier to write unit tests, as dependencies can be easily mocked and injected.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -38,6 +38,9 @@ public class JobImageBuilder {
   @Value("${datajobs.git.url}")
   private String gitRepo;
 
+  @Value("${datajobs.aws.assumeIAMRole:}")
+  private boolean useRoleCredentials;
```
tozka, line 42: I'd encapsulate this behind AWSServiceAccountCredentialsService and it will return credentials, if assumeIAMRole is false it would return session token as empty.

tozka: Also name both the property (assumeIAMRole) and the field name (useRoleCredentials) with the same name please.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSServiceAccountCredentialsService.java
```diff
@@ -0,0 +1,64 @@
...
+            .build();
+
+    AssumeRoleRequest assumeRequest =
+        new AssumeRoleRequest()
+            .withRoleArn(awsRoleArn)
+            .withRoleSessionName(
+                "control-service-session-" + UUID.randomUUID().toString().substring(0, 4))
+            .withDurationSeconds(sessionDuration);
+
+    STSAssumeRoleSessionCredentialsProvider credentialsProvider =
```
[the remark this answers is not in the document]
tozka: possible hint: this is likely solved by moving most of the logic here in the constructor. As by default Spring services are singleton, this class would be a single instance. 

The only gotcha could be that this method `getTemporaryCredentials` can be called by multiple threads (e.g multiple depoy request) , the aws doc say it's thread safe so should be fine.

tozka (changes requested):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSServiceAccountCredentialsService.java
```diff
@@ -0,0 +1,64 @@
...
+  @Value("${datajobs.aws.RoleArn:}")
+  private String awsRoleArn;
+
+  @Value("${datajobs.aws.defaultSessionDurationSeconds:}")
+  private int sessionDuration;
+
+  @Value("${datajobs.aws.region:}")
+  private String awsRegion;
+
+  public BasicSessionCredentials getTemporaryCredentials() {
```
tozka, line 37: The name of the getTemporaryCredentials method implies that it returns the same set of credentials each time it is called. However, the method actually returns a new set of temporary credentials each time it is called. Might be better to rename the method to createTemporaryCredentials or similar.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSServiceAccountCredentialsService.java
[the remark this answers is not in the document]
tozka: possible hint: this is likely solved by moving most of the logic here in the constructor. As by default Spring services are singleton, this class would be a single instance.

The only gotcha could be that this method getTemporaryCredentials can be called by multiple threads (e.g multiple depoy request) , the aws doc say it's thread safe so should be fine.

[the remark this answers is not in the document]
mrMoZ1: It's not closable in the version I'm using - https://javadoc.io/doc/com.amazonaws/aws-java-sdk-sts/1.11.103/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.html, though I agree that it could benefit from a refactor.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSCredentialsService.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Getter private boolean assumeIAMRole;
+
+  @Getter private String awsSecretAccessKey;
+
+  @Getter private String awsAccessKeyId;
+
+  private STSAssumeRoleSessionCredentialsProvider credentialsProvider;
+
+  public AWSCredentialsService(
```
tozka, line 50: I think we need documentation for the method.

Particularly what is the difference between awsSecretAccessKey and serviceAccountSecretKey and such. 

Also, 10 arguments for method /constructor is a lot. 
Maybe use [ConfigutationProperties](https://www.baeldung.com/configuration-properties-in-spring-boot)

It should be something like 

```
@ConfigutationProperties("datajobs.aws")
AWSCredentialsServiceConfig
{
@Getter
private String awsRegion = "";
{
```

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -124,6 +118,23 @@ public boolean buildImage(
...
+      // to use when authenticating against ECR
+      var credentials = awsCredentialsService.createTemporaryCredentials();
+      builderAwsSecretAccessKey =
+          credentials.get(AWSCredentialsService.AWS_SERVICE_ACCOUNT_SECRET_ACCESS_KEY);
+      builderAwsAccessKeyId =
+          credentials.get(AWSCredentialsService.AWS_SERVICE_ACCOUNT_ACCESS_KEY_ID);
+      builderAwsSessionToken =
+          credentials.get(AWSCredentialsService.AWS_SERVICE_ACCOUNT_SESSION_TOKEN);
+    }
+
```
tozka, line 137: Isn't it better to put this logic in AwsCredentialsService that is responsible for building AWS credentials. For example method called "getCredentials" would return the correct ones depending on the flag

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeployTemporaryCredsIT.java
```diff
@@ -0,0 +1,174 @@
...
+    Assertions.assertEquals(true, jobDeployment.getEnabled());
+    Assertions.assertEquals(DataJobMode.RELEASE, jobDeployment.getMode());
+    Assertions.assertEquals(true, jobDeployment.getEnabled());
+    // by default the version is the same as the tag specified by datajobs.vdk.image
+    // for integration test this is registry.hub.docker.com/versatiledatakit/quickstart-vdk:release
+    Assertions.assertEquals("release", jobDeployment.getVdkVersion());
+    Assertions.assertEquals("user", jobDeployment.getLastDeployedBy());
+    // just check some valid date is returned. It would be too error-prone/brittle to verify exact
+    // time.
+    DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(jobDeployment.getLastDeployedDate());
```
tozka, line 161: Most of those verification seem redundant for the purposes of this test .

You need only to verify 
A) A deploy has finished successfully.  which `    Assertions.assertEquals(testJobVersionSha, jobDeployment.getJobVersion());` would be enough 
B) The deployment has created correctly an image in ECR. I would add a check (if one could be done that is fairly short and easy) that verifies an image exists in ECR.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeployTemporaryCredsIT.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  @AfterEach
+  public void cleanUp() throws Exception {
```
tozka, line 165: That would not clean up the image in ECR.

tozka (approved):
Looks good to me. 

I made some suggestions but they are not blockers.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSCredentialsService.java
```diff
@@ -0,0 +1,117 @@
...
+          "Please configure all appropriate properties for aws service account pattern in"
+              + " application.properties.",
+          null);
+    }
+    AWSSessionCredentials serviceAccountCredentials = credentialsProvider.getCredentials();
+    var accessKeyId = serviceAccountCredentials.getAWSAccessKeyId();
+    var secretAccessKey = serviceAccountCredentials.getAWSSecretKey();
+    var sessionToken = serviceAccountCredentials.getSessionToken();
+
+    return Map.of(
```
mivanov1988, line 109: Why map and not a DTO?

PR #1800, opened 2023-03-30
control-service: update helm charts for service account credentials

mrMoZ1:
what: Updated the helm chart to include variables for the service account pattern introduced:
https://github.com/vmware/versatile-data-kit/pull/1787

why: this is a necessary change for this PR - https://github.com/vmware/versatile-data-kit/pull/1787
which introduces abilities to generate temporary credentials.


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -185,12 +185,33 @@ proxyRepositoryURL: "012285273210.dkr.ecr.us-west-2.amazonaws.com/taurus/dp"
...
 ## [Required if deploymentDockerRegistryType=ecr]
+## These values also set the default Amazon ENV variables by convention.
 ## AWS ECR Credentials and repository to store and read images of the deployed data jobs
 ## The configuration should be set in case deploymentDockerRegistryType is ecr
 deploymentEcrAwsRegion: ""
 deploymentEcrAwsAccessKeyId: ""
 deploymentEcrAwsAccessKeySecret: ""
 
+##
+deploymentEcrAwsAssumeIamRole: false
```
mivanov1988, line 196: nit: maybe the name should indicate that it is boolean?

mivanov1988 (approved):

mivanov1988 (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -185,12 +185,33 @@ proxyRepositoryURL: "012285273210.dkr.ecr.us-west-2.amazonaws.com/taurus/dp"
...
+## stored in a secure location, such as an encrypted file, on the EC2 instance.
+## The IAM user is then given specific permissions to only access the ECR registry - read/write images
+## and create repositories.
+## More information: # https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
+## These properties need to be stored in different env variables than the ones not belonging to the
+## service user - e.g. defined above, because if the data jobs K8S cluster is hosted on a EKS
+## instance, some API calls to the instance made by the control-service could fail since we set
+## the default AWS environment variable names above which are used by default on some k8s api calls.
+deploymentEcrAwsServiceAccountAccessKeyId: ""
+deploymentEcrAwsServiceAccountSecretAccessKey: ""
```
tozka, line 211: what is the difference between teh above "deploymentEcrAwsAccessKeyId: ""
deploymentEcrAwsAccessKeySecret" 

And this onees ?Why do we need both?

mrMoZ1: These credentials belong to the service user and the ones you mentioned are meant to be used when long term credentials are present. Also the long term credentials set the default AWS environment variables, which are read by the API when making API calls to the k8s cluster, meaning that if the cluster is hosted on a EKS instance these will be used along the kubeconfig to authenticate API calls. E.g when testing locally if the default AWS env variables are unset (or are set to the ones belonging to the service user) API calls made by the k8s service within control-service will fail due to authorization issues. This could also be the case when control-service runs on a different AWS account than the ECR registry accessed by a service user. I know this from testing locally against the test k8s cluster and ECR repository on a different account.

tozka (approved):

PR #1804, opened 2023-03-30
vdk-frontend: update frontend docs

murphp15:
# Why
We want the frontend VEP to describe all major decisions in the implementation. We now include details about docker and helm.


specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -138,6 +140,16 @@ supports customers that want to:
 1. Use the shared component libraries in their projects.
 2. Build their own docker image for data-pipelines
 
+#### Docker
+A Docker image will be published on docker hub under the user https://hub.docker.com/u/versatiledatakit.
```
ivakoleva, line 144: nit: may rephrase the statement, so we don't need to update it upon making this change -> or alternatively have the ticket linked, so we don't forget to update it?

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -138,6 +140,16 @@ supports customers that want to:
...
+It is built by the CI on any changes to the frontend codebase.
```
ivakoleva, line 145: How does the frontend application image requirements differ from the frontend test image already produced?

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -138,6 +140,16 @@ supports customers that want to:
...
+
+#### Helm(or install helpers)
+There is no use case for having just the frontend installed.
+Because of this there are no helpers for installing the frontend.
+Instead, the frontend will be installed as part of the contrl-service helm chart
```
ivakoleva, line 150: nit: may link to the chart

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -138,6 +140,16 @@ supports customers that want to:
 1. Use the shared component libraries in their projects.
 2. Build their own docker image for data-pipelines
 
+#### Docker
+A Docker image will be published on docker hub under the user https://hub.docker.com/u/versatiledatakit.
+It is built by the CI on any changes to the frontend codebase.
+
+#### Helm(or install helpers)
+There is no use case for having just the frontend installed.
```
ivakoleva, line 148: Is there? And an use case where only the backend is installed?

specs/vep-1507-vdk-operations-ui/README.md
murphp15: The frontend test image includes browsers and testing utils. 
This is completely needless bloat for the production system. 
In the production image we just include the distribution folder from the build directory and nginx server.

specs/vep-1507-vdk-operations-ui/README.md
murphp15: I think there is an argument for only installing the backend 
1. if customers wanted to use their own UI (seems very unlikely)
2. if customers wanted to use just the cli tools, vdk-cli,curl etc... 



We can make the frontend ui configurable similar to how cockroach db is configurable.

specs/vep-1507-vdk-operations-ui/README.md
ivakoleva: May add the requirements known (about the image) to the tickets for creating the runtime image.

PR #1806, opened 2023-03-30
control-service: Add python_version to Control Service API

doks5:
As part of the effort to provide users with the ability to specify the python release version, which they want their data jobs to be deployed with, we need to expose a property in the Control Service's API and ensure that the configuration is stored in the database and the kubernetes objects.

This change exposes a `python_version` property in the api schemas of the DataJobDeployment and DataJobDeploymentStatus components, and propagates the property to the database model and kubernetes resources. All of this is done as one change, to avoid unexpected behaviour or regressions.

VEP for Reference: https://github.com/vmware/versatile-data-kit/tree/main/specs/vep-1739-multiple-python-versions

Testing Done: New and modified existing tests.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -2582,6 +2583,7 @@ private Optional<JobDeploymentStatus> mapV1CronJobToDeploymentStatus(
       if (annotations != null) {
         deployment.setLastDeployedBy(annotations.get(JobAnnotation.DEPLOYED_BY.getValue()));
         deployment.setLastDeployedDate(annotations.get(JobAnnotation.DEPLOYED_DATE.getValue()));
+        deployment.setPythonVersion(annotations.get(JobAnnotation.PYTHON_VERSION.getValue()));
```
ivakoleva, line 2586: What's the value expected range - is it optional or required?

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/JobImageDeployerTest.java
```diff
@@ -155,6 +157,7 @@ public void testScheduleLabelsAnnotations() {
       Assertions.assertEquals("scheduled/runtime", startedBy);
       Assertions.assertEquals("scheduled", executionType);
       Assertions.assertEquals("false", unscheduled);
+      Assertions.assertEquals("3.9", pythonVersion);
```
ivakoleva, line 160: What about tests scenarios of create and update where no explicit version is set?

ivakoleva (approved):
LGTM;
May you link to PR description for convenience, the VEP-1739 described API and use cases?

For example, having a pinned python version corresponds to an exact python version  requirement. Is there a use case of applying a lower bound only, so you still get your data job automatically upgraded to latest?

mivanov1988:
python_version validation will be added in a separate PR, right?

mivanov1988:
Shouldn't we return a default python version in case of empty paython-version annotation?

projects/control-service/projects/model/apidefs/datajob-api/api.yaml
```diff
@@ -1077,6 +1077,10 @@ components:
           description: Job version (can be Git commit)
           type: string
           example: 11a403ba
+        python_version:
+          description: A python release version to use for job deployments
+          type: string
+          example: 3.9
         mode:
```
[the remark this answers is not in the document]
doks5: Can you elaborate?

projects/control-service/projects/model/apidefs/datajob-api/api.yaml
```diff
@@ -1110,6 +1114,10 @@ components:
           description: Job version (can be Git commit)
           type: string
           example: 11a403ba
+        python_version:
+          description: A python release version to use for job deployments
+          type: string
+          example: 3.9
```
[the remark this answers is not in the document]
doks5: No defaults. The is dynamically changed, so we cannot have default at the API spec level. If no python_version passed, the service will use the defaultPythonVersion set by the administrators of the service deployment.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
doks5: The expected range is the supportedPythonVersions. The python version will either be provided by the user, or the defaultPythonVersion set by the service deployment administrators will be used.

doks5:
> python_version validation will be added in a separate PR, right?

Added in this PR.

doks5:
> Shouldn't we return a default python version in case of empty python-version annotation?

In case of empty python-version, we set the defaultPythonVersion.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DeploymentModelConverter.java
```diff
@@ -101,6 +102,10 @@ public static JobDeployment mergeDeployments(
         newDeployment.getVdkVersion() != null
             ? newDeployment.getVdkVersion()
             : oldDeployment.getVdkVersion());
+    mergedDeployment.setPythonVersion(
+        newDeployment.getPythonVersion() != null
+            ? newDeployment.getPythonVersion()
+            : oldDeployment.getPythonVersion());
```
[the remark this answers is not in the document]
doks5: I am not sure I completely understand. If there is no oldDeployment, the method itself would not have been called in the first place.

projects/control-service/projects/model/apidefs/datajob-api/api.yaml
```diff
@@ -1110,6 +1114,10 @@ components:
           description: Job version (can be Git commit)
           type: string
           example: 11a403ba
+        python_version:
+          description: A python release version to use for job deployments
```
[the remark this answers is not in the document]
doks5: I've updated the description to clarify that only python versions supported by the Control Service are to be used. Thanks for the review

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DeploymentModelConverter.java
```diff
@@ -101,6 +105,15 @@ public static JobDeployment mergeDeployments(
         newDeployment.getVdkVersion() != null
             ? newDeployment.getVdkVersion()
             : oldDeployment.getVdkVersion());
+
+    if (oldDeployment.getPythonVersion() == null && newDeployment.getPythonVersion() == null) {
+      mergedDeployment.setPythonVersion(new SupportedPythonVersions().getDefaultPythonVersion());
```
tozka, line 110:
```suggestion
      mergedDeployment.setPythonVersion(new supportedPythonVersions.getDefaultPythonVersion());
```

doks5: Hi, @tozka

This will not work, as the context is static, but the methods in the SupportedPythonVersions class are not.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -141,6 +144,8 @@ public void updateDeployment(
     try {
       log.info("Starting deployment of job {}", jobDeployment.getDataJobName());
       deploymentProgress.started(dataJob.getJobConfig(), jobDeployment);
+      validatePythonVersionIsSupported(jobDeployment.getPythonVersion());
```
mivanov1988, line 147: The validation cannot be done at the `updateDeployment` level since it is asynchronous and in case of an incorrect python version, it won't return an error to the client.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -193,6 +202,24 @@ private void saveDeployment(DataJob dataJob, JobDeployment jobDeployment) {
...
+   * As pythonVersion is optional, we need to check if it is passed. And if it is, we need to
+   * validate that the python version is supported by the Control Service. If it is not, we need to
+   * fail the operation, as we don't have sufficient information for the user's intent to deploy the
+   * data job.
+   *
+   * @param pythonVersion The python version to be used for the data job deployment.
+   */
+  private void validatePythonVersionIsSupported(String pythonVersion) {
+    if (pythonVersion != null && !supportedPythonVersions.isPythonVersionSupported(pythonVersion)) {
+      throw new ApiConstraintError(
```
mivanov1988, line 215: It would be better if you introduce a separate exception class like - DataJobAlreadyRunningException.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DeploymentModelConverter.java
mivanov1988: The `new SupportedPythonVersions()` approach won't work as well since the Spring context won't be initialized and the default version will be null. This is not the right place to enforce the default version which is business logic and should be moved to the DeploymentService.

mivanov1988:
Have you seen the Codacy issues?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DeploymentModelConverter.java
tozka: +1 what Miro said. The method purpose is pretty clear "Merge two deployments with preference of fields from newDeployment over oldDeployment" 

Setting the default version should be done in another place.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
tozka: +1 validation should happen on the synchronous path so that API request fail immediately. 

On the async path. If it happens that the configured python version is not part of the supported version (it could happen if administrators change the supported python versions -e.g let's say admins remove for 3.7 after end of life support) , but user has configured 3.7 

What do we do? 
My suggest was 

A) To log a warning so the administrator is aware (as much as it's expected that admins look at warnings ...) 
B) Do not fail the deployment for certain. 
  B.1) It's possible to continue using the python version (but as it's not supported there's risk the image would not exist and generally I do not favour that option 
  B.2) Find closest supported version and use it.  
C) After B) it might be a good idea to send a notification to user as well so they are aware their job was "forcefully" upgraded  

I know we discussed it briefly during the vep discussion but I do not recall if we reached any consuliton on that. 

It is perfectly fine if you decide to tacke that part of the problem in separate PR.

tozka: Alternatively we can add the validate not just on patch but also on any deploy operation , 
So the next time user tries to deploy they'd get an error that tells them to update the version. 

I prefer that we handle this for the user. Generally from UX point of view, reducing things users need to think about is always preferable. But it's an option.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -193,6 +202,24 @@ private void saveDeployment(DataJob dataJob, JobDeployment jobDeployment) {
...
+          "python_version",
+          "not in the list of supported python versions.",
+          supportedPythonVersions.getSupportedPythonVersions(),
```
tozka, line 218: Perhaps say explicitly "choose one of the following supported python versions: [list of supported versions].

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DeploymentModelConverter.java
doks5: Moved the logic to the service. Thanks for the reviews

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
doks5: Done, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
doks5: Created a completely new exception type with more meaningful message.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -193,6 +202,24 @@ private void saveDeployment(DataJob dataJob, JobDeployment jobDeployment) {
...
+          "Pass one of the supported python versions to the python_version property.");
```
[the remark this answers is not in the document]
doks5: Done, thank you

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
doks5: Good points, moved the validation to the Deployment Controller.

> B.2) Find closest supported version and use it.
I am not a big fan of such approach. How do you define "closest supported version"? If we have ["3.7", "3.8", "3.9"] and the user passes "3.10", then sure the closest would be 3.9. However, if we have ["3.8", "3.10", "3.12"] and the user passes "3.9", what would be the closest - 3.8 or 3.10?
Also, there is no limitation on the format, so the admins can set ["3.8.12", "3.9rc2", "3.10"] and these would still be valid choices. The parsing logic to find the closest version would be a nightmare and prone to corner case issues.

doks5:
> Have you seen the Codacy issues?

As discussed offline, these are fixed now. Thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -193,6 +199,21 @@ private void saveDeployment(DataJob dataJob, JobDeployment jobDeployment) {
...
+   * validate that the python version is supported by the Control Service. If it is not, we need to
+   * fail the operation, as we don't have sufficient information for the user's intent to deploy the
+   * data job.
+   *
+   * @param pythonVersion The python version to be used for the data job deployment.
+   */
+  public void validatePythonVersionIsSupported(String pythonVersion) {
+    if (pythonVersion != null && !supportedPythonVersions.isPythonVersionSupported(pythonVersion)) {
+      throw new UnsupportedPythonVersionException(
+          pythonVersion, supportedPythonVersions.getSupportedPythonVersions().toString());
```
mivanov1988, line 213: Better `supportedPythonVersions.getSupportedPythonVersions().toString()` move that to the UnsupportedPythonVersionException and pass the list.

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
doks5: Good point, fixed. Thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
tozka: You will not be parsing it yourself. You will be using a library like https://github.com/vdurmont/semver4j . 
The logic should not be complicated. sort them all and find the next closest.

tozka (approved):
The change looks good to me. 

But I'd like to discuss following questions (at some point before the feature is considered implemented) 
: 
1.  What happens when administrators change the supported versions? 

What is the right behaviour ? 

2. the same goes for the default version. What happens when administrators change teh default version ?

 I don't think we've thought about and discussed it.

PR #1811, opened 2023-03-31
vdk-plugins: test only oldest and newest supported python version

tozka:
As part of my effort to stabilize our Nightly builds and the CICD cluster I am removing testing middle versions.

We test with the oldest supported version to make sure we do not use a new feature added lately which would break compatability.

We test with newest version to make sure that it runs with newest version. Although the python itself is very likely to be backwords compatible (for minor release), there are sometimes small changes that are not and same cannot be told for 3th party dependecies we may use which may not support latest version fully.

Testing Done: this CI

### DIFF
--- a/projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
+++ b/projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
@@ -31,18 +31,6 @@ build-py37-quickstart-vdk:
   extends: .build-quickstart-vdk
   image: "python:3.7"
 
-build-py38-quickstart-vdk:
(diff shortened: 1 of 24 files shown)


gageorgiev (approved):

ivakoleva (approved):
What are the chances for bringing back those deleted ci jobs?
For example, the possibility of having resources updated or increased in accordance with our current needs (since onboarding of new release cycles). Or, not on every change, but occasionally checking the compliance with all minor versions?
In case the chance is good, then, we may consider a configurable or feature flag approach towards this.

murphp15 (approved):

PR #1812, opened 2023-03-31
[DRAFT] vdk-frontend: improve npm build stability.

murphp15:


projects/frontend/cicd/install_shared.sh
```diff
@@ -5,7 +5,7 @@
 echo "Logging npm engines version..."
 npm version
 echo "Installing dependencies..."
-npm install
+npm install --max_old_space_size=1000
```
murphp15, line 8: @gorankokin 
You want me to put 
```
node --max-old-space-size=1000 `which npm` install
```
Here?

gorankokin: yes

gorankokin:
I think this PR is not needed anymore, since we found that k8s kublets are evicting pods due to low ephemeral storage that exceeds requested quota of 0.
Otherwise we found that nodejs v8 engine default heap size limit is 2GB by default, so there isn't any problem for the time being that requires increasing it.
We will monitor ci/cd pipelines in near future and if there is a need for increasing it, we know the way to go.

ivakoleva (changes requested):
I agree, we may close the ticket until further notice

PR #1816, opened 2023-04-01
vdk-lineage: support for latest version sqllineage library

tozka:
latest version of sqllineage is removing `statements_parsed` method We are adopting our vdk-lineage plugin to support the latest release. This means we won't detect query type but that property was never used anywhere by vdk-lineage.

Testing Done: existing unit tests

### DIFF
--- a/projects/vdk-plugins/vdk-lineage/src/vdk/plugin/lineage/sql_lineage_parser.py
+++ b/projects/vdk-plugins/vdk-lineage/src/vdk/plugin/lineage/sql_lineage_parser.py
@@ -45,11 +45,11 @@ def get_table_lineage_from_query(
 
     runner = LineageRunner(query)
 
-    if len(runner.statements_parsed) == 0:
+    if len(runner.statements()) == 0:
         # log.debug("No statement passed")
         return None
 
-    if len(runner.statements_parsed) > 1:
+    if len(runner.statements()) > 1:
         raise RuntimeError(
             "Query with more than one statement is passed. "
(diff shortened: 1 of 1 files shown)


projects/vdk-plugins/vdk-lineage/src/vdk/plugin/lineage/sql_lineage_parser.py
```diff
@@ -74,7 +74,7 @@ def get_table_lineage_from_query(
 
     return LineageData(
         query=query,
-        query_type=runner.statements_parsed[0].get_type(),
+        query_type="not implemented",
```
DeltaMichael, line 77: Is there no way to get the query type now?

DeltaMichael (approved):

projects/vdk-plugins/vdk-lineage/src/vdk/plugin/lineage/sql_lineage_parser.py
tozka: Not an obvious and easy way. I could use sqlparse library (which sqlilineage used previously) to get it but as it's not needed I opted not to do it now.

DeltaMichael: Fair enough. You should open a ticket for it if you think it's something we'll really need in the future. I'm ok with resolving this either way. 🙂

gageorgiev (approved):

PR #1819, opened 2023-04-01
vdk-control-cli: refactor output printing with printer class

tozka:
Currently, if a command needs to print similar types of data in multiple places
or formats, the developer would have to duplicate the printing code for each location or format. This would lead to a lot of redundant code, which is difficult to maintain and prone to errors. This is making the code more "DRY".

Also, users and devs are limited to a fixed set of output formats provided by the application. This could be restricting if the devs needs to print data in a format that is not supported by the application. E.g I wanted to add `rich` or `streamlit` type potentially.

By introducing support for customizable output formats with the Printer class and related functions, users can define their own output formats and register them with the application using the printer decorator. This allows users to print data in any format they desire, providing greater flexibility and customization options.
This is of course an application of Factory and Strategy design patterns. 

This is adopted in a single command here only.  In a separate PR, I will adopt this printer in the other commands.

Testing Done: unit tests (incl new ones)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ __pycache__/*
 .cache/*
 .*.swp
 */.ipynb_checkpoints/*
+out
 .DS_Store
 .eggs
 build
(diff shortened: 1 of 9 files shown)


gageorgiev (approved):

yonitoo (approved):
Awesome! LGTM!

PR #1820, opened 2023-04-02
vdk-control-cli: fix circular import dependecy

tozka:
There is a circular import issue between the
vdk.internal.control.configuration.vdk_config and
vdk.internal.control.command_groups.version_group.version modules. Specifically, vdk_config is importing version and version is importing cli_utils, which is importing vdk_config. This creates a cycle that cannot be resolved, which is causing the ImportError under certain circumstances.

To fix this issue, I need to refactor the code to remove the circular import. I moved the methods for determining verison in version utils which is used by both vdk_config and version_group.version modules.

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/utils/version_utils.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/utils/version_utils.py
@@ -1,0 +1,27 @@
+# Copyright 2023-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from pkg_resources import DistributionNotFound
+from pkg_resources import get_distribution
+
+# https://packaging.python.org/guides/single-sourcing-package-version/#
+try:
+    # Change here if project is renamed and does not equal the setuptools metadata.name
+    dist_name = "vdk-control-cli"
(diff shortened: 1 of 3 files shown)


projects/vdk-control-cli/src/vdk/internal/control/utils/version_utils.py
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: Apache-2.0
+from pkg_resources import DistributionNotFound
+from pkg_resources import get_distribution
+
+# https://packaging.python.org/guides/single-sourcing-package-version/#
+try:
+    # Change here if project is renamed and does not equal the setuptools metadata.name
+    dist_name = "vdk-control-cli"
+    __version__ = get_distribution(dist_name).version
+    pass
```
DeltaMichael, line 11: What's the point of this `pass` statement?

DeltaMichael (approved):

projects/vdk-control-cli/src/vdk/internal/control/utils/version_utils.py
tozka, line 11:
```suggestion
```

PR #1837, opened 2023-04-03
Simplify and clarify VDK purpose in the README

zverulacis:
This change aims to provide a better, simpler, and easier-to-read way to understand what VDK does.  

The image was one of the proposed ways to show it. I can also do the same with text and some complex formatting. Open for feedback.

The way it's intended to be more clear - short sentences
More readable - white space, formatting
Simple - the division between two components.


README.md
```diff
@@ -18,6 +18,10 @@
         <img alt="YouTube Channel Subscribers" src="https://img.shields.io/youtube/channel/subscribers/UCasf2Q7X8nF7S4VEmcTHJ0Q?style=social"></a>
 
 <!-- TODO: code coverage -->
+### Versatile Data Kit consists of two main components
+
+![versatile-data-kit-componenets](https://user-images.githubusercontent.com/11227374/229588579-7c8d1d51-206e-48d4-8957-a6c9d2190c94.svg)
```
DeltaMichael, line 23: This might not be the best image to start the README. It conveys important information, but I don't think it's the absolute first thing users should see. Maybe it can have its own section below.

DeltaMichael: Also, maybe you can add it in the repo and not use the github CDN. It's much easier to see the diff between images if you change it in the future.

ivakoleva (commented):
I think the image is meant for a dark theme. I've taken the conversation offline.
Basically, I really like the way it was before, seems more simplistic to me.

README.md
```diff
@@ -22,20 +22,30 @@
...
+- run,
+- and manage data workloads, aka [data jobs](https://github.com/vmware/versatile-data-kit/wiki/dictionary#data-job)
 
-VDK is built for resiliency and can recover in mid-process or restart entirely from the start.
+Its Lego-like design consists of lightweight Python modules installed via `pip` package manager. All VDK plugins are easy to combine.
+
+VDK CLI can generate a data job and run your Python code and SQL queries.
+
+**VDK SDK makes your code shorter, more readable, and faster to create.**<br>
+**Ready-to-use data ETL/ELT patterns make Data Science with VDK efficient.**
```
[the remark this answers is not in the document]
ivakoleva: I would like to have more patterns implemented & ready to use, or templates. For example, tools to help in data modelling, and data access. I would agree we need work in that direction -> and therefore keeping a realistic statement about VDK is the right thing to do. 

> Data Scientist [...] cleans, massages, and organizes (big) data.

Do you think VDK now is effectively helping to a person that cleans, massages and organises big data?

murphp15 (approved):
I really like that change

DeltaMichael (approved):

PR #1839, opened 2023-04-04
vdk-meta-jobs: exec job with arguments

### LINKED ISSUE #1502: Executing job from another job (synchronously) with arguments
I want to execute a job from another job synchronously with specifying arguments. Then I need to be able to get the result (i.e. success or failure).

**Suggested solution**
It'd be similar to the meta job plugin - use execute API and track the result.

**Additional context**
I would use this to extract reusable critical section logic.

yonitoo:
What:
Introduce execution of jobs with arguments (synchronously) via DAGs.

The way a job dict looks now:
{
        "job_name": "name-of-job",
        "team_name": "team-of-job",
        "fail_meta_job_on_error": True or False,
        **"arguments": \<json\>**,
        "depends_on": [name-of-job1, name-of-job2]
}

GitHub Issue: [https://github.com/vmware/versatile-data-kit/issues/1502](https://github.com/vmware/versatile-data-kit/issues/1502)

Testing Done: existing tests + new unit test

Closes #1502


tozka:
Please update your PR with latest changes. There is  a lot of unrelated changes in the PR.

yonitoo:
Done. It is ok now.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/dag_validator.py
```diff
@@ -141,6 +149,28 @@ def _validate_fail_meta_job_on_error(self, job: Dict):
...
+    def _validate_arguments(self, job: Dict):
+        if "arguments" in job:
+            if not isinstance(job["arguments"], Dict):
+                self._raise_error(
+                    list(job["job_name"]),
+                    ERROR.TYPE,
+                    "The type of the job dict key arguments is not str.",
+                    f"Change the Data Job Dict value of arguments. "
+                    f"Current type is {type(job['arguments'])}. Expected type is JSON-formatted str.",
+                )
```
tozka, line 161: You are checking if job["arguments"} is Dict, but then saying it must be a string. 

It should be a python dictionary by the way. Not a string. It's a not a nice thing to force user to format their arguments as a string. In Python of course we should be using pythonic objects. 

Better validation is to check if it can be serialized as json. 

So as a validation running `json.dumps` and not loads .

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
```diff
@@ -22,7 +22,12 @@
 
 
 class MetaJobsDag:
-    def __init__(self, team_name: str, meta_config: MetaPluginConfiguration):
+    def __init__(
+        self,
+        team_name: str,
+        meta_config: MetaPluginConfiguration,
+        arguments: Dict = None,
```
tozka, line 29: This looks wrong. 

Arguments must be per job and not for the whole Dag

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -49,12 +49,14 @@ def __init__(
...
         team_name: str,
         rest_api_url: str,
+        arguments: dict = None,
         timeout: int = 5,  # TODO: Set reasonable default
         **kwargs,
     ) -> None:
         self.job_name = job_name
         self.team_name = team_name
         self._rest_api_url = rest_api_url
+        self.arguments = arguments
```
tozka, line 59: should be private . All fields here should be marked as private per https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#clarification-on-access-modifiers-in-python

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
```diff
@@ -327,6 +332,11 @@ def test_meta_job_duplicate_jobs(self):
     def test_meta_job_not_allowed_job_key(self):
         self._test_meta_job_validation("meta-job-not-allowed-job-key")
 
+    def test_meta_job_wrong_job_arguments_json(self):
```
tozka, line 335: wher is the happy path tested ? That you successfully passed arguments to the job ?

You should test a few happy paths : 
* Where arguments are missing - None (which is already tested by previous test
* Where arguments is empty {} 
* Norma {"key: value" , "key2": value}

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
```diff
@@ -11,9 +12,10 @@
 
 class IDataJobExecutor(abc.ABC):
     @abc.abstractmethod
-    def start_job(self, job_name: str, team_name: str):
+    def start_job(self, job_name: str, team_name: str, arguments: Dict):
```
tozka, line 15: must be Optional as arugments can be missing (none) 
Set default 
```suggestion
    def start_job(self, job_name: str, team_name: str, arguments: Optional[Dict] = None):
```

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
```diff
@@ -23,6 +23,7 @@ def run_job(context: JobContext) -> None:
         meta_job_runner.META_CONFIG = MetaPluginConfiguration(
             context.core_context.configuration
         )
+        meta_job_runner.ARGUMENTS = context.job_args.get_arguments()
```
tozka, line 26: What is this?

tozka (changes requested):

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
```diff
@@ -49,12 +49,14 @@ def __init__(
         job_name: str,
         team_name: str,
         rest_api_url: str,
+        arguments: dict = None,
```
tozka, line 52: Should be optional 
```suggestion
        arguments: Optional[Dict] = None,
```

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_dag.py
yonitoo: I got misled by the vdk flag --arguments thinking that if you pass an argument to a DAG job, it would propagate it to all the orchestrated jobs. Consequently, this was brought to the game so that when the runner invokes _vdk run <job> --arguments <args>_, the DAG would pass a default value to all the jobs if such is missing. If you think that such scenario won't he ever needed to be handled, I will remove this logic.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: Related to https://github.com/vmware/versatile-data-kit/pull/1839#discussion_r1157123874
This way we get the args, passed by --arguments flag.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta.py
yonitoo: Done.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
yonitoo: Added them but forgot an log assertion that the arguments are there. Will address this and push later.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
yonitoo: Thanks! Changed those 4 to private.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/dag_validator.py
yonitoo: Makes sense, my bad, changed it.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job.py
yonitoo: Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: Removed it.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
yonitoo: Added the assertions.

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
```diff
@@ -327,6 +330,48 @@ def test_meta_job_duplicate_jobs(self):
...
+            self.runner = CliEntryBasedTestRunner(plugin_entry)
+            result = self._run_meta_job("meta-job-empty-arguments")
+            cli_assert_equal(0, result)
+            job2_post_req = [
+                req
+                for req, res in self.httpserver.log
+                if req.method == "POST"
+                and req.path.split("/jobs/")[1].split("/")[0] == "job2"
+            ]
+            job2_arguments = job2_post_req[0].json["args"]
```
tozka, line 371: extraxft as a helper function _get_job_arguments(httpserver) for example. 

The same code is used in two places.

tozka (approved):

projects/vdk-plugins/vdk-meta-jobs/tests/test_meta_job.py
yonitoo: Done.

tozka:
I suppose it's in separate issue to update the README?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/remote_data_job_executor.py
```diff
@@ -10,9 +11,11 @@
 
 
 class RemoteDataJobExecutor(IDataJobExecutor):
-    def start_job(self, job_name: str, team_name: str):
+    def start_job(self, job_name: str, team_name: str, arguments: Dict = None):
```
ivakoleva, line 14: nit: this seems shady - I can recall there is IJobArguments interface, and there should be usages within vdk-core since all data jobs have arguments. Why does not this apply to dags..?

yonitoo: Well, I guess I missed it. Thanks! Applied it now.

projects/vdk-plugins/vdk-meta-jobs/README.md
```diff
@@ -38,7 +39,8 @@ def run(job_input):
 When defining a job to be run following attributes are supported:
 * **job_name**: required, the name of the data job
 * **team_name:**: optional, the team of the data job. If omitted , it will use the meta job's team
-* **fail_meta_job_on_error**: optional, default is true. if true, the meta job will abort and fail if the orchestrated job fails, if false, meta job won't fail and continue.
+* **fail_meta_job_on_error**: optional, default is true. If true, the meta job will abort and fail if the orchestrated job fails, if false, meta job won't fail and continue.
+* **arguments**: optional, the arguments that are passed to the data job. The arguments could be used for parameter substitution in queries.
```
tozka, line 43: `The arguments could be used for parameter substitution in queries.` 

That seems like it might be confusing here. In what queries in the DAG? In context of a dag. Just passing argument to underlying orchestrated data job seem enough

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
```diff
@@ -16,6 +18,7 @@ class SingleJob:
     job_name: str
     team_name: str = None
     fail_meta_job_on_error: bool = True
+    arguments: IJobArguments = None
```
tozka, line 21: We use primitive types here only. IJobArguments is interface for arguments to be read from Job Input. 

The arguments itself are of type "dict" , we even validate there are of type dict. While IJobArguments is another interface that is not compatible with dict.

projects/vdk-plugins/vdk-meta-jobs/README.md
yonitoo: Now that I look at it, it might be confusing, yes, changing it.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
yonitoo: Makes sense. Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/dag_validator.py
```diff
@@ -141,6 +149,27 @@ def _validate_fail_meta_job_on_error(self, job: Dict):
                 f" is {type(job['fail_meta_job_on_error'])}. Expected type is bool.",
             )
 
+    def _validate_arguments(self, job: Dict):
```
ivakoleva, line 152: nit: It is strange to have `validate_arguments` that accepts job in favour of the arguments to validate. It is also unexpected to have data job object of type dict. 

When designing method contracts, the presumption is to require for passing the minimum needed. Job object seems too broad of a scope to use, for simply validating its arguments. For example, why is the entire job object requested, while using one job field in particular?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/dag_validator.py
```diff
@@ -141,6 +149,27 @@ def _validate_fail_meta_job_on_error(self, job: Dict):
...
+        if "arguments" in job:
+            if not isinstance(job["arguments"], dict):
```
ivakoleva, line 154: nit: were the arguments refactored to IJobArguments?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
```diff
@@ -16,6 +16,7 @@ class SingleJob:
     job_name: str
     team_name: str = None
     fail_meta_job_on_error: bool = True
+    arguments: dict = None
     depends_on: List[str] = field(default_factory=list)
```
ivakoleva, line 20: nit: What is the commonality of SingleJob and IStandaloneDataJob? It is strange that the DagJob isn't already relying on vdk-core API. What's the data job job top-level interface?
That question is out of scope for the current PR, yet I would like to revise it, in terms of refining the domain model.

ivakoleva (approved):

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
yonitoo: Well, the commonality seems to be the job name and job arguments. SingleJob has the team name fail_meta_job_on_error and the depends_on list, meaning that SingleJob, IMO, could rely on IStandaloneDataJob. I will discuss this topic with the DAG-involved personas.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/dag_validator.py
yonitoo: The entire object is requested because when we raise the error, we log the job name. Another solution for this method would be to pass the args and job name. I can change it if this makes more sense.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/dag_validator.py
yonitoo: Yes, but in the base class SingleJob the arguments are required to be of primitive type (dict in this case) and that's why we expect the arguments to be of type dict here.

ivakoleva (approved):
LGTM!

gageorgiev (approved):
looks good, I would just ask that you rename the test job from 'meta-job-arguments' to 'dag-arguments' and the step in it from 'meta.py' to 'dag.py', rest looks fine

PR #1840, opened 2023-04-04
control-service: revert job builder python version

mrMoZ1:
what: reverted to previously used python version for the job builder container and bumped version.txt to trigger new release.

why: as a part of this PR https://github.com/vmware/versatile-data-kit/pull/1810 job builder version was bumped, however this introduces problems with the awscli library used in the builder - it now fails with a missing module. : 
`...
  File "/usr/lib/python3.10/site-packages/dateutil/tz/tz.py", line 19, in <module>
    import six
ModuleNotFoundError: No module named 'six'`

Reverting to python 3.7 seems to fix the issue.

testing: created image locally and tested that aws commands in container work as expected.


mivanov1988 (approved):

mivanov1988:
Should be a possible solution to install `six` package?

doks5 (approved):

tozka (approved):

PR #1842, opened 2023-04-04
vdk-notebook: add hook for saving error information into json file

duyguHsnHsn:
What:
Added a hook that saves error information from job_steps into error.json file.
Also, for more informative error logs caused from cells from .ipynb file, added the id of the failing cell to the logs. (which will be used later from the Jupyter UI for marking the cell)

Why: The main reason for the changes is that in the Jupyter UI we want to introduce dialogs for job run status which consist of shorter message for errors (not all of the logs). But in general, it is a good option to have a place where the reason for the failure is kept (the user does not have to go over all of the logs to see which step is failing).


projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
```diff
@@ -22,6 +26,22 @@ def initialize_job(self, context: JobContext):
...
                 Notebook.register_notebook_steps(file_path, context)
 
+    @hookimpl(hookwrapper=True)
+    def run_job(self, context: JobContext) -> None:
+        out: HookCallResult
+        out = yield
+        result: ExecutionResult = out.get_result()
+        step_results = result.__dict__["steps_list"]
+        for step_result in step_results:
+            if step_result.__dict__["status"] == ExecutionStatus.ERROR:
```
tozka, line 36: why the use of `__dict__` ? result.steps_list ? and step_result.status

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
```diff
@@ -22,6 +26,22 @@ def initialize_job(self, context: JobContext):
...
+                error_info = {
+                    "step_name": step_result.__dict__["name"],
+                    "blamee": step_result.__dict__["blamee"].__str__(),
+                    "details": step_result.__dict__["details"],
+                }
+                with open("error.json", "w") as outfile:
```
tozka, line 42: do not hard code the file name or path and certainly not based on relative path.

 This could end up anywhere. Have a well known location (e.g home dir (~/.vdk/error/job-name/ for example) that can be overridden by configuration if needed.

tozka (changes requested):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
duyguHsnHsn: Wouldn't it be better to have it in the job directory?

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
duyguHsnHsn: I get error "AttributeError: 'ExecutionResult' object has no attribute 'steps_lists'"

tozka: That doesn't make sense, you can see that right? Here is the definition of the class : https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py#L46 

Clearly, the attribute is there. Please, If something doesn't make sense you really need to work harder to understand why it doesn't make sense. 

Hmmm,  if the error message you pasted is correct, the error is due to a typo in the attribute name (the error message says "**steps_lists**")  .  you You should set it to **steps_list** (without the 's' at the end).

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
tozka: Inside the job directly - probably not or it can be deployed alongside the job accidentally. 

That's why the keytab is not being kept there. 

But it could be one level up . If we have a job directory "/home/jobs/xxx-job/ 
Maybe we can keep it at /home/jobs/.error.json 

Also make it hidden file (with a dot) by default. 


But the issue is that the current code will save it in the current working directory which can change easily depending on context and how the program was started. We need more predictable and fixed place.

duyguHsnHsn: I moved it to the parents directory and made it hidden

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
```diff
@@ -7,8 +7,11 @@
 class Cell:
     """
     A class that represents Jupyter code cell
+    See Jupyter cell in his json schema:
```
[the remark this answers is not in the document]
duyguHsnHsn: I didn't see the type, will fix it now. For visualisation. It is not visualised yet, I will have another PR which will be using these changes to visualise the error in the Jupyter UI

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -76,6 +76,7 @@ def register_notebook_steps(file_path: Path, context: JobContext):
                             file_path=file_path,
                             job_dir=context.job_directory,
                             code=cell.source,
+                            cell_id=cell.id,
```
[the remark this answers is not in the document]
duyguHsnHsn: You are right, I will change the "code" to "source"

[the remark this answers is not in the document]
duyguHsnHsn: done

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook_plugin.py
duyguHsnHsn: fixed

tozka (approved):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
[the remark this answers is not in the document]
ivakoleva: nit: may as well keep the cell_source, common to cell_id, it is meaningful and provides context for the reader for what `source` is

PR #1852, opened 2023-04-05
vdk-control-cli: use common output printer

tozka:
Adopting the new output_printer to print all outputs of all commands that were introduced in https://github.com/vmware/versatile-data-kit/pull/1819

Some commands required small refactoring - to make some fields private.

Testing Done: existing tests to verify no regressions

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
@@ -1,14 +1,12 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import glob
-import json
 import logging
 import os
 from typing import Optional
 
 import click
 import click_spinner
-from tabulate import tabulate
 from taurus_datajob_api import ApiException
 from taurus_datajob_api import DataJob
 from taurus_datajob_api import DataJobConfig
@@ -21,6 +19,7 @@ from vdk.internal.control.job.job_archive import JobArchive
 from vdk.internal.control.job.job_config import JobConfig
 from vdk.internal.control.rest_lib.factory import ApiClientFactory
 from vdk.internal.control.rest_lib.rest_client_errors import ApiClientErrorDecorator
+from vdk.internal.control.utils import output_printer
 from vdk.internal.control.utils.cli_utils import get_or_prompt
 from vdk.internal.control.utils.output_printer import OutputFormat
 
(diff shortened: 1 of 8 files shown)


projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
```diff
@@ -31,14 +30,16 @@ class JobDeploy:
...
-    def __init__(self, rest_api_url: str, output):
+    def __init__(self, rest_api_url: str, output: str):
         self.deploy_api = ApiClientFactory(rest_api_url).get_deploy_api()
         self.jobs_api = ApiClientFactory(rest_api_url).get_jobs_api()
         self.job_sources_api = ApiClientFactory(rest_api_url).get_jobs_sources_api()
         # support for multiple deployments is not implemented yet so we can put anything here.
         # Ultimately this will be user facing parameter (possibly fetched from config.ini)
         self.__deployment_id = "production"
         self.__job_archive = JobArchive()
+        self.__output = output
```
DeltaMichael, line 41: `self.__output` just holds the output format, right? If that's the case, it might be a good idea to rename it to `__output_format`. Furthermore, if it's not used for anything else, I suggest we just make it a class field of the printer instead.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
```diff
@@ -31,14 +30,16 @@ class JobDeploy:
     ZIP_ARCHIVE_TYPE = "zip"
     ARCHIVE_SUFFIX = "-archive"
 
-    def __init__(self, rest_api_url: str, output):
+    def __init__(self, rest_api_url: str, output: str):
```
DeltaMichael, line 33: This might be a discussion for another time, but I'm also not sure why the output format is injected through the constructor. I don't think components should be concerned with their output format. That should be configured externally somehow, if there is a convenient way.

DeltaMichael (approved):

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
yonitoo: +1 for the renaming of "output"

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/deploy_cli_impl.py
tozka: That's a good point. It would be better if the Printer is initialized outside and directly injected. Inversion of control/dependency injection and all that. 

There are some complications for some CLI commands (like `vdk deploy --update`) we have a custom message with instructions for users which is very specific for that command and is really needed for text output .

Ultimately we could apply the MVC pattern for CLI  (with this class being split into DeployCommandController and the actual printed output being done in DeployCommandView classes) but the code is not that complex in most commands currently to go in that direction.

PR #1858, opened 2023-04-06
frontend: Omit optional dependencies when testing

DeltaMichael:
Why?

The CICD scripts pull data-pipelines and shared packages from registry instead of using the linked packages. This happens because optional dependencies are not omitted on running npm install

What?

Add the omit flag to npm install for both scripts

How was this tested?

1. Ran tests locally
2. CI/CD

Bugfix, non-breaking

### DIFF
--- a/projects/frontend/cicd/install_data_pipelines.sh
+++ b/projects/frontend/cicd/install_data_pipelines.sh
@@ -9,6 +9,19 @@
 # lints them to verify format, builds the UI application and runs the unit tests.
 ###
 
+link_shared () {
+  shared_dist_dir="../../shared-components/gui/dist/shared"
+  if [ -d "$shared_dist_dir" ]
+  then
+    echo "Linking the shared-components dist rebuild found..."
+    npm link $shared_dist_dir
+    npm link @versatiledatakit/shared
+  else
+    echo "No shared-components dist rebuild found."
+    exit 1
+  fi
+}
+
 if ! which npm >/dev/null 2>&1 ; then
   echo "ERROR:"
   echo "Please install npm 8.5.5+. Install cannot continue without it."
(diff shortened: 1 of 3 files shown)


projects/frontend/cicd/install_data_pipelines.sh
```diff
@@ -33,7 +33,7 @@ else
 fi
 
 echo "Install Dependencies..."
-npm install
+npm install --omit=optional
```
tozka, line 36: How did you verify it works ? Can you point out (in testing done section of the description) the relevant logs or whatever you used to verify the expected behaviour in. 


I am wondering because I see that in data pipelines/gui/package.json versatiledatakit/shared is not declared as optional so wouldn't still be installed from registry ...

DeltaMichael: I tried adding shared as optional too, but it broke when I ran the install command. This tells me that the flag works, but I'll dig deeper into it. 🙂

murphp15:
how did you notice this? was it not testing changes in shared till this PR?

murphp15 (approved):

DeltaMichael:
Never mind, I checked with `npm ls`, it's linking correctly, provided there's a dist dir in shared. If there isn't, it will pull from registry.

dist dir present

```
├── @versatiledatakit/data-pipelines@0.0.0 -> ./dist/data-pipelines
├── @versatiledatakit/shared@0.0.0 -> ./../../shared-components/gui/dist/shared
```

no dist dir

```
├── @versatiledatakit/data-pipelines@0.0.0 -> ./dist/data-pipelines
├── @versatiledatakit/shared@1.0.828653799
```

Closing the PR.

PR #1860, opened 2023-04-07
control-service: Remove support for very old k8s apiVersion

murphp15:
# Why
Shortly we will be adding the frontend to the control-plane helm chats. 
This is a little clean up before making the changes. 

# What 
1. Our docs on the supported k8s versions were contradictory. Within the integration tests readme we said that we supports as far back as 1.15 when this is not true. 
So in this PR we now point to the control-plane readme where the information is more detailed and actually accurate. 
The fact that it is declared only in a single file now makes it more likely to be accurate in the future. 

2. remove support for `networking.k8s.io/v1beta1` in the helm-chart in all versions of k8s which we support `networking.k8s.io/v1` will be present.

# How has this been tested
ran `vdk server --install` and everything looks good.  

this has the ingress flag set to true. so this is tested.


yonitoo (approved):
LGTM.

duyguHsnHsn (approved):

PR #1862, opened 2023-04-07
frontend: create dockerfile for vdk operations ui

DeltaMichael:
## Why?

Part of the wider effort to add the ui to quickstart-vdk. The ui should be shippable as a docker image.

## What?

Create dockerfile + nginx configuration for the ui The dockerfile spins up an nginx server on port 8091 and copies over the contents of dist/ui to the server's html dir

## How has this been tested?

Tested locally by running

```
docker build -t vdk-ui .
docker run --name vdk-ui-container -d -p 8080:8091 vdk-ui
```

and making sure that the ui is served on https://localhost:8080

Feature, non-breaking


projects/frontend/data-pipelines/gui/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nginxinc/nginx-unprivileged:stable-alpine
```
ivakoleva, line 1: Looking for VEP-1507 description of the change in accordance?

murphp15 (approved):
Looks good.

tozka:
I think the Dockerfile to be in a separate directory with its own README (which would explain its [purpose, configuration, and usage](https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#readme-files)) (like those in the [control service here](https://github.com/vmware/versatile-data-kit/tree/main/projects/control-service/projects/job-base-image-secure) )

tozka:
How is it build as part of the CICD? Can you quickly write what are next steps planned

projects/frontend/data-pipelines/gui/config/nginx.conf
```diff
@@ -0,0 +1,11 @@
+server {
+    listen 8091;
+    root /usr/share/nginx/html;
+    large_client_header_buffers 4 32k;
+    add_header Strict-Transport-Security "max-age=0; includeSubDomains" always;
+
+    location / {
+        try_files $uri $uri/ /index.html =404;
+        add_header Strict-Transport-Security "max-age=0; includeSubDomains" always;
```
tozka, line 9: That's a duplicate. It should not be necessary since it's already defined above at server level.

DeltaMichael:
> I think the Dockerfile should be in a separate directory with its own README (which would explain its [purpose, configuration, and usage](https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#readme-files)) (like those in the [control service here](https://github.com/vmware/versatile-data-kit/tree/main/projects/control-service/projects/job-base-image-secure) )

I don't think this needs a separate directory, readme, etc. It's not part of a complex setup like we have in control service. What's in the VEP should be enough. @tozka Let me know if you still feel strongly about this, I'm just saying there isn't really much to document here outside the VEP.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -492,6 +493,17 @@ New versions are released by changing the version in
...
+
+The VDK Operations UI ships with `quickstart-vdk`. The CI/CD pipeline builds the
+image on every change to `data-pipelines` and `shared`. The image is uploaded to
+dockerhub and can then be used in the helm charts for `quickstart-vdk`.
+
+The image contains:
+
+1. An nginx server, configured to listen on port 8091
+2. The Angular UI application running on the server
+
```
ivakoleva, line 506: May describe the diff between the E2e test and Quickstart images?
The FROM instruction is obvious, and a person may track whatever the base is (hopefully). The suggestion here is, to explain the requirements towards the UI distro build, before effectively copying it onto those both images? 
The reader may wonder why is another image needed? Any details about image specifics, should be present in both images manifests, so their purpose may be defined.

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -492,6 +493,17 @@ New versions are released by changing the version in
...
 [registry.hub.docker.com/versatiledatakit/vdk-cicd-base-gui](https://hub.docker.com/r/versatiledatakit/vdk-cicd-base-gui)
 
+##### Quickstart vdk image
+
+The VDK Operations UI ships with `quickstart-vdk`. The CI/CD pipeline builds the
+image on every change to `data-pipelines` and `shared`. The image is uploaded to
+dockerhub and can then be used in the helm charts for `quickstart-vdk`.
+
+The image contains:
+
```
ivakoleva, line 503: nit: may link to the Dockerfile

ivakoleva (approved):
LGTM, please see my comments?

DeltaMichael:
> How is it build as part of the CICD? Can you quickly write what are next steps planned (I am missing the big picture)

It's meant to be a deliverable, same as the data-pipelines package we push to npm, for example. See https://github.com/vmware/versatile-data-kit/pull/1872

specs/vep-1507-vdk-operations-ui/README.md
```diff
@@ -469,6 +470,10 @@ packages. Merging changes into `main` requires that these pipelines pass
 successfully.
 
 ##### e2e test image
+
+Used for running end-to-end tests for the `data-pipelines` package. Exposes all
+`data-pipelines` features in order to run tests. Not user-facing.
+
```
ivakoleva, line 476: I see the changes, you may update the stacked PR as well.
I don't understand how "exposes all data-pipelines features" differs from quickstart-vdk image, I suppose it also exposes these features?

May explain what Frontend artifacts come built-in in e2e test image, and what of those is installed after, and why?

ivakoleva (approved):

specs/vep-1507-vdk-operations-ui/README.md
DeltaMichael: I'm not sure at the moment what we'll hide and what we'll end up hiding and exposing for the UI image. They're technically the same right now. Off the top of my head, the UI image will have:
- csp authentication disabled
- the Explore page disabled
- download keytab button disabled

But it might not be limited to those, so I'd rather not add description for stuff that is not currently implemented. That should be in the scope of another ticket.

murphp15: You can also point out here that the e2e image contains a browser and webpack. 
Neither of these are in the production image.

DeltaMichael: I thought some more about this. When we do e2e testing, we're testing the application as a black box. The testing environment is irrelevant to the application itself. E2e testing and shipping an artefact are two different things. E2e testing requires scaffolding, for example. We see that with bundling Chrome in the docker image. There may be feature flags passed, etc. 

The point is that the app artefact itself doesn't really change, even though configs, environments, and packaging change. So I don't think we should explicitly do e2e testing on the prod Docker image. The artefact itself doesn't change between building the e2e tests image and the production image.

PR #1864, opened 2023-04-07
control service: Dynamically set job base image in builder

doks5:
Currently, the job base image used in job builders is set in the `deploymentDataJobBaseImage` application property. This is set once at Control Service deploy and requires re-deploy of the service to be changed.

As part of the initiative to support multiple python versions for data job deployments, a new application property, `supportedPythonVersions`, was introduced in https://github.com/vmware/versatile-data-kit/commit/967898fddec3e0515f1e5817bc2211cebc1b2d92. Additionally, a new API property, `python_version` was introduced in https://github.com/vmware/versatile-data-kit/commit/5086fa3d2fde0bb9db38e3ca1928764c95cf3900 to allow users to set which python version to be used for job deployments.

With all this done, this change updates the JobImageBuilder to use the above-mentioned configuration to set the job base image dynamically, without requiring Control Service redeploy.

Testing Done: New and existing tests.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
@@ -141,8 +141,7 @@ public class JobImageBuilder {
     // TODO: Remove when deploymentDataJobBaseImage deprecated.
     if (jobDeployment.getPythonVersion() == null && deploymentDataJobBaseImage == null) {
       log.debug(
-              "Missing pythonVersion and deploymentDataJobBaseImage. Data Job cannot be deployed."
-      );
+          "Missing pythonVersion and deploymentDataJobBaseImage. Data Job cannot be deployed.");
       return false;
     }
 
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
@@ -140,7 +140,7 @@ public class JobImageBuilder {
 
     // TODO: Remove when deploymentDataJobBaseImage deprecated.
     if (jobDeployment.getPythonVersion() == null && deploymentDataJobBaseImage == null) {
-      log.debug(
+      log.warn(
           "Missing pythonVersion and deploymentDataJobBaseImage. Data Job cannot be deployed.");
       return false;
     }
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
@@ -85,19 +85,22 @@ public class JobImageBuilder {
   private final DeploymentNotificationHelper notificationHelper;
   private final KubernetesResources kubernetesResources;
   private final AWSCredentialsService awsCredentialsService;
+  private final SupportedPythonVersions supportedPythonVersions;
 
   public JobImageBuilder(
       ControlKubernetesService controlKubernetesService,
       DockerRegistryService dockerRegistryService,
       DeploymentNotificationHelper notificationHelper,
       KubernetesResources kubernetesResources,
-      AWSCredentialsService awsCredentialsService) {
+      AWSCredentialsService awsCredentialsService,
+      SupportedPythonVersions supportedPythonVersions) {
 
     this.controlKubernetesService = controlKubernetesService;
     this.dockerRegistryService = dockerRegistryService;
     this.notificationHelper = notificationHelper;
     this.kubernetesResources = kubernetesResources;
     this.awsCredentialsService = awsCredentialsService;
+    this.supportedPythonVersions = supportedPythonVersions;
   }
 
   /**
@@ -253,6 +256,17 @@ public class JobImageBuilder {
   private Map<String, String> getBuildParameters(DataJob dataJob, JobDeployment jobDeployment) {
     String jobName = dataJob.getName();
     String jobVersion = jobDeployment.getGitCommitSha();
+    String baseImage = null;
+    String pythonVersion = jobDeployment.getPythonVersion();
+
+    // TODO: Remove this part when datajobs.deployment.dataJobBaseImage is deprecated.
+    if (deploymentDataJobBaseImage != null) {
+      baseImage = deploymentDataJobBaseImage;
+    } else if (pythonVersion != null && supportedPythonVersions.isPythonVersionSupported(pythonVersion)) {
+      baseImage = supportedPythonVersions.getJobBaseImage(pythonVersion);
+    } else {
+      baseImage = supportedPythonVersions.getDefaultJobBaseImage();
+    }
 
     return Map.ofEntries(
         entry("JOB_NAME", jobName),
@@ -260,7 +274,7 @@ public class JobImageBuilder {
         entry("GIT_COMMIT", jobVersion),
         entry("JOB_GITHASH", jobVersion),
         entry("IMAGE_REGISTRY_PATH", dockerRepositoryUrl),
-        entry("BASE_IMAGE", deploymentDataJobBaseImage),
+        entry("BASE_IMAGE", baseImage),
         entry("EXTRA_ARGUMENTS", builderJobExtraArgs),
         entry("GIT_SSL_ENABLED", Boolean.toString(gitDataJobsSslEnabled)));
   }
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
@@ -138,6 +138,14 @@ public class JobImageBuilder {
       }
     }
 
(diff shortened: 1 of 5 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -253,14 +256,26 @@ public boolean isBuildingJobInProgress(String dataJobName) {
...
+
+    // TODO: Remove this part when datajobs.deployment.dataJobBaseImage is deprecated.
+    if (deploymentDataJobBaseImage != null) {
+      baseImage = deploymentDataJobBaseImage;
+    } else if (pythonVersion != null
+        && supportedPythonVersions.isPythonVersionSupported(pythonVersion)) {
+      baseImage = supportedPythonVersions.getJobBaseImage(pythonVersion);
+    } else {
+      baseImage = supportedPythonVersions.getDefaultJobBaseImage();
+    }
```
ivakoleva, line 270: nit: this logic may be encapsulated, so that the base image is determined with 1 call, via dedicated functionality to fetch the effective base image provided; then you may add a unit test on it

ivakoleva (approved):
I'm trying looking at the VEP-1739 for:
* what is the deliverable of this change (what components are expected to be changed)
* how is the API changing (configurations before and after)

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mivanov1988, line 270:
```suggestion
    baseImage = deploymentDataJobBaseImage != null ? deploymentDataJobBaseImage : supportedPythonVersions.getJobBaseImage(pythonVersion);
```

Isn't it the same?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -253,14 +256,26 @@ public boolean isBuildingJobInProgress(String dataJobName) {
   private Map<String, String> getBuildParameters(DataJob dataJob, JobDeployment jobDeployment) {
     String jobName = dataJob.getName();
     String jobVersion = jobDeployment.getGitCommitSha();
+    String baseImage = null;
```
mivanov1988, line 259:
```suggestion
    String dataJobbaseImage = null;
```
How about that or something similar?

projects/control-service/projects/pipelines_control_service/src/main/resources/application-prod.properties
```diff
@@ -58,6 +58,8 @@ datajobs.kadmin_password=${KADMIN_PASSWORD}
 # This is the full image name of the data job builder
 datajobs.builder.image=registry.hub.docker.com/versatiledatakit/job-builder:1.1
 datajobs.deployment.dataJobBaseImage=registry.hub.docker.com/versatiledatakit/data-job-base-python-3.7:latest
+datajobs.deployment.supportedPythonVersions=${SUPPORTED_PYTHON_VERSIONS}
```
mivanov1988, line 61: move them to the application.properties and add default values

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/JobImageBuilderTest.java
```diff
@@ -232,4 +236,101 @@ public void buildImage_jobFailed_failure()
             true);
     Assertions.assertFalse(result);
   }
+
+  @Test
+  public void buildImage_pythonVersion() throws InterruptedException, ApiException, IOException {
```
mivanov1988, line 241:
```suggestion
  public void buildImage_deploymentDataJobBaseImageNullAndSupportedPythonVersion_shouldCreateCronJobUsingSupportedPythonVersion() throws InterruptedException, ApiException, IOException {
```

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/JobImageBuilderTest.java
```diff
@@ -232,4 +236,101 @@ public void buildImage_jobFailed_failure()
...
+
+    Map<String, String> capturedEnvs = captor.getValue();
+    Assertions.assertEquals("test-base-image", capturedEnvs.get("BASE_IMAGE"));
+
+    verify(kubernetesService).deleteJob(TEST_BUILDER_JOB_NAME);
+    Assertions.assertTrue(result);
+  }
+
+  @Test
+  public void buildImage_pythonVersion_deploymentDataJobBaseImage()
```
mivanov1988, line 290: Rename it

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
mivanov1988: I think the @ivakoleva suggestion is better. In other words, incorporate this logic in `supportedPythonVersions.getJobBaseImage()`.

doks5: Done. Thank you for the review

doks5: Moved to SupportedPythonVersions. Thank you for the review

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/JobImageBuilderTest.java
doks5: Renamed, thanks

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/JobImageBuilderTest.java
doks5: Renamed and added new test. Thanks

projects/control-service/projects/pipelines_control_service/src/main/resources/application-prod.properties
doks5: They are already in application.properties I added them in prod properties, as I wasn't sure if they wouldn't be overwritten. Nevertheless, I've removed them from prod properties.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
doks5: Changed, thanks

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/JobImageBuilderTest.java
```diff
@@ -232,4 +236,140 @@ public void buildImage_jobFailed_failure()
...
+
+    Map<String, String> capturedEnvs = captor.getValue();
+    Assertions.assertEquals("test-base-image", capturedEnvs.get("BASE_IMAGE"));
+
+    verify(kubernetesService).deleteJob(TEST_BUILDER_JOB_NAME);
+    Assertions.assertTrue(result);
+  }
+
+  @Test
+  public void buildImage_shouldCreateCronjobUsingDeploymentDataJobBaseImage()
```
mivanov1988, line 291: nit: follow the same naming convention as the other ones.

doks5: Renamed, thanks

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -135,6 +138,13 @@ public boolean buildImage(
       }
     }
 
+    // TODO: Remove when deploymentDataJobBaseImage deprecated.
+    if (jobDeployment.getPythonVersion() == null && deploymentDataJobBaseImage == null) {
```
murphp15, line 142: why would a jobDeployment have a null python version?

murphp15:
At the moment, before this branch is merged do we support multiple python versions in the actual data job but there is no way to build data jobs of any version except 3.9?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
doks5: In cases when no python_version is passed to the request, or edge cases where either the supportedPythonVersions or defaultPythonVersion are not properly set. We need to add configuration validation for the properties, but this would be done in a separate PR.

doks5:
> At the moment, before this branch is merged do we support multiple python versions in the actual data job but there is no way to build data jobs of any version except 3.9?

Not yet. There are service deployments that rely on the old way of setting the configurations through vdkImage and deploymentDataJobBaseImage, and if we enable the new functionality that would break them. Even with this change merged, the old way of configuration would still take precedence.

PR #1872, opened 2023-04-10
frontend: publish docker image for ui

DeltaMichael:
## Note

Stacked on top of https://github.com/vmware/versatile-data-kit/pull/1862

## Why?
    
Part of the wider effort to add the UI to quickstart-vdk
    
## What?
    
Build and publish a docker image for the UI as part
of the release step in the CI/CD pipeline. Image tags
follow the same versioning schema as the data-pipelines
package in the npm registry
    
## How has this been tested?
    
Ran the gitlab script locally

https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/4148986837
    
### DIFF
--- a/projects/frontend/cicd/.gitlab-ci.yml
+++ b/projects/frontend/cicd/.gitlab-ci.yml
@@ -1,6 +1,15 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+.frontend_publish_docker_image:
+  image: docker:23.0.1
+  variables:
+    DOCKER_DRIVER: overlay2
+    DOCKER_TLS_CERTDIR: ''
+    DOCKER_HOST: tcp://localhost:2375
+  services:
+    - docker:23.0.1-dind
+
 .frontend_shared_components_changes: &frontend_shared_components_locations
   - "projects/frontend/cicd/**/*"
   - "projects/frontend/shared-components/**/*"
@@ -8,6 +17,11 @@
 .frontend_data_pipelines_changes: &frontend_data_pipelines_locations
   - "projects/frontend/cicd/**/*"
   - "projects/frontend/data-pipelines/**/*"
+
+.frontend_paths_do_not_trigger_e2e: &frontend_do_not_trigger_e2e
+  - "projects/frontend/data-pipelines/README.md"
+  - "projects/frontend/data-pipelines/gui/Dockerfile"
+  - "projects/frontend/data-pipelines/gui/config/nginx.conf"
 
 frontend-data-pipelines-build:
   image: "versatiledatakit/vdk-cicd-base-gui:2.0.0"
@@ -39,6 +53,7 @@ frontend-data-pipelines-build:
       - projects/frontend/data-pipelines/gui/node_modules/
     expire_in: 1 week
 
+# TODO: Add the cicd path back to the rule once e2e tests are stabilized
 frontend-data-pipelines-e2e-tests:
   image: "versatiledatakit/vdk-cicd-base-gui:2.0.0"
   stage: pre_release_test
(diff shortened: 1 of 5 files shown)


projects/frontend/cicd/publish_image_dockerhub.sh
```diff
@@ -0,0 +1,25 @@
...
+patch_version="$3"
+
+[ ! -d "$ui_path/dist/ui" ] && echo "ERROR: dist/ui directory does not exist under path $ui_path" && exit 1
+
+version_tag=$(awk -v id=$patch_version 'BEGIN { FS="."; OFS = "."; ORS = "" } { gsub("0",id,$3); print $1, $2, $3 }' $ui_path/version.txt)
+
+image_repo="$VDK_DOCKER_REGISTRY_URL/$name"
+image_tag="$image_repo:$version_tag"
+
+docker build -t "$image_tag" -t "$image_repo:latest" $ui_path
```
ivakoleva, line 23: The publish script actually creates the docker image release candidate? This wouldn't be the production image, since not tested.

The suggestion here is, keep the variety of artifacts (-like cicd test and prod docker images) to a bare minimum needed. All of those should be tested, then the exact **tested artifacts needs to be published** (rather than publishing something partially tested due it is built on top of some priorly tested stack). 
The concept is, to release an atomic artifact that was priorly verified in its entirety.
See https://github.com/vmware/versatile-data-kit/blob/main/projects/frontend/data-pipelines/gui/package.json#L19


1. [build and unit test] release candidates -> NPM artifacts, dist/ 
2. [build docker image] rc with installed release candidate stack
3. [test docker image in cicd env running e2e tests] with installed release candidates
4. [release] all 2.0.0-rc.12345 to 2.0.0

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -82,6 +82,28 @@ frontend-data-pipelines-release:
...
+    DOCKER_HOST: tcp://localhost:2375
+  services:
+    - docker:19.03.15-dind
+  stage: release
+  before_script:
+    - cd projects/frontend/
+  script:
+    - docker login --username "${VDK_DOCKER_REGISTRY_USERNAME}" --password "${VDK_DOCKER_REGISTRY_PASSWORD}" "${VDK_DOCKER_REGISTRY_URL}"
+    - ./cicd/publish_image_dockerhub.sh vdk-operations-ui ./data-pipelines/gui $CI_PIPELINE_ID
+  rules:
```
murphp15, line 99: How I always test these is by changing the rule in a commit. then when it runs successfully I revert that change. 

Can you do that please.

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -82,6 +82,28 @@ frontend-data-pipelines-release:
       - projects/frontend/data-pipelines/gui/package-lock.json
     expire_in: 1 week
 
+frontend_publish_ui_image:
+  image: docker:19.03.15
```
murphp15, line 86: This is a very old version of the docker image. 
Can you use the newer one that we use in the control-plane gitlab files. 
Can you also extract all this docker config like we do in control-plane and extend it anywhere it is needed.

projects/frontend/cicd/publish_image_dockerhub.sh
murphp15: @ivakoleva 
What you are suggesting is not possible. 
Conceivably the docker image could be broken if there is something wrong with webpack.
However that exact same problem exists for the control plane image if there was a problem in the jdk.


Issues here are only caught in the heartbeat checks preformed in the vdk playground.


In the very near future we are going to add automated testing to vdk-quickstart. when this happens the docker images will be tested.

murphp15 (approved):
Looks good to me. 
A few very small changes requested and then I'm happy

projects/frontend/cicd/publish_image_dockerhub.sh
ivakoleva: For having end to end testing, the deliverable artifact being tested, needs to be operational as usual. The rest is e2e tests orchestration or wiring.

murphp15: we aren't testing an artifact in e2e tests. we are testing files in a source directory.

ivakoleva (approved):

PR #1831, opened 2023-04-03
vdk-dag: Rename vdk-meta-jobs to vdk-dag

gageorgiev:
This PR renames the vdk-meta-jobs plugin to vdk-dags as per the agreement. All references to "Meta Job" have been replaced by "DAG" in appropriate ways. Most importantly, the user-facing API has changed from `MetaJobInput().run_meta_job(jobs)` to `DagInput().run_dag(jobs)`.

Next steps are to correct the example, both in the repo and the wiki, and correct the blogpost.

Testing done: pipelines

PR #1873, opened 2023-04-10
vdk-meta-jobs: improve DAGs code documentation 

yonitoo:
What:
Improve the DAGs code documentation by ensuring all fields and public methods have documentation. 
Add code docs where it was missing.

Testing Done: not needed as only code docs were added

### DIFF
--- a/projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
+++ b/projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
@@ -10,7 +10,13 @@ from typing import List
 @dataclass
 class SingleJob:
     """
-    TODO
+    This class represents a single job to be executed.
+
+    :param job_name: The name of the job.
+    :param team_name: The name of the team that owns the job.
+    :param fail_dag_on_error: A boolean flag indicating whether the job should be executed.
+    :param arguments: A JSON-serializable dictionary of arguments to be passed to the job.
+    :param depends_on: A list of names of jobs that this job depends on.
     """
 
     job_name: str
@@ -22,18 +28,25 @@ class SingleJob:
 @dataclass
 class MetaJob(SingleJob):
     """
-    TODO
+    This class represents a DAG Job, which is a single job itself and consists of single jobs - the orchestrated ones.
+
+    :param jobs: A list of the orchestrated jobs
     """
 
     jobs: List[SingleJob] = field(default_factory=list)
 
 
 class IMetaJobInput(abc.ABC):
+    """
+    This class is responsible for the DAG job run.
+    """
+
     @abstractmethod
     def run_meta_job(self, meta_job: MetaJob):
         """
-        TODO
-        :param meta_job:
+        Runs the given DAG job.
+
+        :param dag: the DAG job to be run
         :return:
         """
         pass
--- a/projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_configuration.py
+++ b/projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_configuration.py
@@ -18,29 +18,64 @@ META_JOBS_MAX_CONCURRENT_RUNNING_JOBS = "META_JOBS_MAX_CONCURRENT_RUNNING_JOBS"
 
 class MetaPluginConfiguration:
     def __init__(self, config: Configuration):
+        """
+
+        :param config: the DAG plugin configuration
+        """
         self.__config = config
 
     def meta_jobs_delayed_jobs_min_delay_seconds(self):
+        """
+        Returns the minimum delay time for a delayed job to be executed in seconds.
+
+        :return: the number of seconds for the minimum delay of a delayed job
+        """
         return self.__config.get_value(META_JOBS_DELAYED_JOBS_MIN_DELAY_SECONDS)
 
     def meta_jobs_delayed_jobs_randomized_added_delay_seconds(self):
+        """
+        Returns the additional randomized delay time in seconds to the minimum delay time of a delayed job.
+
+        :return: the number of seconds for the additional randomized delay of the delayed jobs
+        """
         return self.__config.get_value(
             META_JOBS_DELAYED_JOBS_RANDOMIZED_ADDED_DELAY_SECONDS
         )
 
     def meta_jobs_dag_execution_check_time_period_seconds(self):
+        """
+        Returns the frequency at which the system checks a DAG execution's status.
+
+        :return: the frequency in seconds at which the system checks a DAG execution's status
+        """
         return self.__config.get_value(
             META_JOBS_DAG_EXECUTION_CHECK_TIME_PERIOD_SECONDS
         )
 
     def meta_jobs_time_between_status_check_seconds(self):
+        """
+        Returns the time interval in seconds between status checks for a job.
+
+        :return: the number of seconds between status checks for a job.
+        """
         return self.__config.get_value(META_JOBS_TIME_BETWEEN_STATUS_CHECK_SECONDS)
 
     def meta_jobs_max_concurrent_running_jobs(self):
+        """
+        Returns the limit of concurrent running jobs.
+
+        :return: the number of maximum concurrent running jobs
+        """
         return self.__config.get_value(META_JOBS_MAX_CONCURRENT_RUNNING_JOBS)
 
 
 def add_definitions(config_builder: ConfigurationBuilder):
+    """
+    Defines what configuration settings are needed for the DAGs plugin with reasonable defaults.
+
+    :param config_builder: the builder used to add the configuration variables
+    :return:
+    """
     config_builder.add(
         key=META_JOBS_DELAYED_JOBS_MIN_DELAY_SECONDS,
         default_value=30,
--- a/projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
+++ b/projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
@@ -13,9 +13,19 @@ from vdk.plugin.meta_jobs.meta_configuration import MetaPluginConfiguration
 
 
 class MetaJobsPlugin:
+    """
+    This module is responsible for managing the interaction between the DAGs plugin and VDK commands.
+    """
+
     @staticmethod
     @hookimpl
     def run_job(context: JobContext) -> None:
+        """
+        Sets the DAGs plugin configuration and other parameters.
+
+        :param context: the job context
+        :return:
+        """
         # TODO: there should be less hacky way
         meta_job_runner.TEAM_NAME = context.core_context.configuration.get_value(
             JobConfigKeys.TEAM
@@ -27,9 +37,22 @@ class MetaJobsPlugin:
     @staticmethod
     @hookimpl
     def vdk_configure(config_builder: ConfigurationBuilder) -> None:
+        """
+        Adds the DAG-related configuration variables.
+
+        :param config_builder: the builder used to add the configuration variables
+        :return:
+        """
         add_definitions(config_builder)
 
 
 @hookimpl
 def vdk_start(plugin_registry: IPluginRegistry, command_line_args: List):
+    """
+    Loads the DAGs plugin.
+
+    :param plugin_registry: plugin registry used to register the plugin and its hooks
+    :param command_line_args: a list of command line args passed to VDK
+    :return:
+    """
     plugin_registry.load_plugin_with_hooks_impl(MetaJobsPlugin(), "MetaJobsPlugin")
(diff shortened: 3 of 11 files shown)


projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
```diff
@@ -36,6 +36,12 @@ def __init__(
         self._dequeue_timeout_seconds = dequeue_timeout_seconds
 
     def enqueue(self, element):
+        """
+        Adds an element to the queue.
+
+        :param element: the element to be added to the queue
```
duyguHsnHsn, line 42: Maybe add some info for what type of element is expected

duyguHsnHsn (approved):

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
```diff
@@ -27,9 +37,22 @@ def run_job(context: JobContext) -> None:
...
 @hookimpl
 def vdk_start(plugin_registry: IPluginRegistry, command_line_args: List):
+    """
+    Loads the DAGs plugin.
+
+    :param plugin_registry: plugin registry used to register the plugin and its hooks
+    :param command_line_args: a list of command line args passed to VDK
+    :return:
+    """
     plugin_registry.load_plugin_with_hooks_impl(MetaJobsPlugin(), "MetaJobsPlugin")
```
ivakoleva, line 58: nit: My recommendation is, not to spend redundant effort in documenting trivial code blocks, so that it's less noisy. In case anything in particular has to be said, then providing documentation is excellent.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
```diff
@@ -13,9 +13,19 @@
 
 
 class MetaJobsPlugin:
+    """
+    This module is responsible for managing the interaction between the DAGs plugin and VDK commands.
```
ivakoleva, line 17: nit: It is obvious that each `*Plugin` class or `plugin` file in `vdk-plugins/`, is meant to self-register & parse configurations if any, then hook to be evaluated in data job runtime.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
```diff
@@ -13,9 +13,19 @@
 
 
 class MetaJobsPlugin:
```
ivakoleva, line 15: nit: the `plugin_entry.py` contains MetaJobsPlugin class, and a static method; why not using `<feature>-plugin.py` naming convention that is common to probably most of the plugins we do?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
```diff
@@ -10,7 +10,13 @@
...
 class SingleJob:
     """
-    TODO
+    This class represents a single job to be executed.
+
+    :param job_name: The name of the job.
+    :param team_name: The name of the team that owns the job.
+    :param fail_dag_on_error: A boolean flag indicating whether the job should be executed.
+    :param arguments: A JSON-serializable dictionary of arguments to be passed to the job.
+    :param depends_on: A list of names of jobs that this job depends on.
```
ivakoleva, line 19: Are job_name, team_name and arguments applicable data jobs defined in vdk-core?
What's unique about a DAG, is depends_on and fail_dag_on_error?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
```diff
@@ -22,18 +28,25 @@ class SingleJob:
...
     """
 
     jobs: List[SingleJob] = field(default_factory=list)
 
 
 class IMetaJobInput(abc.ABC):
+    """
+    This class is responsible for the DAG job run.
+    """
+
```
ivakoleva, line 43: What's the commonality between `IMetaJobInput` and `IJobInput`?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
```diff
@@ -10,7 +10,13 @@
 @dataclass
 class SingleJob:
     """
```
ivakoleva, line 12: What's the diff between IStandaloneDataJob and SingleJob?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
yonitoo: Yes, indeed. The difference between DAG and other jobs are the dependencies and the fail_dag_on_error flag.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: you are right, it seems like a good idea to change it to dags_plugin.py

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: I agree I might have overkilled it slightly. Should I remove the already written kinda obvious docs or take it as a recommendation for similar future efforts?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: yep, should I remove the accompanying documentation?

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
yonitoo: I don't think there is anything in common.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
yonitoo: The expected type of element depends on the use case. In the current implementation, we only use it for the DAG's delayed starting jobs but I am not sure if we should add a particular type because we might need the queue for some other objects in the future.

tozka: Duygu has a point. 

We shoudl use type hints everywhere. For such generic class time TimeBasedQueue (similarly to List, queue, deque, etc) we can use 
A) typing.Any to indicate it can be any
B) Better option -> use generics then user would pick the type `queue = TimeBasedQueue[int]` for example) 

For now A) is good enough.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_configuration.py
```diff
@@ -18,29 +18,64 @@
...
+        """
+        Returns the minimum delay time for a delayed job to be executed in seconds.
+
+        :return: the number of seconds for the minimum delay of a delayed job
+        """
         return self.__config.get_value(META_JOBS_DELAYED_JOBS_MIN_DELAY_SECONDS)
 
     def meta_jobs_delayed_jobs_randomized_added_delay_seconds(self):
+        """
+        Returns the additional randomized delay time in seconds to the minimum delay time of a delayed job.
```
tozka, line 37: Refer to the detailed configuration description. You can use seealso in reStructuredText I think: 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
ivakoleva: nit: This may be a separate design topic to this PR.

Looking at the names, they are both job input interfaces and one is for a DAG, and the other one is for a composite data job. 

From _API perspective_, however, those are now used differently -> the `JobInput` object is **managed by VDK**, and provided to the user via run method parameters; while `DagJobInput` instance is expected to be managed (created and handled) by the user.
And that's why, IDagJobInput does not extend IJobInput, from _OOP perspective_.

So now, `file_based_step.py` supports:
```
@staticmethod
    def run_python_step(step: Step, job_input: IJobInput) -> bool:
```

And, as a user extending VDK, I would appreciate extensibility of file steps of type python. 😊 
So that, having 
```
def run(dag_job_input: IDagJobInput):
```
is also a valid step of type Python.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/meta_configuration.py
yonitoo: Done although I'm not sure this is the best way to do these directives.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/time_based_queue.py
yonitoo: Alright. Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: Done.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
yonitoo: I think I understand your point. We can take this topic to a separate discussion - share more thoughts about the cons of the current approach and the benefits of having such common interface and the implementation details around this. should we open a GitHub issue about it? What do you think?

tozka:
Once https://github.com/vmware/versatile-data-kit/pull/1831 is merged you will need to "cross port" your changes there.

yonitoo:
Yes! We discussed this with @gageorgiev and I will adapt this MR to the changes once the one you mentioned is merged.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
ivakoleva: In summary of VDK Core API, 

(1) There is one (primary) **abstraction of what a Data Job is**, and that is `DataJob` class defined run built-in plugin https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py#L233
* It has name, source directory, context, and also could be run.

(2) Then...we have IStandaloneDataJob defined by https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/data_job.py#L11

* It has name, source directory, job args, and could be entered.

**The problem is**, those two abstractions obviously describe the same stuff - a Data Job - however, from _OOP perspective_, those interfaces don't have commonalities. That's the gap, that needs to be addressed.

(3) _And, again, this may be a separate design discussion than this PR_,
however it is related, because now we have another MetaJob abstraction, that is actually a Data Job as well. And, again, the same problem of lacking any link to a definitive Data Job interface. What a Data Job is, remains API-incoherent still, and we should address it.

ivakoleva (approved):
I think less is more, so I'm in favour of removing rather than cluttering with docs. If there is something unordinary that needs to be documented, we may keep the comment to help the reader focus and understand what's important.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/api/meta_job.py
yonitoo: Wow, thanks for the detailed information! I totally agree that this should be addressed and that we should find a way to unify the logic instead of dealing with 3 abstractions. I suggest we take this topic into a separate task.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: Removed some obvious docs.

projects/vdk-plugins/vdk-meta-jobs/src/vdk/plugin/meta_jobs/plugin_entry.py
yonitoo: Removed it.

PR #1885, opened 2023-04-12
control-service: add the frontend to helm

murphp15:
# What
Within the introduction of the frontend docker image we want it to appear when ppl visit the host url.

# How
In this PR we introduce:
1. a new deployment and service to support running the frontend. 
2. an ingress rule to send requests from the root of the host to the frontend and an ingress rule to send any requests for /data-jobs to the control-service. 
3. Support for reading the frontend docker image from a private repo which is something we need within VMWare. 
4. Update the control-service to the a newer version
5. add support for basic configuration of the frontend service e.g. how much cpu the deployment is given. 

What is not in this PR:
1. passing configuration to the frontend via properties or volumeMounts.

# How has this been tested
Locally

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl
@@ -88,6 +88,29 @@ Also, we can't use a single if because lazy evaluation is not an option
 {{- end -}}
 {{- end -}}
 
+
+{{/*
+Return the proper operations ui image name
+*/}}
+{{- define "operations-ui.image" -}}
+{{- $globalRegistryOverrideName := .Values.image.globalRegistryOverride -}}
+{{- $registryName := .Values.operationsUi.registry -}}
+{{- $repositoryName := .Values.operationsUi.repository -}}
+{{- $tag := .Values.operationsUi.tag | toString -}}
+{{/*
+Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
+but Helm 2.9 and 2.10 doesn't support it, so we need to implement this if-else logic.
+Also, we can't use a single if because lazy evaluation is not an option
+*/}}
+{{- if $globalRegistryOverrideName }}
+    {{- printf "%s/%s:%s" $globalRegistryOverrideName $repositoryName $tag -}}
+{{- else if .Values.global.imageRegistry }}
+    {{- printf "%s/%s:%s" .Values.global.imageRegistry $repositoryName $tag -}}
+{{- else -}}
+    {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+{{- end -}}
+{{- end -}}
+
 {{/*
 Return the proper Control Service deployment builder image name
 */}}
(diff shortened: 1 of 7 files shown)


ivakoleva:
> Before this PR, if ingress was enabled the control-service was mapped to the root of the host url.
> After this PR,
> 
> frontend-ui is mapped to the root of the host url.
> control-service is mapped to /data-jobs

May you also elaborate on Swagger UI?

projects/control-service/projects/helm_charts/pipelines-control-service/templates/_helpers.tpl
```diff
@@ -244,6 +267,10 @@ Image Pull Secret in json format
     {{ include "buildImagePullSecretJson" (list .Values.image.registry .Values.image.registryUsernameReadOnly .Values.image.registryPasswordReadOnly) }}
 {{- end }}
 
+{{- define "operationsUiPullSecretJson" }}
+    {{ include "buildImagePullSecretJson" (list .Values.operationsUi.registry .Values.operationsUi.username .Values.operationsUi.password) }}
+{{- end }}
+
```
ivakoleva, line 273: I find those configs like `.Values.image.registryUsernameReadOnly` to be more understandable (due username or password are not suggesting what those are about)

murphp15: @ivakoleva I have now updated it to be `image.username`. Do you think that is descriptive enough? 
 
In fact I think my one is more descriptive. 

`.Values.image.registryUsernameReadOnly` it is unclear what image this username would be pulling. 


Whereas with my it is obvious that the username is for the image operationsUi. 

Mine is also more consistent with the rest of the image usernames and passwords. e.g 


```
{{ include "buildImagePullSecretJson" (list .Values.deploymentBuilderImage.registry .Values.deploymentBuilderImage.username .Values.deploymentBuilderImage.password) }}
```

projects/control-service/projects/helm_charts/pipelines-control-service/templates/ingress.yaml
```diff
@@ -16,7 +16,14 @@ spec:
...
       paths:
       - path: {{ .Values.ingress.path | default "/" }}
-        pathType: {{ .Values.ingress.pathType | default "ImplementationSpecific" }}
+        pathType: "ImplementationSpecific"
+        backend:
+          service:
+            name: {{ .Release.Name }}-ui
+            port:
+              number: 8091
+      - path: {{ .Values.ingress.path | default "/data-jobs" }}
```
DeltaMichael, line 25: So what's the full path to control service going to be exactly? I'd just like to avoid a situation where the url is `{host}:{port}/data-jobs/data-jobs`

DeltaMichael (approved):

murphp15:
> > Before this PR, if ingress was enabled the control-service was mapped to the root of the host url.
> > After this PR,
> > frontend-ui is mapped to the root of the host url.
> > control-service is mapped to /data-jobs
> 
> May you also elaborate on Swagger UI?

@ivakoleva  there is no impact on the swagger url as that starts with /data-jobs. 
In fact no endpoint urls for the control-plane are changed so I am going to remove the breaking-change section from the description.

projects/control-service/projects/helm_charts/pipelines-control-service/templates/ingress.yaml
murphp15: Yeah it will just be a single /data-jobs.

ivakoleva (approved):

PR #1912, opened 2023-04-18
vdk-server: quickstart vdk now includes the UI

murphp15:
# Why
As part of releasing the UI we want to include it in the quickstart vdk process. 

# What
small changes to config and log url to the UI

# How has this been tested
locally


DeltaMichael (approved):

yonitoo (approved):

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
```diff
@@ -673,6 +674,8 @@ def __helm_install_command(self, git_server_ip):
             "--set",
             "ingress.enabled=true",
             "--set",
```
ivakoleva, line 676: I may have missed it, did we update the docs on this?
VEP-1507 and quickstart-vdk ones

ivakoleva (approved):

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
murphp15: reverted.

PR #1946, opened 2023-04-24
Control-service: fix python api release

murphp15:
# Why
We are failing to build the python api client in the main branch. 
There are two issues.
The first is that we are trying to install python using apk. 
This is not needed and is causing errors. 

Also we are running out of memory generating the api client.
Upgrading to the latest version of the dependency has fixed this. 
I was also able to delete .mustache files that we needed in our repo as a workaround to issues from using an older version of openapi-generator. 

# How has this been tested.
Tested manually by making a commit where the client is built for any change. 
Also integration tests


dakodakov (approved):

ivakoleva (approved):
LGTM, thanks for looking into this.

PR #1957, opened 2023-04-25
frontend: Auth configurations organized

ivakoleva:
An appConfig may contain auth, and other configurations.

Refactored the auth-related config into one `auth` container. Traced the configs, and documented their purpose and usages.

Testing done: verified locally; cicd


yonitoo (approved):
LGTM!

DeltaMichael (approved):

PR #1958, opened 2023-04-25
frontend: Toggleable auth

ivakoleva:
We need a feature flag for skipping OAuth2, so the frontend can 
operate with a Control Service where security is disabled. It is the
default use case when having quickstart-vdk for local demo or dev
installation.

This is a stacked PR onto auth configurations organized from
https://github.com/vmware/versatile-data-kit/pull/1957
Introduced authSkipped flag defaulting to false, due security standards.
The top right corner with username and logout dropdown are rendered 
in case auth is enabled.

Testing done: verified both cases locally; added testing


projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
```diff
@@ -30,7 +30,12 @@ export class AppConfigService {
         return this.appConfig;
     }
 
+    getSkipAuth(): boolean {
+        return this.appConfig.auth.skipAuth;
+    }
+
     getAuthCodeFlowConfig(): AuthConfig {
+        if (this.getSkipAuth()) return new AuthConfig();
```
gorankokin, line 38: nit: My personal opinion when we write typescript is to have curly brackets after if statement even if it's without else for better readability, because after transpile to javascript it will undergo minify uglify and will get optimized for Browser execution.
Our main goal should be to make it user friendly as much as possible.
This comment is valid for all places where there are no curly brackets because there is no else.
e.g.

```
if (this.getSkipAuth()) {
    return new AuthConfig();
}

File: projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.ts [hunk 2]
```diff
@@ -24,33 +24,42 @@ export class AppComponent implements OnInit {
 
     constructor(
         private readonly appConfigService: AppConfigService,
-        private readonly oauthService: OAuthService,
+        @Optional() private readonly oauthService: OAuthService,
```
gorankokin, line 27: Because `OAuthModule` is always imported in the `AppModule` (import happens in build time), service `OAuthService` will be always provided from `OAuthModule` (source code - https://github.com/manfredsteyer/angular-oauth2-oidc/blob/3ba7303b5a6fdbae5ca214764c9191637cd00126/projects/lib/src/provider.ts#L23) and Angular dependency injection will always inject that service into `AppComponent` so `@Optional` annotation won't change how it is resolved in runtime.
It is not wrong, but doesn't make any change in runtime, except for the unit tests purposes in the tests where service stub is not provided.

gorankokin (approved):

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.spec.ts
```diff
@@ -19,25 +19,55 @@ import {
...
 describe('AppComponent', () => {
     let routerServiceStub: jasmine.SpyObj<RouterService>;
-    let oAuthServiceStub: jasmine.SpyObj<OAuthService>;
     let navigationServiceStub: jasmine.SpyObj<NavigationService>;
     let viewContainerRefStub: jasmine.SpyObj<ViewContainerRef>;
     let dynamicComponentsServiceStub: jasmine.SpyObj<DynamicComponentsService>;
     let confirmationServiceStub: jasmine.SpyObj<ConfirmationService>;
     let urlOpenerServiceStub: jasmine.SpyObj<UrlOpenerService>;
     let appConfigServiceStub: jasmine.SpyObj<AppConfigService>;
+    const configureTestingModule = (providersAdditional: any[]) => {
```
gorankokin, line 34: You could use here for type `Provider[]`

gorankokin: Most of the linters are configured to WARN when any type is used.

ivakoleva: Yup, I'm familiar with a Provider (I've browsed the data model before to understand it better).
However, the field providers that we do assign to, is of type any[], that's the contract.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.ts
ivakoleva: That's true, the optional is a left-over from my prior attempt to have a conditional modules injection, and optimize the application context to only what's needed.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.html
```diff
@@ -3,8 +3,10 @@
    ~ SPDX-License-Identifier: Apache-2.0
   -->
 
-<clr-main-container [attr.data-automation]="!!idToken ? 'vdk' : null">
-    <ng-container *ngIf="idToken">
+<clr-main-container
+    [attr.data-automation]="!!skipAuth && !!idToken ? 'vdk' : null"
```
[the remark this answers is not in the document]
ivakoleva: It is correct (mind it's skipAuth rather than enable, due it's a good design practice to rather have a negating toggle). Also, I have again tested both use cases (auth enabled, and not) locally, and I am able to successfully auth and see the panel expected. For example
Screenshot 2023-04-26 at 18 48 16

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
ivakoleva: nit: I prefer having a one liner in case possible like here, it's shorter and more readable.

DeltaMichael (approved):

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.ts
gorankokin: Conditional module injection in a standard Angular application without some 3rd party module management, I think would be very hard to achieve, because Module Class decorator is resolved during build time and adding dynamic flags there is not an option.
You could check this article - https://mokkapps.de/blog/manually-lazy-load-modules-and-components-in-angular/ on a first check looks promising to load module in e.g. AppComponent

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
gorankokin: There is no problem with minifiers on one linears, but I think is more readable having curly brackets especially in a bigger methods that do multiple things.

PR #1960, opened 2023-04-25
frontend: Improve visibility of 'User error' messages

hzhristova:
Add new 'details' buttons to make error messages more visible. Note that these messages are not very frequent
Changes are needed in three pages:
1. Manage Data Jobs grid
2. Data Job Executions tab grid
3. Details page - timeline (on both manage and explore sections)

In manage page grid and executions grid, 'details' buttons will appear when the execution status is 'User Error', 'Platform Error' or 'Skipped'. 

When displaying in details page timeline, I show the entire error message without using signpost.

Testing done: verified locally

image
image
image


gorankokin (approved):

tozka (approved):
Thanks for the screenshots . They are the primary thing I reviewed. 

Looks good to me. 

Note that these kind of messages would become more informative and detailed over time. Let's imagine it's 4 times as long as the current text. Would it still look OK?

hzhristova:
Thanks for the notice, Antoni. If the message is longer, this wont be a problem for the signposts. 
For the timelines, the first 40 words will be shown and if the message is longer, a 'Show more' button will appear and the user will be able to expand it all
image
image

tozka:
> Thanks for the notice, Antoni. If the message is longer, this wont be a problem for the signposts. For the timelines, the first 50 words will be shown and if the message is longer, a 'Show more' button will appear and the user will be able to expand it  

Thanks for checking. Looks good

PR #1978, opened 2023-04-27
frontend: visibility of app components is configurable

DeltaMichael:
## Why?

As part of releasing the UI to the community,
we should be able to configure the visibility of
certain components that are not intended for
external use

## What?

Component visibility for explore page is configurable through AppConfigService
Routes are configurable through AppConfigService

## How was this tested?

Locally plus component specs

## What type of change is this?

Feature, breaking


projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.spec.ts
```diff
@@ -63,24 +53,42 @@ describe('AppComponent', () => {
...
+        const fixture = TestBed.createComponent(AppComponent);
+        const app = fixture.componentInstance;
+        expect(app).toBeTruthy();
+        expect(app.explorePageVisible).toBeTrue();
+    });
+
+    it('should create the app with explore page ignored', () => {
+        appConfigServiceStub.getConfig.and.returnValue({
+            auth: {},
+            ignoreRoutes: [],
```
ivakoleva, line 85: May you elaborate the purpose of ignoreRoutes?

DeltaMichael: if you ignore just the explore page component in the html, it won't be visible. However, if you navigate to `/explore/data-jobs` in the browser, you will still get the page. We have to ignore the route to prevent this from happening.

It works locally, but I couldn't figure out a good way to test it in the spec. Any ideas are welcome. 😕

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.ts
```diff
@@ -54,6 +54,10 @@ export class AppComponent implements OnInit {
         return this.oauthService.getIdentityClaims() ? this.getIdentityClaim('username') : 'N/A';
     }
 
+    get explorePageVisible(): boolean {
+        return !this.appConfigService.getConfig().ignoreComponents.includes('explorePage');
```
ivakoleva, line 58: seems you are now verifying the value of appConfig.json propagates to the component, when you should rather verify the value with the functional expectation

DeltaMichael: I don't understand what you're think should be changed here.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.spec.ts
DeltaMichael: Added the test in cypress. It passes locally. Let's see if it will pass in CI.

projects/frontend/data-pipelines/gui/e2e/integration/router/router.spec.js
```diff
@@ -0,0 +1,28 @@
...
+        it('navigates to get-started page when explore page route is ignored', () => {
+            BasePagePO.executeCypressCommand(
+                'appConfigInterceptorDisableExploreRoute'
+            );
+            // wait for login
+            BasePagePO.wireUserSession();
+            BasePagePO.initInterceptors();
+            BasePagePO.navigateTo();
+            // go to explore page url
+            cy.visit('/explore/data-jobs');
```
murphp15, line 21: that's slick.

murphp15 (approved):
Looks good.
Really like the look of the tests.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.ts
ivakoleva: nit: We now have a proper toggle able visibility testing in e2e tests, right? 
That's very cool.

I'll try to explain, 
in `projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.spec.ts`
on L101 and L102 you have disabled ignore, then
on L112 you evaluate the code here L67 that reads the config and returns the value.


So, the test in L112 named 'should create the app with no components ignored'
does actually perform a test if the config value propagated empty? Why? It tests the config property, rather than actually testing if page is visible, is what I'm saying.


The test looks like:
```
explorePageVisibleConfig = empty
expect(explorePageVisibleConfig) is empty
```
do we need it?

ivakoleva (approved):
LVGTM

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
```diff
@@ -8,22 +8,28 @@ import { HttpBackend, HttpClient } from '@angular/common/http';
...
 @Injectable({
     providedIn: 'root'
 })
 export class AppConfigService {
     private httpClient: HttpClient;
     private appConfig: AppConfig;
+    private router: Router;
 
-    constructor(private httpBackend: HttpBackend) {
+    constructor(private httpBackend: HttpBackend, private angularRouter: Router) {
```
gorankokin, line 21: Why declaring one class field `angularRouter` of type `Router` that is after that assigned to class field `router` of the same type `Router`, and then using the second one?
Adding modifier of type `public|protected|private` in class constructor in Typescript will create class field with the same name that later could be use in the class lifecycle for any type of Typescript class (Component|Service|Directive|pure Typescript/JavaScript class|etc...)

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.module.ts
```diff
@@ -68,7 +69,7 @@ export function lottiePlayerLoader() {
     ],
     declarations: [AppComponent, GettingStartedComponent],
     providers: [
-        { provide: AppConfigService, useClass: AppConfigService, deps: [HttpBackend] },
+        { provide: AppConfigService, useClass: AppConfigService, deps: [HttpBackend, Router] },
```
gorankokin, line 72: Why are we providing `HttpBackend` instead of `HttpClient`?

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
```diff
@@ -8,22 +8,28 @@ import { HttpBackend, HttpClient } from '@angular/common/http';
...
         this.httpClient = new HttpClient(httpBackend);
+        this.router = angularRouter;
     }
 
-    loadAppConfig(): Promise<void> {
-        return firstValueFrom(this.httpClient.get<AppConfig>('/assets/data/appConfig.json').pipe()).then((data) => {
-            this.appConfig = data;
+    async loadAppConfig(): Promise<void> {
```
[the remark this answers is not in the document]
DeltaMichael: It's the same thing we were doing before, just better syntax, imo. We need to return a promise. If you call the `async` method without `await`, it will not wait for the promise to resolve and just return it. This is what we want for the APP_Initializer as per https://angular.io/api/core/APP_INITIALIZER

projects/frontend/data-pipelines/gui/e2e/integration/router/router.spec.js
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// <reference types="cypress" />
+
+import { BasePagePO } from '../../support/pages/base/base-page.po';
+
+describe('Routing for pages', () => {
+    describe('smoke', { tags: ['@smoke'] }, () => {
+        it('navigates to get-started page when explore page route is ignored', () => {
```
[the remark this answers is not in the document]
DeltaMichael: I figured we didn't need it since we have a spec for the whole explore page. It's enabled by default.

https://github.com/vmware/versatile-data-kit/blob/person/mdilan/dynamic-routes-app-init/projects/frontend/data-pipelines/gui/e2e/integration/explore/data-jobs/data-jobs.spec.js

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
DeltaMichael: I'll remove the class field. I did this in the interest of testability. I was hoping to test the dynamic routes using an angular spec. I ended up moving it to the cypress tests, so no need for making the router testable.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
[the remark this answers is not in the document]
gorankokin: I think chaining the promises would give better flow control.
e.g.
```
loadAppConfig(): Promise<boolean> {
    const data = await firstValueFrom(this.httpClient.get<AppConfig>( '/assets/data/appConfig.json').pipe());
        this.appConfig = data;
        const localRoutes = routes.filter((route: { path: string }) => {
            return !data.ignoreRoutes.includes(route.path);
        });
        this.router.resetConfig(localRoutes);

        return firstValueFrom(this.httpClient.get<AppConfig>( '/assets/data/appConfig.json'))
                          .then((data) => {
                                  this.appConfig = data;

                                  try {
                                        const localRoutes = routes.filter((route: { path: string }) => {
                                                   return !data.ignoreRoutes.includes(route.path);
                                        });

                                        this.router.resetConfig(localRoutes);

                                        return true;
                                  } catch(e) {
                                         console.error('Failed to reset Router config');

                                         // when exception occurs you can either re-throw the exception and stop application bootstrap
                                         throw e;

                                         // or return boolean true and continue with application bootstrap and default config
                                         // return true;
                                  }
                          });
}
```

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.component.ts
DeltaMichael: I don't think we can test if the html renders using an Angular spec. Maybe I'll move this scenario to cypress as well.

gorankokin: It's possible to test html rendering in unit tests in Angular.
You could look into `placeholder.service.spec.ts` for examples, but in short we access `debugElement` from fixture or some other queried element and using `query` method we could select something from the html nodes.
e.g.

```
const gridDebugElement: DebugElement = fixture.debugElement.query(By.directive(ClarityDataGridComponentStub));
gridComponent = gridDebugElement.componentInstance;
```

Predicate provided to `query` method could be something of `By.directive` where we provide class reference of Component of Directive or `By.css` where we provide name of the html node (e.g. `div`, `p`, `span`, `lib-data-jobs-explore-grid`, etc...)

DeltaMichael: I tried doing it using cypress, but it increased the runtime beyond the 2 hour limit. Maybe what @gorankokin suggested is the way to go. Adding it as tech debt, because I've already spent too much time on it. 

https://jira.eng.vmware.com/browse/TAUR-6037

projects/frontend/data-pipelines/gui/e2e/support/commands.js
```diff
@@ -310,3 +310,51 @@ Cypress.Commands.add(
...
+            },
+            refreshTokenConfig: {
+                start: 500,
+                remainingTime: 360,
+                checkInterval: 60
+            }
+        },
+
+        ignoreRoutes: ['explore/data-jobs', 'explore/data-jobs/:team/:job'],
+        ignoreComponents: ['explorePage']
```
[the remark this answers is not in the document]
DeltaMichael: Added as tech debt. https://jira.eng.vmware.com/browse/TAUR-6038

projects/frontend/data-pipelines/gui/projects/ui/src/app/app.module.ts
DeltaMichael: Wiring the client itself doesn't work for some reason. I tested it locally.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts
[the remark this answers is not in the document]
DeltaMichael: Done.

projects/frontend/data-pipelines/gui/projects/ui/src/assets/data/appConfig.json
```diff
@@ -36,5 +36,7 @@
       "remainingTime": 360,
       "checkInterval": 60
     }
-  }
+  },
+  "ignoreComponents": [],
+  "ignoreRoutes": []
```
ivakoleva, line 41: Do we include any default configurations on hidden pages/routes, for example Explore Data Jobs feature?
I don't see any documented examples on potential configurations of these new options provided?

DeltaMichael: I'm not sure what the suggestion is here. We have examples as part of the tests. These will also be included in the helm chart config.

ivakoleva: Can you please link to the example on how to configure those settings, for hiding the Explore Data Jobs feature?

gorankokin (approved):

PR #2004, opened 2023-05-02
control-service: custom app config values can we passed to helm.

murphp15:
# Why
We want to be able to change the frontend config depending on the situation it is deployed in. 
This lets users provide their own appConfig file to the frontend. 
An example of this is toggling auth on and off. 

# What
Use k8s volumemounts to load in the config file.

# How has this been tested?
Deployed the helm locally.


yonitoo (approved):
Looks great to me!

duyguHsnHsn (approved):

PR #2009, opened 2023-05-03
vdk-dag: fix errors in the api responses

murphp15:


gageorgiev (approved):

projects/vdk-plugins/vdk-dag/tests/test_dag.py
```diff
@@ -327,6 +340,8 @@ def test_dag_concurrent_running_jobs_limit(self):
                         )  # assert that max concurrent running jobs is not exceeded
                     if request.method == "GET":
                         execution = json.loads(response.response[0])
+                        if isinstance(execution, list):
+                            execution = execution[0]
```
yonitoo, line 344: Is that check needed? We are parsing the response on the previous line by taking the first (and only) element of the response.response list, which is JSON. Therefore, isn't the type of "execution" always going to be dict?

yonitoo (commented):
Looks good overall! Left just one comment.

projects/vdk-plugins/vdk-dag/tests/test_dag.py
murphp15: Yeah this is needed some of the request return a list of executions and some return only a single one.

PR #2013, opened 2023-05-04
control-service: logs url can include team name

murphp15:
# Why
Often the URL where the logs are stored will need the team name. 
An example of this is when using our own api to serve the logs.

# What
team name can now be used in the URL 

# How has this been tested 
Updated unit tests. 
In the next two days I will be incorporating this change into quickstart vdk. 

I can wait till it is incorporated in that to merge this if ppl don't feel comfortable with the unit test.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionLogsUrlBuilder.java
```diff
@@ -50,6 +51,7 @@ public class JobExecutionLogsUrlBuilder {
   static final String EXECUTION_ID_VAR = "execution_id";
   static final String OP_ID_VAR = "op_id";
   static final String JOB_NAME_VAR = "job_name";
+  static final String TEAM_NAME_VAR = "team_name";
```
ivakoleva, line 54: Please update docs on config usages, for example https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml#L980

ivakoleva (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionLogsUrlBuilder.java
murphp15: Done! thanks

zverulacis (approved):

PR #2016, opened 2023-05-05
control-service: remove unused parameters

murphp15:
# Why 
I am trying to improve the stability of the integration tests. I have seen a way of improving them but it requires some code changes. 

Before making those code changes I want to push this clean up. as it is very hard to build on the functions that are currently there. 

I removed 4 parameters that are never used. 
I also removed podLabels which is used but it is always an empty map. So it never did anything.
 
# How was this tested
Unit and integration tests.


ivakoleva (commented):
May you provide context from GitHub history about why were those added?
Also, you can use stacked PRs to unblock further work

murphp15:
> May you provide context from GitHub history about why were those added? Also, you can use stacked PRs to unblock further work

These properties have existed since the inception of the repo. Do you want me to find their history in VMwares internal repos?

DeltaMichael:
This API is horrible. There should be no world in which we have to pass 11 parameters, 3 of which are collections to start a k8s job. Can we have a spike ticket to improve it in the future?

DeltaMichael (approved):
This API is horrible. There should be no world in which we have to pass 11 parameters, 3 of which are collections to start a k8s job. Can we have a spike ticket to improve it in the future?

Otherwise, I'm happy with the changes.

doks5 (approved):

ivakoleva (approved):

PR #2022, opened 2023-05-07
control-service: Fix backwards-compatibility issues

doks5:
Currently, there are some backwards-compatibility issues with Control Service deployments that do not integrate support for multiple python versions and move to introduce it. The issues surface, when service administrators decite to add multiple supported python versions in their private values.yaml configuration files. The new configurations are ignored and it is not possible to cleanly migrate to support multiple python versions, as only the old configurations through `datajobs.vdk.image` and `datajobs.deployment.dataJobBaseImage` are recognised.

This change relaxes the service behaviour, so that if the `datajobs.deployment.supportedPythonVersions` and `datajobs.deployment.defaultPythonVersion` configuration options are set, they would take precedence over the old configurations. Additionally, a bug is also fixed, where enabling or disabling data jobs, deployed before the support for multiple python versions was introduced, was failing after the new configurations are set, due to the default python_version being added to JobDeployment at patch request, which should not happen.

Testing Done: New and existing tests. Additionally, the whole setup was tested locally against a kind k8s cluster, and a stand-alone CockroachDB instance.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/DeploymentServiceTest.java
+++ b/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/DeploymentServiceTest.java
@@ -357,37 +357,42 @@ public class DeploymentServiceTest {
     jobDeployment.setEnabled(true);
     jobDeployment.setPythonVersion("3.10");
 
-    ApiConstraintError error = assertThrows(ApiConstraintError.class, () -> { deploymentService.patchDeployment(testDataJob, jobDeployment); });
+    ApiConstraintError error =
+        assertThrows(
+            ApiConstraintError.class,
+            () -> {
+              deploymentService.patchDeployment(testDataJob, jobDeployment);
+            });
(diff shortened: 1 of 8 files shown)


mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/deploy/DeploymentServiceTest.java
```diff
@@ -345,6 +349,50 @@ public void enableDeployment_sameEnabledStatus_updateSkipped() throws ApiExcepti
         .createCronJob(any(), any(), anyString(), anyBoolean(), any(), any(), any(), any());
   }
 
+  @Test
+  public void patchDeployment_setPythonVersion_failPatch() throws ApiException, ApiConstraintError {
```
mivanov1988, line 353:
```suggestion
  public void patchDeployment_notValidPythonVersion_shouldFail() throws ApiException, ApiConstraintError {
```

?

dakodakov (approved):

PR #2023, opened 2023-05-07
vdk-control-cli: Add support for python_version in config

doks5:
As part of the initiative to support multiple python versions for data job deployments, we need to make sure that users can configure their data jobs to use specific python versions, by setting the `python_version` property in the job's **config.ini** file.

This change introduces functionality in the vdk-control-cli client to read the python_version configuration set in the job's config.ini file, and deploy the job through `vdk deploy ...`

It also adds logic to print the python version of data job deployments as part of the `vdk deploy --show ...` and `vdk execute --show ...` commands.

Testing Done: New and existing tests.


projects/vdk-control-cli/tests/resources/test-job-python-version/config.ini
```diff
@@ -0,0 +1,10 @@
+[owner]
+team = test-team
+[job]
+schedule_cron = 16 11 * * *
+python_version = 3.8
```
ivakoleva, line 5: Are there any other use cases known? For example, what's the expectation in case of a typo / non-configured valid version ?

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/execute.py
```diff
@@ -50,6 +50,8 @@ def __model_executions(executions, output_format: OutputFormat) -> str:
         def transform_execution(e: DataJobExecution):
             d = e.to_dict()
             d["job_version"] = e.deployment.job_version
+            if e.deployment.python_version:
+                d["python_version"] = e.deployment.python_version
             del d["deployment"]
             return d
```
ivakoleva, line 57: nit: may add PR description of what's the change in the projects/vdk-control-cli/tests/resources/job-zip/test-job-python-version.zip ?

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_execute.py
```diff
@@ -1,7 +1,6 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import json
-import os
 from unittest.mock import patch
```
ivakoleva, line 5: May you link if described in VEP-1739, what happens in case of local execution of a data job, with added `python_version` in config.ini - in case of mismatch with local python environment?

ivakoleva (approved):

mivanov1988 (approved):

projects/vdk-control-cli/tests/resources/test-job-python-version/config.ini
doks5: If the python version set in the config.ini is not part of the supported python versions, the Control Service returns 403 Bad Request, with a message indicating what python version was passed, and what python versions are supported.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/execute.py
doks5: Updated the PR description.

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_execute.py
doks5: It is not yet described in the proposal. Will update the vep in a separate PR.
As to the question of what happens if the python_version set in the config is different from the one of the local environment, the answer is, nothing. The `python_version` property (if set, as it is optional) is not taken into account until the deployment of the job is initiated.
Do you think we need some form of validation that the `python_version` is the same as the local environment?

projects/vdk-control-cli/tests/resources/test-job-python-version/config.ini
ivakoleva: @doks5 is this scenario test-covered?

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_execute.py
ivakoleva: I think it may be very useful to alert the user in case of a mismatch between (cloud) python version targeted and local python env setup (used for development). This may prevent a lot of confusion.

doks5: This is a good idea. I'll add this functionality in a separate PR, as the change would need to be made in vdk-core.
Thank you for the review

projects/vdk-control-cli/tests/resources/test-job-python-version/config.ini
doks5: We have tests covering python_version mismatches in the Control Service test suit. I don't think it would be useful to have such tests in the cli, as it does not really validate anything, it's just a medium that passes data to the Control Service.

PR #2024, opened 2023-05-08
vdk-examples: Change Meta Jobs to DAGs in examples

gageorgiev:
This PR corrects two examples to bring them in line with the changed name of VDK DAGs. It also includes a note in the VEP to mention this change, however the VEP remains unchanged otherwise.


ivakoleva (approved):

examples/dag-with-args-example/README.md
```diff
@@ -124,6 +124,7 @@ def run(job_input: IJobInput):
     data_job_dir = pathlib.Path(job_input.get_job_directory())
     data_file = data_job_dir / "data.json"
 
+    db_catalog = job_input.get_arguments().get("db_catalog")
```
yonitoo, line 127: The default value for the catalog is "memory" so IMO we should remove it in order to help people with the idea that they could switch to another DB if they want to and not necessarily stick to Trino. Moreover, it is not passed as argument anymore to the JOBS_RUN_ORDER dict so it would have no effect.

examples/dag-with-args-example/README.md
```diff
@@ -510,7 +511,7 @@ from vdk.plugin.dag.dag_runner import DagInput
 
 JOBS_RUN_ORDER = [
     {
-        "job_name": "ingest-job-table-one",
+        "job_name": "ingest-job-table-one ",
```
yonitoo, line 514: nit: why is this space needed?

examples/dag-with-args-example/README.md
```diff
@@ -588,8 +589,9 @@ team = my-team
 
 [vdk]
 dags_max_concurrent_running_jobs = 2
-dags_delayed_jobs_min_delay_seconds = 1
+
```
yonitoo, line 592: nit: why is this space needed? If it's to emphasize on the functionality of dags_max_concurrent_running_jobs, then it's ok.

yonitoo (commented):
Overall looks very good, left a few comments.

examples/dag-with-args-example/README.md
gageorgiev: dropped it

examples/dag-with-args-example/README.md
gageorgiev: dropped

examples/dag-with-args-example/README.md
gageorgiev: dropped this

yonitoo (approved):
Awesome! Ship it!

PR #2028, opened 2023-05-09
control-service: fix oom tests

murphp15:
# Why
In this PR I fix a very unstable oom test. 
The test is very unstable for a number of reasons. 

1. sometime the job actually succeeds, it seems to be able to compelte with in the memory limits after a number of retries. 
We can put the memory requirement any lower to ensure an error because 6mi is the lower limit supported by k8s. 
2. on of the assertions we were making is that the logs are there. However often if the jobs is killed at the very start there are no logs. 

# What 
I fix it by using a cronjob template which doesn't let the job restart. This means if it fails once it will report as failed. 

# How has this been tested?
integration tests.


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobInitContainerOOMIT.java
```diff
@@ -12,15 +12,22 @@
...
 import org.springframework.test.context.TestPropertySource;
 
+import static com.vmware.taurus.datajobs.it.common.JobExecutionUtil.*;
+
 @Slf4j
 @TestPropertySource(
     properties = {
       "datajobs.deployment.initContainer.resources.requests.memory=6Mi",
       "datajobs.deployment.initContainer.resources.limits.memory=6Mi",
+      // This is a standard cron job template except restartPolicy is set to never so that when a
```
murphp15, line 27: load my new cron job template.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobInitContainerOOMIT.java
```diff
@@ -40,7 +47,23 @@ public void testDataJob_causesOOM_shouldCompleteWithUserError(
     String executionId = executeDataJobResult.getRight();
 
     // Check the data job execution status
-    JobExecutionUtil.checkDataJobExecutionStatus(
+    testDataJobExecutionRead(
```
murphp15, line 50: call most of the checks explicitly instead of using the helper method we were using previously so I can skip the logs exist assertion.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/JobExecutionUtil.java
```diff
@@ -150,7 +150,7 @@ public static ImmutablePair<String, String> executeDataJob(
     return ImmutablePair.of(opId, executionId);
   }
 
-  private static void testDataJobExecutionRead(
+  public static void testDataJobExecutionRead(
```
murphp15, line 153: Make methods public. 
They are clean and I think it makes sense for them to be public. I need them in the test class above.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -332,16 +332,10 @@ private V1beta1CronJob loadConfigurableV1beta1CronjobTemplate() {
...
-    File datajobTemplateFile = new File(datajobTemplateFileLocation);
-    if (!datajobTemplateFile.isFile()) {
-      log.warn("Datajob template location '{}' is not a file.", datajobTemplateFileLocation);
-      return null;
-    }
-
     try {
       // Load the configurable datajob template file.
-      return loadV1beta1CronjobTemplate(datajobTemplateFile);
+      return loadV1beta1CronjobTemplate(
```
murphp15, line 337: there was a bug in the code where we weren't loading in files correctly.

mivanov1988 (approved):

mivanov1988:
Good job!

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -332,16 +332,10 @@ private V1beta1CronJob loadConfigurableV1beta1CronjobTemplate() {
...
+          new ClassPathResource(datajobTemplateFileLocation).getFile());
```
ivakoleva, line 338: Why using ClassPathResource for loading the data job template file, in favour of FileSystemResource?

about [classpathresource](https://docs.spring.io/spring-framework/docs/3.2.x/spring-framework-reference/html/resources.html#resources-implementations):

> supports resolution as java.io.File if the class path resource resides in the file system, but not for classpath resources which reside in a jar and have not been expanded (by the servlet engine, or whatever the environment is) to the filesystem

Going into that direction, I wonder why using the File outdated paradigm in favour of the Path API introduced with v1.7 from 2011?

ivakoleva (approved):

PR #2031, opened 2023-05-09
control-service: remove a test that is testing behaviour that doesn't exist

murphp15:
# Why

Currently, the patch method merges fields without checking for changes, which is not accurately reflect the desired behavior of this test. I suggest deleting the test.

# What
The reason that is passes is that they check it never calls createCronJob and updateCronJob.
However there are two implementations of those functions and it is actually calling the other one. 

# How has this been tested? 
unit tests


mivanov1988:
Hi @murphp15, 

Currently, the patch method merges fields without checking for changes, which is not accurately reflect the desired behavior of this test. I suggest deleting the test.

gageorgiev (approved):

doks5 (approved):

mivanov1988 (approved):

PR #2038, opened 2023-05-10
control-service: delete unused method

murphp15:
# Why
Obviously deleting unused methods is a good idea. 
However in this case it is even more important. 

There are tests asserting that create or update are not called. 
But they are checking the wrong methods and so before this PR those assertions are meanigless. 

# How has this been tested
unit tests

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -827,42 +827,6 @@ public abstract class KubernetesService implements InitializingBean {
       boolean enable,
       V1Container jobContainer,
       V1Container initContainer,
-      List<V1Volume> volumes)
-      throws ApiException {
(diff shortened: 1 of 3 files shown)


gageorgiev (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -820,42 +820,6 @@ public Set<String> listCronJobs() throws ApiException {
...
-          volumes,
-          Collections.emptyMap(),
-          Collections.emptyMap(),
-          List.of(""));
-    }
-  }
-
   public void createCronJob(
       String name,
       String image,
```
ivakoleva, line 825: nit: so you are deleting those overloaded methods, and tracing to rewrite aall the old 7parameters-long usages to the 10params-long usages?

Isn't it shorter, to adapt the 7parameters-long method to the 10par one directly (fix the impl to shortcut and reuse the broader one)? No need to rewrite each and every usage...

ivakoleva (approved):

PR #2039, opened 2023-05-10
heartbeat: add test that pings the frontend url

DeltaMichael:
Note: I'll add the test to the CI in a separate PR

## Why

The frontend is packaged in the quickstart-vdk helm chart. Adding a
smoke test to check if the frontend starts correctly provides
an additional layer of release confidence.

## What

Split test templates into HeartbeatTest inherits from HeartbeatBaseTest
HeartbeatBaseTest has the setup, execute_test and clean_up methods.
Its run_test method sets up, executes and cleans up on errors

HeartBeatTest does the same, but in addtion, it

- Creates a Data Job using VDK CLI for managing jobs
- Deploys the data job using the same Versatile Data Kit SDK
- Run database test to verify the job works correctly in cloud runtime.
- Deletes the Data Job using VDK CLI

Test templates now have run_test methods. Run_test methods contain the logic, which
was previously in heartbeat.py for HeartbeatTest and plain logic, which
does not include data job creation for HeartbeatBaseTest

Add a test that does a simple get request to the
frontend URL and expects a success status

## How was this tested

Tested locally by running the quickstart-vdk and running the
trino test against a local instance of trinodb

Tested locally by running the ping frontend test against
a local instance of the frontend

Ran tests included in heartbeat-vdk

CI

Feature, breaking

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
@@ -3,9 +3,7 @@
 import logging
 
 from vdk.internal.heartbeat.config import Config
-from vdk.internal.heartbeat.heartbeat_test import create_test_instance
-from vdk.internal.heartbeat.heartbeat_test import HeartbeatTest
-from vdk.internal.heartbeat.job_controller import JobController
+from vdk.internal.heartbeat.heartbeat_base_test import HeartbeatBaseTest
 from vdk.internal.heartbeat.reporter import TestDecorator
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


murphp15:
BARE_TEST can you rename that to SKIP_CONTROL_PLANE_TESTS?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
murphp15, line 1: Maybe your one could be a completely separate class?

projects/vdk-heartbeat/src/vdk/internal/heartbeat/config.py
```diff
@@ -109,6 +109,12 @@ def __init__(self, config_file=None):
             "CONTROL_API_URL", "VDK_HEARTBEAT_CONTROL_SERVICE_URL", is_required=False
         )
 
+        # The VDK Frontend URL
+        # If no value is found, argument will not be passed and default will be used
+        self.frontend_url = self._get_atleast_one_value(
+            "FRONTEND_URL", "VDK_HEARTBEAT_FRONTEND_URL", is_required=False
```
DeltaMichael, line 115: @murphp15 

>I would think that you shouldn't need the frontend_url. it will always be the control plane url too.
>The control plane will be hit by including /data-jobs in the url.


We have this in the post deploy test config.

```
CONTROL_API_URL=http://cicd-control-service-svc:8092
```

Not sure how it will work exactly with the port as part of the URL. I'd like to avoid parsing it if possible.

projects/vdk-heartbeat/src/vdk/internal/heartbeat/hearbeat.py
```diff
@@ -28,6 +28,12 @@ def __init__(self, config: Config):
     @TestDecorator()
     def run(self):
         run_test = create_test_instance(self._config)
+        if self._config.bare_test:
+            self._run_bare_test(run_test)
```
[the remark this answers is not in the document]
DeltaMichael: Maybe we can call it a base test and the other one can be a control plane test. I was having trouble picking the name as well, so suggestions are welcome. I was thinking of doing a decorator pattern, but figured it might be too complex.

ivakoleva:
It seems confusing, that the proposal for vdk-heartbeat is to be configurable towards Control Service tests only. At the same time, having the Frontend tests always running.

I am not convinced this is the entire VDK heartbeat API change needed. The reason is, that the Control Service is a hard prerequisite as-is while the Frontend service can be used in variety of ways (not necessarily using the service from the helm chart, for example like a NPM library). This is visualised with the Data Framework components https://github.com/vmware/versatile-data-kit#what-vdk-can-do

May you elaborate on the configurational changes you propose, before and after -> covering both services?

murphp15 (approved):

tozka (commented):
Please update README.md with instructions on how to run the new test.

tozka (approved):

PR #2049, opened 2023-05-11
frontend: shareable links with query params for Data Jobs grids

hzhristova:
Save filter changes from Data Jobs Explore and Manage grids as queryParams:

Changes :
1. Add new GridFilter interface that contains possible filters
2. Save initial query params if any
3. Save every filter change as query param and update the link

Testing: local


gorankokin (approved):

projects/frontend/data-pipelines/gui/e2e/integration/manage/data-jobs/data-jobs.spec.js
```diff
@@ -221,7 +221,12 @@ describe(
                     .should('deep.equal', {
                         pathSegment: '/manage/data-jobs',
                         queryParams: {
-                            search: testJobsFixture[0].job_name
+                            search: testJobsFixture[0].job_name,
+                            jobName: testJobsFixture[0].job_name.substring(
+                                0,
+                                20
```
ivakoleva, line 227: nit: why was 20 chosen in particular?

gorankokin: Generic data jobs names, created in context of every e2e tests execution match this pattern `cy-e2e-vdk-lib-{{pipeline_id}}-v12` (e.g. **cy-e2e-vdk-lib-1a4d2540515640d3-v12**) and in order to have multiple job in the list 20 characters length was chosen, and it will search for something like **cy-e2e-vdk-lib-1a4d2**

gorankokin: in the past we used extra long strings for some test but we had problem with Cypress older versions.
there were intermittent failures, because framework was failing to enter the whole text, entered text was truncated.
and to be on the safe side we started using shorted strings, and this was looking promising because will job names will contain part of the data pipelines id in order to have less jobs in the explore or manage data jobs page and avoid intermittent tests failures

PR #2057, opened 2023-05-12
vdk-jupyter: UI failing cell marks

duyguHsnHsn:
What:
More informative and easy to understand Run message on notebook failure are introduced:
On failure return more detailed error message (where the failure happened, what is the error) + a button which allows the user to navigate to the failing cell
Here are some screenshots of the newly introduced cell marks:
Screenshot 2023-05-11 at 16 17 36
Screenshot 2023-05-11 at 16 17 51

Here is a video how it works: 

https://github.com/vmware/versatile-data-kit/assets/87015481/5adbb894-b2c5-42b5-a448-6b747a66536e

Why: This makes working with errors that are produced from notebook cells such more easier. 

Tests: jest tests


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -156,3 +157,57 @@ def create_deployment(
...
+            x
+            for x in pathlib.Path(job_path).iterdir()
+            if (x.name.lower().endswith(".ipynb"))
+        ]
+        for notebook_file in notebook_files:
+            with open(notebook_file) as f:
+                notebook_json = json.load(f)
+                cell_index = 0
+                for jupyter_cell in notebook_json["cells"]:
+                    if jupyter_cell["id"] == failing_cell_id:
```
ivakoleva, line 184: nit: why looping over the entire list of cells, while having the exact cell_id at hand?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
```diff
@@ -98,17 +104,23 @@ export const trackVdkTags = (
           vdkCellIndices.push(cellIndex);
         cellIndex++;
       }
-
+      // this case covers the use case when the notebook is loaded for the first time
+      if (vdkCellIndices.length == 0) {
+        vdkCellIndices = await getVdkCellIndices(
```
ivakoleva, line 109: nit: `if (not vdkCellIndices)`? (I see the codacy warning of 0)

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
```diff
@@ -5,16 +5,20 @@
...
+export const addNumberElement = (number: Number, node: Element) => {
   const numberElement = document.createElement('div');
   numberElement.innerText = String(number);
-  numberElement.classList.add('jp-vdk-cell-num');
-
+  if (node.classList.contains('jp-vdk-failing-cell')) {
+    numberElement.classList.add('jp-vdk-failing-cell-num');
+  } else {
+    numberElement.classList.add('jp-vdk-cell-num');
+  }
```
ivakoleva, line 17: nit: there is a shorter way to deduplicate `numberElement.classList.add(` to once, and have the conditional arg

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -113,3 +113,73 @@ export async function jobdDataRequest(): Promise<void> {
...
+    };
+    const data = await requestAPI<Array<Number>>('vdkIndices', {
+      body: JSON.stringify(dataToSend),
+      method: 'POST'
+    });
+    return data;
+  } catch (error) {
+    await showErrorMessage('Encountered an error. Error:', error, [
+      Dialog.okButton()
+    ]);
```
ivakoleva, line 182: nit: I can frequently see a similar code block:
```
await showErrorMessage('Encountered an error. Error:', error, [
      Dialog.okButton()
    ]);
```
May extract commons, to deduplicate

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -113,3 +113,73 @@ export async function jobdDataRequest(): Promise<void> {
     );
   }
 }
+
+/**
+ * Sent a POST request to the server to get more information about the notebook that is failing the job run
+ * Returns the path to the notebook file and the index of the cell that is failing in the notebook
+ */
```
ivakoleva, line 120: I was looking into PR's description, to find out where the error message is being displayed. May you add a screenshot of the stracktrace/error visualised?

ivakoleva, line 120: May you elaborate the function documentation, regarding failingCellId argument? For example, what happens with different values of failingCellId passed? Like, empty, non-failing-cell, failing-cell?

ivakoleva (approved):
LGTM

ivakoleva (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -167,3 +185,5 @@ def add_handler(handler, endpoint):
     add_handler(CreateJobHandler, "create")
     add_handler(LoadJobDataHandler, "job")
     add_handler(CreateDeploymentHandler, "deploy")
+    add_handler(GetFailingNotebookHandler, "failingNotebook")
+    add_handler(GetVdkIndicesHandler, "vdkIndices")
```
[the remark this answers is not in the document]
duyguHsnHsn: @ivakoleva those are cell indices of cells that are tagged with 'vdk'

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: @ivakoleva if you are talking about the job run error message, you can see the screen recording that I have added to the PR, it is visualised after the run and there I click "Get failing cell"

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -113,3 +113,73 @@ export async function jobdDataRequest(): Promise<void> {
...
+export async function getFailingNotebookInfo(failingCellId: string): Promise<{
```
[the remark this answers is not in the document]
duyguHsnHsn: @ivakoleva it is not numerical, you can see the screen recording there is an Error message displayed with cell_id 
if you stop at 0:10, you can read it

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
ivakoleva: I remember now I saw it in Fri, the video looks very good 😊

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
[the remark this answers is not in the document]
ivakoleva: So it is "job2_10" where 10 is the number of the cell starting from 0, right (I remember reading the algorithm for counting the index)? What is "job2"?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: It is not possible to get the cell with the specified ID without looping through the cells in the notebook JSON object.
This is because the notebook JSON object stores cells as a list, and there is no direct way to access a specific cell in the list without iterating over the cells. The cell ID is not an index in the list, but rather an identifier that is stored within each cell's metadata.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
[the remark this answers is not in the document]
ivakoleva: vdkCellIndices?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
ivakoleva: I don't understand why looping over aall the notebooks in potential existence is needed, to understand what notebook the failing cell comes from?

May you provide an example of two short notebook JSON cells?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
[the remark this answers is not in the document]
duyguHsnHsn: "job" is the name of the notebook, if the notebook is named "nb" the steps would be named nb_1, nb_2, etc. (one step is one jupyter cell) 
the cell id is something different, it is an id that jupyter generates for the cell

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: @ivakoleva yes, a lo file is created with all the logs, so if the user wants to see more information they can just check the log file

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -156,3 +157,57 @@ def create_deployment(
...
+                        notebook_path = str(notebook_file).replace(os.getcwd(), "")
+                        return {
+                            "path": str(notebook_path),
+                            "failingCellIndex": str(cell_index),
+                        }
+                    cell_index += 1
+        return {"path": "", "failingCellIndex": ""}
+
+    @staticmethod
+    def get_vdk_indices(notebook_path: str):
```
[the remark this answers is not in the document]
duyguHsnHsn: changed

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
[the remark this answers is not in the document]
duyguHsnHsn: changed

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: but it is just 1 function would it get any better if we extract it?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
duyguHsnHsn: done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
```diff
@@ -71,41 +71,139 @@ export default class RunJobDialog extends Component<IJobPathProp> {
...
+            </p>
+          </div>
+        ),
+        buttons: [Dialog.okButton()]
+      });
+    } else {
+      message = 'ERROR : ' + message;
+      const errorMessage = new VdkErrorMessage(message);
+      if (
+        !docManager ||
```
DeltaMichael, line 98: Nit: It might be better to export this condition to a variable with a proper name, so it's clear what it does. Decomposing the condition will also help.

```
const errorsHandled = docManager && (await handleErrorsProducedByNotebookCell(errorMessage, docManager))

if (!errorsHandled) {
  // do stuff
}
```

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
```diff
@@ -71,41 +71,139 @@ export default class RunJobDialog extends Component<IJobPathProp> {
...
+          buttons: [Dialog.okButton()]
+        });
+      }
+    }
   }
 }
+
+export const findFailingCellId = (message: String): string => {
+  const regex = /cell_id:([0-9a-fA-F-]+)/;
+  const match = message.match(regex);
```
DeltaMichael, line 121: Nit: Consider destructuring to make the code clearer. It's not clear why you're returning `match[1]` and not `match[0]`, for example. You can use proper names for whatever matches instead of `firstMatch` and `secondMatch`.

```
const [firstMatch, secondMatch] = message.match(regex);
return secondMatch ? secondMatch : ''
```

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
```diff
@@ -71,41 +71,139 @@ export default class RunJobDialog extends Component<IJobPathProp> {
...
+    const { path: nbPath, cellIndex: failingCellIndex } = await getNotebookInfo(
+      failingCellId
+    );
+    if (nbPath) {
+      const navigateToFailingCell = async () => {
+        const notebook = docManager.openOrReveal(nbPath);
+        if (notebook) {
+          await notebook.revealed; // wait until the DOM elements are fully loaded
+          const children = Array.from(notebook.node.children!);
+          if (children) {
```
DeltaMichael, line 144: You don't need this check. Correct me if I'm wrong, but you'll always get an array from the above line, even if it's an empty one. Then you'll run the forEach on an empty array, which will do nothing.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
```diff
@@ -71,41 +71,139 @@ export default class RunJobDialog extends Component<IJobPathProp> {
...
+            children.forEach(async element => {
+              if (element.classList.contains('jp-Notebook')) {
+                const cells = element.children;
+                const inx = Number(failingCellIndex);
+                if (inx < 0 || inx > cells.length) {
```
DeltaMichael, line 149: Is it possible to get this kind of garbage number for the failing cell index? Why does it happen?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
```diff
@@ -71,41 +71,139 @@ export default class RunJobDialog extends Component<IJobPathProp> {
...
+  const regex = /cell_id:([0-9a-fA-F-]+)/;
+  const match = message.match(regex);
+  if (match) return match[1];
+  return '';
+};
+
+/**
+ * Seperate handling for notebook errors - option for the user to navigate to the failing cell when error is produced
+ */
+export const handleErrorsProducedByNotebookCell = async (
```
DeltaMichael, line 129: Some general feedback on this function: It's a little hard to understand what's going on. There's a lot of room for improvement.

1. You can export the dialog templates to constants.
2. You can do away with at least some of the checks
3. You can avoid so much nesting by exporting some of the logic to separate functions.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -113,3 +113,71 @@ export async function jobdDataRequest(): Promise<void> {
...
   }
 }
+
+/**
+ * Sent a POST request to the server to get more information about the notebook that includes a cell with the given id
+ * Returns the path to the notebook file and the index of the cell with the spicific id
+ * If no such notebook in the current directory or no notebook with a cell with such an id is found return empty strings
+ */
+export async function getNotebookInfo(cellId: string): Promise<{
+  path: string;
```
DeltaMichael, line 123: It might be a good idea to define this as a type somewhere.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
DeltaMichael: @duyguHsnHsn You can shorten this by using ternary if you're in the mood for code golf. Not sure if it will be more readable, though. Use your best judgement.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
```diff
@@ -98,17 +112,23 @@ export const trackVdkTags = (
...
+      }
       if (
+        vdkCellIndices.length > 0 &&
         notebookTracker.activeCell &&
         notebookTracker.activeCell.parent &&
         notebookTracker.activeCell.parent.node.children
       ) {
         addVdkCellDesign(
-          notebookTracker.activeCell.node,
           Array.from(notebookTracker.activeCell.parent.node.children!),
```
DeltaMichael, line 128: You can simplify this with the elvis operator, then you won't need any checks. Something like

```
Array.from(notebookTracker?.activeCell?.parent?.node?.children)
```

Not sure about warnings, though, you might have to experiment a bit.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
ivakoleva: I see, the notebook is a composite of cells. However, having a cell_id str, why do we still need to loop over aall the notebooks and cells to find where it belongs to? 
What if the cell_id contains the information about the notebook it belongs to, so we don't have to parse alll the notebooks' cells?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
ivakoleva: Is there a link in the log, where I can click and navigate to the failint cell in the notebook accordingly?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -156,3 +157,57 @@ def create_deployment(
...
+        notebook_files = [
+            x
+            for x in pathlib.Path(pr_path).iterdir()
+            if (x.name.lower().endswith(".ipynb"))
+        ]
+        for notebook_file in notebook_files:
+            with open(notebook_file) as f:
+                notebook_json = json.load(f)
+                cell_index = 0
+                for jupyter_cell in notebook_json["cells"]:
```
DeltaMichael, line 183: Nit: Consider using `enumerate` here and getting rid of the cell_index variable.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -156,3 +157,57 @@ def create_deployment(
...
+            if not os.path.exists(job_notebook):
+                return []
+        with open(notebook_path) as f:
+            notebook_json = json.load(f)
+            vdk_cells = []
+            cell_index = 0
+            for jupyter_cell in notebook_json["cells"]:
+                if "vdk" in jupyter_cell["metadata"].get("tags", {}):
+                    vdk_cells.append(cell_index)
+                cell_index += 1
```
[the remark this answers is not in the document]
DeltaMichael:
```
for index, jupyter_cell in enumerate(notebook_json["cells"]):
    # do index stuff
    # do cell stuff
```

This might be even better.

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
duyguHsnHsn: @DeltaMichael I get codify warnings bc of them, I used to write them like that and stopped bc of it 
Should I go back to them?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
duyguHsnHsn: @DeltaMichael the problem is that in the backend we use the notebook json to get the failingCellIndex while in the front end we use the DOM to get the cells there.
Since the loading of the DOM elements in the front end is not really synchronised this might lead to problems 
For example, cells.lenght can be 0 (tho it is really unlikely it is still possible) just because the rendering takes more time for one or another reason even though there are cells in the json schema 
I think I have fixed it with  await notebook.revealed but still want to have it in any case

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: no, the log file is just an txt, consisting of the standard logs we get on the CLI nothing additional is added there

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
DeltaMichael: Can you check the behaviour if you use elvis and something is missing? Like if activeCell.parent is undefined, for some reason. Sorry, I'm not entirely sure what happens in this case either when you pass it to `Array.from()`. 😕

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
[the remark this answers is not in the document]
duyguHsnHsn: done

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: I do not think we can get all that information from the front-end, since Jupyter uses another path system (it does not show the full path but only part of it) it becomes pretty complex and I think this way is a lot simple

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
duyguHsnHsn: changed it with ternary

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: I extracted it into a different function

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/RunJob.tsx
duyguHsnHsn: refactored

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/vdkTags.ts
duyguHsnHsn: @DeltaMichael in this case nothing happens, the cells remain in their default Jupyter design but I still get warnings, I will keep it like that for now

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: ready

PR #2070, opened 2023-05-16
vdk-notebook: make it main distribution of all jupyter plugins

tozka:
Currently in order to get the full notebook integration one need to install 3 plugins - vdk-notebook, vdk-ipython and vdk-jypyter. This kind of separation is very good both for development purposes and to allow users to choose what they need.

For first time users though we'd like them to have access to everything and by default to be able to use all functionality. In that case it's better to have a single distribution that install all.

I played around with introducing aggregation type of plugin (like quickstart-vdk) but ultimately seems like too many plugins and we can re-use one of the existing ones. I choose vdk-notebook because hte name is the most generic one and can be easily described to include jupyter and ipython integraiton.

The change improves the README accordingly.


projects/vdk-plugins/vdk-notebook/README.md
```diff
@@ -1,19 +1,38 @@
...
+
+After installation:
+- users will have access to the job_input interface, which can be used to execute templates, ingest data, and perform other VDK operations.
+- users will be able to run vdk run on data job with notebook files along side python and SQL files as well.
+- To separate your notebook into VDK and non-VDK cells, use the cell metadata feature in Jupyter. You can set the "vdk" key to true for cells that are part of the data job, and false for cells that are not.
+
 
 ### Example
+
+![Example](./images/example-notebook.jpg)
```
duyguHsnHsn, line 34: I cannot find the image? Could you send a link

projects/vdk-plugins/vdk-notebook/README.md
```diff
@@ -1,19 +1,38 @@
-# vdk-notebook
+# Versatile Data Kit Notebook
```
duyguHsnHsn, line 1: I like the change and think it is a good idea
I do not think that this README changes are enough and think we should include more information both here and in the VEP but this can be done in a separate effort.

projects/vdk-plugins/vdk-notebook/README.md
tozka: I forgot to push it. It should be available now.

duyguHsnHsn (approved):

ivakoleva (approved):

tozka:
I end up not needing this. And also I am not sure if it's a good idea to aggregate them here.   plugin vdk-notebook has specific purpose. It's better to have separate distribution which aggregates them together (like quickstart-vdk does)

PR #2073, opened 2023-05-16
quickstart-vdk: ignore explore page and widgets in frontend

DeltaMichael:
## Why?

The explore page and widgets are not relevant to users outside of VMWare This also fixes the bug where the frontend does not load on running quickstart-vdk due to missing configuration

## What?

Ignore the explore page and widgets in the control service helm chart

## How has this been tested?

Configuration tested locally
Helm chart tested in CI

Bug fix (non-breaking change which fixes an issue) or a cosmetic change/minor improvement

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -43,7 +43,7 @@ operationsUi:
     requests:
       memory: 250Mi
       cpu: 25m
-  config: '{"auth": {"skipAuth": true}}'
+  config: '{"auth": {"skipAuth": true}, "ignoreComponents": ["explorePage", "widgetsComponent"], "ignoreRoutes": ["explore/data-jobs", "explore/data-jobs/:team/:job"]}'
 
 
 # The image configuration for builder jobs used during deployment of data jobs.


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -43,7 +43,7 @@ operationsUi:
     requests:
       memory: 250Mi
       cpu: 25m
-  config: '{"auth": {"skipAuth": true}}'
```
ivakoleva, line 46: So having an empty config is also a valid use case, correct? What I'm saying is, changing the empty config to a non-empty one -is not the bug fix.

DeltaMichael: It's more like an immediate mitigation for the frontend not loading correctly issue. The actual problem is we don't get a default value if the `ignoreRoutes` config is missing altogether. I'll add that info to the bug ticket we have.

murphp15 (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
DeltaMichael: Added info to https://jira.eng.vmware.com/browse/TAUR-6040

PR #2089, opened 2023-05-18
frontend: introduce default values for appConfig

DeltaMichael:
## Why?

Omitting keys in the appConfig.json file causes the application to crash

## What?

Hardcode default values for the appConfig in app-config.model.ts Values are overridden by the values in appConfig.json

## How has this been tested?

Locally
Playground: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLYE8CS5rycgbwChlTkZgoBnMALmRqlAHMAaEsqiBAexABN6jFkQC+RIrxA1k-CPACuAGzABhPhWb0M2SLEQoAvIQ6kK1OsgBEcxSoBilGlfZkG3PoOu24ysAGUPASsxCSkZHgA3aCY5dRBNZGNiN3MaeisomOA5RwsXUMk+GQwAeQAjACtuSx0cfXxk02QAOjafP3jNVzI2lqyoWIgu4GZC8J4lCBalHmYACjKqmoBKIA

Bug fix (non-breaking change which fixes an issue) or a cosmetic change/minor improvement


projects/frontend/data-pipelines/gui/projects/ui/src/assets/data/appConfig.json
```diff
@@ -1,42 +1 @@
...
-    },
-    "refreshTokenConfig": {
-      "start": 500,
-      "remainingTime": 360,
-      "checkInterval": 60
-    }
-  },
-  "ignoreComponents": [],
-  "ignoreRoutes": []
-}
```
ivakoleva, line 42: I don't like this approach, because it was easier to fetch the default appConfig.json file and modify it (before the change), and now there is no sampled JSON a user can use OOTB -since hardcoded.

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.model.ts
```diff
@@ -30,3 +30,42 @@ export interface RefreshTokenConfig {
...
+            allowedUrls: ['https://console-stg.cloud.vmware.com/', 'https://gaz-preview.csp-vidm-prod.com/', '/data-jobs'],
+            sendAccessToken: true
+        },
+        refreshTokenConfig: {
+            start: 500,
+            remainingTime: 360,
+            checkInterval: 60
+        }
+    },
+    ignoreComponents: [],
```
ivakoleva, line 69: What's the actual problem? 
```
ignoreRoutes?: string[];
```
The variables are defined to be optional. The code, depending on AppConfig should be aware of that, and handle the use-case when those optional variables are not configured.

projects/frontend/data-pipelines/gui/projects/ui/src/assets/data/appConfig.json
DeltaMichael: Is appConfig supposed to be available to the end user?

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.model.ts
ivakoleva: > For every new optional variable you introduce, you would have to edit the app-config.service.ts file and add a check for it.

Nope, that's not quite true. When adding new configurations, no explicit declarations are needed in app-config.service.ts. All newly added properties, become available via the root object:
```
getConfig(): AppConfig {
        return this.appConfig;
    }
```

DeltaMichael: @ivakoleva Yes, they're available through the object, that's not the point. You have to add an undefined check every time you want to use them.

```
if (appConfig.myNewOption) {
    if (appConfig.myNewOption == "myDesiredValue") {

    }
}
```

And if you forget the undefined check, the app is likely to break, because there is no default value.

ivakoleva: Again, what is the actual problem?

> What's the actual problem?
> 
> ignoreRoutes?: string[];
> The variables are defined to be optional. The code, depending on AppConfig should be aware of that, and handle the use-case when those optional variables are not configured.

To summarise, no, you don't need to augment app-config.service.ts each time you add a config. 
In your case, both the newly added configs are optional. You need to consider they are optional, when using them.

DeltaMichael: The actual problem is that if `ignoreRoutes` is missing from appConfig.json, the code on this line fails and takes the whole app with it.

https://github.com/vmware/versatile-data-kit/blob/main/projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.service.ts#L29

Correct me if I'm wrong, but what you're suggesting is that we add a check for includeRoutes.

```
    loadAppConfig(): Promise<void> {
        return firstValueFrom(this.httpClient.get<AppConfig>('/assets/data/appConfig.json')).then((data) => {
            this.appConfig = data;
            try {
                const localRoutes = routes.filter((route: { path: string }) => {
                    if (data.ignoreRoutes) {
                        return !data.ignoreRoutes.includes(route.path);
                    } else {
                        return false;
                    }
                });
                this.router.resetConfig(localRoutes);
            } catch (e) {
                console.error('Failed to reset Router config');
                throw e;
            }
        });
    }
```

Ok, fine. But let's say you add another optional config. Will you remember to add a check for it as well? Let's say you're super diligent and you always add checks for the optional configs wherever you use them. Is everyone who will ever touch the codebase as diligent?

The current change avoids this problem altogether by adding defaults for every config.

ivakoleva: I don't certainly agree with the statement

> The actual problem is that if ignoreRoutes is missing from appConfig.json

The `ignoreRoutes` is an optional config, therefore it may be or may be not missing from the config. In case missing, the default value should be considered - instead of breaking the application workflow.

murphp15 (approved):

projects/frontend/data-pipelines/gui/projects/ui/src/app/app-config.model.ts
DeltaMichael: > The ignoreRoutes is an optional config

`ignoreRules` is an **optional field** in the AppConfig interface (field marked with `?`). This means it may or may not be present when you instantiate the object. We haven't really discussed if configs can be optional or not, if they have defaults, etc.

DeltaMichael: @tozka Please let us know what you think.

gorankokin: I can say that both approaches are correct and both have pros and cons.

1. **Default values** -  responsibility shift for every property to the service that parse the raw configuration response into something that application is using.
   - if approach of choice I'll suggest to have two different interfaces or interface and type, one for config.json response in raw format with optional fields and the other one could be either interface or type that make all fields required.
   - then there could be one class of type `ConfigImpl` that will implement the interface or type with required fields and one `ConfigBuilder` or some utility parser that will extract values from raw config and then propagate to the `ConfigImpl` instance (if not found from raw, set default values)
   - with such approach of interfaces and classes raw and parsed config could be identical by paths by will be different because the implementation everything will be required and the invokers shouldn't take care if value is there or not. Instead configuration layer will provide everything that is needed.
 e.g.
```typescript
// utils could go to projects/frontend/shared-components/gui/projects/shared/src/lib/utils/model/shorthand-types.ts
type RecursiveRequired<T> = {
    [P in keyof T]-?: T[P] extends object ? RecursiveRequired<T[P]> : T[P];
};

// interface used for config.json response
interface ConfigRaw {
    name?: string;
    someObject?: {
        someNestedObject?: {
            paths?: string[]
        }
    }
}

// type used to make all fields required or it could be interface with suppressed error because empty interfaces are not allowed
type Config = RecursiveRequired<ConfigRaw>;
// eslint-disable-next-line @typescript-eslint/no-empty-interface
interface Config extends RecursiveRequired<ConfigRaw> {};

// actual implementation where all fields required and should be declared, otherwise will throw exception
class ConfigImpl implements Config {
    name: Config['name'];
    someObject: Config['someObject'];

    constructor(obj: Config) {
         Object.assign(this, obj);
    }
}

// some config builder that will extract values from raw config and assign defaults if missing or some utility parser
class ConfigBuilder {
    private _name: Config['name'];
    private _someObject: Config['someObject'];

    setName(name: Config['name']): this {
        this._name = name;

        return this;
    }
    
    setSomeObject(someObject: Config['someObject']): this {
        this._someObject = someObject;

        return this;
    }

    build(): ConfigImpl {
        // create temp config object and assign default values if missing
        const tempConfig: Config = {};

        return new ConfigImpl(tempConfig);
    }
}

// utility config parser
export class ConfigParser {
    static parse(configRaw: ConfigRaw): Config {
        const configTemp: Config = {
            name: configRaw?.name ?? 'defaultName',
            someObject: {
                someNestedObject: {
                    paths: configRaw?.someObject?.someNestedObject?.paths ?? []
                }
            }
        };

        return new ConfigImpl(configTemp);
    }
}

2. **Directly use raw config** - responsibility of every Class/Function (Component/Service/Directive/Pipe etc...) to check optional fields if they are nullish (Nil) or defined, because field is marked as optional and no one will guarantee that it will be there, check before every access should be made.
```

gorankokin: I can say that both approaches are correct and both have pros and cons.

1. **Default values** -  responsibility shift to the service that parse the raw configuration response into something that application is using.
   - if approach of choice I'll suggest to have two different interfaces or interface and type, one for config.json response in raw format with optional fields and the other one could be either interface or type that make all fields required.
   - then there could be one class of type `ConfigImpl` that will implement the interface or type with required fields and one `ConfigBuilder` or some utility parser that will extract values from raw config and then propagate to the `ConfigImpl` instance (if not found from raw, set default values)
   - with such approach of interfaces and classes raw and parsed config could be identical by paths by will be different because the implementation everything will be required and the invokers shouldn't take care if value is there or not. Instead configuration layer will provide everything that is needed.
 e.g.
```typescript
// utils could go to projects/frontend/shared-components/gui/projects/shared/src/lib/utils/model/shorthand-types.ts
type RecursiveRequired<T> = {
    [P in keyof T]-?: T[P] extends object ? RecursiveRequired<T[P]> : T[P];
};

// interface used for config.json response
interface ConfigRaw {
    name?: string;
    someObject?: {
        someNestedObject?: {
            paths?: string[]
        }
    }
}

// type used to make all fields required or it could be interface with suppressed error because empty interfaces are not allowed
type Config = RecursiveRequired<ConfigRaw>;
// eslint-disable-next-line @typescript-eslint/no-empty-interface
interface Config extends RecursiveRequired<ConfigRaw> {};

// actual implementation where all fields required and should be declared, otherwise will throw exception
class ConfigImpl implements Config {
    name: Config['name'];
    someObject: Config['someObject'];

    constructor(obj: Config) {
         Object.assign(this, obj);
    }
}

// some config builder that will extract values from raw config and assign defaults if missing or some utility parser
class ConfigBuilder {
    private _name: Config['name'];
    private _someObject: Config['someObject'];

    setName(name: Config['name']): this {
        this._name = name;

        return this;
    }
    
    setSomeObject(someObject: Config['someObject']): this {
        this._someObject = someObject;

        return this;
    }

    build(): ConfigImpl {
        // create temp config object and assign default values if missing
        const tempConfig: Config = {};

        return new ConfigImpl(tempConfig);
    }
}

// utility config parser
export class ConfigParser {
    static parse(configRaw: ConfigRaw): Config {
        const configTemp: Config = {
            name: configRaw?.name ?? 'defaultName',
            someObject: {
                someNestedObject: {
                    paths: configRaw?.someObject?.someNestedObject?.paths ?? []
                }
            }
        };

        return new ConfigImpl(configTemp);
    }
}

2. **Directly use raw config** - responsibility of every Class/Function (Component/Service/Directive/Pipe etc...) to check optional fields if they are nullish (Nil) or defined, because field is marked as optional and no one will guarantee that it will be there, check before every access should be made.
```

projects/frontend/data-pipelines/gui/projects/ui/src/assets/data/appConfig.json
```diff
@@ -1,42 +1 @@
-{
```
ivakoleva, line 1: I find useful during troubleshooting, to log onto the container, and cat the appConfig.json. Actually used it yesterday.

ivakoleva (changes requested):

DeltaMichael:
Closing in favour of https://github.com/vmware/versatile-data-kit/pull/2193

PR #2098, opened 2023-05-21
control-service: if a test is in a bad state it fails straight away

murphp15:
# Why
If a test goes into a final state that doesn't match the state we are looking for. we can spend a couple of minutes waiting and checking does it go to the state we want. 

examples of this. 
We want a job to end with platform error. 
The job reports user error. 
we will spend 5 minutes checking is it user error. 

Total waste of time. 

Also if the test fails we will give way more information back in the stacktrace because we are using a callable and an assertion. 

so the stacktrace will include the result of the callable. which is very useful. 

# What
I have added in a fail fast clause that says if the job is not in a running state and not in the final state we want then fail fast. 

# How was this tested
Locally


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/JobExecutionUtil.java
```diff
@@ -161,43 +162,46 @@ public static void testDataJobExecutionRead(
...
+        await()
+            .atMost(11, TimeUnit.MINUTES)
+            .with()
+            .pollInterval(15, TimeUnit.SECONDS)
+            .failFast(
+                () -> {
+                  com.vmware.taurus.controlplane.model.data.DataJobExecution status =
+                      dataJobExecutionCallable.call();
+                  return status != null
+                      && !status
```
mivanov1988, line 193: In my opinion, we should exclude `SUBMITTED` as well.

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/JobExecutionUtil.java
murphp15: oh nice spot.

DeltaMichael (approved):

PR #2113, opened 2023-05-23
control-service: print response body on error

murphp15:
# Why
at the moment when the test fails we only know the status code and we have no idea of the details. 
# What 
Now we include the response body in the error message if there is an exception
# How was this tested
locally

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/DataJobDeploymentExtension.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/DataJobDeploymentExtension.java
@@ -135,20 +135,21 @@ public class DataJobDeploymentExtension
               getClass().getClassLoader().getResourceAsStream(JOB_SOURCE_PATH + jobSource));
 
       // Upload the data job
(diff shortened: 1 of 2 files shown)


duyguHsnHsn (approved):

yonitoo (approved):

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/DataJobDeploymentExtension.java
```diff
@@ -142,8 +142,16 @@ public void beforeEach(ExtensionContext context) throws Exception {
                       .with(user(USER_NAME))
                       .content(jobZipBinary)
                       .contentType(MediaType.APPLICATION_OCTET_STREAM))
-              .andExpect(status().isOk())
               .andReturn();
+
+      if (uploadResult.getResponse().getStatus() != 200) {
```
DeltaMichael, line 147: Is it possible to get a 201 here, for example?

murphp15: Nope only 200 is possible.

PR #2117, opened 2023-05-23
control-service: code expected to run in transaction now runs in transaction

murphp15:
# Why

Often the test DataJobTerminationStatusIT fails with the following stacktrace.
```
org.opentest4j.AssertionFailedError: The value of the taurus_datajob_termination_status metrics does not match. It was actually taurus_datajob_termination_status{data_job="it-datajobterminationstatusit-b1725dc6",execution_id="it-datajobterminationstatusit-b1725dc6-1684778794",service="data-jobs",} 5.0 ==> expected: <true> but was: <false>
	at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at app//org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:211)
	at app//com.vmware.taurus.datajobs.it.DataJobTerminationStatusIT.testDataJobTerminationStatus(DataJobTerminationStatusIT.java:131)
	at 
```
I was looking into it and I realised the database operations were not actually taking place in a transaction despite the method being marked as transactional.
It is because calling a method from another method in a the same class stops spring from being able to intercept the function call and start a transaction. 

Please see here for more details: https://stackoverflow.com/questions/3423972/spring-transaction-method-call-by-the-method-within-the-same-class-does-not-wo 

# What
In this PR I create a second class to call the transactional method so it can run in a transaction. 

# How was this tested
Locally, I was not able to reproduce the randomly failing test. But that is not to say I have 100% fixed it. 
But this change is important because the code now does what the original author expected and will avoid cryptic bugs in the future.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/JobExecutionUtil.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/common/JobExecutionUtil.java
@@ -26,7 +26,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.MediaType;
-import org.springframework.mock.web.MockHttpServletResponse;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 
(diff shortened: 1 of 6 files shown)


mivanov1988 (approved):

murphp15:
@tozka  
I did this (could be by looking at logs by setting logging.level.org.springframework.transaction=debug; or debugger or best with unit tests).

PR #2121, opened 2023-05-23
add properties service unit tests

dakodakov:
vdk-control-service: properties service unit tests

add unit tests for the properties service.

Signed-off-by: Dako Dakov <ddakov@vmware.com>

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/properties/PropertiesServiceTest.java
+++ b/projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/properties/PropertiesServiceTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2021-2023 VMware, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.vmware.taurus.properties;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.vmware.taurus.properties.service.JobProperties;
+import com.vmware.taurus.properties.service.PropertiesRepository;
+import com.vmware.taurus.properties.service.PropertiesService;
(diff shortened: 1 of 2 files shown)


projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/properties/PropertiesServiceTest.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+class PropertiesServiceTest {
+
+  @Mock private PropertiesRepository propertiesRepository;
```
murphp15, line 31: but you are never checking that the repository works? 
Why mock it? if you don't mock it you will need to write less code and also test more.

tozka: I tend to agree. 

I prefer (including in unit tests) to use test doubles/test substitute (fully functioning versions of the dependency) when possible and feasible than mock version. They make the test slightly more black box (hence less fragile, less false positives), make them easier to write (since you need not write a ton of logic to mock methods), and are better for complex interactions (that would be hard to mock).

dakodakov: Hello,
I added these tests in the context of vep-1493-vault-integration - as a next step I intend to do some refactoring and extract a common interface for the Properties and the (upcoming) Secrets Service. 

I believe the test you are asking for is already present - DataJobPropertiesIT - however it's an integration test which runs slower and requires more infra setup.

Do you agree?

mivanov1988 (approved):

murphp15 (approved):

PR #2127, opened 2023-05-25
control-service: a clear error message on how to handle the failed pipeline

murphp15:
# Why
Often we get in a situation where the helm chart can't be deployed because the previous deployment is in a pending state. 
There is not a clear way of how to handle this. However we need to tell ppl how to investigate it.  
for the last week all deployments failed because there was a deployment in pending state for 1 week. 

# What 
exit the deployment early if there is a deployment in pending-state and include in stdout help on how to debug it. 

# How was this tested
I pushed the system into an unstable state and the error message printed out:https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/4348854403 
I also tested when the system is in a stable state and it all looks good


duyguHsnHsn (approved):

mivanov1988:
Good improvement!

mivanov1988 (approved):

tozka (approved):
LGTM. 

I am more interested in why is it happening that often?

PR #2137, opened 2023-05-26
control-service: fail tests fast

murphp15:
# Why
we want faster feedback when a test fails and we want to use less resources in the cicd cluster
# What
integration tests kill themselves if they have a single failure
# How was this tested
locally


mivanov1988 (approved):

mivanov1988:
Really good improvement.

DeltaMichael (approved):

PR #2142, opened 2023-05-29
vdk-jupyter: Pick up REST API URL from the env

gageorgiev:
Currently, users are expected to configure the REST API URL manually, which is a bad practice. After this change, the REST API URL will be picked up from the environment, which means it will be configured when deploying the docker image.

Testing done: pipelines

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
@@ -1,6 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import json
+import os
 
 import tornado
 from jupyter_server.base.handlers import APIHandler
@@ -66,7 +67,7 @@ class DeleteJobHandler(APIHandler):
             status = VdkUI.delete_job(
                 input_data[VdkOption.NAME.value],
                 input_data[VdkOption.TEAM.value],
-                input_data[VdkOption.REST_API_URL.value],
+                os.environ["REST_API_URL"],
             )
             self.finish(json.dumps({"message": f"{status}", "error": ""}))
         except Exception as e:
@@ -89,7 +90,7 @@ class DownloadJobHandler(APIHandler):
             status = VdkUI.download_job(
                 input_data[VdkOption.NAME.value],
                 input_data[VdkOption.TEAM.value],
-                input_data[VdkOption.REST_API_URL.value],
+                os.environ["REST_API_URL"],
                 input_data[VdkOption.PATH.value],
             )
             self.finish(json.dumps({"message": f"{status}", "error": ""}))
@@ -114,7 +115,7 @@ class CreateJobHandler(APIHandler):
             status = VdkUI.create_job(
                 input_data[VdkOption.NAME.value],
                 input_data[VdkOption.TEAM.value],
-                input_data[VdkOption.REST_API_URL.value],
+                os.environ["REST_API_URL"],
                 input_data[VdkOption.PATH.value],
                 bool(input_data[VdkOption.LOCAL.value]),
                 bool(input_data[VdkOption.CLOUD.value]),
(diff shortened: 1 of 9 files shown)


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -66,7 +67,7 @@ def post(self):
             status = VdkUI.delete_job(
                 input_data[VdkOption.NAME.value],
                 input_data[VdkOption.TEAM.value],
-                input_data[VdkOption.REST_API_URL.value],
+                os.environ["REST_API_URL"],
```
tozka, line 70: It's not necessary to fetch the environment variable each time a method is called. It would be more efficient and maintainable to fetch the REST_API_URL once and store it as a class variable, or somewhere else where it can be accessed. Then, you can use that stored value in the methods.

Also there seems to be no error handling mechanism in case the REST_API_URL environment variable is not set. In the current code, if os.environ["REST_API_URL"] doesn't exist, it will raise a KeyError. 


Overall my recommendation is not to use directly os.environ. Instead, I recommend using a configuration provider type of class. This class would handle fetching and validating environment variables, enabling the application to separate concerns and focus more on business logic

tozka (changes requested):
Thanks for doing this change. 

This would improve security and maintainability vdk-jypyter. Using environment variables to configure deployment-specific details like the REST API URL is a common practice and allows for more flexibility and easier management.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
tozka: Do consider adding tests that cover what happens if REST API URL is  misconfigured or missing.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -89,7 +90,7 @@ def post(self):
             status = VdkUI.download_job(
                 input_data[VdkOption.NAME.value],
                 input_data[VdkOption.TEAM.value],
-                input_data[VdkOption.REST_API_URL.value],
+                os.environ["REST_API_URL"],
```
murphp15, line 93: Can you not make this a parameter of the methods and instead only load it where it is needed? 
It will make it clearer that it is a global value.

duyguHsnHsn:
Also see jobData.ts, vdkOptions.py, and vdkOptions.ts. Rest api url options should be removed from there,  as well.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -11,6 +12,22 @@
 from .vdk_ui import VdkUI
 
 
+class HandlerConfiguration:
+    def __init__(self):
+        self._rest_api_url = os.environ["REST_API_URL"]
+        if not self._rest_api_url:
+            raise Exception(
+                "Missing environment variable REST_API_URL. "
+                "This is probably caused by a corrupt environment."
```
duyguHsnHsn, line 21: Here you can add more information on what the rest api url  is for (or where it is used) and consequences of the missing variable, to make it more user accessible.

gageorgiev: maybe I'm wrong about this, but I don't think a user should ever be thinking about rest api URLs, it should be seamless; this error isn't something we expect users to be hitting ever

tozka: I think you are absolutely right, that in best case, users should not need to think about rest API URLs and should be seamless. 

But if they hit the error then what ?

In this case we should use clear, concise, and informative error messages to help users understand what went wrong and how to resolve it. (As outlined in https://github.com/vmware/versatile-data-kit/wiki/Design-Principles) 

For that reason our error handling format recommends that a good error message should answer what happened, why it happened , what are the consequences for the user and what are the countermeasures (how the user can resolve it). 
([[link](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/CONTRIBUTING.md#error-handling)]) 
Let's try each error message and exception message to answer those questions.

duyguHsnHsn (approved):
Looks good!

yonitoo (approved):
LGTM

tozka (commented):
One more thing. Almost never use raise Exception. If you see it in the code, we should fix it.

tozka (approved):

PR #2143, opened 2023-05-29
control-service: allow more time to reach a complete state

murphp15:
# Why
when the system is under load jobs often take a little over 5 minutes. 
Because we fail fast we won't waste needless time if the job goes into a failed state. 

# What
Wait longer to get into a successful state 

# How was this tested
CICD pipeline


DeltaMichael (approved):

doks5 (approved):

tozka:
I feel like we shoud investigate why it needs so much time , How much of a load we are creating just with tests? It's not like real production system with 1000 of jobs

PR #2151, opened 2023-05-29
control-service: handle null started by value

murphp15:
# Why
We are seeing errors when the UI when using the latest version of control service
# What
Write a test to check null values for startedby are supported. Also support null startedBy
# How was this tested
Locally

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
@@ -109,8 +109,8 @@ public class JobExecutionService {
       // We default to 'manual'; this also served for the purposes of backwards compatibility
       annotations.put(
           JobAnnotation.EXECUTION_TYPE.getValue(),
-              jobExecutionRequest.getStartedBy() != null &&
-          jobExecutionRequest.getStartedBy().contains("scheduled")
+          jobExecutionRequest.getStartedBy() != null
+                  && jobExecutionRequest.getStartedBy().contains("scheduled")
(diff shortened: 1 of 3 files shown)


tozka (approved):

DeltaMichael (approved):

gorankokin (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/execution/JobExecutionService.java
```diff
@@ -109,7 +109,8 @@ public String startDataJobExecution(
       // We default to 'manual'; this also served for the purposes of backwards compatibility
       annotations.put(
           JobAnnotation.EXECUTION_TYPE.getValue(),
-          jobExecutionRequest.getStartedBy().contains("scheduled")
```
doks5, line 112: Do we have a clue why the value is null, and not a username or some identifier? We hit this in an internal deployment of the Control Service, and it was unexpected to say the least.

murphp15: The frontend doesn't send a startedBy. 
Is this not expected? If it is not expected we need to change the open api definition.

murphp15: This code was changed recently. 
This caused the regression.

doks5: Oopsie, yup, I saw the change. Should've `git blame`'ed the file first. Sorry

PR #2156, opened 2023-05-30
control-service: better error logging and pull private image in private test

murphp15:
# Why
1. When the job fails we want to know the reason and not just the status
2. We are now using github to pull public images from to avoid docker hub throttling. In doing this the private image test was no longer using a private image. Now it is using a private image again. 

# What
1. If the status code is not 200 log the response body
2. pull a private image


gageorgiev (approved):

ivakoleva:
I don't understand what a "private image" is, is there any documented work on the subject so I can quickly check it out?
Where I anticipated to find such info:
* https://github.com/vmware/versatile-data-kit/blob/main/CONTRIBUTING.md
* https://github.com/vmware/versatile-data-kit/tree/main/cicd
* https://github.com/vmware/versatile-data-kit/wiki/Gitlab-CICD

PR #2172, opened 2023-05-31
vdk-control-cli, vdk-jupyter: Allow for creating a job with a notebook step

gageorgiev:
Currently, when creating a job through the jupyter UI, the created job
has a sql step and a python step. We'd like a job created that way to
have only a notebook inside, as steps will be contained within that notebook.

This required changes to both the control cli module, as well as the jupyter
python module. An alternative way was possible which left the control cli
untounched, however that path was not chosen since it would require patching
module namespaces during runtime, which is prone to bugs and is to be avoided
if possible.

This is currently a draft PR, next required work is documenting this in the VEP

Testing done: tested locally by creating a new job through jupyter UI and observing
the notebook step inside

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
@@ -30,7 +30,13 @@ class JobCreate:
 
     @ApiClientErrorDecorator()
     def create_job(
-        self, name: str, team: str, path: str, cloud: bool, local: bool
+        self,
+        name: str,
+        team: str,
+        path: str,
+        cloud: bool,
+        local: bool,
+        is_jupyter: bool = False,
     ) -> None:
         self.__validate_job_name(name)
         if local:
@@ -40,7 +46,7 @@ class JobCreate:
             self.__create_cloud_job(team, name)
         if local:
             job_path = self.__get_job_path(path, name)
-            self.__create_local_job(team, name, job_path)
+            self.__create_local_job(team, name, job_path, is_jupyter)
             if cloud:
                 self.__download_key(team, name, path)
 
@@ -63,8 +69,13 @@ class JobCreate:
             f"Data Job with name {name} created and registered in cloud runtime by Control Service."
         )
 
-    def __create_local_job(self, team: str, name: str, job_path: str) -> None:
-        sample_job = self.__vdk_config.sample_job_directory
+    def __create_local_job(
+        self, team: str, name: str, job_path: str, is_jupyter: bool
+    ) -> None:
+        if is_jupyter:
+            sample_job = self.__vdk_config.jupyter_sample_job_directory
+        else:
+            sample_job = self.__vdk_config.sample_job_directory
         log.debug(f"Create sample job from directory: {sample_job} into {job_path}")
         cli_utils.copy_directory(sample_job, job_path)
         local_config = JobConfig(job_path)
--- a/projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/configuration/vdk_config.py
@@ -137,6 +137,18 @@ class VDKConfig:
             sample_job_dir = os.path.abspath(template_module_path)
         return sample_job_dir
 
+    @property
+    def jupyter_sample_job_directory(self) -> str:
+        jupyter_sample_job_dir = os.getenv("VDK_CONTROL_SAMPLE_JOB_DIRECTORY", None)
+        if not jupyter_sample_job_dir:
+            import vdk.internal.control.job.jupyter_sample_job
+
+            template_module_path = (
+                vdk.internal.control.job.jupyter_sample_job.__path__._path[0]
+            )
+            jupyter_sample_job_dir = os.path.abspath(template_module_path)
+        return jupyter_sample_job_dir
+
 
 class VDKConfigFolder:
     """
--- a/projects/vdk-control-cli/src/vdk/internal/control/job/jupyter_sample_job/10_notebook.ipynb
+++ b/projects/vdk-control-cli/src/vdk/internal/control/job/jupyter_sample_job/10_notebook.ipynb
@@ -1,0 +1,36 @@
+{
+ "cells": [],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 2
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython2",
+   "version": "2.7.6"
+  },
+  "pycharm": {
+   "stem_cell": {
+    "cell_type": "raw",
+    "source": [
+     "\n",
+     "\n"
+    ],
+    "metadata": {
+     "collapsed": false
+    }
+   }
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
--- a/projects/vdk-control-cli/src/vdk/internal/control/job/jupyter_sample_job/README.md
+++ b/projects/vdk-control-cli/src/vdk/internal/control/job/jupyter_sample_job/README.md
@@ -1,0 +1,33 @@
+# My shiny new job
+
+Versatile Data Kit feature allows you to implement automated pull ingestion and batch data processing.
+
+### Create the data job Files
+
+Data Job directory can contain any files, however there are some files that are treated in a specific way:
+
+* SQL files (.sql) - called SQL steps - are directly executed as queries against your configured database;
+* Python files (.py) - called Python steps - are Python scripts that define run function that takes as argument the job_input object;
+* config.ini is needed in order to configure the Job. This is the only file required to deploy a Data Job;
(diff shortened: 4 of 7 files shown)


tozka:
The is_jupyter boolean flag, in this case, is an example of the [control flag anti-pattern](https://refactoring.guru/remove-control-flag) in software design. This approach can lead to more complex, harder-to-understand code over time . 

Also , the is_jupyter flag couples the logic for creating a job with the specific type of sample job (Jupyter). If new  types of sample obs need to be supported in the future (as we would like to). 

So if we need to create a notebook specific sample job. 
We can do that entirely in the jupyter codebase (in vdk-jupyter plugin)
There's no need to do any changes in  vdk-control-cli. 

If we want to add support in vdk-control-cli to be able to provide different types of sample jobs, that's different.
There are some approaches. For example we can extend `vdk create --sample` to have an argument . The arguments could be 
A) local directory 
B) packaged location (e.g vdk.internal.notebook.sample_job)
C) URI (git or http) 

This would allow users to create sample job from anywhere.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -30,7 +30,13 @@ def __init__(self, rest_api_url: str):
...
     @ApiClientErrorDecorator()
     def create_job(
-        self, name: str, team: str, path: str, cloud: bool, local: bool
+        self,
+        name: str,
+        team: str,
+        path: str,
+        cloud: bool,
+        local: bool,
+        sample_job=None,
```
tozka, line 39: use consistent argument names across methods if the semantics are the same. And use different if they have changed 

 . I got confused here . I thought sample_job_module was something different.

tozka, line 39: don't forget the type hints.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -63,15 +69,45 @@ def __create_cloud_job(self, team: str, name: str) -> None:
...
+        This method generates the directory for the sample job which will be created when invoking `vdk create`.
+        Its control flow is as follows:
+         - if the configuration variable VDK_CONTROL_SAMPLE_JOB_DIRECTORY is set, it takes precedent over
+         everything else
+         - next, if a sample_job_module is passed to the method by its invoking class, it takes precedent next; the
+         purpose of this is to allow plugins and extensions to set a new default sample job, but still not overwrite
+         one which was set by the user
+         - finally, the default sample job is used
+
+        :param sample_job_module: a Python module containing the files for a sample job
```
tozka, line 94: I think it would be better to be sample_job_path. 

sample_job_module means the user has to follow our package structure and basically insert their own code in our packages. This might lead to issues (more likely to cause conflicts, harder for them to do, what if we change our package structure). 

sample_job_path also facilitate future use-case -e.g if we support downloading sample job from http or git or creating sample job from another data job :)

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: I've fixed this, I noticed it after pushing

gageorgiev: not completely sure what type hint to include as we pass a module

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: noted

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
tozka: Google (or chatGPT now) has all the answers. 

For example https://stackoverflow.com/questions/53780735/what-is-the-type-hint-for-a-any-python-module :) 

Though if we use path it might be moot point.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -63,15 +69,45 @@ def __create_cloud_job(self, team: str, name: str) -> None:
...
+            return self.__vdk_config.sample_job_directory
+
+        if not sample_job_module:
+            import vdk.internal.control.job.sample_job
+
+            template_module_path = vdk.internal.control.job.sample_job.__path__._path[0]
+            sample_job_dir = os.path.abspath(template_module_path)
+        else:
+            template_module_path = sample_job_module.__path__._path[0]
+            sample_job_dir = os.path.abspath(template_module_path)
```
ivakoleva, line 108: nit: try not to duplicate code, as a general advice

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
ivakoleva: agree, path makes more sense  than file or directory(since a certain type of file), to describe location

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -40,7 +46,7 @@ def create_job(
             self.__create_cloud_job(team, name)
         if local:
             job_path = self.__get_job_path(path, name)
-            self.__create_local_job(team, name, job_path)
+            self.__create_local_job(team, name, job_path, sample_job_directory)
```
ivakoleva, line 49: May consider reusing `__get_job_path()` validation in a way, similar to L48. 
For example, `job_path` arg is already validated to be an actual path.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -63,15 +69,45 @@ def __create_cloud_job(self, team: str, name: str) -> None:
...
         log.debug(f"Create sample job from directory: {sample_job} into {job_path}")
         cli_utils.copy_directory(sample_job, job_path)
         local_config = JobConfig(job_path)
         if not local_config.set_team_if_exists(team):
             log.warning(f"Failed to write Data Job team {team} in config.ini.")
         log.info(f"Data Job with name {name} created locally in {job_path}.")
 
+    def __create_sample_job_dir(self, sample_job_module):
+        """
+        This method generates the directory for the sample job which will be created when invoking `vdk create`.
```
ivakoleva, line 85: Where does the method actually "create/generate" the dir?

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -30,7 +30,13 @@ def __init__(self, rest_api_url: str):
...
     @ApiClientErrorDecorator()
     def create_job(
-        self, name: str, team: str, path: str, cloud: bool, local: bool
+        self,
+        name: str,
+        team: str,
+        path: str,
+        cloud: bool,
+        local: bool,
+        sample_job_directory=None,
```
ivakoleva, line 39: The default value is not None, but the actual default sample_job dir in vdk-control-cli. 
The advice is to keep a clean user-facing api surface, that reflects the actual defaults.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk/internal/control/job/jupyter_sample_job/10_notebook.ipynb
```diff
@@ -0,0 +1,36 @@
...
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython2",
+   "version": "2.7.6"
+  },
+  "pycharm": {
+   "stem_cell": {
+    "cell_type": "raw",
+    "source": [
```
duyguHsnHsn, line 24: The default notebook does not contain any sample code and configurations for vdk-ipython
Will adding the sample code and configurations be a separate effort?

gageorgiev: no, it will be part of this PR

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: resolving this since it was merged in a separate PR

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: resolving this since it was merged in a separate PR

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: resolving this since it was merged in a separate PR

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: resolving this since it was merged in a separate PR

duyguHsnHsn (approved):
Seems good to me

DeltaMichael (approved):

PR #2177, opened 2023-05-31
vdk-control-cli: Allow extensions to specify a sample job

gageorgiev:
The purpose of this change is the following: some extensions to VDK might use vdk-control-cli as a dependency to create new jobs of a certain type through the JobCreate class. Currently, there is a way to specify what job is created, which is the VDK_CONTROL_SAMPLE_JOB_DIRECTORY, however this config variable is aimed at users which want to create a job template which is specific to their environment, and must be in the form of a path in their file system. An extension cannot use this to create a new sample job. This change allows for this by extending the JobCreate class' create method with a new parameter which can be passed when the method is called. This parameter must be a Python module which is to be included in the extension's source code and must contain the necessary job files which will be created when running creating a job.

Testing done: unit test

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import pathlib
+from types import ModuleType
 from typing import Optional
 from typing import Tuple
 
@@ -36,7 +37,7 @@ class JobCreate:
         path: str,
         cloud: bool,
         local: bool,
-        sample_job_directory=None,
+        sample_job_directory: ModuleType = None,
     ) -> None:
         self.__validate_job_name(name)
         if local:
@@ -70,7 +71,7 @@ class JobCreate:
         )
 
     def __create_local_job(
-        self, team: str, name: str, job_path: str, sample_job_directory
+        self, team: str, name: str, job_path: str, sample_job_directory: ModuleType
     ) -> None:
         sample_job = self.__create_sample_job_dir(sample_job_directory)
         log.debug(f"Create sample job from directory: {sample_job} into {job_path}")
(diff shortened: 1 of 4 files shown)


projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -63,15 +69,45 @@ def __create_cloud_job(self, team: str, name: str) -> None:
...
+
+        if not sample_job_directory:
+            import vdk.internal.control.job.sample_job
+
+            template_module_path = vdk.internal.control.job.sample_job.__path__._path[0]
+            sample_job_dir = os.path.abspath(template_module_path)
+        else:
+            template_module_path = sample_job_directory.__path__._path[0]
+            sample_job_dir = os.path.abspath(template_module_path)
+        return sample_job_dir
```
tozka, line 109: that doesn't seem correct?

You've removed draft but it seems still a draft . Have you forgotten to push a commit

tozka:
> Testing done: manual test

I assume you will rewrite "manual test" to "automated test in tests directory" ? 

Or describe in detail what were the manual steps taken and why it's not feasible to automate them.

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
```diff
@@ -30,7 +30,13 @@ def __init__(self, rest_api_url: str):
...
     @ApiClientErrorDecorator()
     def create_job(
-        self, name: str, team: str, path: str, cloud: bool, local: bool
+        self,
+        name: str,
+        team: str,
+        path: str,
+        cloud: bool,
+        local: bool,
+        sample_job_directory=None,
```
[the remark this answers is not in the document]
gageorgiev: it is a python module, not sure what type to specify

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/create.py
gageorgiev: which part does not seem correct?

tozka: hmmm, I thought sample_job_directory would be of type "str" or pathlib.Path ? 

That's why that seems incorrect ...

ivakoleva:
How does this PR relate to https://github.com/vmware/versatile-data-kit/pull/2172?

duyguHsnHsn (approved):
Looks good to me!

tozka:
> This parameter must be a Python module which is to be included in the extension's source code and must contain the necessary job files which will be created when running creating a job


I think the parameter should be a Path (type pathlib.Path)

You can easily get the path of a module . But in the future I want to add functionality that will allow to create sample jobs out of examples in GitHub 

On top of that a Path is generally more intuitive for developers as it simply requires a location in the file system. A Module, on the other hand, requires being properly structured, compiled, and imported, which can potentially introduce extra complexity. Debugging issues with a Module can be more complicated as it requires a deeper understanding of the Python import system

gageorgiev:
@tozka this means we cannot use an absolute path since the path will depend on the user's env. How can we ensure consistency of the relative path of the sample job?

tozka:
> @tozka this means we cannot use an absolute path since the path will depend on the user's env. How can we ensure consistency of the relative path of the sample job?
 
Maybe I am missing something. But I do not see the problem. 
 
So if we have this interface `create_job(..., sample_job_path: pathlib.Path). `
 
 In Notebook plugin we can use it like that: 
 
 ```
path_sample_job = os.path.abspath(vdk.internal.notebook.sample_job.__path__._path[0])
create(..., path_sample_job)
```

In the future if we build github support we can use it like that 
```
path_sample_job = download_from_github_and_save_to_temp("https://github.com/vmware/versatile-data-kit/tree/main/examples/ingest-from-rest-api-example/ingest-from-rest-api-example-job") 
create(..., path_sample_job)
```

gageorgiev:
gotcha, that does seem more extensible, will implement it

PR #2184, opened 2023-05-31
Control-service: better error logging allowing to understand failing test

murphp15:
# Why
The test DataJobTerminationStatusIT fails often with the assertion that even though the job is completed the metrics are saying it is not completed. 
An example of this can be seen in these logs https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/4373575597.

# What 

I have looked through the logs and I'm 99% sure that it is because there are two instances of the job launching.
Which seems like a bug in either production code or test code but I can see which. 

In this PR I increase the amount of logs so that we will be able to debug it better in the future. 
In particular I output all sql statement to the logs to help debug the issue. 

I added a new assertion to make sure only a single job has run. 
I cleaned up the existing assertion to check for the actual execution instance instead of just looking by job name. 

# How was this tested
Locally it is green. when these extra logs when the issue happens in cicd again I should be able to understand the root cause.


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobTerminationStatusIT.java
```diff
@@ -171,6 +179,20 @@ private static Optional<String> findMetricsWithLabel(
...
+  public static int findMetricCountWithLabel(
+      CharSequence input, String metrics, String label, String labelValue) {
+    Pattern pattern =
+        Pattern.compile(
+            String.format("%s\\{.*%s=\"%s\"[^\\}]*\\} (.*)", metrics, label, labelValue),
+            Pattern.CASE_INSENSITIVE);
+    Matcher matcher = pattern.matcher(input);
+    int counter = 0;
+    while (matcher.find()) {
+      counter++;
```
ivakoleva, line 191: nit: may leverage an elegant `matcher.groupCount()`

gageorgiev (approved):

tozka (approved):

PR #2198, opened 2023-06-05
vdk-impala: Introduce checks for insert template

sbuldeev:
Add functionality to the insert processing templates in order to allow quality checks to be made before the data is inserted into the target table.

Currently, the checks done on the insert template processing step are not covering if the semantics of the data is correct. Therefore, bad data could go into the target table which could be unwanted behavior.

More details explained in
https://github.com/vmware/versatile-data-kit/issues/1361

What:
-Adding functionality to handle the insert template behavior if the user provides checks

Tests: provided positive and negative regression tests

Signed-off-by: Stefan Buldeev sbuldeev@vmware.com


tozka:
The title of the PR says snapshot, not insert template :)

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/02-handle-quality-checks.py
```diff
@@ -0,0 +1,79 @@
...
+
+    check = job_arguments.get("check")
+    partition_clause = job_arguments["_vdk_template_insert_partition_clause"]
+    source_schema = job_arguments.get("source_schema")
+    source_view = job_arguments.get("source_view")
+    target_schema = job_arguments.get("target_schema")
+    target_table = job_arguments.get("target_table")
+    insert_query = get_file_content(SQL_FILES_FOLDER, "02-insert-into-target.sql")
+
+    if check:
```
tozka, line 32: You should notice al ot of similarties between this and https://github.com/vmware/versatile-data-kit/pull/2040/files 

I think it's good idea to consider a refactoring change (could be separate PR) where we remove some of the duplcaiton : 

For example we can introduce some helper class ChecksHelper where we can have method like 
```
def perform_quality_checks_and_staging(job_input: IJobInput, check, original_transform_query, [potentially other common args] ): 
``` 

where a lot of the logic about performing the check,c creating the staging tales and so on can be moved. 

What do you think? That would reduce the code in each of those classes by moving it to a common place. And make it easier to maintain and fix in the future

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/02-handle-quality-checks.py
```diff
@@ -0,0 +1,79 @@
...
+            insert_into_target = insert_query.format(
+                source_schema=staging_schema,
+                source_view=staging_table_name,
+                _vdk_template_insert_partition_clause=partition_clause,
+                target_schema=target_schema,
+                target_table=target_table,
+            )
+            job_input.execute_query(insert_into_target)
+        else:
+            raise Exception("The data is not passing the quality checks!")
```
tozka, line 76: Avoid raising base exception . This is called type of [error hiding](https://en.wikipedia.org/wiki/Error_hiding) that causes issues. 

You can either create a new exception type (I think having DataQualityException would be the best). In this exception class you can even add custom fields with more information about the data quality check (context) which might provide the user (likely you in the future) more information on what is going on.

tozka, line 76: An exception message is recommended to  clearly communicate to the user (the one expected to see the exception) what was the error , why it happens and very importantly - what are the consequences and how they can resolve it by themselves (countermeasures) . 

See also [our coding standard error handling guidances](https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#error-handling).

tozka (changes requested):
The functionality looks good. Now I am focusing on some small improvements on maintainability and ease-of-troubleshooting of the change.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/02-handle-quality-checks.py
sbuldeev: Done, added detailed exception message aligned with the VDK standard exception messages

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/02-handle-quality-checks.py
sbuldeev: I'll need some more time to come up with the best possible solution for this common functionality improvement as all of the templates have some specific implementation parts. The common between them is:
1) Staging table name is being calculated
2) Staging table is being aligned with the target one
3) All/part of the data is moved to the staging table
4) Checks are performed on the staging table/staging+target tables
5) Staging data is moved to target

tozka: You do not have to address this comment in this PR. Let's merge it , and you can address in separate PR.

tozka (approved):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/README.md
```diff
@@ -12,6 +12,8 @@ This template can be used to load raw data from Data Lake to target Table in Dat
 - target_table    - Data Warehouse table of DW type table, where target data is loaded
 - source_schema   - Data Lake schema, where source raw data is loaded from
 - source_view     - Data Lake view, where source raw data is loaded from
+- check           - (Optional) Callback function responsible for checking the quality of the data
```
murphp15, line 15: I think that this should say the callable takes in a parameter which is the table name which contains the data to validate .

murphp15 (approved):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/templates/load/fact/insert/README.md
sbuldeev: Yes, makes sense. Improved the check description in scd1, snapshot and insert templates.

PR #2207, opened 2023-06-06
Control-service: remove errors on job deployment

murphp15:
# Why
closes: https://github.com/vmware/versatile-data-kit/issues/2199
At the moment if you deploy a job successfully there are some errors in the logs.  
This is confusing for maintainers of control-service trying to debug issues. 
It is also terrible experience for users of quickstart vdk where they see errors in the logs of the pods if they are looking around. 

The error happens when we deploy the job for the first time.
We check if the cronjob exists if it doesn't we build the image. 
This is a totally standard flow of operations and so shouldn't result in errors at the warn level. 

# What 
In this PR I add different methods for ppl to call if they want to know if a cronjob exists or if they know it exists and want the details about it. 
I also clean up some existing code as it was written in a very non functional style. 

# How was this tested? 
I updated a single integration test to check that the line is not appearing in the logs. I don't want to write an integration test specifically for this. as running assertions on logs output is not somethin that is normally tested.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -452,6 +452,19 @@ public Optional<JobDeploymentStatus> readCronJob(String cronJobName) {
...
+  public boolean cronJobExists(String cronJobName) {
+    try {
+      initBatchV1beta1Api().readNamespacedCronJob(cronJobName, namespace, null);
+    } catch (ApiException e) {
+      try {
+        initBatchV1Api().readNamespacedCronJob(cronJobName, namespace, null);
+      } catch (ApiException e2) {
+        return false;
+      }
+    }
```
tozka, line 464: add logging (it may be debug) or we will wonder if there's an issue in the future . 


It's good idea when logging or re-throwing to wrap it in our KubernetesException since it prints more details (by default ApiException does not print http body where the actual reason for the error is)

tozka: Should we not return false only if the API call returns 404 Not Found. If it's 5xx error, we should re-throw. Otherwise the logic would be faulty, we will think a corn job does not exists while there is network or other infrastructure issue.

tozka (changes requested):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -45,10 +45,20 @@ public class DeploymentService {
...
 
-  public Optional<JobDeploymentStatus> readDeployment(String jobName) {
+  public Optional<JobDeploymentStatus> readOptionalDeployment(String jobName) {
     return jobImageDeployer.readScheduledJob(jobName);
   }
 
+  public JobDeploymentStatus readDeployment(String jobName) {
+    return jobImageDeployer
+        .readScheduledJob(jobName)
+        .orElseThrow(() -> new DataJobDeploymentNotFoundException(jobName));
```
ivakoleva, line 55: nit: may reuse `readDeploymentOptional`; so that in case the implementation changes (readScheduledJob), you only need to edit once method; and also prevent duplicate code

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
```diff
@@ -70,6 +70,10 @@ public Optional<JobDeploymentStatus> readScheduledJob(String dataJobName) {
     return jobDeployment;
   }
 
+  public boolean deploymentExists(String dataJobName) {
+    return dataJobsKubernetesService.cronJobExists(getCronJobName(dataJobName));
```
ivakoleva, line 74: nit: 
(1) unexpected naming; the above method `readScheduledJob` also uses the cronjob, why not scheduledJobExists to keep consistency of terminology? 

(2) why using `getCronJobName`? 
```
// Public for integration testing purposes
  public static String getCronJobName(String jobName) {
    return jobName;
  }
```
(3) Why do we need this delegator to `dataJobsKubernetesService.cronJobExists`? It simply links a feature of another service, why is that needed?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
```diff
@@ -269,8 +274,7 @@ public void deleteDeployment(String dataJobName) {
   }
 
   private boolean deploymentExistsOrInProgress(String dataJobName) {
-    return jobImageBuilder.isBuildingJobInProgress(dataJobName)
-        || readDeployment(dataJobName).isPresent();
+    return jobImageBuilder.isBuildingJobInProgress(dataJobName) || deploymentExists(dataJobName);
```
ivakoleva, line 277: nit

The prior `readDeployment(dataJobName).isPresent()`, after the changes, corresponds to `readDeploymentOptional(dataJobName).isPresent()`. 
Why swapping it with newly introduced `deploymentExists()` that duplicates this logic entirely (ideally, such a method should be reusing readDeploymentOptional.isPresent btw)?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -452,6 +452,19 @@ public Optional<JobDeploymentStatus> readCronJob(String cronJobName) {
     return jobStatus.isPresent() ? jobStatus : readV1CronJob(cronJobName);
   }
```
ivakoleva, line 454: Can you add into the PR description, what was the actual cause of that error message, and how is that problem addressed to not print it? You bring 5 main/java/ class changes with this PR, and the cause and fix remain unknown.

ivakoleva (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
murphp15: This is a really good point. This is actually a bug in the existing code and is not introduced in this PR. 
But I will address it.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentService.java
murphp15: No the deploymentExist method is lighter because it doesn't create needless objects just to check are they present.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
murphp15: updated PR desc: 

```
The error happens when we deploy the job for the first time.
We check if the cronjob exists if it doesn't we build the image. 
This is a totally standard flow of operations and so shouldn't result in errors at the warn level. 
```

tozka (approved):

DeltaMichael:
Closing due to inactivity.

PR #2240, opened 2023-06-13
frontend: disable stable tagging for ui docker images

DeltaMichael:
## Why

Building shared-components and data-pipelines simultaneously in CI causes the UI docker image to break.

More info https://jira.eng.vmware.com/browse/TAUR-6084

## What

Disable the CI job that tags UI docker images as stable. quickstart-vdk uses the stable tag. Disabling tagging until the issue is solved makes sure that quickstart-vdk will not break because of changes we introduce in the meantime

## How was this tested

n/a

Temporary bug mitigation


murphp15 (approved):

tozka (approved):
Is this still needed? I can se it's 3 days. If it was urgent we probably should have merged it by now

PR #2242, opened 2023-06-13
vdk-core: Add python version disparity warning

doks5:
If a user has set a specific python_version in their data job's config.ini file, but they develop the data job in a python environment with a different version, the execution results may be unexpected when the job is deployed.

This change adds a warning message for local data job executions, in case there is a disparity between the configured python version and the python environment that executes the job.

Testing Done: Locally tested with a data job, and verified the message appears when there is a python version disparity.

```bash
warning: 
warning:                         
warning:                   ********************************************************************************
warning:                         Python version (3.9), set in the job's config.ini file, is different
warning:                         from the python version (3.8) used to execute the data job.
warning:                         
warning:                   ********************************************************************************
warning:
```

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
@@ -5,12 +5,16 @@ import logging
 import math
 import os
 import pathlib
+import re
+import sys
 from typing import cast
 from typing import Dict
 from typing import List
 from typing import Optional
 
 import click
+from vdk.internal.builtin_plugins.config import vdk_config
+from vdk.internal.builtin_plugins.config.job_config import JobConfig
 from vdk.internal.builtin_plugins.run import job_input_error_classifier
 from vdk.internal.builtin_plugins.run.data_job import DataJobFactory
 from vdk.internal.builtin_plugins.run.execution_tracking import (
@@ -19,6 +23,7 @@ from vdk.internal.builtin_plugins.run.execution_tracking import (
 from vdk.internal.builtin_plugins.version import version
 from vdk.internal.core import errors
 from vdk.internal.core.context import CoreContext
+from vdk.internal.core.errors import VdkConfigurationError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


dakodakov (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -66,6 +71,45 @@ def __split_into_chunks(exec_steps: List, chunks: int) -> List:
...
+
+            pattern = r"^\d+\.\d+"
+            version_match = re.match(pattern, configured_python_version)
+
+            if version_match:
+                extracted_configured_version = version_match.group()
+                if extracted_configured_version != local_py_version:
+                    log.warning(
+                        f"""
+                        {os.linesep + (' ' * 18) + ('*' * 80)}
```
tozka, line 106: Let's improve the message. 

As a reminder we have a recommendation on what good message should contain: https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#error-handling

doks5: Hi, @tozka 

I've updated the warning message, following the error-handling standard as much as possible. Thanks

gageorgiev (approved):
looks great

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -66,6 +71,55 @@ def __split_into_chunks(exec_steps: List, chunks: int) -> List:
...
+            pattern = r"^\d+\.\d+"
+            version_match = re.match(pattern, configured_python_version)
+
+            if version_match:
+                extracted_configured_version = version_match.group()
+                if extracted_configured_version != local_py_version:
+                    log.warning(
+                        f"""
+                        {os.linesep + (' ' * 20) + ('*' * 80)}
+                        Python version ({configured_python_version}), set in the job's config.ini file, is different
```
tozka, line 107: We do not need to follow the format religiously if it doesn't make things clearer, right? 
We are answering here different questions with similar answers. 

The important thing is to provide the tools and information for users to understand what is the problem and how to solve it themselves. 

I think we can be more concise (less duplcation and unnecessary information) and explicit .

```
What: Python version mismatch between local python and configure python 
Why: The Python version specified in the job's config.ini file ({configured_python_version}) is different from the Python version ({local_py_version}) used to execute the data job.
Consequences: Developing a data job using one Python version and deploying it with a different version can result in unexpected and difficult-to-troubleshoot errors. This inconsistency may cause issues such as compatibility problems, module incompatibilities, or unexpected behavior during execution
Countermeasures: To resolve this issue, ensure that the Python version specified in the python_version property of the config.ini file matches the Python version of your execution environment by either editing the python_version property in config.ini or switching local environment to matching version of python
```

Is this clear ? Can user understand what they need to do ? Since it's a warning it's important to be clearer in the consequences so user can decide for themselves what they want to do.

doks5: Thanks for the review, @tozka 
Opened a new PR with the improved message: https://github.com/vmware/versatile-data-kit/pull/2250

PR #2244, opened 2023-06-13
control-service: add tmp dir path to image deployer's env variables

mrMoZ1:
what: Added the location to temporary directory for data jobs as an env variable in the JobDepoyer class.

why: As discusses in this comment - https://github.com/vmware/versatile-data-kit/pull/2206#discussion_r1223049666

testing- n/a trivial change.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
@@ -51,6 +51,8 @@ public class JobImageDeployer {
   private static final String KEYTAB_FILENAME_ENV = "VDK_KEYTAB_FILENAME";
   private static final String ATTEMPT_ID = "VDK_ATTEMPT_ID";
   private static final String BASE_CONFIG_FOLDER = "VDK_BASE_CONFIG_FOLDER";
+  private static final String VDK_TEMPORARY_WRITE_DIRECTORY = "VDK_TEMPORARY_WRITE_DIRECTORY";
+  private static final String TEMPORARY_WRITE_DIRECTORY_PATH = "/var/tmp";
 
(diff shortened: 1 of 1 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
```diff
@@ -51,6 +51,8 @@ public class JobImageDeployer {
   private static final String KEYTAB_FILENAME_ENV = "VDK_KEYTAB_FILENAME";
   private static final String ATTEMPT_ID = "VDK_ATTEMPT_ID";
   private static final String BASE_CONFIG_FOLDER = "VDK_BASE_CONFIG_FOLDER";
+  private static final String VDK_TEMPORARY_WRITE_DIRECTORY = "VDK_TEMPORARY_WRITE_DIRECTORY";
+  private static final String TEMPORARY_WRITE_DIRECTORY_PATH = "/var/tmp";
```
tozka, line 55: Hmm, Isn't TEMPORARY_WRITE_DIRECTORY_PATH actually "EPHEMERAL_VOLUME_MOUNT_PATH" ?

Do we need separate variable ? Or if we want for clarity sake we can just assign it 
`TEMPORARY_WRITE_DIRECTORY_PATH=EPHEMERAL_VOLUME_MOUNT_PATH`

mivanov1988: Yes, it is the same.

mrMoZ1: Is there a situation where those two can be different folders? (that's why I decided to use a separate variable.) In any case I think we should have two different variables here.

tozka (approved):

mivanov1988 (approved):

PR #2249, opened 2023-06-14
control-service: make kubernetes service easy  to test.

murphp15:
# What
The KubernetesService is by far the biggest class in control-plane. 

It has much too much responsibilities. 

It is responsible for initialing the class that it will use. 
For example it is responsible for creating the k8s client. 
This is against the spring model where each component should be setup with all the components it needs. 
But also it makes it hard to test, because we can't pass in mock clients. 

In this PR I create the client outside of the kuberentesService and autowire it into the service. 
I also do this for the batch clients. 

In a previous PR(https://github.com/vmware/versatile-data-kit/pull/2207#discussion_r1221372123) it was pointed out that there is a bug in the kubernetesService where 500s from the cluster are misclassified as the job not existing. 
To write a test for that bug is not easy because we are creating the clients within the class. 
After this PR is merged it will be much easier to test that bug. 

# How
I have taken the approach that more properties in kuberentes service should be final. 

In the PR it should be obvious that this change is better for tests and production code. 

In the old code we were calling initBatchV1Client() before every call to kuberentes just to create the client again which was totally needless because it should have been a singleton. 

Also in the tests we were often creating a mock to mock a single function to return another mock. All this noise is gone from the tests. 

# How how this tested? 
Local and CICD

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesServiceConfiguration.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesServiceConfiguration.java
@@ -28,69 +28,69 @@ import java.util.concurrent.TimeUnit;
 @Slf4j
 public class KubernetesServiceConfiguration {
 
-    private final UserAgentService userAgentService;
+  private final UserAgentService userAgentService;
 
-    @Autowired
-    public KubernetesServiceConfiguration(UserAgentService userAgentService) {
-        this.userAgentService = userAgentService;
+  @Autowired
+  public KubernetesServiceConfiguration(UserAgentService userAgentService) {
+    this.userAgentService = userAgentService;
+  }
+
+  @Bean
+  public BatchV1Api deploymentBatchV1Api(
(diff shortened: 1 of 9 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesServiceConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+    if (userAgentService != null) {
+      client.setUserAgent(userAgentService.getUserAgent());
+    }
+
+    // Annoying error: Watch is incompatible with debugging mode active
+    // client.setDebugging(true);
+    client.setHttpClient(
+        client.getHttpClient().newBuilder().readTimeout(0, TimeUnit.SECONDS).build());
+    // client.getHttpClient().setReadTimeout(0, TimeUnit.SECONDS);
```
DeltaMichael, line 93: Nit: Why do we need this comment and the above ones?

DeltaMichael (approved):

tozka (approved):

PR #2264, opened 2023-06-15
control-service: use real class instance in test

murphp15:
# Why
Makes tests clearer, easier to write and test more functionality. 


projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
```diff
@@ -444,7 +442,15 @@ public void testCreateV1CronJobFromInternalResource() {
...
-    KubernetesService mock = Mockito.mock(KubernetesService.class);
+    var mock =
+        Mockito.spy(
+            new DataJobsKubernetesService(
+                "default",
+                true,
+                new ApiClient(),
+                new BatchV1Api(),
+                new BatchV1beta1Api(),
+                new JobCommandProvider()));
```
gageorgiev, line 453: why not use the newDataJobKubernetesService function? I noticed that the only difference in the returned object is the boolean, maybe you could refactor that method to accept a param, and default to false so it doesn't require changing the other instances of the method call. Maybe this is overthinking it though.

gageorgiev (approved):

projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
murphp15: > why not use the newDataJobKubernetesService function? I noticed that the only difference in the returned object is the boolean

Yeah your description is spot on. 

Yeah I thought it was overkill because this is the only place we would pass in the parameter as true.

gageorgiev: gotcha, that sounds fine then

duyguHsnHsn (approved):

PR #2271, opened 2023-06-15
vdk-core: Add python_version configuration to config-help

doks5:
As part of the initiative to support multiple python versions for data job deployments, a new configuration property, `python_version` was introduced. This configuration, however, was not initially added to the output of the `vdk config-help` command.

This change adds the `python_version` configuration to the JobConfigIniPlugin, as it is related to the config.ini file setup.

Testing Done:
Tested locally by running `vdk config-help` and verifying that the configuration is present in the output.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
@@ -25,6 +25,7 @@ ATTEMPT_ID = "ATTEMPT_ID"
 EXECUTION_ID = "EXECUTION_ID"
 OP_ID = "OP_ID"
 TEMPORARY_WRITE_DIRECTORY = "TEMPORARY_WRITE_DIRECTORY"
+PYTHON_VERSION = "PYTHON_VERSION"
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


gageorgiev (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -128,6 +129,21 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             " no guarantee that files created during a local data job execution"
             " will  be present or absent for the next execution.",
         )
+        config_builder.add(
+            PYTHON_VERSION,
+            None,
+            False,
+            "The python version, which is to be used for the data job deployment."
+            " This configuration is to be set in the [job] section of the config"
```
tozka, line 137: We should explicitly say, this configuration can only be set in [job] section of the config and will be ignored if set anywhere else (e.g environment variable). 

As this is the current state.

tozka: Let's update JobConfigIniPlugin to set the value.
This way when inspecting (e.g with vdk debug) people can see what value is configured.

doks5: Hi, @tozka 

Updated the description and moved the configuration to the JobConfigIniPlugin.

tozka: I meant that there should be 

```
config_builder.set_value(
                "python_version",
                job_config.get_python_version(),
            )
```

doks5: Ah, I see. In such case, should the definition remain in JobConfigIniPlugin, or should I return it to the CoreConfigDefinitionPlugin plugin?

tozka: It does not matter at this stage too much. 

But Job Config seems more appropriate since it's a python version of a job not just the general SDK.

doks5: I've added the part to set the value.

tozka (approved):

PR #2279, opened 2023-06-19
vdk-notebook: handle job with  mixed .ipynb, .py, .sql  files use-case

duyguHsnHsn:
What:
Added string function to the stepBuilder, which is being called when a new notebook step is added to the job.
This ensures that the steps are executed alphanumerically.

Why:
currently, when we execute "vdk run" firstly all the python and sql files are executed and then the notebook files (which is not expected since we expect vdk run to run files alphanumerically) 


projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
```diff
@@ -82,6 +82,7 @@ def register_notebook_steps(file_path: Path, context: JobContext):
                         )
                         notebook_steps.append(step)
                         context.step_builder.add_step(step)
+                        context.step_builder.sort_steps()
```
ivakoleva, line 85: how many times is the sorting being executed, seems in a for cycle? it should be executed once at the end of init job phase

projects/vdk-core/src/vdk/internal/builtin_plugins/run/step.py
```diff
@@ -49,3 +49,6 @@ def add_step(self, step: Step) -> None:
 
     def get_steps(self) -> list[Step]:
         return self.__steps
+
+    def sort_steps(self) -> None:
+        self.__steps.sort(key=lambda step: step.name)
```
ivakoleva, line 54: An alternative approach to explicit sorting, is a sorted data structure. So, the sorting may not be omitted by accident, nor it should be executed explicitly. It guarantees consistency.

Also, I've noticed the change is within vdk-core internal, and only utilised in the notebook plugin. Why is that?

ivakoleva (approved):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/notebook.py
duyguHsnHsn: fixed

projects/vdk-core/src/vdk/internal/builtin_plugins/run/step.py
duyguHsnHsn: Since, as you said, it is only utilised in the vdk-notebook, I will remove the changes from the class and do the sorting explicitly in the vdk-notebook plugin

gageorgiev (approved):

PR #2269, opened 2023-06-15
control-service: multiple namespaces in testing

murphp15:
# Why
A bug creeped in to the code where we were deploying data jobs to control namespace and we weren't catching it. 
We couldn't catch it because in CICD data jobs and control deploy to the same namespace.
# What
control and data jobs now use differnt namespace in cicd.

PR #2290, opened 2023-06-20
control-service: install kubectl

murphp15:
# Why
The script deploy-testing-pipelines-service.sh needs kubectl installed if you are setting up the env. 
Here I install it so we can use the setup env in this step.

I discovered this because I am trying to create secrets in a new namespace in this PR https://github.com/vmware/versatile-data-kit/pull/2269.

This must be a regression from when we moved away from the aws image.


projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -257,6 +257,9 @@ control_service_deploy_testing_data_pipelines:
     - pip install --upgrade pip && pip install awscli
     - export FRONTEND_TAG=latest
     - export DESIRED_VERSION=v3.12.0 # helm version
+    - curl -LO https://dl.k8s.io/release/v1.27.2/bin/linux/amd64/kubectl
+    - chmod +x ./kubectl
+    - mv ./kubectl /usr/bin/kubectl
     - curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
     - bash -ex ./projects/control-service/cicd/deploy-testing-pipelines-service.sh
```
ivakoleva, line 264: How come we find about this now, can you add it to the PR description? For example, link the cicd job indicating the missing command? 
The cause is with the image change from proum/aws-iam-authenticator in Jan, or something else?

ivakoleva (approved):

duyguHsnHsn (approved):

projects/control-service/cicd/.gitlab-ci.yml
ivakoleva: there should be a cicd job failure as a result from such a regression. did we find such?

tozka: Probably no CICD failure: `kubectl` is used only when `RUN_ENVIRONMENT_SETUP` is set to 'y' (yes) in deploy-testing-pipelines-service.sh

And it's never really been set to yes in the Gitlab. I ran it once when initially setting up the environment in the beginning and I ran it likely from my machine.

PR #2291, opened 2023-06-21
control-service: move cron jobs methods to the data jobs class

murphp15:
# Why
At the moment the KubernetesService class as much to much responsibilities. 
It is extended by DataJobsKubernetesService and controlKubernetesService class.
However it includes the logic that is unique to those individual classes also. 
For example only the datajobsKS creates or deploys cron jobs.
So in this PR I move those external facing functions to the dataJobsclass and fix the tests. 

I have not move alot of the more private functions to the datajobs class yet because alot of other tests need to be changed to achieve this as they are being tested using reflection and I want to keep the PRs small. 

 
# How has this been tested?
Locally. 

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -719,45 +719,6 @@ public abstract class KubernetesService {
         .collect(Collectors.toSet());
   }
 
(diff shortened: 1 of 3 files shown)


gageorgiev (approved):
Just because I dont want to cross-reference 200 loc, the methods are the same, they're just moved to the subclass, correct?

tozka (approved):

murphp15:
> Just because I dont want to cross-reference 200 loc, the methods are the same, they're just moved to the subclass, correct?

Correct!

PR #2292, opened 2023-06-21
vdk-jupyter: pin jupyterlab to 3.6.3 in pyproject.toml

duyguHsnHsn:
What:
pinned jupyterlab version to 3.6.3 in pyproject.toml

Why: not being pinned was cause of some installation issues since the npm package dependencies are pinned

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/pyproject.toml
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.4.0", "jupyterlab>=3.4.7,<4.0.0", "hatch-nodejs-version"]
+requires = ["hatchling>=1.4.0", "jupyterlab==3.6.3", "hatch-nodejs-version"]
 build-backend = "hatchling.build"
 
 [project]


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.4.0", "jupyterlab>=3.4.7,<4.0.0", "hatch-nodejs-version"]
+requires = ["hatchling>=1.4.0", "jupyterlab==3.6.3", "hatch-nodejs-version"]
```
DeltaMichael, line 2: I don't agree with this, it should still be a range

```
>=3.4.7,<=3.6.3
```

duyguHsnHsn: @DeltaMichael in the npm package we pinned all the versions to 3.6.3 and the builders, as well, I ma not sure whether having lower versions here might cause some problem to not

DeltaMichael: My concern is if someone with an older version of Jupyter is using the extension and then tries to update. What's the user experience in this case? Otherwise, I'm ok with pinning it to 3.6.3 if it's pinned everywhere else.

DeltaMichael (approved):
Can you please add the conflicting packages to the des

yonitoo (approved):

PR #2294, opened 2023-06-21
control-service: ability to send authenticated email notifications

mrMoZ1:
what: Refactored email notifications package so that authenticated emails can be sent if proper configuration supplied.

why: As part of increasing the security of VDK we need to be able to send authenticated notification emails.

testing: added unit tests


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/DataJobNotification.java
```diff
@@ -81,11 +87,19 @@ public void notifyJobDeployError(JobConfig jobConfig, String errorName, String e
...
     }
   }
+
+  private void sendNotification(NotificationContent notificationContent) throws MessagingException {
+    if (configuration.isAuthEnabled()) {
+      authenticatedEmailNotification.send(notificationContent);
+    } else {
+      notification.send(notificationContent);
+    }
+  }
```
tozka, line 104: AuthenticatedEmailNotification is an EmailNotification but with added authentication. 

The DataJobNotification class doesn't have to know whether the notification needs to be authenticated or not. It will only depend on an abstraction (EmailNotification interface send(... ), not on any concrete implementation.   

This aligns well with the principles of good object-oriented design specifically -  the dependency inversion principle (depend on abstractions - the EmailNotification interface, not on any concrete implementation) and encapsulation (hide implementation details).

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/AuthenticatedEmailNotification.java
```diff
@@ -0,0 +1,90 @@
...
+    properties.putAll(this.emailPropertiesConfiguration.smtpWithPrefix());
+    session =
+        Session.getInstance(
+            properties,
+            getAuthenticator(
+                emailPropertiesConfiguration.getUsername(),
+                emailPropertiesConfiguration.getPassword()));
+  }
+
+  public void send(NotificationContent notificationContent) throws MessagingException {
```
tozka, line 43: I realize it's a draft. 
So I hope in latest version we are going to remove the duplication. 

But since Iam looking into it, let me share some ideas. 

We can either create one class and handle autneticaion internally n the class or we can still have 2 classes and employ decorator pattern.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailPropertiesConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+package com.vmware.taurus.service.notification;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class EmailPropertiesConfiguration {
```
tozka, line 15: nit: Properties and Configuration are generally the same thing (broadly) or very close. Just "EmailConfiguration" seems good enough ?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailPropertiesConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public String getPassword() {
+    return smtpWithPrefix().get(MAIL_SMTP_PASSWORD);
+  }
+
+  public boolean isAuthEnabled() {
+    return Boolean.parseBoolean(smtpWithPrefix().get(MAIL_SMTP_AUTH));
+  }
+
+  public String getTransportProtocol() {
```
tozka, line 52: add short description,  what is "transport protocol" and what are expected values for transport protocol ?

tozka (approved):
Looks pretty good to me. Thanks . I made a few "clean code" minor suggestions . Feel free to address them as you see fit and merge.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailNotification.java
```diff
@@ -95,11 +70,47 @@ public void send(NotificationContent notificationContent) throws MessagingExcept
...
+    var mimeMessage = prepareMessage(notificationContent);
+    transport.send(mimeMessage, recipients);
+  }
+
+  private void sendAuthenticatedEmail(NotificationContent notificationContent, Address[] recipients)
+      throws MessagingException {
+    Transport transport = session.getTransport();
+    var mimeMessage = prepareMessage(notificationContent);
+    try {
+      transport.connect(
```
murphp15, line 94: do you need to connect every single time? 
You are always using the same session. 
So maybe you only need to connect the one time?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailNotification.java
```diff
@@ -95,11 +70,47 @@ public void send(NotificationContent notificationContent) throws MessagingExcept
...
+    } else {
+      sendUnauthenticatedEmail(notificationContent, recipients);
+    }
+  }
+
+  private void sendUnauthenticatedEmail(
+      NotificationContent notificationContent, Address[] recipients) throws MessagingException {
+    Transport transport = session.getTransport();
+    var mimeMessage = prepareMessage(notificationContent);
+    transport.send(mimeMessage, recipients);
```
murphp15, line 86: in my ide it is saying this is a static method and that you didn't need to create an instance of transport.

murphp15:
if I send emails using authenticated emails instead what difference does it make to the customers. 
Does it make the emails less likely to be marked as spam?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailNotification.java
mrMoZ1: Though this suggestion is certainly a possibility and I have considered it in general I think it's better to use and close the connection every time because SMTP servers will try to kill the session for various reasons - http://www.postfix.org/postconf.5.html#smtp_connection_reuse_time_limit , http://www.postfix.org/postconf.5.html#smtp_connection_cache_reuse_limit .To elaborate SMTP uses socket connection that remain open until explicitly closed - http://www.google.com/codesearch#ojJ6TKlKTCM/trunk/mail/src/java/com/sun/mail/smtp/SMTPTransport.java&type=cs . I also believe that in our case it's better to create a new connection instead of maintaining it open all the time, since we don't send that many emails anyway.

mrMoZ1:
> if I send emails using authenticated emails instead what difference does it make to the customers. Does it make the emails less likely to be marked as spam?

- Notification email is verifiable by receiving users.
- Can also use different providers such as outlook or gmail that require authentication. 
- Can also monitor a sent inbox with outgoing notifications.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailNotification.java
murphp15: makes sense. thanks

murphp15 (approved):

PR #2295, opened 2023-06-21
control-service: use full url for heartbeat tests and heartbeat tests run in multiple namespaces

murphp15:
# Why
We want to catch regressions in control plane if they are related to using the wrong namespace. 
# What 
We now use different namespace for deployment and control. 
# How was this tested?
on a previous commit I enabled heartbeat steps and everything ran well.


tozka:
Maybe we can use the public one https://iaclqhm5xk.execute-api.us-west-1.amazonaws.com ?

Added benefit some tests can run in the gitlab.com runners (instead of ours) . 
I've been wanting to run vdk-hearbeat on windows for example.

projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -39,76 +39,6 @@
     - apk --no-cache add openjdk17-jdk git zip curl zip py-pip
     - pip install --upgrade pip && pip install awscli
 
-control_service_build_image:
```
DeltaMichael, line 42: Why remove all these stages?

murphp15: only doing this for testing.

projects/control-service/cicd/deploy-testing-pipelines-service.sh
```diff
@@ -51,22 +51,21 @@ if [ "$RUN_ENVIRONMENT_SETUP" = 'y' ]; then
...
     kubectl patch serviceaccount default -p '{"imagePullSecrets":[{"name":"'$secret_name'"},{"name":"'$dockerhub_secretname'"}]}'
 
   fi
-
 fi
 
 # this is the internal hostname of the Control Service.
 # Since all tests (gitlab runners) are installed inside it's easier if we use it.
-export CONTROL_SERVICE_URL=${CONTROL_SERVICE_URL:-"http://cicd-control-service-svc:8092"}
+export CONTROL_SERVICE_URL=${CONTROL_SERVICE_URL:-"http://cicd-control-service-svc.cicd.svc.cluster.local:8092"}
```
murphp15, line 68: @tozka 
This value is set but never used as far as i can see. Do you think it can be removed?

murphp15:
> Maybe we can use the public one https://iaclqhm5xk.execute-api.us-west-1.amazonaws.com ?
> 
> Added benefit some tests can run in the gitlab.com runners (instead of ours) . I've been wanting to run vdk-hearbeat on windows for example.
> 
> The downside may be that if Control Service is re-created (deleted and created) the Load Balanced domain name would change and one need to manually update the API gateway (https://github.com/vmware/versatile-data-kit/wiki/Gitlab-CICD#api-gateway)

@tozka  I don't agree with using the external url its needless IO on the network and not the pattern we should encourage.

projects/control-service/cicd/deploy-testing-pipelines-service.sh
tozka: It's used by this line below : 

```
envsubst < $VDK_OPTIONS > $VDK_OPTIONS_SUBSTITUTED
```

This generate vdk-options from this "template" ([control-service/cicd/vdk-options.ini](https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/cicd/vdk-options.ini#L12)) which are passed to Control Service .

tozka:
> 
> @tozka I don't agree with using the external url its needless IO on the network and not the pattern we should encourage.

You make a valid point, when it comes to jobs deployed / jobs runnning on the cloud ("cloud jobs").

Let me clarify the difference between local command runs and cloud runs.

-  Local command runs refer to executing jobs or other CLI commands on the local machine of the user, where the vdk is run on the same system where it's stored.
 - On the other hand, cloud runs involve deploying and executing jobs on a cloud-based platform (VDK Kubernetes runtime) 

For cloud jobs, indeed, minimizing unnecessary network I/O by avoiding external URLs is a more efficient approach as it reduces unnecessary network traffic and latency and instability. 

However  vdk-heartbeat is designed to verify local CLI (Command Line Interface) commands and simulate how user users VDK. For that case, it might be beneficial to use the external URL. The rationale behind this is that these tests simulate the conditions under which end-users would be interacting with our application. Typically, these users would be accessing our services via the external URL. Therefore, using the external URL in tests helps ensure that we're accurately replicating and testing the user experience. 

As part of its test vdk-heartbeat also deploys a job and remotely executes it. And also prints the logs of that remote execution. That remote execution likely would use CONTROL_SERVICE_URL that is internal because it's part of the "cloud" but the rest of the vdk-heartbeat which runs commands outside of the cloud should use external address

tozka:
> @tozka I am only coming back to this PR now.
> 
> I am setting the URL only in ini files which are used internally? Where in this PR do you want me to use the full external URL?

I forgot what was the problem being fixed here? Does it still exists? (perhaps update PR description?)

projects/control-service/cicd/deploy-testing-pipelines-service.sh
```diff
@@ -51,29 +51,36 @@ if [ "$RUN_ENVIRONMENT_SETUP" = 'y' ]; then
 
   if [ -n "$DOCKERHUB_READONLY_USERNAME" ]; then
     dockerhub_secretname='secret-dockerhub-docker'
-    kubectl create secret docker-registry "$dockerhub_secretname" \
+    kubectl create secret docker-registry secret-dockerhub-docker \
                          --namespace="cicd-deployment" \
                          --docker-server="https://index.docker.io/v1/" \
-                         --docker-username="$DOCKERHUB_READONLY_USERNAME" \
+                         --docker-username="vmwaivanov" \
```
tozka, line 57: I'd prefer the username to be configurable as well DOCKERHUB_READONLY_USERNAME

tozka (approved):

DeltaMichael:
@murphp15 Will this affect the frontend deployment in any way?

https://github.com/vmware/versatile-data-kit/blob/main/projects/frontend/cicd/.gitlab-ci.yml#L201
https://github.com/vmware/versatile-data-kit/blob/main/projects/frontend/cicd/ping_frontend_heartbeat_config.ini#L5

DeltaMichael (approved):

duyguHsnHsn (approved):

murphp15:
@DeltaMichael  it won't effect the frontend in anyway because deployments never make requests to the frontend.

PR #2312, opened 2023-06-23
frontend: use versatiledatakit/shared rc-builds for data-pipelines

DeltaMichael:
## Why?

Linking shared-components to data-pipelines using npm fails to produce a working UI docker image. This is a CI-only issue.

## What?

Introduce rc-versions for @versatiledatakit/shared
On a successful build of shared, and rc-version is pushed to npm registry
The data-pipelines build pulls the latest rc-version when running, instead of linking to the local installation of shared

Versioning schema follows semver

{major}.{minor}.{patch}-rc.{rc-number}

The rc-number is incremented based on the latest rc-number for the pipeline-id. This helps avoid conflicts between different pipelines running builds for shared

Regular shared versions are tagged with latest in npm registry
Rc versions are tagged with rc, so they don't get pulled accidentally when running npm install

## How was this tested?

Ran a build in CI and tested that the docker image starts locally as part of quickstart-vdk by changing the deployment tag

https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/909682172

Docker image tag pushed by pipeline `5eb2c00`
To check it locally, run quickstart-vdk and edit the deployment to use this tag.
Screenshot 2023-06-26 at 12 17 49

Feature/non-breaking

### DIFF
--- a/projects/frontend/cicd/.gitlab-ci.yml
+++ b/projects/frontend/cicd/.gitlab-ci.yml
@@ -38,7 +38,10 @@ frontend-data-pipelines-build:
   before_script:
     - cd projects/frontend/
   script:
-    - ./cicd/install_data_pipelines.sh
+    # Note: if frontend-shared-components-build is not part of the pipeline and no rc-version exists, $SHARED_RC_VERSION will be empty.
+    # If $SHARED_RC_VERSION is empty, calling install_data_pipelines.sh will install the latest @versatiledatakit/shared package from registry.
+    - export SHARED_RC_VERSION=$(../../cicd/get_latest_shared_rc_version.sh $CI_PIPELINE_ID)
+    - ./cicd/install_data_pipelines.sh $SHARED_RC_VERSION
   coverage: /^TOTAL\s+\d+\s+\d+\s+(\d+\%)$/
   retry: !reference [.retry, retry_options]
   rules:
@@ -93,7 +96,7 @@ frontend-data-pipelines-release:
   before_script:
     - cd projects/frontend/
   script:
-    - ./cicd/publish_package_npm.sh data-pipelines $CI_PIPELINE_ID $NPM_TOKEN $NPM_REGISTRY
+    - ./cicd/publish_package_npm.sh data-pipelines $CI_PIPELINE_ID $NPM_TOKEN $NPM_REGISTRY latest
   retry: !reference [.retry, retry_options]
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule"'
@@ -124,6 +127,7 @@ frontend_publish_ui_image:
       changes: *frontend_data_pipelines_locations
   extends: .frontend_publish_docker_image
 
+# TODO: Enable this when https://github.com/vmware/versatile-data-kit/issues/2332 is resolved
 # frontend_tag_ui_image_stable:
 #   stage: release
 #   before_script:
@@ -211,6 +215,9 @@ frontend-shared-components-build:
     - cd projects/frontend/shared-components/gui
   script:
     - ../../cicd/install_shared.sh
+    # Note: if this is the first time the pipeline runs and no rc-version exists, $SHARED_RC_INCREMENT will equal 1.
+    - export SHARED_RC_INCREMENT=$(../../cicd/get_next_shared_rc_version_number.sh $CI_PIPELINE_ID)
+    - ../../cicd/publish_package_npm.sh shared-components $CI_PIPELINE_ID-rc.$SHARED_RC_INCREMENT $NPM_TOKEN $NPM_REGISTRY rc
   retry: !reference [.retry, retry_options]
   rules:
     - if: '$CI_COMMIT_BRANCH == "main" || $CI_PIPELINE_SOURCE == "external_pull_request_event"'
@@ -234,7 +241,7 @@ frontend-shared-components-release:
   before_script:
     - cd projects/frontend/
   script:
-    - ./cicd/publish_package_npm.sh shared-components $CI_PIPELINE_ID $NPM_TOKEN $NPM_REGISTRY
+    - ./cicd/publish_package_npm.sh shared-components $CI_PIPELINE_ID $NPM_TOKEN $NPM_REGISTRY latest
   retry: !reference [.retry, retry_options]
   rules:
     - if: '$CI_PIPELINE_SOURCE == "schedule"'
--- a/projects/frontend/cicd/install_data_pipelines.sh
+++ b/projects/frontend/cicd/install_data_pipelines.sh
@@ -9,6 +9,8 @@
 # lints them to verify format, builds the UI application and runs the unit tests.
 ###
 
+shared_version=$1
+
 if ! which npm >/dev/null 2>&1 ; then
   echo "ERROR:"
   echo "Please install npm 8.5.5+. Install cannot continue without it."
@@ -26,8 +28,14 @@ rm -f "package-lock.json"
 shared_dist_dir="../../shared-components/gui/dist/shared"
 if [ -d "$shared_dist_dir" ]
 then
-  echo "Linking the shared-components dist rebuild found..."
-  npm link "$shared_dist_dir"
+  if [ -n "$shared_version"]
+  then
+    echo "Installing the latest shared-components rc-version"
+    npm install --no-save "@versatiledatakit/shared@$shared_version"
+  else
+    echo "Linking the shared-components dist rebuild found..."
+    npm link "$shared_dist_dir"
+  fi
 else
   echo "No shared-components dist rebuild found."
 fi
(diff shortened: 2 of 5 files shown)


projects/frontend/cicd/extract_shared_rc_increment.sh
```diff
@@ -0,0 +1,16 @@
...
+# Extracts the latest rc-version number for the pipeline id that's passed
+# and returns the incremented version number
+if [ $# -eq 0 ]
+  then
+    echo "ERROR: No argument for pipeline id supplied"
+    exit 3
+fi
+
+out=$(npm view @versatiledatakit/shared versions --json | grep $1-rc | cut -d \" -f 2 | tail -1 | awk -F '.' '{print $4}')
+out=$((out+1))
```
murphp15, line 15: what does this line do?

projects/frontend/cicd/install_data_pipelines.sh
```diff
@@ -26,8 +28,14 @@ rm -f "package-lock.json"
...
 if [ -d "$shared_dist_dir" ]
 then
-  echo "Linking the shared-components dist rebuild found..."
-  npm link "$shared_dist_dir"
+  if [ -n "$shared_version"]
+  then
+    echo "Installing the latest shared-components rc-version"
+    npm install --no-save "@versatiledatakit/shared@$shared_version"
+  else
+    echo "Linking the shared-components dist rebuild found..."
```
murphp15, line 36: when would we want to use this? 
might be better just to delete?

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -124,6 +125,7 @@ frontend_publish_ui_image:
       changes: *frontend_data_pipelines_locations
   extends: .frontend_publish_docker_image
 
+# TODO: Enable this when https://jira.eng.vmware.com/browse/TAUR-6077 is resolved
```
murphp15, line 128: can you make a github issue and link to that.

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -38,7 +38,8 @@ frontend-data-pipelines-build:
   before_script:
     - cd projects/frontend/
   script:
-    - ./cicd/install_data_pipelines.sh
+    - export SHARED_RC_VERSION=$(../../cicd/extract_shared_rc_version.sh $CI_PIPELINE_ID)
```
murphp15, line 41: what happens is shared didn't run in this pipeline? 
What value will it use for the shared rc version?

murphp15 (approved):

tozka:
> locally as part of quickstart-vdk by changing the deployment tag

can you provide more details on the testing? Provide such details that I can replicate the manual test done from the description only.

projects/frontend/cicd/.gitlab-ci.yml
DeltaMichael: It will just pass blank. If the pipeline hasn't run for shared, then there will be nothing in the registry and we'll get nothing when we pull the versions and filter. The install script will then build data-pipelines with the latest version. I'll test again just in case.

projects/frontend/cicd/extract_shared_rc_increment.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
```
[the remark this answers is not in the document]
DeltaMichael: I was struggling with this as well. This script just gives you the rc number incremented by 1, e.g. `8`. The other one gives you the current version, e.g. `1.3.13515124-rc.4`. I'll think about it some more and rename both scripts to be less confusing.

DeltaMichael:
@tozka Added more details on how to test.

projects/frontend/cicd/.gitlab-ci.yml
murphp15: can you document that here.

projects/frontend/cicd/extract_shared_rc_version.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+
+# Extracts the latest rc-version for the pipeline id that's passed
```
tozka, line 7: why not named the file "extract_latest_shared_rc_version" to be more explicit.

projects/frontend/cicd/extract_shared_rc_version.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ $# -eq 0 ]
+  then
+    echo "ERROR: No argument for pipeline id supplied"
+    exit 3
+fi
+npm view @versatiledatakit/shared versions --json | grep $1-rc | cut -d \" -f 2 | tail -1
```
tozka, line 13: Instead of grep and cut may be we can use jq which is a tool for parsing json: `jq -r '.[] | select(contains("909652143-rc"))'` 

Now imagine that npm view doesn't return the json formatted with new lines .

projects/frontend/cicd/extract_shared_rc_increment.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Extracts the latest rc-version number for the pipeline id that's passed
+# and returns the incremented version number
+if [ $# -eq 0 ]
+  then
+    echo "ERROR: No argument for pipeline id supplied"
+    exit 3
+fi
+
+out=$(npm view @versatiledatakit/shared versions --json | grep $1-rc | cut -d \" -f 2 | tail -1 | awk -F '.' '{print $4}')
```
tozka, line 14: How do we know for certian in what order `npm view` returns the versions? We seem to rely on that to take the last version. But i could not see anything reading `npm help view` .

DeltaMichael: It's true, it's not mentioned anywhere in the docs. However, based on experiments I've done, it does sort them by version and not by string. So it will do something like

```
  "1.1.909410501-rc.0",
  "1.1.909410501-rc.1",
  "1.1.909410501-rc.2",
  "1.1.909410501-rc.11",
  "1.1.909410501-rc.12",
  "1.1.909410501-rc.13",
```

which is what we want, but I haven't seen it do

```
  "1.1.909410501-rc.0",
  "1.1.909410501-rc.1",
  "1.1.909410501-rc.11",
  "1.1.909410501-rc.12",
  "1.1.909410501-rc.13",
  "1.1.909410501-rc.2",
```

or having them in random order. Again, it's not mentioned anywhere, but logically, it should sort them. Otherwise, this whole command is pretty useless.

projects/frontend/cicd/extract_shared_rc_version.sh
DeltaMichael: A good argument against this is that npm is a third-party tool and we're adding another third-party tool with jq. The current script uses unix built-ins, so we know they'll always behave the same way. Yes, npm is likely to change, but so is jq, so we're increasing the odds of something changing.

DeltaMichael: Also, not sure where to add `jq` if we decide to use it instead. The pipeline currently uses `versatiledatakit/vdk-cicd-base-gui:2.0.0` and I have no idea where it comes from and where it gets built. 😕

projects/frontend/cicd/.gitlab-ci.yml
DeltaMichael: Also worth noting: we do something similar in the increment script when we extract the version and increment it. If it's an empty string, it will still increment it and return 1, because bash considers empty string equivalent to zero when doing math.

projects/frontend/cicd/.gitlab-ci.yml
DeltaMichael: [linked to github issue](https://github.com/vmware/versatile-data-kit/issues/2332)

projects/frontend/cicd/.gitlab-ci.yml
DeltaMichael: Documented it in both places

projects/frontend/cicd/extract_shared_rc_increment.sh
tozka: OK

projects/frontend/cicd/extract_shared_rc_version.sh
tozka: I also did not know , but searching git repo shows it's produced by here https://github.com/vmware/versatile-data-kit/blob/a374d42c5e9d5800701d9892ade3edb23ee89317/projects/frontend/cicd/.gitlab-ci.yml#L196

If you can address it with sed, that's also fine. You can decide to leave it as it is even.
The risk is not that big perhaps but there's some. 

I will leave it up to you how to address it or not

DeltaMichael: https://github.com/vmware/versatile-data-kit/issues/2337

PR #2317, opened 2023-06-26
vdk-core: adapt to recent pluggy changes

dakodakov:
The new version of pluggy's _result cannot be automagically converted to a dictionary which causes issues.

Added a custom string converter.


dakodakov:
> 

Yes, it is backwards compatible.

tozka:
Thanks for fixing this! I saw our nightly builds started fail because of this last week and was planning to raise it but skip my mind. We are planning to have an "EO" on duty person monitoring CICD (incl. nightly builds) so hopefully we will fix those sooner in the future .

tozka (approved):

mivanov1988 (approved):

PR #2318, opened 2023-06-26
vdk-core: add vdk sdk secrets api - part I

dakodakov:
Incremental change for adding the secrets capability.

Since we have a somewhat circular dependecy with the vdk-test-utils plugin more changes - full secerts implementation and tests will follow in the next MR.

### DIFF
--- a/projects/vdk-core/src/vdk/api/job_input.py
+++ b/projects/vdk-core/src/vdk/api/job_input.py
@@ -33,6 +33,26 @@ class IProperties:
         pass
 
 
+class ISecrets:
+    """
+    Allows for Data Job to store and retrieve sensitive data.
+    Secrets are solution for the following use cases
+    * Keeping API keys and passwords necessary to connect to different systems
+    """
+
+    @abstractmethod
+    def get_secret(self, name: str, default_value: Any = None) -> str:
+        pass
+
+    @abstractmethod
+    def get_all_secrets(self) -> dict:
+        pass
+
+    @abstractmethod
+    def set_all_secrets(self, secrets: dict):
+        pass
+
+
 class IJobArguments:
     """
     Allows for users to pass arguments to data job run.
@@ -332,7 +352,9 @@ class ITemplate:
         pass
 
 
-class IJobInput(IProperties, IManagedConnection, IIngester, ITemplate, IJobArguments):
+class IJobInput(
+    IProperties, IManagedConnection, IIngester, ITemplate, IJobArguments, ISecrets
+):
     @abstractmethod
     def get_name(self) -> str:
         """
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -77,13 +77,72 @@ def set_properties_factory_method(
...
+        :param secrets: dictionary with the new job's secrets
+        :return secrets: dictionary with the secrets handled
+        """
+        pass
+
+
+ISecretsFactory = Callable[[], ISecretsServiceClient]
+
+
+class ISecretsRegistry(ABC):
```
murphp15, line 119: will there be multiple implementations of this?
I wouldn't have thought this needed to be an interface.

murphp15 (approved):

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
dakodakov: Yes, it's pretty much copied from the Properties API where we have "the real" one as well as test/dummy/in-memory implementations... Additionally it allows for extensibility.

mivanov1988 (approved):

tozka (approved):

projects/vdk-core/src/vdk/api/job_input.py
```diff
@@ -33,6 +33,26 @@ def set_all_properties(self, properties: dict):
...
 
+class ISecrets:
+    """
+    Allows for Data Job to store and retrieve sensitive data.
+    Secrets are solution for the following use cases
+    * Keeping API keys and passwords necessary to connect to different systems
+    """
+
+    @abstractmethod
+    def get_secret(self, name: str, default_value: Any = None) -> str:
```
tozka, line 44: nit: name of what? I know it's also not documented in get_property but I was left wondering what is the expected format of name , just any string. 

Btw. what are the allow value formats. They are also not documented explicitly in IProperties but there are checks. Perhaps they should have been

It's a nit (feel free to ignore comment).

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -77,13 +77,72 @@ def set_properties_factory_method(
...
+    """
+    Secrets registry to enable register implementation Secrets service
```
tozka, line 121: nit: that sounds awkward to me 

```suggestion
        Registry to enable the registration of Secrets service implementations.
```

projects/vdk-core/src/vdk/api/job_input.py
dakodakov: will address them it in the upcoming MR

PR #2319, opened 2023-06-26
vdk-jupyter: installation and build

duyguHsnHsn:
What:
Pinned all the dependencies to a specific version and add package-lock.json file. The purpose is to lock the installed dependencies to a specific version number so that subsequent installations can replicate the exact same dependency tree in the future, regardless of any new package releases.

Why: the package has really unstable build

Tested: cicd script is added for installing and building the package

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
+++ b/projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
@@ -17,9 +17,8 @@ build-vdk-jupyterlab-extension:
     - apk --no-cache add npm
     - apk add rust cargo
     - apk add libffi-dev
-    - pip install pytest-cov
-  script:
     - pip install jupyter
+  script:
     - echo "Install Dependencies..."
     - npm ci
     - echo "Install Vdk JupyterLab Extension..."
--- a/projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
+++ b/projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
@@ -1,0 +1,34 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+.retry:
+  retry_options:
+    max: 1
+    when:
+      - always
+
+
+build-vdk-jupyterlab-extension:
+  stage: build
+  image: "python:3.8-alpine"
+  before_script:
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+  stage: build
+  image: "python:3.8-alpine"
+  before_script:
+    - cd projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension || exit 1
+    - pip install -U pip setuptools pre-commit
+    - apk --no-cache add npm
+    - apk add rust cargo
+    - apk add libffi-dev
+    - pip install pytest-cov
+  script:
```
murphp15, line 21: Is these steps ppl might want to do locally? 
if it is then it might be best to do it in a script.

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -0,0 +1,34 @@
...
+build-vdk-jupyterlab-extension:
+  stage: build
+  image: "python:3.8-alpine"
+  before_script:
+    - cd projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension || exit 1
+    - pip install -U pip setuptools pre-commit
+    - apk --no-cache add npm
+    - apk add rust cargo
+    - apk add libffi-dev
+    - pip install pytest-cov
```
murphp15, line 20: Why did you decide to install pytest-cov in the before_script section and then jupyter in the script section?

murphp15 (approved):

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
duyguHsnHsn: @murphp15  in macOS these are not required and when building locally I had no problems with those

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
duyguHsnHsn: @murphp15 actually, I will delete the pytest-cov now since it is not needed here we only check the installation and the build, will move the jupyter installation

tozka (approved):

PR #2325, opened 2023-06-26
vdk-core: add vdk sdk secrets api - part III

dakodakov:
Incremental change for adding the secrets capability.

Add secrets implementation and tests in vdk-core.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/base_secrets_impl.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/base_secrets_impl.py
@@ -2,17 +2,18 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-def check_valid_property(k: str, v: str, supported_types=None) -> None:
+def check_valid_secret(k: str, v: str, supported_types=None) -> None:
     """
-    Check if property key and value are valid
+    Check if secret key and value are valid
     """
+
     if supported_types is None:
         supported_types = []
 
     if str != type(k) or k.strip() != k or "".join(k.split()) != k:
         msg = (
-            f"Property {k} is of unsupported type or has unsupported name. "
-            f"Only string properties with no whitespaces in the name are supported."
+            f"Secret {k} is of unsupported type or has unsupported name. "
+            f"Only string secrets with no whitespaces in the name are supported."
         )
         raise ValueError(msg)
 
@@ -21,7 +22,7 @@ def check_valid_property(k: str, v: str, supported_types=None) -> None:
 
     if type(v) not in supported_types:
         msg = (
-            f"Value {v} for property {k} is of unsupported type {type(v)}. "
+            f"Value for secret  {k} is of unsupported type {type(v)}. "
             f"Only int, float, str, list, and NoneType types are supported. "
         )
         raise ValueError(msg)
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/cached_secrets.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/cached_secrets.py
@@ -1,0 +1,32 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.job_input import ISecrets
+
+
+class CachedSecretsWrapper(ISecrets):
+    """
+    Wraps any ISecrets so that get_* calls are cached until set_* is called
+    Since secrets are rarely updated, it's better to have them cached.
+    """
+
+    def __init__(self, secrets_impl: ISecrets):
+        self.provider = secrets_impl
+        self.cached_dict = (
+            None  # when None then cache needs to be refreshed on next get_*()
+        )
+
+    def get_secret(self, name, default_value=None):
+        res = self._get_cached_dict().get(name, default_value)
+        return res
+
+    def get_all_secrets(self):
+        return self._get_cached_dict()
+
+    def set_all_secrets(self, secrets):
+        self.provider.set_all_secrets(secrets)
+        self.cached_dict = None
+
+    def _get_cached_dict(self):
+        if self.cached_dict is None:
+            self.cached_dict = self.provider.get_all_secrets()
+        return self.cached_dict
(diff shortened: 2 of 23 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/base_properties_impl.py
```diff
@@ -1,13 +1,14 @@
...
+def check_valid_property(k: str, v: str, supported_types=None) -> None:
     """
     Check if property key and value are valid
     """
+    if supported_types is None:
+        supported_types = []
+
     if str != type(k) or k.strip() != k or "".join(k.split()) != k:
         msg = (
             f"Property {k} is of unsupported type or has unsupported name. "
```
murphp15, line 14: there is a bug on line 24 of this file. it always says f"Only int, float, str, list, and NoneType types are supported. ". 
Even if they are not the types in the supportes_types list.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/base_properties_impl.py
```diff
@@ -1,13 +1,14 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
-from typing import List
-from typing import Type
 
 
-def check_valid_property(k: str, v: str, supported_types: List[Type] = []) -> None:
+def check_valid_property(k: str, v: str, supported_types=None) -> None:
```
murphp15, line 5: why did you removed the type info? 
Also why do you set the default to None. But then the first thing you do if it is None is set it to empty list.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/base_secrets_impl.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+
+def check_valid_secret(k: str, v: str, supported_types=None) -> None:
```
murphp15, line 5: is the only difference between this and check_valid_property the logging?

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/cached_secrets.py
```diff
@@ -0,0 +1,32 @@
...
+
+class CachedSecretsWrapper(ISecrets):
+    """
+    Wraps any ISecrets so that get_* calls are cached until set_* is called
+    Since secrets are rarely updated, it's better to have them cached.
+    """
+
+    def __init__(self, secrets_impl: ISecrets):
+        self.provider = secrets_impl
+        self.cached_dict = (
```
murphp15, line 14: why is there brackets here ?

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/datajobs_service_secrets.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+
+    __VALID_TYPES = [int, float, str, list, dict, type(None)]
+
+    def __init__(
+        self,
+        job_name: str,
+        team_name: str,
+        secrets_service_client: ISecretsServiceClient,
+        write_preprocessors: Optional[List[ISecretsServiceClient]] = None,
```
murphp15, line 34: Could you give an example of what a preprocessor might do.
Just curious.

murphp15: Its too confusing using one interface for different purposes.

murphp15 (approved):

tozka (commented):
There's a lot of duplication between properties and job_secrets. 

Do you plan to refactor it? 

Sometimes duplication may be ok . If two modules will have different reasons to change and are likely to diverge. 

But this is not true for large part of the code here.

tozka (approved):
The PR looks good. But we will need to include refactoring at some point of this initiative.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/base_properties_impl.py
dakodakov: Codacy check suggestion

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/base_properties_impl.py
dakodakov: Codacy check suggestion

tozka: To be more precise , the codacy warning is because Python evaluates the default argument once and then uses the same "instance" of that argument every time the function is called (if the argument is not provided). 

Basically acts as a global variable. 

See: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/dangerous-default-value.html

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/base_secrets_impl.py
dakodakov: Yes, as Antoni suggested, we can try to refactor this, however I would first like to get it running as I had similar intentions in the Control Service, which ultimately didn't come to fruition as the differences were far too many to extract sensible common classes.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/datajobs_service_secrets.py
dakodakov: I believe this is part of the extensibility mechanism which allows users to plug-in write pre-processor which modify secrets or properties before they are stored.

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/cached_secrets.py
dakodakov: fixed.

PR #2338, opened 2023-06-28
support: slack notification on pipeline failure

DeltaMichael:
## Why?
    
In order to set up CI on-call for vdk, visible and reliable notifications are required.
The notifications should alert on pipeline failure and provide the necessary information for troubleshooting.
    
## What?
    
Call an incoming webhook for a custom slack application on pipeline failure. This sends a message to the corresponding channel.
    
## How was this tested?
    
Ran in pipeline and sent notification to debug channel.
Note that the env variable for the notification hook will have to be changed after this is merged.

https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/914912201

Feature/non-breaking

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,3 +38,14 @@ end:
   only:
     refs:
       - external_pull_requests
+
+notify_failure:
+  stage: end
+  script:
+    - ./cicd/notify.sh "The pipeline for the $CI_COMMIT_BRANCH branch has failed.\nPlease take the necessary steps to fix it. https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/$CI_PIPELINE_ID" $VDK_SLACK_NOTIFICATION_HOOK
+  allow_failure: true
+  rules:
+    - if: '$CI_COMMIT_BRANCH != "main"'
+      when: never
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      when: on_failure
--- a/cicd/notify.sh
+++ b/cicd/notify.sh
@@ -1,0 +1,5 @@
+#!/bin/bash -e
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+curl -X POST -H "Content-type: application/json" --data "{\"text\":\"$1\"}" $2


mivanov1988:
Have you looked at https://docs.gitlab.com/ee/user/project/integrations/gitlab_slack_application.html?

DeltaMichael:
@mivanov1988 

> Have you looked at https://docs.gitlab.com/ee/user/project/integrations/gitlab_slack_application.html?

Yes, there are a couple of problems with it.

1. I requested the GitLab app to be added to our Slack workspace, but there was no response. I ended up making my own app and it got approved pretty quickly. I find it hard to believe I'm the first person in VMWare who has requested this, so maybe there's a reason behind it. 😕 We can dig a bit deeper, although I'm not sure where to start.

2. We're setting up on-call around the pipeline. This will most likely involve other notifications, e.g. when the on-call is rotating. The official GitLab app is not sufficient for this use case.

murphp15 (approved):

tozka:
> @mivanov1988
> 
> > Have you looked at https://docs.gitlab.com/ee/user/project/integrations/gitlab_slack_application.html?
> 
> Yes, there are a couple of problems with it.
> 
> 1. I requested the GitLab app to be added to our Slack workspace, but there was no response. I ended up making my own app and it got approved pretty quickly. I find it hard to believe I'm the first person in VMWare who has requested this, so maybe there's a reason behind it. 😕 We can dig a bit deeper, although I'm not sure where to start.
> 2. We're setting up on-call around the pipeline. This will most likely involve other notifications, e.g. when the on-call is rotating. The official GitLab app is not sufficient for this use case.

I belive for now use case #2 is not a priority . 

I am not sure I get why simple notifications are not possible with the slack app . But let's start with this now. We can always improve (if needed) later.

.gitlab-ci.yml
```diff
@@ -37,3 +38,13 @@ end:
   only:
     refs:
       - external_pull_requests
+
+notify_failure:
+  stage: notify
+  script:
```
tozka, line 44: The script must never fail.

.gitlab-ci.yml
```diff
@@ -37,3 +38,13 @@ end:
   only:
     refs:
       - external_pull_requests
+
+notify_failure:
+  stage: notify
```
tozka, line 43: I don't think we need a new stage. "end" stage is the end. Nothing else happens there (it's actually empty job). So we can put this there

cicd/notify.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+curl -X POST -H 'Content-type: application/json' --data "{\"text\":\"$1\"}" $2
```
tozka, line 5: fix the codacy warning, please. Seem relevant and important enough.

.gitlab-ci.yml
DeltaMichael: It doesn't matter if we add allow_failure. The script is always in the last stage of the pipeline. It only runs if one of the previous stages has failed. There is no scenario where the pipeline is successful, but the notify stage fails. Either the pipeline fails and the notify stage runs and is successful, or the pipeline fails and the notify stage fails as well.

I'll add the flag in case this scenario changes in the future.

PR #2348, opened 2023-07-03
control-service: split build job base image CI/CD step

mivanov1988:
Split the CI/CD step for Python and base job image publishing
due to the long-running execution (more than an hour).
Also, added native Oracle dependencies to the base image
which are needed for oracledb in Thick mode.

Testing Done:
https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/921583615


.gitlab-ci.yml
```diff
@@ -15,6 +15,8 @@ include:
 
 stages:
   - build
+  - publish_python_image
```
DeltaMichael, line 18: You probably don't need separate stages for this. You can just make two jobs in the publish_artifacts stage. If the order of jobs is important, you can make one depend on the other using the `needs` keyword.

```
 needs:
    - job: .control_service_publish_job_base_image_secure
```
https://docs.gitlab.com/ee/ci/yaml/index.html#needs

projects/control-service/projects/python-image-secure/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This directory provides the source of some base images for standard python versions.
+It's used by secured installation of VDK.
+
+The current base image installs supporting libraries for some native bindings necessary for installing from source
+some python packages which user may specify for their data job.
+
+## Build
+
+To build the job_base images run `./publish-job-base` which will publish new base image to versatiledatakit container registry.
```
DeltaMichael, line 13: Shouldn't it be `./publish-job-base-image` ?

murphp15:
Where is the python image used?

murphp15 (approved):

.gitlab-ci.yml
```diff
@@ -15,6 +15,8 @@ include:
...
+  - publish_job_base_image
   - publish_artifacts
```
tozka, line 20: Is there reason for adding new stags. Why not use either publish_artifacts stage? 

If there are some specific dependencies we can use https://docs.gitlab.com/ee/ci/yaml/#needs

Since stages are global for all projects CI I would like us to have good justification for adding or changing them.
Since more stages make CICD harder to understand for people. 

 Also the name of the stages are very specific for the Control Service. I think needs is the better solution here if dependencies are required.

tozka:
Why do we need Oracle dependencies to the base image
which are needed for oracledb in Thick mode.? 

Isn't thin mode enough

.gitlab-ci.yml
mivanov1988: Agreed, I will switch to the `needs`, thanks.

mivanov1988:
> Why do we need Oracle dependencies to the base image which are needed for oracledb in Thick mode.?
> 
> Isn't thin mode enough

Unfortunately, the Thin mode is not enough. It turned out that we have a lot of internal customers that use older Oracle versions.

projects/control-service/projects/python-image-secure/README.md
mivanov1988: Correct, I will fix it.

.gitlab-ci.yml
DeltaMichael: I was going to suggest this initially, but seeing the amount of jobs that would have to be chained with `needs` seemed like it would make the configuration file too verbose. 😕 I'm ok with whatever you decide @mivanov1988 , just think about this trade-off as well.

DeltaMichael (approved):

PR #2356, opened 2023-07-04
frontend: Fix navigation in Data Jobs

gorankokin:
Changed navigation in Data Jobs list from `navigateToUrl` to `locationToURL`, because telemetry events PageView are sent when Angular router navigation occurs. We want to stop sending telemetry PageView events on every filter apply.
Fix for Browser pop states while User is in Data Jobs list and going backward/forward and state from URL is propagated to the component.
Removed icon **new** from Executions tab in Data Job view.
Fix for some e2e tests.

More info about `navigateToUrl` and `locationToURL`:
- `navigateToUrl` initiates Browser navigation through Angular `Router` service which on the other hand triggers router change and component resolving through Angular Router mechanism that at the end fires `NavigationEnd` event and sends telemetry event PageView
- `locationToURL` initiates Browser navigation through Angular `Location` service, it updates Browser URL but doesn't trigger route change through Angular `Router` and because of that it doesn't send telemetry event PageView.
  - I can "say" that for telemetry it is still the same page, not new one
  - context for the page (components) is just fine-tuned, but global context is the same e.g. Manage Data Jobs and the component list that shows all the Data Jobs in the system

More info about `"We want to stop sending telemetry PageView events on every filter apply"`:
- We make it to be consistent with the other pages in VDK Data Pipelines UI and also with the other pages where VDK is consumed in Supercollider
- Telemetry library consumes Angular `NavigationEnd` events from Angular Router and sends them as PageView events to the REST Api
- Query params changes initiated from filter changes, shouldn't be logged as PageView from our perspective and for consistency
- With the new change we will lose filters information in PageView events url, but we will still have filters information in the other events like click, right_click, etc..

### DIFF
--- a/projects/frontend/data-pipelines/gui/projects/data-pipelines/src/lib/components/data-jobs-explore/components/grid/data-jobs-explore-grid.component.spec.ts
+++ b/projects/frontend/data-pipelines/gui/projects/data-pipelines/src/lib/components/data-jobs-explore/components/grid/data-jobs-explore-grid.component.spec.ts
@@ -178,12 +178,12 @@ describe('DataJobsExploreGridComponent', () => {
 
     describe('urlUpdateStrategy', () => {
         it('should verify the behaviour of _doUrlUpdate when urlUpdateStrategy is default (updateRouter)', () => {
-            const navigateToUrlSpy = spyOn(component.urlStateManager, 'navigateToUrl').and.returnValue(Promise.resolve(true));
+            const locationToUrlSpy = spyOn(component.urlStateManager, 'locationToURL').and.callFake(CallFake);
 
             // @ts-ignore
             component._doUrlUpdate();
 
-            expect(navigateToUrlSpy).toHaveBeenCalled();
+            expect(locationToUrlSpy).toHaveBeenCalled();
         });
 
         it('should verify the behaviour of _doUrlUpdate when urlUpdateStrategy is changed (updateLocation)', () => {
@@ -198,6 +198,10 @@ describe('DataJobsExploreGridComponent', () => {
     });
 
     describe('urlStateManager', () => {
+        beforeEach(() => {
+            fixture.detectChanges();
+        });
+
         it('should verify will invoke default urlStateManager (locally created)', () => {
             // Given
             const setQueryParamSpy = spyOn(component.urlStateManager, 'setQueryParam').and.callFake(CallFake);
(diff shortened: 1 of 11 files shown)


tozka:
> We don't want telemetry to be send when filters are applied in the URL as query string.

I am not sure I understand that. Why do not we want telemetry to send when filters are applied ? That seems like useful information to understand what filters are people most often using , how are they changing them ?

gorankokin:
@tozka the answer is complex, if you want we could make a quick call and I can explain you what we want to achieve with this changes here, in metadata and in supercollider.

gorankokin:
> > We don't want telemetry to be send when filters are applied in the URL as query string.
> 
> I am not sure I understand that. Why do not we want telemetry to send when filters are applied ? That seems like useful information to understand what filters are people most often using , how are they changing them ?

@tozka the answer is complex, if you want we could make a quick call and I can explain you what we want to achieve with this changes here, in metadata and in supercollider.

tozka (approved):

projects/frontend/data-pipelines/gui/projects/data-pipelines/src/lib/components/data-jobs-explore/components/grid/data-jobs-explore-grid.component.spec.ts
```diff
@@ -333,10 +344,14 @@ describe('DataJobsExploreGridComponent', () => {
...
     describe('search', () => {
+        beforeEach(() => {
+            fixture.detectChanges();
+        });
+
         it('makes expected calls', () => {
             spyOn(component, 'loadDataWithState').and.callThrough();
             component.search('searchValue');
-            expect(component.searchQueryValue).toBe('searchValue');
+            expect(component.clrGridUIState.search).toBe('searchValue');
```
VladimirPetkov1, line 354: I see many magic string literals in the PR, is it good idea to be defined as constants?

VladimirPetkov1 (approved):

hzhristova (approved):

projects/frontend/data-pipelines/gui/projects/data-pipelines/src/lib/components/data-jobs-explore/components/grid/data-jobs-explore-grid.component.spec.ts
gorankokin: They are numbers to acquire specific intercepted request from the framework and I don't agree with the static analyzer.

e.g.
```
expect(setQueryParamSpy.calls.argsFor(11)).toEqual([QUERY_PARAM_SEARCH, 'search test value 2', 0]);
```

with `argsFor(11)` we will get parameters for call with id number 11

PR #2359, opened 2023-07-05
frontend: quickstart-vdk operability tests using cypress

DeltaMichael:
## Why?

We test that the docker image is starting correctly, but we don't actually test that the frontend is operational. This has resulted in docker images where the docker image starts, but the frontend is not operational with no way of detecting the problem.

## What?

- Move regular e2e tests into their own directory
- Introduce quickstart-smoke script
- Create basic cypress smoke test for the frontend
- Run smoke test after deployment to test environment

## How was this tested?

Ran quickstart-vdk locally and ran cypress against frontend CI on localhost:8092

Ran smoke tests and e2e tests in CI https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/921430730

Feature/non-breaking


DeltaMichael:
Follow-up ticket https://github.com/vmware/versatile-data-kit/issues/2365

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -148,6 +148,10 @@ frontend_publish_ui_image:
 
 frontend_deploy_testing_data_pipelines:
   stage: pre_release_test
```
tozka, line 150: That is clearly from previous change but do you know why is "frontend_deploy_testing_data_pipelines" part fo pre_release_test stage? It should be part of pre_release stage where the actual "pre-releases" like testing deployment are done ?

DeltaMichael: Yeah, makes sense. I must have missed it when I did the initial change.

tozka (approved):

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -191,6 +195,30 @@ frontend_heartbeat_test:
     - if: '$CI_COMMIT_BRANCH == "main"'
       changes: *frontend_data_pipelines_locations
 
+frontend_quickstart_operability_test:
+  stage: pre_release_test
+  needs:
+    - job: frontend_heartbeat_test
```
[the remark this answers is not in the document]
DeltaMichael: No point doing operability tests if the docker image doesn't start.

projects/frontend/cicd/.gitlab-ci.yml
```diff
@@ -191,6 +195,30 @@ frontend_heartbeat_test:
...
+      optional: false
+      artifacts: false
+    - job: frontend-data-pipelines-build
```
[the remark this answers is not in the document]
DeltaMichael: We need the artefacts. They get lost along the way somehow, maybe due to switching of the runner image.

projects/frontend/cicd/.gitlab-ci.yml
DeltaMichael: Looking at it again, here is the stage order.

```
  - build
  - publish_artifacts
  - pre_release
  - pre_release_image
  - pre_release_test
  - release
  - release_image
  - end
```

It's a bit strange, but it's set up so the image is published and tagged as `latest` in the `pre_release_image` stage. So it makes sense to deploy the image tagged with `latest` in the next stage. To be fair, it might be a better idea to move the initial publishing to `publish_artifacts` or `pre-release` and introduce an rc-tag. We can deploy and then tag with `latest` on subsequent stages, once we know the image is deployable, at least. This seems out of scope for this change, so I'll open a separate ticket.

tozka (approved):

projects/frontend/cicd/.gitlab-ci.yml
DeltaMichael: https://github.com/vmware/versatile-data-kit/issues/2377

PR #2361, opened 2023-07-05
vdk-jupyter:  fix server error in jupyter ui and remove unneeded code

duyguHsnHsn:
What:
1.removed some unneeded code
2. fixed some issues with deployment from jupyter
3. fixed server fail on not present config.ini 

Why: 
1. there was an error in the server that was failing when config.ini is not present, I put that in try catch 
2. the deploy was failing because of enabled flag

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
@@ -12,7 +12,6 @@ from vdk.internal.control.command_groups.job.delete import JobDelete
 from vdk.internal.control.command_groups.job.deploy_cli_impl import JobDeploy
 from vdk.internal.control.command_groups.job.download_job import JobDownloadSource
 from vdk.internal.control.utils import cli_utils
-from vdk.internal.control.utils.cli_utils import get_or_prompt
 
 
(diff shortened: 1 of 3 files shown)


gageorgiev (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -152,22 +151,27 @@ def create_deployment(name: str, team: str, path: str, reason: str, enabled: boo
         :param enabled: flag whether the job is enabled (that will basically un-pause the job)
         :return: output string of the operation
         """
-        output = ""
+        output = "text"
         cmd = JobDeploy(RestApiUrlConfiguration.get_rest_api_url(), output)
-        path = get_or_prompt("Job Path", path)
```
DeltaMichael, line 157: Why are we not prompting the user for these anymore?

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: @DeltaMichael in the front end I already send him some messages about those, so doing it here is meaningless

PR #2366, opened 2023-07-06
frontend: bump cicd-base-gui image version

DeltaMichael:
## Why?

The previous version (2.0.0) had outdated toolchain dependencies

## What?

Bump vdk-cicd-base-gui image version to 2.1.2

Downgrade Cypress and angular-cli to 9.7.0 and 13.3.7 respectively.

## How was this tested?

CI

Feature/non-breaking


murphp15 (approved):

projects/frontend/shared-components/gui/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "shared-wrapper",
   "version": "0.0.0",
   "engines": {
-    "node": "^14.15.0 || >=16.10.0",
-    "npm": ">=8.5.5"
+    "node": "^14.15.0 || >=18.12.0",
+    "npm": ">=9.7.2"
```
duyguHsnHsn, line 6: Is it necessary to change the npm version as well?

duyguHsnHsn (approved):

projects/frontend/shared-components/gui/package.json
DeltaMichael: We build and ship with a Docker image using this specific version, so it should also be reflected here. One of the core principles when we were open-sourcing the CI was that building locally and in CI should be the same.

PR #2368, opened 2023-07-06
vdk-impala: support also pydantic 1.0

tozka:
Ensure vdk-impala works with pydantic 1.0 as this is still used and libraries do not support the new version.

In hindsight, I should have pinned to 1.0 only. But I did not and it's a good idea to support the new version otherwise at some point it will cause conflicts and problems as other libraries adopt 2.0 of pydantic

Testing Done: installed locally pydantic 1.0 . The Gitlab CI is testing with 2
 - `pip install pydantic==1.*`
 - ../build-plugin.sh (from vdk-impala dir) 
- verified the test exited successfully. 

I toyed with the idea of adding a test for pydanitc 1.0 but it would require some refactoring of hte plugin framework to allow that and it's not that urgent as the changes are fairly small.

### DIFF
--- a/projects/vdk-plugins/vdk-impala/requirements.txt
+++ b/projects/vdk-plugins/vdk-impala/requirements.txt
@@ -1,12 +1,5 @@
 
 # testing requirements
-click
 docker-compose
-impyla
-pyarrow
-pydantic
 pytest-docker
-tabulate
-vdk-core
-vdk-lineage-model
 vdk-test-utils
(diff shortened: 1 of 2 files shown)


doks5 (approved):
Maybe add a note to the PR description how the change was tested?

tozka:
> Maybe add a note to the PR description how the change was tested?

Explained, thanks for the reminder.

gageorgiev (approved):

Issue #2041, opened 2023-05-10
control-service: logs endpoint hangs and eventually throws an error if the job is still running


murphp15:
**Describe the bug**
if a job is running and we make a request for its logs the control service hangs and after a very long time returns an error. 

**Steps To Reproduce**
1.  install a quicksstart-vdk instance
2. create a job which would be long running, maybe add a sleep of 10 minutes in the job.
3. create and deploy a job in the newly created instance
4. execute the job
5. try look at the logs while the job is running

**Expected behavior**
Ideally it would return the generated logs so far. 
But a second best option would be to return a line saying "job still running please check back later."

PR #2370, opened 2023-07-07
control-service: logs endpoint doesn't hang

murphp15:
# Why 
Closes: https://github.com/vmware/versatile-data-kit/issues/2041 
if a job is running and we make a request for its logs the control service hangs and after a very long time returns an error.

# What
The error was happening because we were reading the logs for kubernetes using an inputStream function. 
Kuberenetes was streaming the logs to the control plane and doesn't close the input stream till the jobs is completed. 

Prevoiusly we were using 
```java
new PodLogs(client).streamNamespacedPodLog(...)
```
Within this function Kubernetes calls
```java
new CoreV1Api(client).readNamespacedPodLogCall(... , follow=true)
```

Instead of calling `PodLogs.streamNamespacedPodLog` instead I call   `CoreV1Api.readNamespacedPodLogCall` directly and I set follow=false. 
Now the function returns immediately. 

# How has this been tested
Integration test.
I have created a job that runs for 20 minutes. I query for its logs and make sure that it returns within 5 seconds.  


projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobViewLogsIT.java
```diff
@@ -0,0 +1,87 @@
+/*
```
doks5, line 1: I am probably missing something, but I don't see cleanup of the test job being performed. Does that mean that the job just runs for 20 minutes, and is then left in the namespace?

doks5 (approved):

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobViewLogsIT.java
murphp15: DataJobDeploymentExtension clean it up when the test finished.

PR #2375, opened 2023-07-10
frontend: Fix router event handling in base class

gorankokin:
* Fix router event handling in DataJobsBaseGridComponent
  * Distinguish between Browser popped state and changed Team
  * fix will affect only case when there is external Team selector dependency like `DataPipelinesConfig['manageConfig'].selectedTeamNameObservable`
* Added new SystemEvent with id `"SE_Location_Change"` bound to constant `SE_LOCATION_CHANGE` that is publicly exposed through `"@versatiledatakit/shared"`
  * with the new core SystemEvent there is provided interface for its payload
  * event is fired whenever method locationToURL in URLStateManager is invoked and condition allows location change
* Small refactor: introduced interface for SystemEvent payload about event with id `"SE_Navigate"`

Tested with Supercollider in local dev environment and everything works as expected.
Test scenarios:
1. In Data Jobs manage there are applied multiple filters in the grid and there is external dependency that team A is selected.
   1.  whenever external dependency notify that team was changed to B, content will be reloaded and Data Jobs for team B would be loaded and all filters that are applied in the grid will be preserved and used to load Data Jobs for team B
   2. before the fix, on Team change filters were deleted and default filter by status enabled in manage wasn't preserved
2. Tests that the newly introduced event could be consumed through `SystemEventHandler` in supercollider codebase and expected payload was received with attached debugger, and afterward processing is done for the needs of supercollider.
3. Automation tests are not added, because in public VDK UI there is no concept for external Team dependency (global Team selector) because it lays on Taurus foundation, where there wasn't concept of Teams but rather than that it was built around deployment per Tenant. Adding parametrized public VDK UI just for the test (where exist some abstract Team selector exist) is additional development cost for which we don't find its benefits right now.

### DIFF
--- a/projects/frontend/data-pipelines/gui/projects/data-pipelines/src/lib/components/base-grid/data-jobs-base-grid.component.ts
+++ b/projects/frontend/data-pipelines/gui/projects/data-pipelines/src/lib/components/base-grid/data-jobs-base-grid.component.ts
@@ -318,6 +318,7 @@ export abstract class DataJobsBaseGridComponent
      */
     onModelInit(): void {
         let initializationFinished = false;
+        let previousState: RouteState;
 
         this.subscriptions.push(
             this.routerService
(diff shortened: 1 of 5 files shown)


hzhristova (approved):

tozka:
> Fix router event handling in DataJobsBaseGridComponent.
> Distinguish between Browser popped state and changed Team.

What is the problem here? Please can you explain in the PR description. 

So is it when I navigate back to data jobs page (e.g click back) or when I change team in the UI, this causes problem.... Is that correct ? And what problem? Without understanding the actual problem it's hard to review properly. 


> Tested with Supercollider in local dev environment and everything works as expected.

Though I cannot comment on the testing until I am sure I understand the problem, description is not enough. Please provide information on exact manual steps so that if I have to re-do the the test myself I can do that without help. Also why aren't automated test in the codebase possible ?

thechefo (approved):

PR #2380, opened 2023-07-10
vdk-server: url in swagger ui is correct

murphp15:
# Why
When using quickstart VDK the swagger endpoint thinks that the base url should be http://localhost when infact it should be http://localhost:8092. 
The reason it gets confused is because we map the kind cluster port 80 to 8092 on the local machine. 

# What
The only fix available to us is to hard code the url in the quickstart vdk installer class. 

# How was this tested
In a previous commit I release and image and tested quickstart vdk locally with the change


projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
```diff
@@ -700,9 +700,13 @@ def __helm_install_command(self, git_server_ip):
             "--set",
             "extraEnvVars.SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI=http://localhost",
             "--set",
+            "extraEnvVars.SWAGGER_UI_HOSTNAME=http://localhost:8092",
+            "--set",
             "datajobTemplate.template.spec.successfulJobsHistoryLimit=5",
             "--set",
             "datajobTemplate.template.spec.failedJobsHistoryLimit=5",
+            "--set",
+            "image.tag=swaggertest",
```
DeltaMichael, line 709: Is this left over from testing?

DeltaMichael (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+import io.swagger.v3.oas.models.servers.Server;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+import javax.annotation.PostConstruct;
+import java.util.List;
+
+/**
+ * This class exists so that when running quickstart vdk the swagger ui has the correct base url.
```
tozka, line 18: This is not a problem unique to quickstart vdk. But to any deployment that is going to be similar to quickstart vdk. I would generalize the description since and use quickstart vdk as an example for one such case.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+ * <p>This is not actually correct! We are mapping port 80 from the cluster port 8092 on the local
+ * machine. There is no way for the service to know this mapping is happening so we need to provide
+ * it with the correct url.
+ */
+@Configuration
+public class SwaggerConfiguration {
+
+  private final OpenAPI openAPI;
+
+  @Value("${swagger-ui.hostname:}")
```
tozka, line 31: Currently, we document all variables in application-prod.yaml 
It could be just 

```
# This is ... 
# ... 
# ...
swagger-ui.hostname=${SWAGGER_UI_HOSTNAME:""}


It has the following benefits 
-  For people not familiar with spring auto wiring from env variables it makes it clear what is the env variable name that is used.
- It acts as a single place to document all variables used by the Control service.


As part of the variable documentation, please explain when (in what deployment topology) a user should use it (set it). It is currently not clear to me.

**Review by tozka — CHANGES_REQUESTED** [review]

File: projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java (hunk 2 above)
Comment (murphp15) [reply projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java:18]: Well it is unique to kind. 
When running behind a reverse proxy/ingress it can actually look at the forward-x headers. 
Even if service is running behind multiple layers of ingress it will most likely get the url correct. 


I don't see other setting it up in kind except when using quckstart vdk.

File: projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py (hunk 1 above)
Comment (murphp15) [reply projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py:709]: Yeah

File: projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.annotation.PostConstruct;
+import java.util.List;
+
+/**
+ * This class exists so that when running quickstart vdk the swagger ui has the correct base url.
+ *
+ * <p>Without this in qucikstart vdk the swagger will think the url should be localhost:80/.
+ *
+ * <p>This is not actually correct! We are mapping port 80 from the cluster port 8092 on the local
```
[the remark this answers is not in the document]
murphp15: It is suppose to say that the first statement is not correct. But I will remove that.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java
murphp15: WDYT?

tozka: My advice is not to assume how people would set it up. You can never know. Likelyhood is a lot of people would do something we do not expect. 

So my advice is to explain (to the best of your knowledge) under what setup would this need to be set. If there's no accurate way to describe it (there's a lot of uncertainty) . Or you can just say - Used if swagger URL is not autodetected correctly for any reason just set it using this variable. 
 
Another setup that is not autodetected correctly is https://iaclqhm5xk.execute-api.us-west-1.amazonaws.com/data-jobs/swagger-ui/index.html#/ 

You would notice the Swagger URL is the non secure (http) one instead of the API gateway provided URL. Things work because the non-secure one is also public at the moment.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java
murphp15: details added to the properties file. 
Let me know if you are happy with it.

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
```diff
@@ -700,6 +700,8 @@ def __helm_install_command(self, git_server_ip):
             "--set",
             "extraEnvVars.SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI=http://localhost",
             "--set",
+            "extraEnvVars.SWAGGER_UI_HOSTNAME=http://localhost:8092",
```
tozka, line 703: nit: I discourage making changes in multiple components in the same PR. 

I prefer using stacked PRs (https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/) . But I am not insistent on it. I just recommend it :) 

I think it's better for easier reversibility , clear commit log and tracing, release notes are clearer, currently release notes would only note the change in vdk-server but not the one is control-service

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/SwaggerConfiguration.java
murphp15: Please resolve if you think it looks better now.

tozka (approved):
Thanks! Looks good. I left general recommenation comment which you can ignore for this change if you want.

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
murphp15: I was actually planning on doing multiple PRs at first. 

If you look in my commit history. you can see I was only making the control-plane changes. 


But then to test it I needed to make these changes. 

Then given the size of the PR it just felt right to put it in the one PR.

tozka: > But then to test it I needed to make these changes.

Stacked PRs enabled that - https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/

PR #2381, opened 2023-07-10
control-service: push to multiple registries

tozka:
To avoid issues with docker limits sometimes it may be beneficial to use
ghcr.io instead for example in our CICD or in vdk-server .

This change publishes some images to both docker.io and ghcr.io

The GitLab CI configuration is changed to include a shared library for common CI/CD tasks. 
Two new scripts have been added for parsing and pushing Docker images. 
These scripts are now being used in the build process to push images to multiple registries in.

Testing Done: Temporarily changed the pipeline to push images on PR and verified images are pushed to both correctly. See it in https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/927292918

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,6 +4,7 @@
 image: "python:3.7"
 
 include:
+  - "cicd/.gitlab-ci-lib.yml"
   - "projects/vdk-control-cli/.gitlab-ci.yml"
   - "projects/vdk-core/.gitlab-ci.yml"
   - "projects/frontend/cicd/.gitlab-ci.yml"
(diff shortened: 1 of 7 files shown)


murphp15 (approved):

DeltaMichael:
> This change publishes some images to both docker.io and ghcr.io

We should also apply this to the frontend runner base images at the very least.

tozka:
> > This change publishes some images to both docker.io and ghcr.io
> 
> We should also apply this to the frontend runner base images at the very least. If it's going to be in this PR, please open a separate ticket for it.

Better I'd just open a PR. https://github.com/vmware/versatile-data-kit/pull/2390

DeltaMichael (approved):

PR #2406, opened 2023-07-12
vdk-jupyter: convert job operation

duyguHsnHsn:
What: 
This change includes: 

- implementing the actual transform operation
- adding an archive of the old job to the parent directory of the job
- adding a handler that captures the request from the front-end for the transform operation
- adding the necessary UI components

More info for the operation can be found: https://github.com/vmware/versatile-data-kit/wiki/VDK-Jupyter-Integration-Convert-Job-Operation

Testing: python unit tests, TBD: jest tests

Example of how transformed job looks like: 
Screenshot 2023-07-14 at 14 27 21
Screenshot 2023-07-14 at 14 27 30
Screenshot 2023-07-14 at 14 28 07

This is how a transformed job would look like, no UI changes are introduced in this PR


tozka:
Please provide screenshot highlighting any UI changes

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -206,9 +216,10 @@ def add_handler(handler, endpoint):
     add_handler(RunJobHandler, "run")
     add_handler(DeleteJobHandler, "delete")
     add_handler(DownloadJobHandler, "download")
-    add_handler(ConvertJobToNotebookHandler, "convertJobToNotebook")
+    add_handler(TransformJobHandler, "convertJobToNotebook")
```
yonitoo, line 219: Didn't we decide to rename the 'transform job' to 'convert job to notebook'? 
If there was another discussion later, whatever the chosen one is, I'd suggest sticking to it and renaming all the rest occurrences of the other one. As of now, adding this change would mix them and introduce inconsistencies.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx
```diff
@@ -52,8 +55,11 @@ function add_command(commands: CommandRegistry, schemaNaming: string, label: str
           runningVdkOperation = true;
           jobData.set(VdkOption.PATH, workingDirectory);
           await jobdDataRequest();
```
murphp15, line 57: why do you need to run this here?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
```diff
@@ -60,6 +71,9 @@ export async function showConvertJobToNotebookDialog() {
     if (confirmation.button.accept) {
       let { message, status } = await jobConvertToNotebookRequest();
```
murphp15, line 72: use const instead of let.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
```diff
@@ -60,6 +71,9 @@ export async function showConvertJobToNotebookDialog() {
...
       if (status) {
+        const transformjobResult = JSON.parse(message);
+        const notebookContent = initializeNotebookContent(transformjobResult["code_structure"], transformjobResult["removed_files"])
+        createTranformedNotebook(notebookContent, commands, fileBrowser, notebookTracker);
```
murphp15, line 76: you need an await here?

murphp15 (approved):
documentation on the classes is v good

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx
duyguHsnHsn: @murphp15 if there is a config file in the current directory it gathers the information about the job from it and sets default values to the inputs

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
duyguHsnHsn: @murphp15 i will be fixing those, i got some warnings

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
duyguHsnHsn: We discussed and we will be renaming everything to convert

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
duyguHsnHsn: Actually i realised we redefine message, so we cannot make it const

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
duyguHsnHsn: added

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/convert_job.py
```diff
@@ -0,0 +1,88 @@
...
+        parent_dir = self.get_parent_dir()
+        return os.path.join(parent_dir, f"{dir_name}_archive")
+
+    def archive_folder(self):
+        shutil.make_archive(self.get_archive_name(), "zip", self._dir_path)
+
+
+class ConvertJobDirectoryProcessor:
+    """
+    The TransformJobDirectoryProcessor class provides functionality to process a directory,
```
DeltaMichael, line 44: Nit: It was renamed to ConvertJobDirectoryProcessor, so should be reflected in the doc as well.

duyguHsnHsn: @DeltaMichael right, I forgot to change them there ;(

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/convert_job.py
```diff
@@ -0,0 +1,88 @@
...
+        all_files = sorted(glob.glob(os.path.join(self._dir_path, "*")))
+        self._archiver.archive_folder()
+
+        for file in all_files:
+            if file.endswith(".sql"):
+                with open(file) as f:
+                    content = f.read()
+                self._code_structure.append(f'job_input.execute_query("""{content}""")')
+                self._removed_files.append(os.path.basename(file))
+                os.remove(file)
```
DeltaMichael, line 74: What if the process is interrupted and you end up with half the files removed and an incomplete notebook? Sure, the user could restore their job from the archive, but it's a bit of an annoying experience, imo.

Suggestion is to do this kind of cleanup at the very end, once we know that the notebook was converted correctly. Maybe have it as a separate method that deletes all the files in `_removed_files` and call it after the `processor.process_files`

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -111,16 +112,20 @@ def post(self):
...
+    to a notebook type data job
     """
 
     @tornado.web.authenticated
     def post(self):
-        # TODO fix this as part of the implementation
-        print("Successfully connected to the Convert Job To Notebook handler!")
+        input_data = self.get_json_body()
+        try:
+            message = json.dumps(VdkUI.convert_job(input_data[VdkOption.PATH.value]))
```
DeltaMichael, line 125: It might be a good idea to put these `json.dumps()` calls inside the `convert_job` and `finish` methods. Unless those methods are used for something else, this would make the code much more readable.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/convert_job.py
```diff
@@ -0,0 +1,88 @@
...
+import shutil
+
+
+def validate_dir(dir_path):
+    if not os.path.isdir(dir_path):
+        raise ValueError(f"{dir_path} is not a valid directory.")
+    return True
+
+
+class DirectoryArchiver:
```
DeltaMichael, line 15: I don't see the point in this being a class. You just initialize it in the processor and call it's archive_folder method. Unless you absolutely have to re-use it somewhere, I suggest you move that `validate_dir()` call to the `archive_folder()` method, put `archive_folder()` inside the processor class and get rid of the class.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
duyguHsnHsn: @DeltaMichael it is used in all the other handlers too so i think making one different is not good but idk

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/convert_job.py
duyguHsnHsn: @DeltaMichael yes, good idea I will add it

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/convert_job.py
duyguHsnHsn: @DeltaMichael I did it because i didn't really want to make the processor class so broad and having too different methods, because it is a totally different thing i can even call the archiver outside that class and do not include it in the class, i think i will do it like that instead so we can purely use the processor for processing and not to archive

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
DeltaMichael: Please open an improvement ticket for it, like we discussed.

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
duyguHsnHsn: @DeltaMichael https://github.com/vmware/versatile-data-kit/issues/2438

PR #2407, opened 2023-07-12
vdk-jupyter: publish image to pip registry

murphp15:
# Why
We need to publish the image to a pip registry so others can consume it and setup is easy and doesn't require cloning the source code

# What
Most plugins are installed using the same command. which is declared in a common gitlab file. 
For building this plugin I didn't think it made sense to reuse that code as building jupyter is a little more complicated. 
So the release is totally new code. 

# How was this tested
I published to the registry and installed locally. 

Commands: 
```
pip install -U vdk_jupyterlab_extension
jupyter lab
```
Make sure the UI has the correct panels
Screenshot 2023-07-13 at 12 32 42

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
@@ -17,7 +17,9 @@ for the frontend extension.
 ## Install and run
 
 ```bash
+# install the extension
 pip install vdk-jupyterlab-extension
+# run jupyterlab
 jupyter lab
 ```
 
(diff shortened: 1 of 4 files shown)


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
```diff
@@ -14,16 +14,11 @@ for the frontend extension.
...
-To install the extension,first navigate to the project directory and then execute:
+## Install and run
 
 ```bash
-# Once in the project directory, use the npm ci command to install
-# the exact versions of the dependencies specified in the package-lock.json
-npm ci
-# Install the extension
 pip install vdk-jupyterlab-extension
+jupyter lab
```
duyguHsnHsn, line 21: Here add a comment # run jupyterlab

duyguHsnHsn (approved):
Looks good to me

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/README.md
murphp15: Done

PR #2409, opened 2023-07-13
vdk-control-cli: remove set-secret for properties

dakodakov:
Remove the set-secret option for the properties command as we now have separate secrets command for handling sensitive data.
Update example/documentation.
Update tests.

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/properties.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/properties.py
@@ -29,7 +29,6 @@ class PropertyOperation(Enum):
     """
 
     SET = "set"
-    SET_SECRET = "set_secret"  # nosec
     GET = "get"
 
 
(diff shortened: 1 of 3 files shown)


tozka (approved):

tozka:
Please explain why the change is needed in the description

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/properties.py
```diff
@@ -327,13 +314,9 @@ def properties_command(
             )
 
 
-def _get_key_value_pairs(set, set_secret):
+def _get_key_value_pairs(set):
```
doks5, line 317: Maybe this could be renamed to something like `set_property`?

doks5 (approved):

projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/properties.py
dakodakov: let's leave it as it is for now... pending refactoring.

PR #2415, opened 2023-07-14
control-service: errors should be returned to client correcctly

murphp15:
# Why
At the moment if the kubernetes client throws and error then we assume that the job is not there. 
This is obviously not correct. We should only consider the job not there if it is a 404. 

# What
Now we check to see the status to decide what to do. 

# How was this tested
New unit test which mock k8s responses.


gageorgiev (approved):

tozka (approved):

murphp15:
Not ready to be merged.

tozka (approved):

DeltaMichael (changes requested):
Don't merge this before all the tests run.

PR #2430, opened 2023-07-19
control-service: Update contributing.md with correct java requirements

danail-georgiev:
Changing Java minimal version to 17 since the project can not build with Java 11


tozka (approved):
Thanks for the fix. And sorry about the caused confusion. We should have corrected that earlier.

doks5 (approved):

PR #2445, opened 2023-07-24
vdk-plugins: fix build of multiple plugins

tozka:
There were overall 3 issues that needed to be fixed in the build of those   plugins: 

The first one is 2.0 major release of pytest-docker (https://pypi.org/project/pytest-docker/#history) which we use for starting docker container in some plugin tests . This impacted us because 

1. pytest-docker change default command from `docker-compose` to `docker compose` (no dash) 
2. But the library we use (docker-compose) installs only `docker-compose` command (with a dash)

--
The other problematic library is the installation of docker-compose itself :). 
docker-compose depends on PyYaml and PyYAML is having the following issue: https://github.com/yaml/pyyaml/issues/724 (in short a recent release of Cython 3.0 broke PyYAML)

As a workaround, we can pin `PyYAML==5.3.1` And also remove eager (transitive) upgrades when installing the package with pip to reduce the chance of accidentally upgrading it.

--
Finally removed all non- testing dependencies from requirements.txt. In one of the plugin was failing to install because it was missing install requirements in setup.py and was not caught during testing because requirements.txt was installing that dependency. 

------

An alternative approach to fix it might have been to switch to use [testcontainer](https://pypi.org/project/testcontainers/) (like vdk-trino uses) instead of pytest-docker + docker-compose. But that would require refactoring of the tests so I opted out for now.

### DIFF
--- a/projects/vdk-plugins/vdk-greenplum/.plugin-ci.yml
+++ b/projects/vdk-plugins/vdk-greenplum/.plugin-ci.yml
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+
 
 .build-vdk-greenplum:
   variables:
--- a/projects/vdk-plugins/vdk-impala/.plugin-ci.yml
+++ b/projects/vdk-plugins/vdk-impala/.plugin-ci.yml
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+
 
 .build-vdk-impala:
   variables:
--- a/projects/vdk-plugins/.plugin-common.yml
+++ b/projects/vdk-plugins/.plugin-common.yml
@@ -9,12 +9,12 @@
 
 .build-plugin:
   stage: build
-  before_script:
+  # before_script: leaving it empty so that script can be extended
+  script:
     - cd projects/vdk-plugins/
     - pip install -U pip
     - pip install ./vdk-test-utils # TODO: REMOVE THIS AT SOME POINT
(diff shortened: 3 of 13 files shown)


projects/vdk-plugins/vdk-greenplum/.plugin-ci.yml
```diff
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+
```
doks5, line 4: nit: Is the extra line necessary (same comment for all the places with extra added lines)?

doks5 (approved):

projects/vdk-plugins/vdk-impala/.plugin-ci.yml
```diff
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
```
murphp15, line 4: what happens when you remove this image? It just uses a default one from gitlab? 
If so what is it? 
I'm just curious, I will still approve.

murphp15 (approved):

projects/vdk-plugins/vdk-impala/.plugin-ci.yml
tozka: There's a default one specified in the root gitlab-ci.yaml (https://github.com/vmware/versatile-data-kit/blob/main/.gitlab-ci.yml#L4)


There is a potential problem in specifying the same "global" key in multiple yaml files, it gets overwritten.
And that is global key. It is not scoped to anything (unlike per-job image which is scoped to a specific gitlab ci job)

The problem can surface because the override is not scoped to the file. So depending on the import order you may get surprising values of an image in some cases in some files. E.g you are running python tests and all of a sudden

PR #2449, opened 2023-07-24
control-service: fix helm chart

dakodakov:
Fix helm chart to properly substitute vault secret name.


doks5 (approved):

murphp15:
Is it just formatting?

tozka:
How was the issue caught? 

How was it tested ? If it was tested manually please provide clear steps (that anyone can follow to re-test if so wants)

 And can we add some automated testing.

There is `helm lint` being executed in CI but it did not catch that issue.

PR #2450, opened 2023-07-24
[bug fix] vdk-core: Add job name to state

doks5:
Currently, when a data job is executed and notifications are enabled, the resulting email does not contain the data job name and instead says `None`. Example:
```
[success][data job run] None

Dear Versatile Data Kit user,
Last run of your data job *None* has succeeded.
You can find full logs of the job execution here.
The sender mailbox is not monitored, please do not reply to this email.
Best,

```

This change fixes this behaviour by setting the Notification Plugin to attach to the
finalize_job hook, get the job name from the JobContext provided by the hook, and
save as an instance variable. The name of the job is then passed to the vdk_exit hook,
which is responsible for sending the notifications.

Testing Done: Locally executed a test data job and verified that the email notification contains the data job name.

*Additional test configuration that needs to be set as env vars:*

```
VDK_ENABLE_ATTEMPT_NOTIFICATIONS=true
VDK_LOG_CONFIG=CLOUD
VDK_NOTIFICATION_ENABLED=true
VDK_NOTIFIED_ON_JOB_SUCCESS=<target-email-address>
VDK_NOTIFICATION_SENDER=<sender-email-address>
```

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/notification/notification.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/notification/notification.py
@@ -15,6 +15,7 @@ from vdk.internal.builtin_plugins.notification.notification_configuration import
     SmtpConfiguration,
 )
 from vdk.internal.builtin_plugins.run.execution_state import ExecutionStateStoreKeys
+from vdk.internal.builtin_plugins.run.job_context import JobContext
 from vdk.internal.core import errors
 from vdk.internal.core.config import ConfigurationBuilder
 from vdk.internal.core.context import CoreContext
@@ -29,7 +30,7 @@ def __get_list(value: str) -> List[str]:
     return result
 
 
-def _notify(error_overall, user_error, configuration, state):
+def _notify(error_overall, user_error, configuration, state, job_name):
     notification_cfg = NotificationConfiguration(configuration)
 
     if (
@@ -41,7 +42,6 @@ def _notify(error_overall, user_error, configuration, state):
         f"Prepare to send notification if necessary: error_overall: {error_overall}"
     )
 
-    job_name = state.get(ExecutionStateStoreKeys.JOB_NAME)
     op_id = state.get(CommonStoreKeys.OP_ID)
 
     job_log_url_template = notification_cfg.get_job_log_url_pattern()
@@ -107,9 +107,16 @@ def _notify(error_overall, user_error, configuration, state):
 
 
 class NotificationPlugin:
+    def __init__(self):
+        self._job_name = None
+
     @hookimpl(tryfirst=True)
     def vdk_configure(self, config_builder: ConfigurationBuilder):
         notification_configuration.add_definitions(config_builder)
+
+    @hookimpl
+    def finalize_job(self, context: JobContext):
+        self._job_name = context.name
 
     @hookimpl
     def vdk_exit(self, context: CoreContext, exit_code: int):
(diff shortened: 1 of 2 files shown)


mivanov1988 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -253,6 +253,7 @@ def __init__(
             )
         self._name = data_job_directory.name if name is None else name
         self._data_job_directory = data_job_directory
+        core_context.state.set(ExecutionStateStoreKeys.JOB_NAME, self._name)
```
tozka, line 256: This seems like a patch solution, and it may not address the root cause of the problem. 

If we do not address the root cause, it will surface in another way (with similar problem) . 

Let's think about what is the root cause? 

The context state is updated by [ExecutionTrackingPlugin](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_tracking.py#L15) 
during initialize_job hook. 

While the [notifications](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/notification/notification.py#L115) are sent during vdk_exit (much later hook)

So why is not the state propagated properly? 

I think the issue might be in that , core_context.state is hierarchical. 

WHen a job is started a child context is created , and the job name is in the child context, then when the job finishes, teh child context is "poped" and the parent context knows nothing about the job state. 

I would change the [notification plugin](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/notification/notification.py#L115) to hook also into some hook during the [Job Run](https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-plugins#data-job-run-execution-cycle) and record as classic instance variable the needed information (job name, whatever else is needed) and use it during vdk_exit.

doks5: I was hesitant to attach to yet another hook for this, as it seemed as too much of an overhead. But, yeah, I agree that in this case it is better to use a hook to avoid unexpected issues.

murphp15 (approved):

tozka (changes requested):
Please, please, don't forget and skip automated tests. 
Including some tests to demonstrate that the changes actually fix the described issue, and to prevent future regressions is pretty important. 

And it's not hard. Adding an assert statement in test_run_notifications.py (test_run_successfull) would be enough to cover it: 
 A) It would reproduce the issue, since the job name is missing the message there as well
 B) After the fix is applied the would  pass verifying the fix and preventing future regressions.

tozka (approved):
Otherwise the change looks good to me. Providing tests are added, the change is approved from me.

doks5:
> Please, please, don't forget and skip automated tests. Including some tests to demonstrate that the changes actually fix the described issue, and to prevent future regressions is pretty important.
> 
> And it's not hard. Adding an assert statement in test_run_notifications.py (test_run_successfull) would be enough to cover it: A) It would reproduce the issue, since the job name is missing the message there as well B) After the fix is applied the would pass verifying the fix and preventing future regressions.

Updated test.

doks5:
> Otherwise the change looks good to me. Providing tests are added, the change is approved from me.

Thanks for the review

Issue #2451, opened 2023-07-24
VEP-2448: Define Motivations, Goals and Non-Goals


DeltaMichael:
**What is the feature request? What problem does it solve?**

Fill out the Motivations and Goals sections of VEP-2448

**Suggested solution**

Base text of off user research

**Additional context**
https://github.com/vmware/versatile-data-kit/issues/2399
https://github.com/vmware/versatile-data-kit/issues/2398

Issue #2448, opened 2023-07-24
Create Run Logs Initial VEP Document


DeltaMichael:
**What is the feature request? What problem does it solve?**
A PR with an initial VEP is created

**Suggested solution**
Follow. https://github.com/vmware/versatile-data-kit/blob/main/specs/README.md

Issue #2408, opened 2023-07-13
VDK Run logs: Simplified and readable


sabadzhiev:
**_Business case:_**
Provide robust error troubleshooting and recovery mechanisms for ingestion/transformations.
Easier to discover source of errors in ingestion/transformations:
- improve error messages
- better traceability

The improvements are focusing the needs of the data practitioners working on both VDK OSS and the Internal VMWare deployment(s).

_Note_: A special consideration are the logs in combination with VDK DAGs.

_Note_: Feedback materials are available:

_**Metrics:**_
Improve readability score (for example Flesch-Kincaid, Gunning Fog Index, or the Simple Measure of Gobbledygook (SMOG)) two times.
Measure "Error Spotting Time" 2x times at least (evaluated with user testing or telemetry from SC).
How many times user try to access the logs, How long are using looking at logs (reduce time)

PR #2456, opened 2023-07-24
specs: VEP-2448: VDK Run Logs: Simplified And Readable

DeltaMichael:
# Why?

Part of the run logs improvement initiative https://github.com/vmware/versatile-data-kit/issues/2408

https://github.com/vmware/versatile-data-kit/issues/2451
https://github.com/vmware/versatile-data-kit/issues/2448

# What?

Create VEP document and fill out Summary, Motivation and Goals sections

# How has this been tested?

n/a

docs

### DIFF
--- a/specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
+++ b/specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
@@ -1,0 +1,251 @@
+
+# VEP-2448: VDK Run Logs: Simplified And Readable
+
+* **Author(s):** Dilyan Marinov (mdilyan@vmware.com), Antoni Ivanov (aivanov@vmware.com)
+<!-- * **Status:** draft | implementable | implemented | rejected | withdrawn | replaced -->
+* **Status:** draft
+
+
+To get started with this template:
+
+- [x] **Create an issue in Github (if one does not exists already)**
+- [x] **Make a copy of this template directory.**
+  Copy this template into the specs directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) created above.
+- [ ] **Fill out this file as best you can.**
+  There are instructions as HTML comments.
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+- [ ] **Create a PR for this VEP.**
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the VEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+<!-- Provide table of content as it's helpful. -->
+
+- [VEP-2448: VDK Run Logs: Simplified And Readable](#vep-2448-vdk-run-logs-simplified-and-readable)
+  - [Summary](#summary)
+  - [Glossary](#glossary)
+  - [Motivation](#motivation)
+    - [Logs on failures](#logs-on-failures)
+    - [Logs on success](#logs-on-success)
+    - [Debug mode](#debug-mode)
+    - [Progress Tracking](#progress-tracking)
+    - [VDK exception handling and  categorization](#vdk-exception-handling-and--categorization)
+  - [Requirements and goals](#requirements-and-goals)
+  - [High-level design](#high-level-design)
+  - [API design](#api-design)
+  - [Detailed design](#detailed-design)
+  - [Implementation stories](#implementation-stories)
+  - [Alternatives](#alternatives)
+
+## Summary
+
+<!--
+Short summary of the proposal. It will be used as user-focused
+documentation such as release notes or a (customer facing) development roadmap.
+The tone and content of the `Summary` section should be
+useful for a wide audience.
+-->
+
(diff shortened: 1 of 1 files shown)


murphp15 (approved):

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -0,0 +1,244 @@
...
+<!--
+It tells **what** is it trying to achieve?
+List the specific goals (functional and nonfunctional requirements)? How will we
+know that this has succeeded?
+
+Specify non-goals. Clearly, the list of non-goals can't be exhaustive.
+Non-goals are only features, which a contributor can reasonably assume were a goal.
+One example is features that were cut during scoping.
+-->
+1. Data job run logs should keep the noise to a minimum and highlight useful information
```
tozka, line 128: Let's try to make them more specific and detailed. But not verbose - in the same time we should be concise. 
Here is where we enter the functional and nonfunctional requirements. 

That article seems to have some good ideas: https://olive.app/blog/how-to-write-a-good-requirement/

Could you take a look and try to see if you can rephrase them a bit?

tozka (changes requested):

DeltaMichael:
@tozka I've re-worded the requirements section and added a TODO to the Summary section. Please, review again.

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -0,0 +1,251 @@
...
+error, data error, library issue, or platform error, as very important
+visibility of the steps or components where failures occur so people can quickly
+find out the principal issue
+
+Logging and error handling should be improved in ways that adress all these
+challenges. This will result in less time spent troubleshooting, fewer support
+queries, more data job runs that are successful and an overall better user
+experience.
+
+## Requirements and goals
```
tozka, line 120: Now that I think about it, I am not sure we are 100% ready for the requirements section. It's something we write after we have some good idea on the solution and finalize the scope only. 

So let's revisit it later after the solution research?

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -0,0 +1,251 @@
...
+<!--
+It tells **what** is it trying to achieve?
+List the specific goals (functional and nonfunctional requirements)? How will we
+know that this has succeeded?
+
+Specify non-goals. Clearly, the list of non-goals can't be exhaustive.
+Non-goals are only features, which a contributor can reasonably assume were a goal.
+One example is features that were cut during scoping.
+-->
+1. Data job run logs provide useful and readable information
```
tozka, line 130: Avoid non specific and subjective words like `useful` in requirements and goals. Goals should be specific, measurable, achievable, relevant, timebound :) 

"readable" is also subjective. 

A bit more specific and measureable goal could be: Reduce Noise in Logs  by reducing log volume by about 75%. Of which 20% are garbage and should be deleted (like all the repeatable things we have) and  the rest of 55% are perhaps useful for further investigation and not a first thing user see , but a second step.
  
I think we need to consider "cloud" and "local" workflows separately also. 
 
 In Cloud workflows you need information like execution id, job name, attempt id, op id printed on each line so that you can search by them because they are mixed with a lot of other logs. 
 
 Locally you do not need that information logged on each line. 

In Both cases debug and verbose logs (especially stack traces) can be valuable and should not disappear but should be hidden (like in npm when something happens, it tells you to check detailed logs in file XXX.) 

What does hidden mean in cloud case, perhaps we need to A) allow users to filter them out easily (and by default they should be filtered out) B) allow users admin to control what is logged easier

In Local case, we can use the npm approach. 

Those are at least 2 use case where cloud and local workflow slightly differs. So we need to account for that in our requirements.

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -0,0 +1,251 @@
...
+   1. Data job run logs provide progress tracking information
+      1. current step
+      1. ingested data
+      1. the amount of resources currently consumed
+      1. progress bars for local runs
+   1. User logs are highlighted
```
tozka, line 136: What is "highlighted" ? You do not mean it literally I suppose.

tozka (approved):
Looks good to me. Thanks for driving this initiative. 

I suggest we finish the goals and requirements section after we do some research on the solutions.

That way we would know better what we want to do (what is in scope and and what is not). 
Depending on the cost of the different solutions some things may go out of scope because they are too costly or UX is too bad or some other reason.

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
DeltaMichael: VEPs should be improved in iterations, so we can absolutely revisit it. I expect our requirements to evolve as we do more research and brainstorming.

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
DeltaMichael: I meant they should stand out in some way. Not sure if there's a way to force color in terminal, but it's an option. I'll change it to "stand out".

PR #2466, opened 2023-07-25
versatile-data-kit: remove global image from Gitlab CI

tozka:
In GitLab CI, you can specify certain settings globally, affecting the entire CI/CD pipeline. These are called "global keys. There is a global key that defines a specific Docker image (image:)

Now, the problem arises when we define the same global key in multiple YAML files within the repository. Since global keys are not scoped to specific files, they can clash and cause conflicts. When GitLab processes the CI configuration, it reads all the YAML files and applies the settings in the order they are imported.

If the same global key appears in different files, the last occurrence of that key in the import order will overwrite the previous ones. This behavior can lead to unintended consequences. For example, we might encounter unexpected Docker images being used for certain jobs.

To prevent this from happening I am removing all global `image:` keys except in root level gitlab ci file to avoid any potential issues.

Specific images needed are set on job-level which is scoped to a job. This is the best practice that we should follow and not rely on defaults.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+# default image used by all pipelines if not overridden per job 
 image: "python:3.9"
 
 include:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+image: "python:3.9"
 
 include:
   - "cicd/.gitlab-ci-lib.yml"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # default image used by all pipelines if not overridden per job
+# it is better not to rely on default image and specify image explicitly for each job
 image: "python:3.9"
 
 include:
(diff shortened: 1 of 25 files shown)


murphp15 (approved):

.gitlab-ci.yml
```diff
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+image: "python:3.9"
```
murphp15, line 4: Can you add a comment here explaining that this will be used by all pipelines.

tozka, line 4:
```suggestion
# default image used by all pipelines if not overridden per job 
image: "python:3.9"
```

.gitlab-ci.yml
```diff
@@ -1,7 +1,8 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+# default image used by all pipelines if not overridden per job
+image: "python:3.9"
```
tozka, line 5:
```suggestion
# it is better not to rely on default image and specify image explicitly for each job
image: "python:3.9"
```

doks5 (approved):

projects/vdk-plugins/vdk-control-api-auth/.plugin-ci.yml
```diff
@@ -1,7 +1,7 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-image: "python:3.7"
+
```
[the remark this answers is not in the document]
tozka: Maybe not, but it was easier to do copy and replace this way .

PR #2470, opened 2023-07-25
control-service: prevent integer translation in helm chart

dakodakov:
Prevent integer translation to a fraction in helm chart.


tozka (approved):
I think that we need more robust testing of the helm chart in this repo

mivanov1988 (approved):

PR #2471, opened 2023-07-25
vdk-ipython: package ipython correctly

murphp15:
# Why
There is a bug in the current release of vdk-ipython. It is not including the files at the root of the source folder. 
# What 
Remove the line from the setup.py so files at the root are included. 
# How was this tested?
Packaged locally and all looks good
Steps:
1. run `python setup.py sdist --formats=gztar`
2. in the generated dist folder there will be a zipped file. uncompress this. 
3. In the uncompressed folder make sure the python file is present.


tozka:
The PR include changes not just in vdk-ipython. Maybe you need to rebase ? 

Is there related Github issue to the PR ? Can you describe in a bit more details the bug?

murphp15:
@tozka  rebased. sorry about that.

projects/vdk-plugins/vdk-ipython/setup.py
```diff
@@ -16,7 +16,6 @@
     long_description_content_type="text/markdown",
     install_requires=["vdk-core", "iPython"],
     package_dir={"": "src"},
-    packages=setuptools.find_namespace_packages(where="src"),
```
tozka, line 19: I don't understand the fix. Removing it seems like bad idea. We keep sources in "src" folder so we use setuptools.find_namespace_packages  to automatically find all the Python packages within the directory

murphp15:
@tozka there are no packages in that python project. There is just a single python file at the root of source.

projects/vdk-plugins/vdk-ipython/setup.py
tozka: Oh, that's not following the plugin templates. 

The source code should be in in "src" directory in package starting with vdk.plugin.<plugin-name>.... 

That's the structure of all other plugins . It's the one setup if a plugin is used using the cookiecutter - https://github.com/tozka/cookiecutter-vdk-plugin/tree/main

But I don't get the problems? Which files are not included? And why do we need them?

murphp15: vdk_ipython.py is at the root of the source folder. It is not in a package so if that line is in the setup.py then this file is dropped. 

@duyguHsnHsn  do you think we could put the plugin in a package or would it make it less usable for customers do you think?

tozka:
> Packaged locally and all looks good

I favour automated tests? Why isn't our Gitlab CI not catching the issue. 

But for any manual tests, please describe them in such a way that I (or anyone) can reproduce them. It's useful since 
A) I sometimes reproduce them to verify they work in another environment 
 B) in the future looking at previous commits if there was a manual tests I can reuse it for similar change

murphp15:
@duyguHsnHsn  do you have tests on the juypter UI, why did they not catch this issue?

projects/vdk-plugins/vdk-ipython/setup.py
tozka: I think I got the problem.

 `find_namespace_packages` doesn't include files that are not in a directory (aka package), and those that are at the root level specified are ignored. I find this a bit strange to be honest. And slightly problematic. 
 
 
The current approach you are doing would also have a namespace/package. If you unzip the tar you will see the file is inside a "src" folder . So a user would need to do 
```
from src import vdk_ipython
```

you can test it yourself by installing `pip install dist/vdk-ipython-0.1.0.tar.gz` 

The test pass because they assume "src" structure is kept and set src on the PythonPath and ignore setup.py

PR #2474, opened 2023-07-25
vdk-jupyter: vdk jupyter deploy job fix

duyguHsnHsn:
What:
Changed how deploy job operation is handled in Jupyter UI: 
1. The output of the operation is now changed to give deployment: 
Screenshot 2023-07-25 at 16 10 24

2. Removed enable flag from the Jupyter dialog: 

Screenshot 2023-07-25 at 16 46 53

Why:
1. Currently the deployment from Jupyter is not working correctly because of the new feature with multiple python versions. 

When we do deploy, we get message that the deployment was successful but it is not  actual and in the cli  where jupyter lab is run we get this:

Screenshot 2023-07-25 at 23 18 44

2.  The team agreed on removing the option and to work with the default enable=true since this may cause user confusions 


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -143,37 +143,36 @@ def create_job(name: str, team: str, path: str, local: bool, cloud: bool):
...
         :param path: the path to the job's directory
         :param reason: the reason of deployment
-        :param enabled: flag whether the job is enabled (that will basically un-pause the job)
         :return: output string of the operation
         """
+        import io
+        import sys
+
+        buffer = io.StringIO()
+        sys.stdout = buffer
```
tozka, line 159: Why is the redirect necesary. It's not very clearly explained in the PR description. And I need to understand that before I can review if this solution is would be suitable.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/__tests__/deploy-job-component.spec.ts
```diff
@@ -38,19 +38,6 @@ describe('#render()', () => {
   });
 });
 
-describe('#onEnableClick', () => {
```
tozka, line 41: You may want to just add a simple test or extend some assert /verification to ensure that the new default behavior works

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
tozka, line 159: Note that generally Redirecting sys.stdout is considered less than ideal practice and is saved for last resort. 

* sys.stdout is a global variable, and redirecting it affects the entire application. This can lead to unexpected results if different parts of the codebase are assuming different things about where stdout is going.
* In multi threads , which I think we might have a case here, redirecting stdout can cause problems, as threads might interfere with each other's output
 * If an error occurs after redirecting stdout but before it's reset back to its original value, well that would be an issue. 
  
  It's risky and you should be always careful when doing it. 
  
  In this case a better approach is to pass it another printer 
  
  You can create new printer easily : just a class decorated with `@printer` (as [seen here](https://github.com/vmware/versatile-data-kit/blob/ffdee8bac07e3718299eb10e024c7319179b3b93/projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py#L63) ) 
  
  And then you can set the output = "memory" for example (something that stores the results in memory instead of printing to stdout ) for example.

murphp15 (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: I have created a memory printer please check it

duyguHsnHsn: this is no more actual with the latest changes, I will resolve this conversation

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -143,37 +143,30 @@ def create_job(name: str, team: str, path: str, local: bool, cloud: bool):
...
-        return f"Job with name {name} and team {team} is deployed successfully!"
+        cmd.create(
+            name=name,
+            team=team,
+            job_path=path,
+            reason=reason,
+            vdk_version=None,
+            enabled=True,
+        )
+        result = cmd._JobDeploy__printer.get_memory()
```
tozka, line 165: You are accessing a private field. You should not do that. 

Python has the concept of name mangling precisely to remind developers that hey this is private - if they accidentally do `cmd.__printer` , they'd get an error

So 
A) Either make the attribute public
B) Employ dependency injection/inversion of control (pass the Output printer class)

I vote for B.

tozka (changes requested):

projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
```diff
@@ -130,3 +161,4 @@ class OutputFormat(str, Enum):
 
     TEXT = "TEXT"
     JSON = "JSON"
+    FILE = "MEMORY"
```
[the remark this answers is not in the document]
duyguHsnHsn: Yeah, I was firstly trying to print within temp file it remained from that i will change it quickly

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: this problem will be solved in https://github.com/vmware/versatile-data-kit/pull/2477 and all the unnecessary changes in this pr will be removed

tozka (approved):

PR #2475, opened 2023-07-25
control-service: remove default vault token

dakodakov:
Remove the default value of the vault integration token, to reduce confusion when configuring vault.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
+++ b/projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
@@ -271,5 +271,5 @@ featureflag.vault.integration.enabled=false
 vdk.vault.uri=http://localhost:8200/v1/
 vdk.vault.approle.roleid=
 vdk.vault.approle.secretid=
-vdk.vault.token=root
+vdk.vault.token=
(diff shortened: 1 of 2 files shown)


doks5 (approved):

mivanov1988 (approved):

mivanov1988 (approved):

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -271,5 +271,5 @@ featureflag.vault.integration.enabled=false
 vdk.vault.uri=http://localhost:8200/v1/
 vdk.vault.approle.roleid=
 vdk.vault.approle.secretid=
-vdk.vault.token=root
+vdk.vault.token=
```
tozka, line 274: nit: maybe we should consider seting roleid/secretid with higher priority? 

 if roleID and secretID is not empty then use it else use if token is not empty use token, else error? or just nothing (can vault be used without auth, I doubt that)

That would prevent in the future some developer from changing this back to root for convenience and causing problems.

dakodakov: fixed.

PR #2477, opened 2023-07-26
vdk-control-cli: pass the printer class in JobDeploy creation and add additional memory printer

duyguHsnHsn:
What:Introduced an enhanced feature, the MemoryPrinter, it conserves text for future use, instead of instant printing. Along with this, adapted the JobDeploy class to receive a printer object instead of an OutputFormat enum.

Why:
The inclusion of MemoryPrinter caters to use cases where an output needs to be stored for later representation, such as in a Jupyter UI. This modification in the JobDeploy class enables the printer object's use beyond JobDeploy: the printer can be initialised first, utilised within JobDeploy, and then accessed again to retrieve the stored output. This stops the need for accessing private fields within JobDeploy and promotes better coding practices.

### DIFF
--- a/projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
+++ b/projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
@@ -11,6 +11,7 @@ from typing import Optional
 
 import click
 from tabulate import tabulate
+import io
 
 
 class Printer(abc.ABC):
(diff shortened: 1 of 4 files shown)


projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
```diff
@@ -107,6 +108,40 @@ def print_dict(self, data: Dict[str, Any]) -> None:
             click.echo("{}")
 
 
+class MemoryPrinter(Printer):
+    output_buffer = io.StringIO()
```
tozka, line 112: output_buffer is currently a class attribute (a static variable) which makes it shared across all instances of the class. It might be better to move it to an instance attribute so that each instance of MemoryPrinter has its own buffer. 
 
 In Python, this happens by defining a constructor using` __init__` 
 
 Read over https://docs.python.org/3/tutorial/classes.html

projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
```diff
@@ -107,6 +108,40 @@ def print_dict(self, data: Dict[str, Any]) -> None:
             click.echo("{}")
 
 
+class MemoryPrinter(Printer):
```
tozka, line 111: Maybe named it InMemoryTextPrinter 

Because the format is still in text

duyguHsnHsn: yes, sure

projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
duyguHsnHsn: fixed

projects/vdk-control-cli/src/vdk/internal/control/utils/output_printer.py
```diff
@@ -109,16 +109,17 @@ def print_dict(self, data: Dict[str, Any]) -> None:
 
 
 class InMemoryTextPrinter(Printer):
-    output_buffer = io.StringIO()
+    def __init__(self):
+        self.output_buffer = io.StringIO()
```
tozka, line 113: Make this private : https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#clarification-on-access-modifiers-in-python

tozka (approved):

yonitoo (approved):
LGTM!

Issue #2467, opened 2023-07-25
VDK Plugins tests - migrate to testcontainers from pytest-docker


antoniivanov:
**Describe the bug**
 Because of https://github.com/vmware/versatile-data-kit/pull/2445 we cannot use PyYAML>=5.4 with docker-compose in tests.
So we had to downgrade to 5.3.1 to be able to run the tests sucessfully
 
Downgrading to PyYAML 5.3.1 trigger security alert https://github.com/vmware/versatile-data-kit/security/dependabot/21 
 
And there is no version between 5.3.1 and 5.4 

**Suggested solution** 

I would suggest that we switch to testcontainers entirely. They are some advantages of using testcontainers over pytest-docker as explained when we used them to stabilize vdk-trino tests in https://github.com/vmware/versatile-data-kit/pull/1677 

**Additional context**
Add any other context about the problem here. Consider attaching logs if there are any.

PR #2503, opened 2023-07-31
vdk-impala, vdk-postgres: adopt testcontainers

antoniivanov:
See #2467 for details


doks5 (approved):

projects/vdk-plugins/vdk-postgres/tests/conftest.py
```diff
@@ -44,17 +42,34 @@ def docker_compose_file(pytestconfig):
...
+        runner = CliEntryBasedTestRunner(postgres_plugin)
+        wait_for(partial(_is_responsive, runner))
+        # wait 2 seconds to make sure the service is up and responsive
+        # might be unnecessary but it's out of abundance of caution
+        time.sleep(2)
+        print(
+            f"Postgres service started on port {container.get_exposed_port(port)} and host {container.get_container_host_ip()}"
+        )
+    except Exception as e:
+        print(f"Failed to start Postgres service: {e}")
```
murphp15, line 65: I don't like the way you are logging the exception and also throwing it. 
This always results in lots of noise in our code.

murphp15: I think its best to create a new exception which includes the logs. 
I think its best to remove  `print(f"Failed to start Postgres service: {e}")`

antoniivanov: Make sense.

projects/vdk-plugins/vdk-impala/tests/conftest.py
```diff
@@ -16,18 +20,13 @@
...
-    except:
-        return False
-
-
-@pytest.fixture(scope="session")
-def docker_compose_file(pytestconfig):
-    return os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "docker-compose.yml"
-    )
+        result: Result = runner.invoke(["impala-query", "--query", "SELECT 1"])
```
murphp15, line 23: do you really need the try catch here? 
Would be cleaner to just rmeove it and you aren't really adding much by wrapping it.

projects/vdk-plugins/vdk-postgres/tests/conftest.py
```diff
@@ -19,17 +22,12 @@
...
-    except:
-        return False
-
-
-@pytest.fixture(scope="session")
-def docker_compose_file(pytestconfig):
-    return os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "docker-compose.yml"
-    )
+    except Exception as e:
```
murphp15, line 25: Same applies here. I think its cleaner if we remove the try catch.

projects/vdk-plugins/vdk-impala/tests/conftest.py
```diff
@@ -39,18 +38,31 @@ def docker_compose_file(pytestconfig):
...
+        )
+    except Exception as e:
+        raise RuntimeError(
+            f"Failed to start Impala service: {e}\n"
+            f"Container logs: {container.get_logs()}"
+        ) from e
+
+    def stop_container():
+        container.stop()
+        print("Impala service stopped")
```
DeltaMichael, line 64: You should log this instead of printing.

projects/vdk-plugins/vdk-impala/tests/conftest.py
```diff
@@ -39,18 +38,31 @@ def docker_compose_file(pytestconfig):
...
+    container = DockerContainer("cpcloud86/impala:latest").with_bind_ports(port, port)
+
+    try:
+        container.start()
+        runner = CliEntryBasedTestRunner(impala_plugin)
+        wait_for(partial(_is_responsive, runner))
+        # wait 2 seconds to make sure the service is up and responsive
+        # might be unnecessary but it's out of abundance of caution
+        time.sleep(2)
+        print(
```
DeltaMichael, line 53: You should log this instead of printing.

antoniivanov: Yes, I agree. But as I am refactoring code. I am not 100% sure why we are using print instead of log. There should be a good reason.  I need to research that and would do that in a separate PR.

projects/vdk-plugins/vdk-impala/tests/conftest.py
antoniivanov: Yes, because I need to re-wrap it in ConnectionError or the setup would fail (the wait_for waits only for ConnectionError and TimeOut error)

projects/vdk-plugins/vdk-postgres/tests/conftest.py
antoniivanov: Same answer as above. Let me see if I can re-write it to be clearer. Maybe add a coment

projects/vdk-plugins/vdk-impala/tests/conftest.py
antoniivanov: I figured alternative way. I think it's more readable as well this way.

murphp15 (approved):

PR #2506, opened 2023-08-01
vdk-jupyter: automate create options

duyguHsnHsn:
What:
Changed create job option to create local job automatically and cloud if REST_API_URL is set. 

Currently there are two options for created job result: 
Created both on cloud and locally: 
Screenshot 2023-08-08 at 14 46 03

Created locally only because of lack of url: 
Screenshot 2023-08-08 at 14 42 32
Screenshot 2023-08-08 at 14 42 42

The dialog for create now looks like this: 
Screenshot 2023-08-04 at 12 18 13

Why:
Currently, when the user wants to create a job from the Jupyter UI he gets dialog asking whether it will be a local job, a cloud job or both. We discussed and got on an agreement that the options should be removed from the dialog. 

Creating job should be: 
- always creating locally
- creating in the cloud only if REST_API_URL is set 


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -118,30 +118,35 @@ def download_job(name: str, team: str, path: str):
...
         )
-
-        cmd.create_job(name, team, path, cloud, local, jupyter_job_dir)
-
-        return f"Job with name {name} was created."
+        try:
+            cmd = JobCreate(RestApiUrlConfiguration.get_rest_api_url())
+            cli_utils.check_rest_api_url(RestApiUrlConfiguration.get_rest_api_url())
+            cmd.validate_job_path(path, name)
+            cmd.create_job(name, team, path, True, True, pathlib.Path(jupyter_job_dir))
```
DeltaMichael, line 136: This doesn't seem right. You removed the boolean arguments from `create_job`, right? It should take 3 arguments, but here, you're passing in 6.

DeltaMichael (approved):
I disagree with the concept. This is not a good user experience. What if I want to create a job just locally, but I was creating cloud jobs previously. Do I have to unset the env variable and then set it again to get back to what I was doing?

I wasn't part of the discussions, however, which is completely my fault. I'll approve and we can possibly raise it again down the line.

antoniivanov:
> I disagree with the concept. This is not a good user experience. What if I want to create a job just locally, but I was creating cloud jobs previously. Do I have to unset the env variable and then set it again to get back to what I was doing?
> 
> I wasn't part of the discussions, however, which is completely my fault. I'll approve and we can possibly raise it again down the line.

The concept of local and cloud doesn't really exist in the user's mind. It's something we are forcing on them. And we should not really. The fewer things the user needs to know and understand the better their experience would be.

As far as the user is concerned a job is a job. They want to be able to create it and develop it and when it's ready schedule it /deploy it so it runs regularly and monitor. 

The less things you make the user think about the better their life will be. Because they do not know (nor need to care about the concept of local vs cloud), they would not really care if job is created in the cloud (by created we really mean the name is reserved in most cases).

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -118,30 +118,35 @@ def download_job(name: str, team: str, path: str):
...
             os.path.join(os.path.dirname(__file__), "jupyter_sample_job")
         )
-
-        cmd.create_job(name, team, path, cloud, local, jupyter_job_dir)
-
-        return f"Job with name {name} was created."
+        try:
+            cmd = JobCreate(RestApiUrlConfiguration.get_rest_api_url())
+            cli_utils.check_rest_api_url(RestApiUrlConfiguration.get_rest_api_url())
+            cmd.validate_job_path(path, name)
```
antoniivanov, line 135: That's redundant , JobCreate already validates job path

antoniivanov (changes requested):

DeltaMichael:
> The less things you make the user think about the better their life will be. Because they do not know (nor need to care about the concept of local vs cloud), they would not really care if job is created in the cloud (by created we really mean the name is reserved in most cases).

I disagree, because we're actually making the user think about more things. We're making them think about an environment variable that also has a side effect when set. It determines both the control service URL and **if** we talk to the control service.

Let's say the user really doesn't need to know about that. Then they'll create a bunch of local jobs they have no intention of deploying, look at their cloud environment and discover that it's filled with garbage entries. Then they'll wonder why and will have to think about the env variable. It's quite a confusing user experience.

Now compare that to a checkbox that says "Cloud".

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: this is a class defined in the cli which accepts cloud and local flags 
in the jupyter extension we just make use of it so i still need to pass the flags accordingly

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -118,30 +118,35 @@ def download_job(name: str, team: str, path: str):
         return f"Downloaded the job with name {name} to {path}. "
 
     @staticmethod
-    def create_job(name: str, team: str, path: str, local: bool, cloud: bool):
+    def create_job(name: str, team: str, path: str):
```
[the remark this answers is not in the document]
duyguHsnHsn: fixed

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: fixed

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -118,30 +118,35 @@ def download_job(name: str, team: str, path: str):
...
+            result = f"Job with name {name} was created locally and in the cloud."
+        except ValueError as e:
+            cmd = JobCreate("")
+            cmd.validate_job_path(path, name)
+            cmd.create_job(name, team, path, False, True, pathlib.Path(jupyter_job_dir))
+            result = (
+                f"Job with name {name} was created only locally. "
+                f"If you are not using the Control Service the next lines should not concern you!"
+                f"We tried to create it in the cloud but come up to:"
+                f"{e}"
```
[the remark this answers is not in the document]
duyguHsnHsn: Addressed https://github.com/vmware/versatile-data-kit/issues/2535

PR #2510, opened 2023-08-02
vdk-sqlite: support whitespace in column names

DeltaMichael:
## Why?

Trying to create columns with whitespace in the name and inserting data in them using vdk-sqlite resulted in error.

This was most evident in examples like csv-ingestion

https://github.com/vmware/versatile-data-kit/wiki/Ingesting-local-CSV-file-into-Database

## What?

- Add double quotes around column names with whitespace in them when the table is created.
- Replace named placeholders in insert query with question marks
- Convert payload from dicts to lists to pass it to above query
- Check if payload headers match column names before it's passed

## How has this been tested?

Unit tests

## What type of change is this?

Bugfix

### DIFF
--- a/projects/vdk-plugins/vdk-sqlite/src/vdk/plugin/sqlite/ingest_to_sqlite.py
+++ b/projects/vdk-plugins/vdk-sqlite/src/vdk/plugin/sqlite/ingest_to_sqlite.py
@@ -89,30 +89,13 @@ class IngestToSQLite(IIngesterPlugin):
     def __ingest_payload(
         self, destination_table: str, payload: List[dict], cur: Cursor
     ) -> None:
-        fields, query = self.__create_query(destination_table, cur)
-
-        for obj in payload:
+        values, query = self.__create_query(destination_table, payload, cur)
+        for obj in values:
             try:
                 cur.execute(query, obj)
                 log.debug("Payload was ingested.")
             except Exception as e:
-                if collections.Counter(fields) != collections.Counter(obj.keys()):
-                    errors.log_and_rethrow(
-                        errors.ResolvableBy.USER_ERROR,
-                        log,
-                        "Failed to sent payload",
-                        f"""
-                        One or more column names in the input data did NOT
-                        match corresponding column names in the database.
-                           Input Table Columns: {list(obj.keys())}
-                        Database Table Columns: {fields}
-                        """,
-                        "Will not be able to send the payload for ingestion",
-                        "See error message for help ",
-                        e,
-                        wrap_in_vdk_error=True,
-                    )
-                elif isinstance(e, ProgrammingError):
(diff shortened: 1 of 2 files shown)


projects/vdk-plugins/vdk-sqlite/src/vdk/plugin/sqlite/ingest_to_sqlite.py
```diff
@@ -198,12 +202,16 @@ def __create_table_if_not_exists(
...
         """
-        Creates table from give list of columns and table name
+        Creates table from given list of columns and table name
         """
-        columns_as_sql_expression = ",".join(
-            [f"{col_name} {col_type}" for col_name, col_type in columns.items()]
-        )
-        sql = f"CREATE TABLE IF NOT EXISTS {destination_table} ( {columns_as_sql_expression} )"
+        names = [
+            f"{col_name} {col_type}"
```
murphp15, line 208: why not just quote everything?

projects/vdk-plugins/vdk-sqlite/src/vdk/plugin/sqlite/ingest_to_sqlite.py
```diff
@@ -169,18 +152,39 @@ def __table_columns(
...
+                    Database Table Columns: {fields}
+                    """,
+                    "Will not be able to send the payload for ingestion",
+                    "See error message for help ",
+                )
+
+        # the query fstring evaluates to 'INSERT INTO dest_table (val1, val2, val3) VALUES (?, ?, ?)'
+        # assuming dest_table is the destination_table and val1, val2, val3 are the fields of that table
+        values = [[obj.get(field) for field in fields] for obj in payload]
+        fields = [field if " " not in field else f'"{field}"' for field in fields]
```
murphp15, line 185: again is it not easier just to quote everything?

murphp15 (approved):

projects/vdk-plugins/vdk-sqlite/src/vdk/plugin/sqlite/ingest_to_sqlite.py
DeltaMichael: https://www.sqlite.org/draft/quirks.html -> Point 8

> The SQL standard requires double-quotes around identifiers and single-quotes around string literals. For example:
"this is a legal SQL column name"
'this is an SQL string literal'
SQLite accepts both of the above. But, in an effort to be compatible with MySQL 3.x (which was one of the most widely used RDBMSes when SQLite was first being designed) SQLite will also interpret a double-quotes string as string literal if it does not match any valid identifier.

I'd like to avoid this behaviour on the off chance we get it somehow. It would be extremely hard to debug.

PR #2511, opened 2023-08-02
vdk-impala: Handle errors on refresh/invalidate metadata

doks5:
Currently, if a data job executes a query, which fails due to metadata issue, the ImpalaErrorHandler
logic tries to refresh or invalidate the metadata, and re-try the query. If the job has only read
access to the table, an authorization error is raised and the job failes with User Error. This should
not happen, as the refresh/invalidate operation is initiated by vdk, not the user.
    
This change updates the error handler logic by adding try/except blocks around refresh/invalidate
metadata operations, so that errors that arise from these operations do not cause job failures.
    
Testing Done: Added test

### DIFF
--- a/projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
+++ b/projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
@@ -151,13 +151,21 @@ class ImpalaErrorHandler:
                 )
                 # Try refreshing the table metadata several times,
                 # and if the issue is not fixed invalidate the metadata.
-                if recovery_cursor.get_retries() > 3:
-                    recovery_cursor.execute(
-                        "invalidate metadata `" + database + "`.`" + table + "`"
-                    )
-                else:
-                    recovery_cursor.execute(
-                        "refresh `" + database + "`.`" + table + "`"
+                try:
+                    if recovery_cursor.get_retries() > 3:
+                        recovery_cursor.execute(
+                            "invalidate metadata `" + database + "`.`" + table + "`"
+                        )
+                    else:
+                        recovery_cursor.execute(
+                            "refresh `" + database + "`.`" + table + "`"
+                        )
+                        time.sleep(
+                            2 ** recovery_cursor.get_retries() * self._backoff_seconds
+                        )  # exponential backoff 30s, 60s, 2m, 4m, 8m
+                except Exception as e:
+                    self._log.info(
+                        f"Refresh/Invalidate metadata operation failed with error: {e}"
                     )
                     time.sleep(
                         2 ** recovery_cursor.get_retries() * self._backoff_seconds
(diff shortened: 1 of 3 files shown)


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -228,6 +228,12 @@ def _handle_metadata_exception_with_backoff(
                 classname_with_package="impala.error.OperationalError",
                 exception_message_matcher_regex=".*ImpalaRuntimeException: Error making 'updateTableColumnStatistics'.*",
             )
+            or errors.exception_matches(
+                exception,
+                classname_with_package="impala.error.HiveServer2Error",
+                exception_message_matcher_regex="AuthorizationException: User .*does not have privileges to "
```
antoniivanov, line 234: The issue with Impala sometimes taking time to propagate new authorization settings and rasing AuthorizationException is not related to invalidate metadata /refresh . 

but to any query. So why not re-try on any query?

I

antoniivanov:
I do not fully comprehend the problem and it clashes of my  (independent) understanding of the overall problem. 

One problem is 

A)Impala is eventually consistent and sometimes with raise AuthorizationException even if job user has been authorized recently. So re-trying with back-off make sense while waiting for new authorization rules to sync 

B)Impala recovery mechanism when querying a table and it detects that may it's out of sync, tries to execute invalidate metadata on that table before -retrying the original query. But if it doesn't have write permission to that table, invalidate metadata fails . No amount of re-trying would prevent that since the permission would not change. It still would not have write permission.  

This seems like a solution for A) but the description seems more like B) So I am confused. 

Is there option C I am not seeing ?

doks5:
> I do not fully comprehend the problem and it clashes of my (independent) understanding of the overall problem.
> 
> Related problems I know about are
> 
> A)Impala is eventually consistent and sometimes raises AuthorizationException even if the data job user has been authorized recently. So re-trying with back-off makes sense while waiting for new authorization rules to sync
> 
> B)Impala recovery mechanism when querying a table and it detects that the table metadata is out of sync, tries to execute invalidate metadata on that table (to sync it) before retrying the original query. But if it doesn't have "write" permission to that table, invalidate metadata fails. No amount of re-trying would prevent that since the permission would not change. It still would not have "write" permission.
> 
> The code seems like a solution for A) but the description seems more like B) So I am confused.
> 
> Is there option C I am not seeing ?

Hi, @antoniivanov 

When an authorization error is raised by impala, the message gives indication what type of operation has been attempted.
The case that I'm trying to attempt at solving is when a data job, that has been granted only `read` access to a table, executes a `SELECT * FROM some_table`, the query fails with metadata inconsistency error (maybe some other job which has write access has modified the table just before the "select" query was issued), and the error handling tries to `refresh/invalidate` the metadata. Because the job has only read access, the `refresh/invalidate` query would fail with authorization error and trigger a job failure. By just sleeping for a bit and re-trying the initial `select` query, we can avoid job failures if, in the meantime, the metadata has managed to sync across the impala coordinators.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
doks5: Refactored, as discussed offline

doks5: It was agreed that the previous change to single out `Authorization errors` on refresh/invalidate metadata and handle them, was too limiting, and instead a different approach is needed. The refresh/invalidate metadata queries are issued by the error handling logic, not by the user, therefore we need to swallow these errors, and just retry the initial query. This way, if the initial query fails again, it would be somewhat expected, and the log message for the error would make more sense to the user, instead of receiving a notification for an error related to refresh/invalidate metadata, which would not make any sense.

doks5:
Hi, @antoniivanov 

I've addressed the feedback and updated the comments and PR description.

antoniivanov (approved):

antoniivanov:
Looks good to me. Thanks. 

I updated the title. The prefix is just the name of the plugin (vdk-impala) without [vdk-plugins]

dakodakov (approved):

PR #2514, opened 2023-08-03
vdk-control-cli: add --set-prompt option for secrets

dakodakov:
Add the --set-prompt option for the secrets command to allow users to set secrets values via a prompt.


yonitoo (approved):
LGTM!

murphp15 (approved):

antoniivanov (commented):
If we have `vdk secrets --set-prompt` I think we should have `vdk properties --set-prompt`

Because I think they should be consistent interfaces

But also Properties can be used to save sensitive data (if secrets is disabled or if sensitive information is state (e.g user ids) and need to be updated often or some other case)> So it make sense to have it.

PR #2516, opened 2023-08-03
vdk-jupyter: add first version of getting started page

duyguHsnHsn:
What: Created a "Getting Started" page for VDK Jupyter integration.

Why: 
A "Getting Started" page serves as a foundational guide for newcomers. By providing clear, step-by-step instructions, we ensure that users can efficiently navigate and leverage the VDK Jupyter integration right from the outset, minimizing potential roadblocks or uncertainties.

Note: You can find the diagrams here: https://miro.com/app/board/uXjVMwyYjto=/?share_link_id=860687647726


yonitoo:
Would it be possible to add some images here?

antoniivanov:
To reduce the image sizes: 
* Reduce the size of each image. Do we need 1900x1000 image? Considering it's part of a notebook where we have only xbyY size anyway. Maybe approximately 1000x500 would be enough 
* JPEG (instead of png) format  would generally result in smaller images . Alternatively, use indexed PNG . Also I personally have not really used WebP but it's a modern format that may provide more optimal images. 
* Alternatively SVG might be better if feasible. 
* Consider gifs - which would allow you to combine multiple images in one animation. That may be better user experience ? 
* There are some websites to optimize images like https://compressor.io/,  https://tinypng.com/

murphp15:
Can you update the images so they don't say locally and in the cloud.

antoniivanov:
How were the images created? 

It is **very** important that anyone can edit those images after you without any help. 

So somewhere (it may be in a README.md) we should have clear instructions and preferably either a link or a source code depending on the tool used which would allow someone else to make changes.

duyguHsnHsn:
> 

I have created a miro board i will include it in the PR when it is ready

sabadzhiev (commented):
Duygu, 

Looks great. Putting me feedback here(both notes bellow are minor though):

#1: Typo: 'At the core of it, it is a directory with different scripts and inside of it.'
#2: There are a couple of places like: 'WARNING! If this is your initial e....'
This sound more like a tip to me, than a warning. May be, change it to: 'TIP: If this is your initial e....' ... or something not that scaring as 'WARNING' :slightly_smiling_face: .

projects/vdk-plugins/vdk-jupyter/getting-started.ipynb
```diff
@@ -0,0 +1,292 @@
...
+    "\n",
+    "   <p>VDK supports having many Python, SQL and Notebook steps in a single Data Job. These steps are carried out in an ascending alphabetical sequence based on the file names, and Notebook steps are run from the beginning to the end of the file. For an effective execution order while retaining informative file names, it's beneficial to prefix your file names with numbers.</p>\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eb565197",
+   "metadata": {},
+   "source": [
+    "<h2>Step 0: Explore VDK's Functionalities</h2>\n",
```
antoniivanov, line 57: We can skip that. People would not be using CLI so !vdk --help would be just confusing since most of functionality is duplicated through the UI or the notebook

antoniivanov (approved):
Looks pretty good to me. We can remove the CLI reference in this getting started

projects/vdk-plugins/vdk-jupyter/getting-started.ipynb
```diff
@@ -0,0 +1,292 @@
...
+    "This guide will delve into each of these options for a comprehensive understanding.\n",
+    "We'll also provide you with information on utilizing VDK within Notebooks.</p>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3e038188",
+   "metadata": {},
+   "source": [
+    "<h2>Step 1 Login </h2>\n",
```
antoniivanov, line 86: Radost suggested that we should add something like the workflow expected almost as a table of content 

E.g Login -> Create 
                 -> Download. -> Develop -> Run -> Deploy -> ...
                 
And people can click on it and go to that section

gageorgiev (approved):

PR #2518, opened 2023-08-03
vdk-control-api-auth: add get_authenticated_username

antoniivanov:
Add to Authentication library a method to get_authenticated_username.

The purpose of this method is ot be used during debugging , in logging and in telemetry use cases.

Testing Done: unit tests

### DIFF
--- a/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
+++ b/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
@@ -53,7 +53,7 @@ class Authentication:
             A flag, indicating if credentials should be cached locally (in a
             file).
         :param possible_jwt_user_keys:
-            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used.
+            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used. It is a list of keys where the first existing key in a JTW token is returned.
             Defaults to some common user keys.
         """
         self._username = username
--- a/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
+++ b/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
@@ -1,10 +1,16 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
+import logging
+from typing import List
+from typing import Optional
+
 from vdk.plugin.control_api_auth.auth_config import InMemAuthConfiguration
 from vdk.plugin.control_api_auth.auth_exception import VDKInvalidAuthParamError
 from vdk.plugin.control_api_auth.autorization_code_auth import RedirectAuthentication
 from vdk.plugin.control_api_auth.base_auth import BaseAuth
 from vdk.plugin.control_api_auth.login_types import LoginTypes
+
+log = logging.getLogger(__name__)
 
 
 class Authentication:
@@ -21,6 +27,7 @@ class Authentication:
         auth_discovery_url: str = None,
         auth_type: str = None,
         cache_locally: bool = False,
+        possible_jwt_user_keys=None,
     ):
         """
         :param username: A user's username in case basic authentication is used.
@@ -45,6 +52,9 @@ class Authentication:
         :param cache_locally:
             A flag, indicating if credentials should be cached locally (in a
             file).
+        :param possible_jwt_user_keys:
+            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used.
+            Defaults to some common user keys.
         """
         self._username = username
         self._password = password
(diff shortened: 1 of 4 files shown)


projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -90,6 +95,39 @@ def authenticate(self) -> None:
...
+            return None
+
+        try:
+            import jwt
+
+            jwt_payload = jwt.decode(self._token, options={"verify_signature": False})
+            if not jwt_payload:
+                return None
+
+            possible_user_keys = ["username", "acct", "preferred_username", "email"]
```
doks5, line 119: Do we want these to be configurable, or we don't expect them to change from one user installation to another?

doks5 (approved):

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -45,6 +52,9 @@ def __init__(
         :param cache_locally:
             A flag, indicating if credentials should be cached locally (in a
             file).
+        :param possible_jwt_user_keys:
+            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used.
```
antoniivanov, line 56:
```suggestion
            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used. It is a list of keys where the first existing key in a JTW token is returned.
```

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -45,6 +52,9 @@ def __init__(
         :param cache_locally:
             A flag, indicating if credentials should be cached locally (in a
             file).
+        :param possible_jwt_user_keys:
+            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used. It is a list of keys where the first existing key in a JTW token is returned.
```
antoniivanov, line 56:
```suggestion
            Used by get_authenticated_username to try to discover correct username if OAuth2 and JWT token is used. It is a list of keys where the first existing key in a JWT token is returned.
```

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -21,6 +27,7 @@ def __init__(
         auth_discovery_url: str = None,
         auth_type: str = None,
         cache_locally: bool = False,
+        possible_jwt_user_keys=None,
```
murphp15, line 30: I would remove unlesss there is a very definite use case for it today.
YAGNI.

murphp15: I still stand by my point here, but also happy to approve!

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
```diff
@@ -90,6 +109,38 @@ def authenticate(self) -> None:
...
+            jwt_payload = jwt.decode(self._token, options={"verify_signature": False})
+            if not jwt_payload:
+                return None
+
+            for key in self._possible_jwt_user_keys:
+                user_id = jwt_payload.get(key)
+                if user_id:
+                    return user_id
+
+        except Exception as e:
```
murphp15, line 138: If a JWT token can't be decoded should we really catch it?
I would actually want to see this error instead of getting none back.

murphp15: Given that customers are very unlikely to have debug enabled this could be a difficult issue to find.

murphp15: Oh I see that it is used for logging and telemetry and probably not worth bubbling an exception for that.

projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/authentication.py
antoniivanov: In the spirit of simplicity and avoiding overengineering, your approach is perhaps cleaner. 

But given the context, when dealing with external systems or components that can have variable behavior (like different authentication providers with different JWT token structures), building in some level of flexibility is almost mandatory if you want to provide robust solution.

projects/vdk-plugins/vdk-control-api-auth/tests/test_get_auth_user.py
```diff
@@ -0,0 +1,49 @@
...
+    assert auth.get_authenticated_username() is None
+
+
+@patch("jwt.decode")
+def test_get_authenticated_username_priority_order_kept(mock_jwt_decode):
+    mock_jwt_decode.return_value = {"sub": "user_from_sub", "acct": "user_from_acct"}
+    auth = Authentication(
+        token="testtoken", possible_jwt_user_keys=["user", "email", "sub", "acct"]
+    )
+    assert auth.get_authenticated_username() == "user_from_sub2"
```
antoniivanov, line 44:
```suggestion
    assert auth.get_authenticated_username() == "user_from_sub"
```

murphp15 (approved):

PR #2522, opened 2023-08-07
vdk-jupyter: add download-key to the VDK menu

yonitoo:
What:
Add a new option in the VDK dropdown menu that enables you to download keytab of a Cloud-deployed data job.

Why: 
Because this functionality is present in the CLI but not in the Jupyter UI.

Testing Done: 
Added tests and all, including the existing ones, pass.

VDK Menu

Download Key Dialog

Success message


duyguHsnHsn:
Can you add the result dialog of the operation, as well

duyguHsnHsn (approved):
Seems good to me!

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/DownloadKey.tsx
```diff
@@ -0,0 +1,61 @@
...
+          label="Job Name:"
+        ></VDKTextInput>
+        <VDKTextInput
+          option={VdkOption.TEAM}
+          value="default-team"
+          label="Job Team:"
+        ></VDKTextInput>
+        <VDKTextInput
+          option={VdkOption.PATH}
+          value={this.props.jobPath}
```
antoniivanov, line 40: The key should not be downloaded inside the job directory but alongside it. 

```
/root
|-- job-dir/
|   |-- config.ini
|   |-- ...
|-- job-key

 ```

yonitoo: changed this to the parent directory

murphp15:
@yonitoo can you update the picture for the download key dialog now that the folder is changed.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/DownloadKey.tsx
murphp15: @antoniivanov  if they save it somewhere else will it use it? 
Or does it need to be saved exactly there? 
maybe they should not be able to choose where to save it to. 
Make it simpler for customers.

antoniivanov: If they save it somewhere else , they could configure it (using [config described here](https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-plugins/vdk-kerberos-auth#configuration) which nobody ever does . By large vdk download-key CLI downloads it in the right location . If for some reason it is not they would move it.

yonitoo:
Updated the screenshot!

murphp15:
IMO, If we take the approach of building a press release first and then build the product after this dialog should 100% not be present!

PR #2524, opened 2023-08-07
control-service: support for pyodbc

mivanov1988:
Why
Currently, the base secure image does not contain the required libraries for `pyodbc` Python package.

What
Installed required native dependencies.

Testing Done
Locally through the following command:

```
docker buildx build -f versatile-data-kit/projects/control-service/projects/job-builder-secure/Dockerfile.python.vdk -t test-image-tag . --build-arg job_githash=localhsh1 --build-arg base_image=registry.hub.docker.com/versatiledatakit/data-job-base-python-3.8-secure:latest --build-arg job_name="data_job_name" --progress=plain
```


dakodakov (approved):

mivanov1988:
> Does the normal image contains unixodbc?

I think It does not.

antoniivanov:
> > Does the normal image contains unixodbc?
> 
> I think It does not.

Then do we need that change there? 

Also how does it work? In Super Collider if I am not mistake the normal image is still used.

PR #2538, opened 2023-08-09
vdk-jupyter: add icons to vdk operation result dialogs

yonitoo:
What: 
Add tick/X icons to resulting dialog messages.

Why: 
Make success/error messages more user-friendly.

Path error on deployment run
Path error 2
Screenshot 2023-08-09 at 12 54 00
Screenshot 2023-08-09 at 12 53 49
Screenshot 2023-08-09 at 12 53 41


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx
```diff
@@ -120,18 +122,36 @@ function add_command(
...
         } else {
-          await showErrorMessage(
-            'Another VDK operation is currently running!',
-            'Please wait until the operation ends!',
-            [Dialog.okButton()]
-          );
+          await showDialog({
+            title: 'Error',
+            body: (
+              <div className="vdk-error-dialog">
```
antoniivanov, line 128: The same error dialog structure is repeated in multiple places. It's better if we create a reusable React component for displaying errors

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/DeployJob.tsx
```diff
@@ -84,19 +85,39 @@ export async function showCreateDeploymentDialog(statusButton?: StatusButton) {
...
         } else {
-          await showErrorMessage(
-            'Encоuntered an error while running the job!',
-            message,
-            [Dialog.okButton()]
-          );
+          await showDialog({
+            title: CREATE_DEP_BUTTON_LABEL,
+            body: (
+              <div className="vdk-error-dialog">
```
antoniivanov, line 91: Be careful of magic strings. If you are using a string in multiple places, consider putting it as a constant

antoniivanov (changes requested):

gageorgiev (approved):
I agree with Antoni's comments, but it looks good besides for those.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx
yonitoo: yeah, this makes a lot of sense to refactor. I will come up with proper showErrorDialog util function and use it where possible.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/DeployJob.tsx
yonitoo: Agreed. Will create a constant about it.

yonitoo: Done.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/commandsAndMenu.tsx
yonitoo: Done.

antoniivanov (approved):

PR #2540, opened 2023-08-09
control-service: add file filter before job upload

mrMoZ1:
what: Refactored current job filtering logic to rely on inheritance and templates, providing more flexibility for future refactoring of required filtering. In addition to the already present allow list - added a filter list that deletes file types matched from a comma separated property before uploading data job source code files to version control/building job the image.

why: Currently users encounter challenges while deploying data jobs due to the presence of compiled Python files (*.pyc) that conflict with the upload validation process. This hinders their ability to redeploy jobs, especially when working with Jupyter Notebooks. To enhance the user experience and streamline the deployment process, we need to implement a file exclusion mechanism similar to .gitignore logic during data job uploads.

testing: added unit test


antoniivanov (changes requested):

projects/control-service/projects/pipelines_control_service/src/main/resources/org.apache.tika.mime/custom-mimetypes.xml
```diff
@@ -0,0 +1,5 @@
+<mime-type type="application/pyc">
```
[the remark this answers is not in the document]
mivanov1988: That's a good observation. If we have to create a new custom type for each file extension, it might not be suitable. Instead of just adding an additional type to the values.yaml, we would also need to write additional code.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadFileFilter.java
```diff
@@ -0,0 +1,86 @@
...
+
+    try (Stream<Path> stream = Files.walk(Paths.get(jobFolder.getPath()))) {
+      stream
+          .filter(Files::isRegularFile)
+          .forEach(
+              file -> {
+                validateFile(file.toAbsolutePath(), jobName);
+              });
+    } catch (IOException e) {
+      log.error("Exception while processing filter list: {}", e);
```
murphp15, line 61: We are actually trying to clean up error logging in our system at the moment. 
We are seeing the same error logged many times. 
Can you please not log it here. 

The new rule is that you should only log an error if you are going to recover from it. 
Sprint will log it at the exit point of the web request so there is no fear that the exception won't get logged. 

It might just be cleaner to let the exception bubble up instead of trying to handle it at all.

projects/control-service/projects/pipelines_control_service/src/main/resources/org.apache.tika.mime/custom-mimetypes.xml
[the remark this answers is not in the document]
mrMoZ1: Tika doesn't support .pyc files (https://github.com/apache/tika/blob/main/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml), which are classified as the most generic binary files - octet-stream, that's why I created this mime type - since this change is focused mainly on filtering the .pyc files produced from Jupiter-notebooks. The test contains a .pyc file.

projects/control-service/projects/pipelines_control_service/src/test/resources/org/apache/tika/mime/custom-mimetypes.xml
```diff
@@ -0,0 +1,5 @@
+<mime-type type="application/pyc">
```
murphp15, line 1: The folder structure here is actually different from the one used in main. 
Here is it folders whereas in main it is a dot separated file name.
 
 Also given that this is the same as the one in main how do you actually know this is being picked up in tests? 
 I would just remove this if it is the same and then you are sure that the tests are testing what will be deployed.

projects/control-service/projects/pipelines_control_service/src/main/resources/org.apache.tika.mime/custom-mimetypes.xml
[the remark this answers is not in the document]
mrMoZ1: Just as a FYI tika supports more than 1000 file types but not .pyc files.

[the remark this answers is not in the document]
antoniivanov: Ok. I am ok if you decide to leave it as it is. 

I feel it might be better to add support for specify extensions. But I am ok if we do not.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUploadFilterListValidator.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Override
+  String[] getValidationTypes() {
+    return this.filterList;
+  }
+
+  @Override
+  String detectFileType(Path filePath) {
+    return FilenameUtils.getExtension(filePath.getFileName().toString());
```
antoniivanov, line 40: detectFileType should be the same for both filtering and allowing 

Otehrwise filterList and allowedList would have different format which would be very error prone. 
Also likely we may want to allow files by extension and filter out files by detected tika type. 

I would suggest to have 2 methods 

detectFileType -> which would return the Tika type 
detectFileExtesion -> which would return the extension. 
Both would be in the parent class. 

then `matchTypes` method would accept (detectedType, detectedExtension). 

We would need 2 properties allowedTypes and allowedExtensions and filterTypes and filterExtensions

as we should be explicit about the format of the values and we cannot do that in one property.

antoniivanov (approved):

projects/control-service/projects/pipelines_control_service/src/test/resources/filter-job/config.ini
```diff
@@ -0,0 +1,56 @@
+; Supported format: https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
+
+; This is the only required file in a Data Job.
+; Read more to understand what each option means:
+
+; Information about the owner of the Data Job
+[owner]
+
+; Team is a way to group Data Jobs that belonged to the same team.
+team = supercollider
```
doks5, line 10: Maybe change the team?

projects/control-service/projects/pipelines_control_service/src/test/resources/filter-job/config.ini
```diff
@@ -0,0 +1,56 @@
...
+; notified_on_job_failure_platform_error=example@example.com; example2@example.com
+notified_on_job_failure_platform_error=
+
+; Semicolon-separated list of email addresses to be notified on job execution success.
+notified_on_job_success=
+
+; Semicolon-separated list of email addresses to be notified of job deployment outcome.
+; Notice that if this file is malformed (file structure is not as per https://docs.python.org/3/library/configparser.html#supported-ini-file-structure),
+;   then an email notification will NOT be sent to the recipients specified here.
+notified_on_job_deploy=mzhivkov@vmware.com
```
doks5, line 56: Does this really need to be here?

doks5 (approved):

PR #2541, opened 2023-08-09
vdk-jupyter: add testing to CI/CD

yonitoo:
What:
Add test stage to the .plugin-ci.yml file of the vdk-jupyter plugin.

Why:
Right now, there is a build and release stage but we would want to run the tests every time to ensure code robustness.

Testing Done:
The Gitlab pipeline is green.

Server tests failure


projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -33,6 +33,41 @@ build-vdk-jupyterlab-extension:
...
+    - jlpm
+    - jlpm test
+
+    - echo "Running Integration tests..."
+    - jlpm install
+    - jlpm build:prod
+    - cd ./ui-tests
+    - jlpm install
+    - jlpm playwright install
+    - jlpm playwright test
```
antoniivanov, line 59: Is there dependencies between the tests? Do we need to run the server tests first then frontend then integration ? 

If not it would better to split them up in 3 jobs , no? 
A) It would be easier to see what fails where. 
B) They would run faster because they run in parallel 


To reuse common setup across the 3 script use gitlab `extends` - https://docs.gitlab.com/ee/ci/yaml/?ref=jsramblings.com#extends

antoniivanov (changes requested):

gageorgiev (approved):

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
murphp15: Completely agree. 
Can you also confirm that all 3 tests suits output reports in junit format?

murphp15:
It is very important to test this with failing cases and make sure the report makes it obvious which tests have failed.

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
yonitoo: Great suggestion! I split the tests into 3 separate stages (server tests, frontend tests and integration tests). Each of them now generates JUnit reports in the vdk-jupyter/vdk-jupyterlab-extension/ directory of the plugin. Tested it locally - it works.

murphp15 (approved):

antoniivanov:
> It is very important to test this with failing cases and make sure the report makes it obvious which tests have failed.

That's a very good point. Let us get into practice for tests to verify the "failing" logs, we should link a fail CI or uploaded failed logs in the Testing Done section so other people can see if they find it easy to understand the problem (since what is easy to see to me often it is not so to someone else).

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -8,29 +8,71 @@
...
-  rules: # we want to trigger build jobs if there are changes to this package,
-    # but not if there are changes to VDK plugins or the main directory
-    - if: '$CI_COMMIT_BRANCH == "main" || $CI_PIPELINE_SOURCE == "external_pull_request_event"'
-      changes:
-        - "projects/vdk-plugins/vdk-jupyter/**/*"
+
+
+server-tests-vdk-jupyterlab-extension:
+  extends: .common-setup-vdk-jupyterlab-extension
+  stage: test
```
antoniivanov, line 39: You may notice from the error https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/962699510

But there's not test stage. Let's use `build` stage.

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -8,29 +8,71 @@
...
+    - jlpm install
+    - jlpm build:prod
+    - cd ./ui-tests
+    - jlpm install
+    - jlpm playwright install
+    - jlpm playwright test
+  artifacts:
+    when: always
+    reports:
+      junit: "projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/integration_tests_results.xml"
```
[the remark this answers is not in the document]
yonitoo: I tried to apply the common setup to "build-vdk-jupyterlab-extension" as well. If we drop its usage, we could do that. Should I?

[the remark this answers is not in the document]
yonitoo: Actually, as the build doesn't create report, the artifacts section would just be ignored and not fail with some error I guess. I will try it.

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -8,29 +8,71 @@
...
+    - pip install -e ".[test]"
+    - pytest -vv -r ap --cov vdk-jupyterlab-extension --junit-xml=server_tests_results.xml
+  artifacts:
+    when: always
+    reports:
+      junit: "projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/server_tests_results.xml"
+
+frontend-tests-vdk-jupyterlab-extension:
+  extends: .common-setup-vdk-jupyterlab-extension
+  stage: test
```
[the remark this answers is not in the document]
yonitoo: "build-vdk-jupyterlab-extension" also extends ".common-setup-vdk-jupyterlab-extension" where the stage is "build". The solution in the previous comment could apply here as well.

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
yonitoo: This would mean that we can add "stage: build" to the common setup, right?

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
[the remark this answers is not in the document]
yonitoo: Due to [the failed pipeline](https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/4866840293), I will revert it - remove the artifacts part from the common setup.

antoniivanov:
> > It is very important to test this with failing cases and make sure the report makes it obvious which tests have failed.
> 
> That's a very good point. Let us get into practice for tests to verify the "failing" logs, we should link a fail CI or uploaded failed logs in the Testing Done section so other people can see if they find it easy to understand the problem (since what is easy to see to me often it is not so to someone else).

Can you please add in testing done links to "failed" (even if faked) runs of gitlab tests so that we can see how easy it is to find the issue. Sort of like testing the error scenario for the CICD.

antoniivanov:
I clicked to update your branch becasue there were some unrelated tests getting started. 

The frontend tests are missing from the test report: https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/978573326/test_report

yonitoo:
Addressed. The Frontend tests are not showing up in the test report.

antoniivanov (approved):
Looks good . Thanks!

Issue #2539, opened 2023-08-09
Ability to disable VDK UI


mivanov1988:
Currently, there is no way to disable VDK UI through the Control Service Helm Chart. As a workaround, we can set `operationsUi.replicaCount: 0`.

PR #2545, opened 2023-08-10
control-service: Ability to disable VDK UI in Helm

antoniivanov:
See Issue #2539

To test this (and future helm changes) more robustly I am adding a new job to which would run on MR and would run the same script to deploy the helm chart in kubernets but as dry run .

### DIFF
--- a/projects/control-service/cicd/.gitlab-ci.yml
+++ b/projects/control-service/cicd/.gitlab-ci.yml
@@ -284,7 +284,7 @@ control_service_publish_api_client:
 # (there are docker KDC servers that can be used for demo? but it will not work with Impala ?)
 # The cockroach storage class is taken from DECC https://devhub.eng.vmware.com/console/resources/namespaces
 # Go to the namespace -> StorageClass
-control_service_deploy_testing_data_pipelines:
+.control_service_deploy_testing_data_pipelines:
   stage: pre_release
   image: docker:23.0.1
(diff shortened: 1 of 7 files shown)


projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -307,6 +310,19 @@ control_service_deploy_testing_data_pipelines:
     - if: '$CI_COMMIT_BRANCH == "main"'
       changes: *control_service_helm_change_locations
 
+control_service_helm_chart_dry_run:
```
murphp15, line 313: we have a linting step already. 
What is the difference between dry run and linting? 
Maybe we should remove the linting step if this dry run will catch all those issues?

murphp15 (approved):

projects/control-service/cicd/.gitlab-ci.yml
antoniivanov: No, both are valuable in my opinion.

* `helm lint` inspects the **entire** chart, checking every file and template for potential style and syntax issues. 
* With `helm install --dry-run` we are rendering the templates for the particular configuration provided by our values.yaml/--set options only.  It will catch runtime issues and issues with rendering the templates which lint cannot. But only on the configured path.

gageorgiev (approved):

PR #2550, opened 2023-08-14
vdk-jupyter: Change root FS position for path parameter

gageorgiev:
Currently, VDK commands in Jupyter require a path parameter which points to the location of the data job relatively to the current location of the user in the Jupyter file system menu on the right.

This is suboptimal, as it makes it less consistent to use and is the cause of at least one bug.

This change streamlines this by making it so the path parameter is always assumed to point relatively to the FS location where Jupyter was instantiated. This means all commands will expect the same path param value for a given job, regardless of where the user is currently located within the Jupyter FS menu.

This also fixes a bug where a job inside a given directory cannot be ran from inside that directory.

Some UI changes were also introduced, now the placeholder for the path contains the whole Jupyter path and the input is adjusting to the size of the placeholder (since we put the whole relative path it can become a long string) 

Screenshot 2023-09-11 at 10 05 47

This UI change gives hint to the user what we expect for the input and makes it proof for user mistakes. 

Testing done: only manual so far, working on an actual test


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -237,3 +236,16 @@ def convert_job(job_dir: str):
...
+def expand_path(path: str):
+    """
+    Expands the path parameter to be rooted at the file system location where Jupyter was initially ran.
+    It also does this in a way where it does not matter whether the user inputs an initial forward slash or not.
+
+    :param path: path parameter provided by the user
+    :return: expanded path
+    """
+    return pathlib.Path(
+        os.getcwd() + "/" + (path[1:] if path.startswith("/") else path)
```
antoniivanov, line 250: Please use Pathlib.Path.joinPath method to concatanate paths and not `+`

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/ui-tests/tests/vdk-jupyterlab-extension.spec.ts
```diff
@@ -145,3 +145,26 @@ test('should create an init cell when opening a new notebook', async ({
     page.getByText(`job_input = VDK.get_initialized_job_input()`)
   ).toBeVisible();
 });
+
+test(
+  'should create a new dir, navigate to it,' +
+    'create a new job, attempt to create a job relative to' +
+    'the original dir and succeed',
+  async ({ page }) => {
```
DeltaMichael, line 153: Shouldn't we be asserting something at the end of this test?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -247,3 +246,13 @@ def convert_job(job_dir: str):
...
+
+
+def expand_path(path: str):
+    """
+    Expands the path parameter to be rooted at the file system location where Jupyter was initially ran.
+
+    :param path: path parameter provided by the user
+    :return: expanded path
+    """
+    return pathlib.Path(os.getcwd()).joinpath(path)
```
duyguHsnHsn, line 258: What happens if the user directly puts the full path, not relative Jupyter path?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/ui-tests/tests/vdk-jupyterlab-extension.spec.ts
gageorgiev: no since the test would fail if any individual step is impossible

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -61,7 +58,7 @@ def run_job(path, arguments=None):
         if len(script_files) == 0:
             return {"message": f"No steps were found in {path}!"}
         with open("vdk_logs.txt", "w+") as log_file:
-            path = shlex.quote(path)
+            path = shlex.quote(str(path))
```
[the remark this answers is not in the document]
gageorgiev: shlex.quote expects a string, not a pathlib.Path object

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
gageorgiev: it would fail

duyguHsnHsn: @gageorgiev don't u think it would be nice to cover both cases ?

gageorgiev: My understanding of the ticket was that we should want that to fail:
```
dditionally, the path input should be fixed so that it always points relatively to the path assumed root by the Jupyter server, i.e. the directory is was ran inside of. This means that users should always provide the full path starting from the root Jupyter dir...
```

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/ui-tests/tests/vdk-jupyterlab-extension.spec.ts
DeltaMichael: > no since the test would fail if any individual step is impossible

Ok, but there's a possible scenario where all the commands pass and the job doesn't get created for some reason. If I deleted the call from the jupyter plugin to the code that actually creates the job, the test wouldn't catch it in the current state.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: I suppose it is all good then but what kind of a message will the user get if it fails like that? both the strings concatenated and saying it is not existing? I think it would look strange

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: @gageorgiev I see, maybe then we should change the labels of all path inputs to explain more that it only accepts jupyter relative paths?

duyguHsnHsn (commented):
Looks good! Please make sure that nothing in the front-end remains that works with the previous implementation! 
For example does this cover the use-case od default job-input which fetches the current directory from the server and if the user does not modify it, it uses it? I am not exactly sure how it was working but please check it. If it seems all fine I will approve the pr

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/VdkTextInput.tsx
```diff
@@ -1,75 +1,121 @@
...
+     * @param prevProps - previous component properties
+     */
+    componentDidUpdate(prevProps: IVdkTextInputProps): void {
+        if (prevProps.value !== this.props.value) {
+            this.adjustInputWidth();
+        }
+    }
+
+    /**
+     * Adjusts the input width based on its placeholder value.
```
antoniivanov, line 58: Please, explain why is it needed in the comment.  Not so much what it does which is clear from the title

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/VdkTextInput.tsx
```diff
@@ -1,75 +1,121 @@
...
+     */
+     adjustInputWidth(): void {
+         const currentInput = this.inputRef.current;
+         if (!currentInput) return;
+
+         let maxWidth = 250;
+
+         jobData.forEach((value) => {
+             const tempSpan = document.createElement('span');
```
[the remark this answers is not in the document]
duyguHsnHsn: @antoniivanov the idea of the method is to change the input size according to the placeholder we have, you can check the photo i have provided in the description. 
The inputs are not responsive by default and if we have a really long path as in the example it would hide the placeholder which looks strange
My idea here was the user to be able to see the whole path so they know what kind of path we expect from them since in the usability tests it was a great issue 
Do you have any other recommendations for this issue? Actually yes manipulating the dom is not bad but in the worst case scenario we would get the default input size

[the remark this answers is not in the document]
duyguHsnHsn: @antoniivanov I have modified the code to use React's state for adjusting the input width. Now we use the DOM only for the tempSpan and if something happens it will use the default 250px value

antoniivanov (approved):
Let's merge it.

I am still not confident adjustInputWidth logic is worth the cost (dev, maintenance, performance)  but it's an usability improvement and I am all for that. So we will evaluate as we see how it's used and if it causes issues.

PR #2552, opened 2023-08-14
vdk-jupyter: reuse authenticated page Oauth2 access-token

antoniivanov:
if the user has already authenticated when entering the notebook server and has generated access token, there's no point in trying to re-login again if the same authentication can be used against VDK Control Service. Best user experience is to resuse the token transparently

The way the design works is operator can specify the token location when installing vdk-jupyter extension.

Currently, this is set as environment variables. But once we get up to speed using jupyter settings capabilities we will translate them there. So for now the environment variables are purposefully not documented in README. 

Once they do we can automatically update the server with the latest access token which can be used in operations against the VDK Control Service

Details in
https://github.com/vmware/versatile-data-kit/blob/main/specs/vep-994-jupyter-notebook-integration/README.md#security-and-permissions


antoniivanov:
I will close for now as it's not a priority.

PR #2557, opened 2023-08-16
control-service: remove hardcoded image pull policy from job deployer

mrMoZ1:
what: removed hardcoded imagePullPolicy from the JobImageDeployer service.

why: 
Currently, the "Always" strategy fetches the latest version of the image from the registry. This results in an unnecessary image pulls.
Switching to the "IfNotPresent" strategy will address this issue. With this strategy, images will be pulled only if they are not already present in the local cache, matching the behavior of our data jobs that use exact tags associated with the current version. This will reduce unnecessary network calls to the Docker Registry.

testing: trivial

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -1000,6 +1000,9 @@ dataJob:
       # In case of negative number it will be subtracted from the {{end_time}}.
       # If left blank, defaults to 0.
       endTimeOffsetSeconds: 0
+  deployer:
+    # imagePullPolicy for data job images deployed to K8S.
+    imagePullPolicy: "IfNotPresent"
   deployment:
     initContainer:
       ## Resources set on Data Job initContainer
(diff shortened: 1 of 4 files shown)


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -1000,6 +1000,9 @@ dataJob:
       # In case of negative number it will be subtracted from the {{end_time}}.
       # If left blank, defaults to 0.
       endTimeOffsetSeconds: 0
+  deployer:
```
mivanov1988, line 1003: What's the rationale behind using "deployer" instead of a construction similar to `dataJob.deployment.jobContainer.imagePullPolicy`?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployer.java
```diff
@@ -42,6 +42,9 @@ public class JobImageDeployer {
   @Value("${datajobs.deployment.readOnlyRootFilesystem:}")
   private boolean readOnlyRootFilesystem;
 
+  @Value("${datajobs.deployer.imagePullPolicy:IfNotPresent}")
```
antoniivanov, line 45: Let's use the same prefix as the above property `datajobs.deployment.jobImagePullPolicy` like `datajobs.deployment.readOnlyRootFilesystem`

antoniivanov (approved):

mivanov1988 (approved):

PR #2558, opened 2023-08-16
specs: vep-2448 detailed design section

DeltaMichael:


specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -315,58 +315,106 @@ Explain how does the system handle API violations.
...
+this.
+
+### Clean Error Handling
+
+Error messages should clearly state the problem without placeholder and repeated
+logs text so that users can directly understand what went wrong. Users should be
+able to see the original exception when it's passed up the call stack and is in
+the user code so they can handle it.
+
+VDK developers should be discouraged to use generic error messages like "An
```
gageorgiev, line 365:
```suggestion
VDK developers should be discouraged from using generic error messages like "An
```
nit

gageorgiev (approved):
looks great

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -315,58 +315,106 @@ Explain how does the system handle API violations.
...
+error occurred". This will give more meaningful feedback to users.
```
antoniivanov, line 366: Another idea potenitally. 

For some common errors we can consider creating an error dictionary/repository that stores all the known errors, their descriptions, and potential solutions.

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -315,58 +315,106 @@ Explain how does the system handle API violations.
...
+
+### Progress Indicators
+
+Data jobs in the CLI should display progress indicators by default instead of
```
antoniivanov, line 370: We should consider implementing an asynchronous mechanism for updating progress indicators. This would make sure that the primary task doesn’t get interrupted or slowed down due to progress bar updates. This is really relevant for ingestion. 

Also as we need progress indicators in multiple places likely we need a common encapsulation for that - in a form of a python module or plugin. 

The progress indicator should providing a notification callback mechanism or similar so it's integratable with Notebook

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
```diff
@@ -315,58 +315,106 @@ Explain how does the system handle API violations.
...
+
+1. Get rid of the log and throw/rethrow pattern
+   - recoverable exceptions should be logged at the warn level and swallowed
+   - unrecoverable exceptions should be either wrapped in a generic error and
+     re-thrown or just thrown and passed to the caller
+2. Create specific error classes for User, Platform and Config errors
+   - possibly extend the classes into more specific errors, e.g. DataQualityException
+   - the logic that determines the type of error and who's responsible for
+     fixing it, e.g. get_blamee should remain in place
+   - developers should have the option to pass the error up to the data job
```
antoniivanov, line 407: There's need for more indepth proposal here. Let's consider a spike for that and think it through.

The catch is we want to propagate an original exception up the call stack . But we also want to keep information about its classification (e.g., "user error", platform error). 

So how do we do that ? 


Options 1 Define Custom Exception class with attributes user to use:


We raise an exception like that 

```
class VDKUserCodeException(Exception):
    def __init__(self, message, original_exception=None):
        super().__init__(message)
        self.original_exception = original_exception
```

Or we can use the native python `__cause__` and exception chaining https://peps.python.org/pep-3134/ 
But we need to make it somewhat mandatory

Specifically for user vs platform error I was thinking having it be attribute might be more appropriate since the same FileNotFound exception for example could be user or platform depending on where it was raised from:  
```
class VDKException(Exception):
    def __init__(self, message, resolvable_class : ResolveableBy=None, original_exception=None):
        super().__init__(message)
        self.resolvable_class = resolvable_class 
        self.original_exception = original_exception
```


And user would use it like that 
```
try: 
   job_input.execue_template() 
except VDKException e: 
   if isinstance(e.original_exception, DataQualityException): 
      do_someting() 
 ```
  This is not very intuitive UX in ptyhon though. 
  
  
  I think users would except something like 
```
try: 
   execute_risky_operation()
except OriginalException as e:
  ... 
```

If we want that while still providing a mechanism to classify the exception we may add custom attributes: 
```
def classify_exception(exception, classification):
    exception.classification = classification
    return exception
```

antoniivanov (approved):
Looks good to me at this stage. There need to be a bit more details into how we are going to do things which we can add down the line in next few weeks I hope.

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
DeltaMichael: Linked to https://github.com/vmware/versatile-data-kit/issues/2572

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
DeltaMichael: Added this as separate ticket https://github.com/vmware/versatile-data-kit/issues/2593

specs/vep-2448-vdk-run-logs-simplified-and-readable/README.md
DeltaMichael: Added to https://github.com/vmware/versatile-data-kit/issues/2571

PR #2559, opened 2023-08-16
frontend: Fix for e2e test

gorankokin:
## Why

`find('[data-cy=data-pipelines-job-executions-type-container]')` causes the test to fail if it doesn't find any job executions. It should be returning an empty list instead.

## What

Introduce logic that returns an empty list

## How was this tested

Ran e2e tests locally
Ran e2e tests in CI

## What type of change is this

Bug fix

 - Fix one bug in e2e tests for Data Job Executions


DeltaMichael (approved):

yonitoo (approved):

antoniivanov (approved):

DeltaMichael:
@antoniivanov Added some more details to description

PR #2590, opened 2023-08-21
vdk-jupyter: add oauth2 authentication implementation

antoniivanov:
This is adding the server part of Oauth2 authentication process.

It adds 1 more APIs: `/login`

When called it without "code" query paramter, it will start the authentication proces as per OAuth2 standard .
We are using only native app workflow with PKCE (RFC 7636) because we cannot really secure the server side so we cannot reliably use client secret.
When called with "code" query paramter it will finish the process and exchange the code for access token (and refresh token) and safe it in VDK storage.

This change add integration with jupyter configuration. This way the extension can be configured more natively using jupyter configuration mechanism.

So now configuration can be set in any of the ways juptyer server supports - https://jupyter-server.readthedocs.io/en/latest/users/configuration.html

E.g 
I've set locally in `~/.jupyter/jupyter_lab_config.py`

```
c.VdkJupyterConfig.oauth2_authorization_url = "https://console.cloud.vmware.com/csp/gateway/discovery"
c.VdkJupyterConfig.oauth2_token_url = "https://console.cloud.vmware.com/csp/gateway/am/api/auth/authorize"
c.VdkJupyterConfig.oauth2_client_id = "xxx"
c.VdkJupyterConfig.rest_api_url = "https://supercollider.vmware.com/"    
```

In future change we can add integration between VDK configuration mechanims and jupyter so that properties set in VDK can be recognized in Jupyter and vice-versa but that's more advanced use-case

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/oauth2.py
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/oauth2.py
@@ -22,6 +22,8 @@ log = logging.getLogger(__name__)
 
 class OAuth2Handler(APIHandler):
     def initialize(self, vdk_config: VdkJupyterConfig):
+        log.info(f"VDK config: {vdk_config.__dict__}")
+
         self._authorization_url = vdk_config.oauth2_authorization_url
         self._access_token_url = vdk_config.oauth2_token_url
         self._client_id = vdk_config.oauth2_client_id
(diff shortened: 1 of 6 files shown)


duyguHsnHsn (approved):
Looks really good to me!

gageorgiev (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/oauth2.py
```diff
@@ -0,0 +1,136 @@
...
+
+        data = (
+            f"code={code}&"
+            + f"grant_type=authorization_code&"
+            + f"code_verifier={code_verifier}&"
+            f"redirect_uri={redirect_uri}"
+        )
+        basic_auth = HTTPBasicAuth(self._client_id, "")
+        try:
+            # TODO : this should be async io
```
DeltaMichael, line 121: Why?

antoniivanov: It's ok to block if you cannot proceed with other operations without authentication. This is not the case here. Only few VDK-specific operations require this authentication. But as it's generally single user server it's probably ok to leave .

PR #2591, opened 2023-08-21
vdk-jupyter: impelement manual oauth2 login workflow

antoniivanov:
When jupyter notebook is deployed in a Multi-Notebook server instance (like jupyterhub) each user instance has a distinct URL. This means we don’t have a stable redirect URL to register with your OAuth2 provider.

User Scenario
1. Every time a user starts a new notebook server, it gets a unique URL to their notebook
2. OAuth2 Challenge: Since the redirect URL is dynamic based on the instance, we can’t pre-register a fixed callback URL with the OAuth provider.

Given this setup there are a couple of solution
A) The authorization provider could allow dynamic URLs (not very likely as it's not very secure)

B) The Notebook parent server could provide stable proxy to the underlyng notebook (so we can use `http://jupyterhub/notebook?user=xxx` and parent server would proxy this call to the user's notebook server). This way only the query paramter is dynamic which is often ok with authorization providers. The registered notebook url woudl be `http://jupyterhub/notebook`

C) If none of the above works , this solution is proposed where we can use Manual Login where user would need to copy one URL into the application.

Here is a short gif showing the process:

</img>


antoniivanov:
Note : this is a stacked PR (https://benjamincongdon.me/blog/2022/07/17/In-Praise-of-Stacked-PRs/) . So will be merged after PR https://github.com/vmware/versatile-data-kit/pull/2590 is merged

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/__tests__/manual-oauth2-flow-component.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    it('should render without crashing', () => {
+        render(<ManualOAuth2Flow/>);
+    });
+
+    it('should open a new authentication window on button click', async () => {
+        (requestAPI as jest.MockedFunction<typeof requestAPI>).mockResolvedValue('http://auth-url.com');
+        const component = render(<ManualOAuth2Flow/>);
+        const button = component.getByText("Click here to start.");
+        fireEvent.click(button);
```
antoniivanov, line 30: I am using fireEvent for consistency. But from what I am reading, overall [recommendations](https://ph-fritsche.github.io/blog/post/why-userevent) are to use user-event instead - https://testing-library.com/docs/user-event/intro/#differences-from-fireevent

duyguHsnHsn:
The design seems a little bit off to me. Maybe changing the style of the listed items would make it look better and maybe the first button should be blue as the other "Ok" and "Continue" buttons from the Jupyter UI.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    render() {
+        return (
+            <div className={LOGIN2_DIALOG_CLASS}>
+                {this.state.authStatus === AuthStatus.NONE &&
+                    <div>
+                        <p>Authenticate process will start. After you clik a new window will be opened.</p>
+                        <ul>
+                            <li>Please <b>wait</b> for the authentication window to redirect you back to 127.0.0.1.</li>
```
duyguHsnHsn, line 104: Why is it specifically 127.0.0.1?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/login.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+  );
+  console.log('open login window with redirect url: ' + redirect_url);
+  const oauthWindow = window.open(
+    redirect_url,
+    'oauthWindow',
+    'width=800,height=500'
+  );
+  if (oauthWindow == null) {
+    console.log('Failed to open OAuth2 login window');
```
duyguHsnHsn, line 22: instead of this you can you ShowErrorMessage() from @jupyterlab/apputils

duyguHsnHsn: If there is no specific reason for only consol logging it

duyguHsnHsn (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
```diff
@@ -0,0 +1,148 @@
...
+                        <p>Authentication failed!</p>
+                    </div>
+                }
+            </div>
+        );
+    }
+}
+
+
+export async function showLogin2Dialog() {
```
DeltaMichael, line 139: I think it's cleaner for components to have a default export. Maybe you can move this function to the file where it's called.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
```diff
@@ -0,0 +1,148 @@
...
+ *    they are redirected back, with an authorization code in the URL.
+ *
+ * 3. Users are instructed to copy and paste this URL into an input field manualy
+ *
+ * 4. This code is then sent to the 'login' endpoint to exchange it for tokens or finalize the authentication process.
+ *
+ * 5. The component's state is updated based on the success or failure of the authentication.
+ *
+ */
+export class ManualOAuth2Flow extends Component<{}, ManualOAuth2FlowState>  {
```
DeltaMichael, line 47: Nit: Do we use class components throughout the jupyter plugin? I'm just curious if there's a reason behind it. Functional components with hooks read a lot better, i.e. you don't have to define a constructor, a render method and you don't have to define the initial state.

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
antoniivanov: In the OAuth2 authorization flow, the redirect URI is where the user is sent after they have authenticated with the authorization server. This URI must be registered with the OAuth2 provider and must match exactly the redirect URI sent in the authorization request.

Using 127.0.0.1 as the redirect URI is a fairly safe option as it always point to your local machine. This address can't be hijacked with DNS changes.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
antoniivanov: I was wondering the same. 
We use class components through the jupyter plugin. So I am sticking to that in this change.

We are using React 15, it seems. Hooks and functional components were introduced in React 16 .
I have not delved into why yet. Duygu may shed more light on that. 

But we may change that for Beta , the very least jupyterlab 4 uses newer react.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
antoniivanov: I am sticking to the convention used in the other components with `showXXXDialog` function in the file.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/Login2.tsx
duyguHsnHsn: @antoniivanov @DeltaMichael we are using react 17, you can check the package.json. As for react 15 as far as i remember we were using a lower version of react when i started working for the integration for sure but not sure whether it was 15 or 16. At that point using components seemed ok, since most of the other extensions as git extension was using the same structure.
But as Antoni said we can think for changes for the beta version

PR #2600, opened 2023-08-23
vdk-kerberos-auth: enable auth to work inside a running asyncio event loop

antoniivanov:
This change adds _run_coroutine function to ensure proper execution of a get_tgt coroutine, even if called within an already running event loop in asyncio (which in principal is not allowed)

This is happening when the plugin is invoked within a jupyter notebook which runs within a async event loop started by tornado server. And this causes following errors:
```
File
"/opt/conda/lib/python3.7/site-packages/vdk/plugin/kerberos/minikerberos_authenticator.py",
line 90, in _kinit
    loop.run_until_complete(get_tgt())
File "/opt/conda/lib/python3.7/asyncio/base_events.py", line 563, in
run_until_complete
    self._check_runnung()
File "/opt/conda/lib/python3.7/asyncio/base_events.py", line 526, in
_check_runnung
    'Cannot run the event loop while another loop is running')
RuntimeError: Cannot run the event loop while another loop is running
```

The way the fix works is
- If the current event loop is running, the function now starts a new thread to run the coroutine, ensuring that it runs to completion without conflict with the existing loop.
- Otherwise it creates a new asyncio event loop in the current thread and runs the couritine as before.

### DIFF
--- a/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
+++ b/projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import os
 import tempfile
+import threading
 
 from minikerberos.common.creds import KerberosCredential
 from minikerberos.common.target import KerberosTarget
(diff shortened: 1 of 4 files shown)


duyguHsnHsn (approved):
Looks good to me!

projects/vdk-plugins/vdk-kerberos-auth/src/vdk/plugin/kerberos/minikerberos_authenticator.py
```diff
@@ -57,8 +64,42 @@ def __exit__(self, *exc):
...
+                    asyncio.set_event_loop(loop)
+                    result = loop.run_until_complete(coroutine)
+
+                thread = threading.Thread(target=run_in_new_thread)
+                thread.start()
+                thread.join()
+            else:
+                result = loop.run_until_complete(coroutine)
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
```
DeltaMichael, line 96: This is meant as a retry, right?

DeltaMichael (approved):

PR #2602, opened 2023-08-23
vdk-ipython: infer correctly job name and add more tests

antoniivanov:
If job name is not explcitly provided we should infer it correctly from the path name.

Also I extended ipython test coverage to cover most of the other untested job input methods

### DIFF
--- a/projects/vdk-plugins/vdk-ipython/src/vdk/plugin/ipython.py
+++ b/projects/vdk-plugins/vdk-ipython/src/vdk/plugin/ipython.py
@@ -41,12 +41,16 @@ class JobControl:
     ):
         path = pathlib.Path(path) if path else pathlib.Path(os.getcwd())
         job_args = json.loads(arguments) if arguments else None
-        self._name = name
+        self._name = path.name if name is None else name
         self._path = path
         self._arguments = job_args
         self._template = template
         self.job = None
         self.job_input = None
+        log.debug(
+            f"Job Control for job with name {self._name} from path {self._path} "
+            f"with arguments {self._arguments} and template {self._template}"
+        )
 
     def get_initialized_job_input(self):
         """
--- a/projects/vdk-plugins/vdk-ipython/src/vdk/plugin/ipython.py
+++ b/projects/vdk-plugins/vdk-ipython/src/vdk/plugin/ipython.py
@@ -41,7 +41,7 @@ class JobControl:
     ):
         path = pathlib.Path(path) if path else pathlib.Path(os.getcwd())
         job_args = json.loads(arguments) if arguments else None
-        self._name = path.name if name is None else name
+        self._name = path.name if not name else name
         self._path = path
         self._arguments = job_args
         self._template = template
--- a/projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
+++ b/projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
@@ -2,8 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 import logging
 import time
+from contextlib import contextmanager
 
 import pytest
+from _pytest._py.path import LocalPath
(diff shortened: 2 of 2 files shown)


duyguHsnHsn (approved):
Looks good!

projects/vdk-plugins/vdk-ipython/src/vdk/plugin/ipython.py
```diff
@@ -41,12 +41,16 @@ def __init__(
     ):
         path = pathlib.Path(path) if path else pathlib.Path(os.getcwd())
         job_args = json.loads(arguments) if arguments else None
-        self._name = name
+        self._name = path.name if name is None else name
```
DeltaMichael, line 44: Nit:

```suggestion
        self._name = path.name if not name else name
```

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -27,6 +30,14 @@ def ip(session_ip):
     session_ip.run_line_magic(magic_name="reset", line="-f")
 
 
+@contextmanager
+def get_vdk_ipython(session_ip, vdk_load_arguments_line=""):
+    session_ip.run_line_magic(magic_name="load_ext", line="vdk.plugin.ipython")
+    session_ip.run_line_magic(magic_name="reload_VDK", line=vdk_load_arguments_line)
+    yield session_ip
```
DeltaMichael, line 37: Is this supposed to be a generator? I don't see the point of using yield if we'll just call it once.

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -111,11 +123,11 @@ async def test_extension_with_ingestion_job(ip, tmpdir):
     )
 
     # get the data that is going to be ingested
-    ip.get_ipython().run_cell("import requests")
+    ip.get_ipython().run_cell("import json")
     ip.get_ipython().run_cell(
-        'response = requests.get("https://jsonplaceholder.typicode.com/todos/1")'
+        """raw = '{ "userId": 1, "id": 1, "title": "delectus aut autem", "completed": false }' """
```
DeltaMichael, line 128: Nit: Might be a good idea to add a comment explaining what this represents (placeholder).

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
antoniivanov: This is yield-based context manager. And it must be called exactly once. See https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager 
 
 This allows me to do later this:
 ```
 with get_vdk_ipython() as ip:
     # code before yield is executed and ip is returned 
 # After going out of scope automatically clean up code after yield is executed. 
 ```

projects/vdk-plugins/vdk-ipython/tests/test_plugin.py
```diff
@@ -73,16 +84,17 @@ def test_calling_get_initialise_job_input_multiple_times(ip, tmpdir):
     assert isinstance(ip.user_global_ns["job_input"], IJobInput)
 
     # check whether first job_input is the same as the second one
-    assert result_job_input == ip.get_ipython().getoutput("job_input")
+    assert result_job_input == ip.user_global_ns["job_input"]
 
 
 # uses the pytest tmpdir fixture - https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture
-async def test_extension_with_ingestion_job(ip, tmpdir):
+def test_extension_with_ingestion_job(ip, tmpdir):
```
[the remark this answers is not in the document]
antoniivanov: It as async by mistake and the test was never getting executed hence

DeltaMichael (approved):

PR #2603, opened 2023-08-24
vdk-jupyter: optimize ConvertToJobNotebook notes

antoniivanov:
Try to split out in one more section the notes after the conversion.

Also make them collapssable so they do not appear as so much text and user can click to see more info if they want.

You can see how it looks here: 
image

And with the text expanded: 

image


duyguHsnHsn:
Can you add the text in the drop downs to the PR descriptions

duyguHsnHsn (approved):
It looks good to me! Maybe just change the "vdk run" to "VDK run" since in some parts it is with lower case and in some with upper. IMO,it is better to keep it consistently  upper.  Only the tag "vdk" should remain with lower.

yonitoo (approved):
Awesome! Ship it!

PR #2617, opened 2023-08-28
vdk-control-cli: download keytab along with downloading job 

duyguHsnHsn:
What: Added download keytab to download-job command

Why: Currently, the download-job command downloads a data job that is deployed in the Cloud without downloading its keytab. This functionality is provided by download-key but instead of doing 2 operations, a simpler approach for a user who wants to execute a Cloud deployed job locally would be to download both with one operation.

Testing: python unit tests


projects/vdk-control-cli/src/vdk/internal/control/command_groups/job/download_job.py
```diff
@@ -38,6 +38,16 @@ def download(self, team: str, name: str, path: str):
...
             log.info(f"Downloaded Data Job in {path}/{name}")
+
+            log.info(f"Downloading keytab...")
+            keytab_file_path = os.path.join(path, f"{name}.keytab")
+            response = self.api.get_jobs_api().data_job_keytab_download_with_http_info(
+                team_name=team, job_name=name, _preload_content=False
+            )
+            with open(keytab_file_path, "wb") as w:
+                w.write(response.raw_data)
+            log.info(f"Saved keytab in {keytab_file_path}")
```
antoniivanov, line 49: encapsulate this into a separate method.

antoniivanov: Add appropriate error handling. 

If the download-key fails we do not want the whole Download job to be considered failure. 
Keys are to mandatory and jobs may work without them. Only for some functionalities and in some cases are they used.

It would be more appropriate to post a warning to user that they do not have key.

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_download_job.py
```diff
@@ -70,6 +82,11 @@ def test_download_job_does_not_exist(httpserver: PluginHTTPServer, tmpdir: Local
         uri="/data-jobs/for-team/test-team/jobs/test-job/sources", method="GET"
     ).respond_with_response(Response(status=404))
 
+    # Mock a 404 response for the keytab endpoint
```
antoniivanov, line 85: write separate tests methods test_download_job_does_not_exist and test_download_key_does_not_exist

Otherwise how do you know which one is handled appropriately

antoniivanov (changes requested):

antoniivanov (approved):

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_download_job.py
```diff
@@ -82,3 +99,32 @@ def test_download_job_does_not_exist(httpserver: PluginHTTPServer, tmpdir: Local
     assert "The requested resource cannot be found" in result.output
     # assert that vdk does not try cleaning up a non-existent archive
     assert "Cannot cleanup archive" not in result.output
+
+
+def test_download_key_does_error(httpserver: PluginHTTPServer, tmpdir: LocalPath):
```
yonitoo, line 104: nit: the name of the test could me a bit more descriptive - what kind of error is raised for example.

yonitoo (approved):
LGTM! Left a small suggestion.

projects/vdk-control-cli/tests/vdk/internal/control/command_groups/job/test_download_job.py
duyguHsnHsn: Here we do not check what the error is, we check whether the operation returns success (which is the expected. behaviour since the download job is done) and leaves a warning with the error about the keytab download
I will leave it as it is

PR #2618, opened 2023-08-28
vdk-jupyter: improve init message

antoniivanov:
It's better not to provide link to git repo source code. Links to source code are not best documentation. But also job_input could be changed through plugins in which case it won't be accurate.

Notebook `help(xxx)` is much more suitable.

Remove the unnecesary warning. it's not critical and even if by mistake they mark the current cell , they'd quickly find out t problem when using vdk run

I also made the cell non-editable. During user testing, users were trying to edit it and this may cause potential issues as it's not expected

output of help(job_input) looks like that: 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ out
 build
 dist
 src/*.egg-info
+.yarn
 
 # Project files
 .ropeproject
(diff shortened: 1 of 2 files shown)


duyguHsnHsn (approved):

duyguHsnHsn:
Please, add a sentence taht explains this cell should be executed to load job_input. We got it as feedback

antoniivanov:
> Please, add a sentence that explains this cell should be executed to load job_input. We got it as feedback

Done. I also made the cell non-editable. During user testing, users were trying to edit it and this may cause potential issues as it's not expected

dakodakov (approved):

PR #2619, opened 2023-08-29
vdk-ipython: add support for %%vdksql magic command

antoniivanov:
To provide a seamless user experience for data professionals using VDK
Notebooks for SQL queries. The feature allows them to execute SQL
queries (and VDK SQL steps in the future) right within the notebook, visualize the
results in an interactive grid, and resuse managed SQL connections
through VDK. 

Leveraged the ipyaggrid package to display SQL query results in a
user-friendly, interactive grid. Also added auto-initialization for VDK
job and managed its connections for SQL execution. Opted for ipyaggrid
over other options like qgrid due to its flexibility and feature set.
Other evaluated were  `itables` (not as rich), `beakers` (could not get
it to work), `qgrid` (could not get it to work)

Ipyaggrid is optional because it's not clear if every notebook installation would support (and vdk-ipython should work not just with jupyterlab or even some versions of jupyterlab may not)

Did minor refactoring. Move all Job control related functionality into `vdk.plugin.ipython.job` module.
The new SQL functionality is in `vdk.plugin.ipython.sql`  while the registration is kept in `vdk.plugin.ipython` module (though `__init__.py` file)

Tested with functional and unit tests.  Manual testing was also performed to confirm the user
experience and feature functionality. See attached screenshots and video

ipyaggrid is optional but if it is installed, the grid result would work like this. 

image

https://github.com/vmware/versatile-data-kit/assets/2536458/3d9425e4-b78b-48ab-902b-633603f5f719

Without ipyaggrid i looks like : 

image

### DIFF
--- a/projects/vdk-plugins/vdk-ipython/requirements.txt
+++ b/projects/vdk-plugins/vdk-ipython/requirements.txt
@@ -1,7 +1,7 @@
+ipyaggrid
 
 
-IPython
-
+# testing dependecies
 pytest
 vdk-core
 vdk-sqlite
--- a/projects/vdk-plugins/vdk-ipython/setup.py
+++ b/projects/vdk-plugins/vdk-ipython/setup.py
@@ -14,7 +14,7 @@ setuptools.setup(
     description="Ipython extension for VDK",
     long_description=pathlib.Path("README.md").read_text(),
     long_description_content_type="text/markdown",
-    install_requires=["vdk-core", "iPython"],
+    install_requires=["vdk-core", "iPython", "pandas"],
     package_dir={"": "src"},
     packages=setuptools.find_namespace_packages(where="src"),
     entry_points={"vdk.plugin.run": ["ipython = vdk.plugin.ipython"]},
--- a/projects/vdk-plugins/vdk-ipython/README.md
+++ b/projects/vdk-plugins/vdk-ipython/README.md
@@ -7,19 +7,44 @@ The command enables the user to load job_input for his current data job and use 
 
 See more about magic commands: https://ipython.readthedocs.io/en/stable/interactive/magics.html
 
+## Installation
+
+To use the extension it must be firstly installed with pip as a python package.
+```
+pip install vkd-ipython
+```
 
 ## Usage
-To use the extension it must be firstly installed with pip as a python package.
 Then to load the extension in Jupyter the user should use:
 ```
 %reload_ext vdk.plugin.ipython
 ```
+### Data Job Python coding cells
 And to load the VDK (Job Control object):
 ```
 %reload_VDK
 ```
-The %reload_VDK magic can be used with arguments such as passing the job's path with --path
-or giving the job a new with --name, etc.
+The %reload_VDK magic can be used with arguments:
+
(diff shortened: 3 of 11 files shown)


duyguHsnHsn:
What happens with the standard run, currently vdk-run only works with standard python so cells which contain this would automatically fail, please change the vdk-notebook to work with those too if we are really going to provide this as functionality

duyguHsnHsn:
If we are nto going to add it as a step now, i would say this might cause a lot of errors since probably they users will try to add it as a step

antoniivanov:
> If we are not going to add it as a step now, i would say this might cause a lot of errors since probably the users will try to add it as a step and there will be difference between development and full execution of a job

I've specified in the description we are going to support `and VDK SQL steps in the future`

antoniivanov:
> If we are not going to add it as a step now, i would say this might cause a lot of errors since probably the users will try to add it as a step and there will be difference between development and full execution of a job

I am not sure that's true. But I do have the change for notebook step ready . But that will be separate Pull Request

antoniivanov:
> > If we are not going to add it as a step now, i would say this might cause a lot of errors since probably the users will try to add it as a step and there will be difference between development and full execution of a job
> 
> I am not sure that's true. But I do have the change for notebook step ready . But that will be separate Pull Request

You can see the change in vdk-notebook in https://github.com/vmware/versatile-data-kit/pull/2622

duyguHsnHsn:
Please add the changes to the README of the plugin and also include a simple example there

duyguHsnHsn (approved):

projects/vdk-plugins/vdk-ipython/requirements.txt
```diff
@@ -1,7 +1,7 @@
+ipyaggrid
```
DeltaMichael, line 1: > Ipyaggrid is optional because it's not clear if every notebook installation would support (and vdk-ipython should work not just with jupyterlab or even some versions of jupyterlab may not)

This doesn't make any sense, we add it to requirements.txt, but then we say it's optional and check for it further down the line. Am I misunderstanding how requirements.txt works?

projects/vdk-plugins/vdk-ipython/setup.py
```diff
@@ -14,7 +14,7 @@
     description="Ipython extension for VDK",
     long_description=pathlib.Path("README.md").read_text(),
     long_description_content_type="text/markdown",
-    install_requires=["vdk-core", "iPython"],
+    install_requires=["vdk-core", "iPython", "pandas"],
```
DeltaMichael, line 17: Same question here.

DeltaMichael (approved):

projects/vdk-plugins/vdk-ipython/setup.py
antoniivanov: I made pandas a required dependency here because the likelihood of not having a pandas in a notebook installation is extremely low. Pandas is a de-facto standard for python based transformations. 

On the other hand, ippygrid is a specific widget that may not work across all versions of jupyter or other notebook solutions. I certainly haven't tested it. And it's better if it's optional

PR #2627, opened 2023-08-31
vdk-jupyter: show status on starting new operation

antoniivanov:
Common usability issue with starting operations is that when you click the operation (e.g Run) nothing happens.

This is fairly cheap temporary fix to at least show the status dialog and ask the user to explicitly close it.

Longer term solution should be planned where we will not only show that operaiton is runnign, but show how long time it reemians, where it is currently (e.g what is in case of vdk run), show logs (if user wants to) and so on. But that's a bigger story.

https://github.com/vmware/versatile-data-kit/assets/2536458/7c35a01c-7e51-42be-b4ae-270ca475910b


DeltaMichael (approved):

antoniivanov:
> Can you add what happens if we click on the timer thing too after clicking ok

The same thing as before

sabadzhiev (approved):

PR #2628, opened 2023-08-31
control-service: add support for multiple jwt issuers

mrMoZ1:
what: added support for multiple jwt issuers. Added an extra env variable to handle the extra issuers since the primary - spring.security.oauth2.resourceserver.jwt.issuer-uri; is a part of spring boot's configuration. Any extra issuers should be listed in datajobs.authorization.oauth2.resourceserver.jwt.extra-issuer.uris;

why: user use case dictates the need for more than a single issuer.

testing: 
Tested locally - manually :
A) If authentication is disabled and issue uri is empty , things work API requests can be made.
B) If auth is disabled but uris is not empty, then again you can do some operation without authentication.
C) if auth is ENABLED and URI is not empty - you are required authentication token - tested with tokens from two different issuers, and with non valid token as well as one of the authentication URI variables empty. Authentication works.
D) if auth is ENABLED and URI is empty - some error should show up.


projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -94,6 +99,8 @@ public SecurityConfiguration(
       FeatureFlags featureFlags,
       @Value("${spring.security.oauth2.resourceserver.jwt.jwk-set-uri:}") String jwksUri,
       @Value("${spring.security.oauth2.resourceserver.jwt.issuer-uri:}") String issuer,
+      @Value("${datajobs.authorization.oauth2.resourceserver.jwt.extra-issuer.uris:}")
+          String[] extraIssuers,
```
antoniivanov, line 103: Should we deprecate `spring.security.oauth2.resourceserver.jwt.issuer-uri` and replace it with one property `spring.security.oauth2.resourceserver.jwt.issuer.uris` which is a list ?

Until removed (in future change) we can concatenate both ?

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -159,6 +168,31 @@ private SecurityFilterChain enableSecurity(
     return http.build();
   }
 
+  @Bean
+  JwtIssuerAuthenticationManagerResolver byIssuer(JwtAuthenticationConverter converter) {
```
antoniivanov, line 172: When authentication is disabled issuer should be ignored (as it is not needed)
If auth is enabled there should be at least one issuer, I think.

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
mrMoZ1: That's my main concern with this change and am happy to do it either way, it's just that this is a standard Spring property, and the second approach is also backwards compatible - users wont't need to change helm charts in already deployed instances with just a single issuer - which should be most common use case.

antoniivanov: My suggestion is to name the new one not 'extra-issue-uris' but simple issue-uris. 

Otherwise for now to keep the previous one (simply write a comment that it is deprecated). 

At some point in the future when we are sure there is no usage of the deprecated one or we make a explicit major release  we can remove it

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
```diff
@@ -159,6 +169,40 @@ private SecurityFilterChain enableSecurity(
...
+      provider.setJwtAuthenticationConverter(converter);
+      managers.put(issuer, provider::authenticate);
+    }
+    return new JwtIssuerAuthenticationManagerResolver(managers::get);
+  }
+
+  @Bean
+  @ConditionalOnExpression(
+      "not '${spring.security.oauth2.resourceserver.jwt.issuer-uri:}'.equals('')")
+  public JwtDecoder jwtDecoder(String issuer) {
```
antoniivanov, line 194: Please make sure to test 
A) If authentication is disabled and issue uri is empty ,  things work - you can do some operation wiht the API
B) If auth is disabled but uris is not empty, then again you can do some operation without authentication 
C) if auth is ENABLED and URI is not empty - you are required authentication token 
D) if uath is ENABLED and URI is empty - some error should show up

antoniivanov (approved):
I will be out most of next week. So I will approve it as it looks good. 

Make sure to test all the cases I outlined please. And of course any other you think is needed.

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -58,7 +58,13 @@ datajobs.security.oauth2.enabled=true
 # used to verify any JSON Web Token (JWT) issued by the authorization server
 # It is required.
 spring.security.oauth2.resourceserver.jwt.jwk-set-uri=${SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI:http://localhost}
+
 # spring.security.oauth2.resourceserver.jwt.issuer-uri=${SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUER_URI}
+# Provide a comma separated list of extra issuer uris to be added to the above.
+# to be used only when additional issuers need to be used. the issuers should be listed
+# in a comma separated list e.g "https://example.org/issuerOne,https://example.org/issuerTwo"
+# the first issuer should not be included here and the above field must be populated extra issuers to work.
+# datajobs.authorization.oauth2.resourceserver.jwt.extra-issuer.uris=${DATAJOBS_AUTHORIZATION_OAUTH2_RESOURCESERVER_JWT_EXTRA_ISSUER_URIS:}
```
doks5, line 67: Would a change to the helm chart be required?

doks5 (approved):

projects/control-service/projects/base/src/main/java/com/vmware/taurus/security/SecurityConfiguration.java
mrMoZ1: Tested locally:

A) If authentication is disabled and issue uri is empty , things work.
B) If auth is disabled but uris is not empty, then again you can do some operation without authentication.
C) if auth is ENABLED and URI is not empty - you are required authentication token - tested with tokens from two different issuers, and with non valid token. Authentication works.
D) if auath is ENABLED and URI is empty - some error should show up.

PR #2631, opened 2023-09-01
vdk-jupyterlab: show status dialog on deploy and create

antoniivanov:
Currently it was not shown on Deploy or Create but those could take some time espeically deploy which can be rather slow.  This is causing quite a confusion to users


duyguHsnHsn:
@antoniivanov did you check the other operations?

antoniivanov:
> @antoniivanov did you check the other operations?

Yes. All the other have status

PR #2641, opened 2023-09-05
control-service: add data job deployment dynamic property source selection mechanism

mrMoZ1:
what:
Added a feature flag mechanism which will be used to allow for writes to both K8S and the database, and providing the flexibility to select the data source (either the database or K8S) when reading properties based on the feature flag configuration.

why: 
This change anticipates future work on: https://github.com/vmware/versatile-data-kit/tree/main/specs/vep-2272-complete-data-job-configuration-persistence#database-data-model-changes. which aims to remove the dependency on the Kubernetes (K8S) cluster for storing part of the configuration of data jobs. By decoupling the configuration from the K8S cluster, we can improve the backup/restore process and overall system performance. Currently, the loss of K8S namespace results in the potential loss of data jobs, requiring a manual and complex restoration process.

testing:
added unit test which checks default configuration


projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -184,6 +184,12 @@ spec:
               value: "{{ .Values.security.authorization.jwtClaimUsername }}"
             - name: DATAJOBS_DEPLOYMENT_JOB_IMAGE_PULL_POLICY
               value: "{{ .Values.dataJob.deployment.jobImagePullPolicy }}"
+            - name: DATAJOBS_DEPLOYMENT_CONFIGURATION_PERSISTENCE_WRITE_TO_K8S
```
doks5, line 187: Do we need two separate configurations? Can't we have one, similar to the "read"? E.g: `DATAJOBS_DEPLOYMENT_CONFIGURATION_PERSISTENCE_WRITE_DATA_SOURCE` with the options "*k8s*", "*db*", "*all*".
I am not sure if it would be ideal, but having so many different configuration variables seems a bit too much, at least in my opinion.

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -1026,6 +1026,17 @@ dataJob:
...
+    configuration:
+      # Configuration persistence are the persisted properties that allow the seamless migration
+      # (re-deployment) of data jobs to a new K8S cluster.
+      persistence:
+        # Variable to enable/disable property writes to Kubernetes
+        writeToK8S: true
+        # Variable to enable/disable property writes to the database
+        writeToDB: true
+        # Variable to select the truth data source for reading properties. Options are DB for database and
+        # K8S for kubernetes.
```
doks5, line 1038:
```suggestion
        # Variable to select the truth data source for reading properties. Options are "DB" for database and
        # "K8S" for kubernetes.
```

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -54,6 +54,14 @@ datajobs.security.kerberos.krb5ConfigLocation=${SECURITY_KERBEROS_KRB5_CONFIG_LO
 # Settings for OAuth2 authentication provider
 datajobs.security.oauth2.enabled=true
 
+# Configuration persistence are the persisted properties that allow the seamless migration
+# of data jobs to a new K8S cluster.
+datajobs.deployment.configuration.persistence.writeToK8S=${DATAJOBS_DEPLOYMENT_CONFIGURATION_PERSISTENCE_WRITE_TO_K8S:true}
+datajobs.deployment.configuration.persistence.writeToDB=${DATAJOBS_DEPLOYMENT_CONFIGURATION_PERSISTENCE_WRITE_TO_DB:true}
+# Variable to select the truth data source for reading properties. Options are DB for database and
+# K8S for kubernetes.
```
doks5, line 62:
```suggestion
# Variable to select the truth data source for reading properties. Options are "DB" for database and
# "K8S" for kubernetes.
```

doks5 (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
mrMoZ1: My thinking here was extensibility - e.g adding data sources if required in the future would be clearer IMO to provide an additional property with true/false, instead of using a csv list to store toggled properties. I'm not so sure now.

PR #2648, opened 2023-09-11
vdk-core: add log stacktrace flag

DeltaMichael:
## Why?

When we log the stack trace, we're logging exceptions that we've already logged during execution. This increases the logs length by about 30%.

## What?

Add flag that disables/enables logging the stack trace on exit code 1

## How has this been tested?

Ran vdk locally with a failing data job

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
@@ -20,6 +20,7 @@ JOB_GITHASH = "JOB_GITHASH"
 LOG_CONFIG = "LOG_CONFIG"
 LOG_LEVEL_VDK = "LOG_LEVEL_VDK"
 LOG_LEVEL_MODULE = "LOG_LEVEL_MODULE"
+LOG_STACK_TRACE_ON_EXIT = "LOG_STACK_TRACE_ON_EXIT"
 WORKING_DIR = "WORKING_DIR"
 ATTEMPT_ID = "ATTEMPT_ID"
 EXECUTION_ID = "EXECUTION_ID"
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -83,6 +84,13 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             "Allowed values: CRITICAL, ERROR, WARNING, INFO, DEBUG. "
             "If not set python default or one set by vdk -v LEVEL is used. ",
         )
+        config_builder.add(
+            LOG_STACK_TRACE,
+            False,
+            True,
+            "Controls whether the full stack trace is displayed again on exit code 1. "
+            "False by default, set to True in production environments for more debug output. ",
```
antoniivanov, line 92: > set to True in production environments for more debug output

That is not up to us. But up to the operator of VDK or the end user.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
```diff
@@ -83,6 +84,13 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             "Allowed values: CRITICAL, ERROR, WARNING, INFO, DEBUG. "
             "If not set python default or one set by vdk -v LEVEL is used. ",
         )
+        config_builder.add(
+            LOG_STACK_TRACE,
```
antoniivanov, line 88: I suggest `LOG_STACK_TRACE_AT_EXIT ` as it's more clear

projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py
DeltaMichael: It was meant to be an imperative. 😄 As in "Set this to true in prod environments if you want more debug output". I'll change it so it's more clear.

antoniivanov (approved):

gageorgiev (approved):

PR #2653, opened 2023-09-12
vdk-core: remove log and re-throw pattern

DeltaMichael:
## Why?

Remove the log and re-throw pattern to reduce the output logs size and increase readability

## What?

Remove loggign from log_and_throw and log_and_rethrow methods in the errors module. Create new methods that just add to the resolvable context and throw or re-throw the exception. Keep the api as is, but make the old methods call the new ones. This prevents plugins that use older versions of vdk-core from breaking.

## How has this been tested?

Before: https://pastebin.com/TCk4qsMC
After: https://pastebin.com/eJ7X4UcE

Ran locally, CI tests

Non-breaking change


antoniivanov:
Can you please upload the logs of a job (with some failures) before this change and after this change ?

gageorgiev (approved):

DeltaMichael:
Closing in favor of https://github.com/vmware/versatile-data-kit/pull/2666, which addresses the same problem.

PR #2655, opened 2023-09-12
control-service: implement Webhook APIs authentication

mivanov1988:
Why
If a data job authenticates against the Control Service via Kerberos, the Control Service won't be able to forward the access token to webhook APIs, because it does not exist.

What
Implemented method for obtaining oAuth2 access token via refresh token.

Testing Done:
Locally against real oAuth2 Authorization Server.

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -402,16 +402,20 @@ webHooks:
    ## 4xx response content is shown to end user as an error message.
    ## 5xx responses will be retried (number of retries is based on configuration, defaults to 3)
    ##
-   ## In case authenticationEnabled is set to true, the Control Service (CS) will transmit the oAuth2 access token
-   ## to the WebHook API. This access token serves the purpose of authenticating the client against the CS.
+   ## In case authenticationEnabled is set to true, the Control Service (CS) will obtain the oAuth2 access token from
+   ## authorizationServerEndpoint using the authorizationRefreshToken. After that the access token will be transmitted to the WebHook API.
    postCreate:
       webhookUri: ""
       internalErrorsRetries: 3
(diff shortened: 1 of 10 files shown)


mrMoZ1 (approved):

dakodakov (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -402,16 +402,20 @@ webHooks:
...
-   ## In case authenticationEnabled is set to true, the Control Service (CS) will transmit the oAuth2 access token
-   ## to the WebHook API. This access token serves the purpose of authenticating the client against the CS.
+   ## In case authenticationEnabled is set to true, the Control Service (CS) will obtain the oAuth2 access token from
+   ## authorizationServerEndpoint using the authorizationRefreshToken. After that the access token will be transmitted to the WebHook API.
    postCreate:
       webhookUri: ""
       internalErrorsRetries: 3
       authenticationEnabled: false
+      authorizationServerEndpoint: ""
+      authorizationRefreshToken: ""
```
antoniivanov, line 412: Document those, please. They would be extemely hard to figure out without context. 

THe first one is really "The URL of the OAuth2 authorization server's token endpoint. This endpoint is used for obtaining access tokens and refreshing tokens." 

The refresh token that is used to obtain a new access token from the authorizationServerEndpoint.

antoniivanov (commented):
Sorry it took me time to take a look. Please look over the comments and address them in separate PR if you agree. If now, let's  discuss them.

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
mivanov1988: Thanks, I will address the comments tomorrow.

PR #2658, opened 2023-09-13
vdk-control-api-auth: [bug-fix] authorization header base64 encoding

doks5:
Currently, when using PKCE authorization, the client_secret is set to None when the base64 string is encoded. This leads to some authorization endpoints to not properly decode the `authorization` header and fail the authorization process.

This change converts the None object to an empty string, when PKCE is used, in order to make the resulting base64-encoded string compatible with more authorization endpoints.

Testing done: Local testing against a production authorization endpoint and a unit test.


antoniivanov (approved):
I think I've hit this as well. I set client secret to "" (empty string). 

But this is much better future-proof solution. Thanks for making it!

mivanov1988 (approved):

PR #2660, opened 2023-09-13
control-service: add amazon rds ca certificates to data job image

mrMoZ1:
what:
Adding amazon rds ca certificates to secure job image

why:
Data job users require ca certificates to connect to amazon databases

testing:
building images locally and inspecting certificate output:
List truncated because of 100+ certificates
`38.45 subject=C = US, O = "Amazon Web Services, Inc.", OU = Amazon RDS, ST = WA, CN = Amazon RDS eu-west-1 Root CA RSA4096 G1, L = Seattle

38.46 subject=C = US, O = "Amazon Web Services, Inc.", OU = Amazon RDS, ST = WA, CN = Amazon RDS eu-central-1 Root CA RSA4096 G1, L = Seattle

38.46 subject=C = US, O = "Amazon Web Services, Inc.", OU = Amazon RDS, ST = WA, CN = Amazon RDS eu-north-1 Root CA RSA2048 G1, L = Seattle
`

### DIFF
--- a/projects/control-service/projects/job-base-image-secure/Dockerfile-data-job-base
+++ b/projects/control-service/projects/job-base-image-secure/Dockerfile-data-job-base
@@ -97,10 +97,8 @@ FROM photon:latest
 # excluding build dependencies.
 COPY --from=build /usr/local/ /usr/local/
 
-# Installs Python dependencies
-RUN yum install libffi-devel -y
-
-# Download amazon AWS RDS certificates
+# Installs Python dependencies and downloads AWS RDS certificates
 RUN : \
+ && yum install libffi-devel -y \
  && echo "Downloading AWS RDS certificate bundle..." \
  && wget https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -O /etc/ssl/certs/global-bundle.pem
--- a/projects/control-service/projects/job-base-image-secure/Dockerfile-data-job-base
+++ b/projects/control-service/projects/job-base-image-secure/Dockerfile-data-job-base
@@ -99,3 +99,8 @@ COPY --from=build /usr/local/ /usr/local/
 
 # Installs Python dependencies
 RUN yum install libffi-devel -y
+
+# Download amazon AWS RDS certificates
+RUN : \
+ && echo "Downloading AWS RDS certificate bundle..." \
+ && wget https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -O /etc/ssl/certs/global-bundle.pem
--- a/projects/control-service/projects/job-builder-secure/Dockerfile.python.vdk
+++ b/projects/control-service/projects/job-builder-secure/Dockerfile.python.vdk
@@ -51,7 +51,9 @@ RUN : \
           && yum autoremove build-essential gcc glibc-devel git unzip -y;  fi \
     && echo "Installing native dependencies ..." \
     && yum install libffi-devel libstdc++ findutils openssl-c_rehash -y \
-    && echo "Refreshing CA certificates ..." \
+    && echo "Downloading AWS RDS certificate bundle..." \
+    && wget https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -O /etc/ssl/certs/global-bundle.pem \
+    && echo "Refreshing CA certificates" \
     && /usr/bin/rehash_ca_certificates.sh \
     && echo "Deleting system packages ..." \
     && yum autoremove shadow toybox openssl-c_rehash -y \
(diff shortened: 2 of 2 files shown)


gageorgiev:
not sure I understand this change, if users need a particular certificate, is it not their responsibility to include it as part of their job or download it?

mrMoZ1:
> not sure I understand this change, if users need a particular certificate, is it not their responsibility to include it as part of their job or download it?

Job users can't download or add anything in the secure images except for the tmp folders.

mivanov1988 (approved):

projects/control-service/projects/job-builder-secure/Dockerfile.python.vdk
```diff
@@ -51,7 +51,9 @@ RUN : \
           && yum autoremove build-essential gcc glibc-devel git unzip -y;  fi \
     && echo "Installing native dependencies ..." \
     && yum install libffi-devel libstdc++ findutils openssl-c_rehash -y \
```
mivanov1988, line 53: I think the best place for this installation is in the `job-base-image-secure` since it is a relatively small file, and we will avoid the dependency on AWS with each job build.

projects/control-service/projects/job-builder-secure/Dockerfile.python.vdk
```diff
@@ -51,7 +51,9 @@ RUN : \
...
-    && echo "Refreshing CA certificates ..." \
+    && echo "Downloading AWS RDS certificate bundle..." \
+    && wget https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -O /etc/ssl/certs/global-bundle.pem \
```
mivanov1988, line 55: I think the best place for this installation is in the job-base-image-secure since it is a relatively small file, and we will avoid the dependency on AWS with each job build.

projects/control-service/projects/job-base-image-secure/Dockerfile-data-job-base
```diff
@@ -99,3 +99,8 @@ COPY --from=build /usr/local/ /usr/local/
 
 # Installs Python dependencies
 RUN yum install libffi-devel -y
```
mivanov1988, line 101: please merge two commands

antoniivanov:
> > not sure I understand this change, if users need a particular certificate, is it not their responsibility to include it as part of their job or download it?
> 
> Job users can't download or add anything in the secure images except for the tmp folders.

That's not 100% true. 

A) They can download it to temp directory and then point REQUESTS_CA_BUNDLE (if they use requests library) or SSL_CERT_DIR (for openssl) to that location. Now they need to do that every execution and it's not very nice and good user experience. 

B) user should install certifi first  It provides Mozilla’s carefully curated collection of Root Certificates that is more extensive that the one that comes with default python version. It should be tried before resorting to adding more custom certificates. 

C) Similarly to certifi if user can install aws-certifi (there is not such but could be created). 

I would definitely want to know if `certifi` has been tried and works before proceeding with this change.

mivanov1988:
> > > not sure I understand this change, if users need a particular certificate, is it not their responsibility to include it as part of their job or download it?
> > 
> > 
> > Job users can't download or add anything in the secure images except for the tmp folders.
> 
> That's not 100% true.
> 
> A) They can download it to temp directory and then point REQUESTS_CA_BUNDLE (if they use requests library) or SSL_CERT_DIR (for openssl) to that location. Now they need to do that every execution and it's not very nice and good user experience.
> 
> B) user should install `certifi` in requrements.txt (https://pypi.org/project/certifi/) first. It provides Mozilla’s carefully curated collection of Root Certificates that is more extensive that the one that comes with default python version. It should be tried before resorting to adding more custom certificates.
> 
> C) Similarly to certifi if user can install aws-certifi (there is not such but could be created).
> 
> I would definitely want to know if `certifi` has been tried and works before proceeding with this change.

In order to speed up the process can you assist us with this matter? A customer is currently facing a blockage related to it.

antoniivanov (approved):

antoniivanov:
> In order to speed up the process can you assist us with this matter? A customer is currently facing a blockage related to it.

I've approved it. But advice the user the try adding `certifi` to see if it would change anything.

mrMoZ1:
> > In order to speed up the process can you assist us with this matter? A customer is currently facing a blockage related to it.
> 
> I've approved it. But advice the user the try adding `certifi` to see if it would change anything.

The customer tried a few things to get this to work - one of them was installing the certifi library.

Here is what the customer tried on their own and it didn't work:

I can’t seem to get this working. The file system is read only and certificates are not an allowed file type that I can add to the data job. Even saving the certificate as a .txt the file gets recognized as a cert.

I even tried adding the certificate as a variable in python and creating an in memory file and it wouldn’t work.

Either we need to allow certificate files or I need these in the container already.

Issue #2572, opened 2023-08-17
[SPIKE] Passing errors to user code


DeltaMichael:
## Overview

We want to check if the original exception thrown from user code, libraries, etc. is propagated up the call stack.

Before #2666 users had to do 

```
try: 
   job_input.execue_template() 
except VDKException e: 
   if isinstance(e.original_exception, DataQualityException): 
      do_someting() 
 ```
This is not very intuitive UX in ptyhon though. 

After #2666 , they should be able to do
```
try: 
   execute_risky_operation()
except OriginalException as e:
  ... 
```

## Acceptance criteria

1. Test if errors can be passed and handled in user code
2. Write functional tests for passing and handling errors in user code

DeltaMichael:
Related https://github.com/vmware/versatile-data-kit/pull/2558#discussion_r1300104139

PR #2666, opened 2023-09-15
vdk-core: separate error logging, error reporting and error classification

DeltaMichael:
## Why

As part of the run logs initiative, we should add classification attributes to exceptions that are handled in vdk. These attributes are used for easier error classification and will help us enrich the vdk exception class family, so we can pass more info to our users.

This is not possible without separating the error logging and error reporting mechanisms, so some refactoring was required.

## What

The main changes are in `errors.py`. All other changes (tests, other modules) are a result of changes in `errors.py`. The changes for specific plugins are out of scope for this PR due to compatibility issues.

### Add attributes to exceptions

Base vdk exceptions now have two new attributes - `to_be_fixed_by` and `type` (names subject to change). These help the error classification functions determine who to blame when an exception is raised. They're also displayed when pretty-printing the vdk exceptions. UserCodeError, VdkConfigurationError and PlatformService error are preserved as convenience classes that hardcode these attributes, based on who we decide is responsible for these exceptions. They can be extended into more specific exceptions in subsequent PRs, related to https://github.com/vmware/versatile-data-kit/issues/2572

### Introduce a reporting API

```
report(error_type, exception: BaseException)
report_and_throw(exception: BaseVdkException)
report_and_rethrow(error_type, exception: BaseException)
```

The report functions adds any base exception (not necessarily a vdk one) to the resolvable context. It determines who to blame for the error and sets the `to_be_fixed_by` and `type` attributes of the exception based on the error classification.

report_and_throw is used for convenience. It adds the vdk exception that was passed to the resolvable context and throws it.

report_and_rethrow just calls report and throws the exception that was passed.

Finally, we have the `log_exception(log, exception: BaseException, *lines)` function. It builds a message from the lines we pass and logs it at the warn level. It then logs the exception that was passed.

The result is that we've effectively decoupled logging from the error reporting and error classification functions. We can choose to report an exception and not log it, or log and exception and not report it, depending on how we want to handle it. This provides a certain amount of freedom if we want to eliminate log statements we consider unnecessary.

### Change the way vdk exceptions are represented

Override the __str__ and __repr__ methods in BaseVdkException. The constructor takes a variable number of lines that are used to create the error message and exception representation. We print a nice box that tells us what kind of exception occurred and who should fix it, along with the details.

The constructor can still take an ErrorMessage or dict instance for compatibility reasons. That logic should be removed when we remove ErrorMessage from the plugins.

### What was removed

Passing `ErrorMessage` to vdk exceptions is not necessary anymore, so the pattern has been removed wherever it's used for instantiation in vdk-core.

Formatting the error message also had separate functions, which are no longer needed, so they've been removed.

The decorator for error messages for some special cases has been removed and the logic added to the only function that actually used it.

DomainError, `log_and_throw` and `log_and_rethrow` are kept for compatibility reasons, but they should be phased out in all plugins that call them as separate PRs.

## Resulting user experience

**Before:** https://pastebin.com/raw/YsgzpUj4

Configuration error is printed with verbose stack trace

**After:** https://pastebin.com/raw/MUsLcjB2

Configuration error is printed with shorter stack trace

**Before:** https://pastebin.com/raw/qbfRfAUn

Exception coming from user code is wrapped in UserCodeError

**After:** https://pastebin.com/raw/1HGfDqye

Exception coming from user code is not wrapped, but attributes are set correctly, e.g. blamee is displayed as user error in the subsequent result.

**Alternative**: https://pastebin.com/raw/AsjDeyHZ

Alternatively, we can wrap non-vdk exceptions on re-throw, but this means we should re-visit some of our logging statements, because the logs become too verbose. Since logging exceptions is now separate from reporting them, we can dive deeper into the issue.

### Folow-up

https://github.com/vmware/versatile-data-kit/issues/2677
https://github.com/vmware/versatile-data-kit/issues/2678
https://github.com/vmware/versatile-data-kit/issues/2679
https://github.com/vmware/versatile-data-kit/issues/2680
https://github.com/vmware/versatile-data-kit/issues/2681
https://github.com/vmware/versatile-data-kit/issues/2682
https://github.com/vmware/versatile-data-kit/issues/2683
https://github.com/vmware/versatile-data-kit/issues/2684

## How was this tested

Ran locally with jobs that cause errors, ran the vdk-core tests locally as well. For plugin tests, we rely on CI.

Potentially breaking change

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
@@ -141,34 +141,34 @@ class IngesterBase(IIngester):
         See parent doc
         """
         if len(column_names) == 0 and destination_table is None:
-            errors.log_and_throw(
-                ResolvableBy.USER_ERROR,
-                log,
-                what_happened="Failed to ingest tabular data",
-                why_it_happened="Either column names or destination table must be specified."
-                "Without at least one of those we cannot determine how data should be ingested and we abort.",
-                consequences="The data will not be ingested and the current call will fail with an exception.",
-                countermeasures="Pass column names or destination table as argument.",
+            errors.report_and_throw(
+                UserCodeError(
+                    "Failed to ingest tabular data",
+                    "Either column names or destination table must be specified."
+                    "Without at least one of those we cannot determine how data should be ingested and we abort.",
+                    "The data will not be ingested and the current call will fail with an exception.",
+                    "Pass column names or destination table as argument.",
+                )
             )
 
         if not ingester_utils.is_iterable(rows):
-            errors.log_and_throw(
-                ResolvableBy.USER_ERROR,
-                log,
-                what_happened="Cannot ingest tabular data",
-                why_it_happened=f"The rows argument must be an iterable but it was type: {type(rows)}",
-                consequences="The data will not be ingested and current call will fail with an exception.",
-                countermeasures="Make sure rows is proper iterator object ",
+            errors.report_and_throw(
+                UserCodeError(
+                    "Cannot ingest tabular data",
+                    f"The rows argument must be an iterable but it was type: {type(rows)}",
+                    "The data will not be ingested and current call will fail with an exception.",
+                    "Make sure rows is proper iterator object ",
+                )
             )
 
         if not isinstance(column_names, Iterable):
-            errors.log_and_throw(
-                ResolvableBy.USER_ERROR,
-                log,
-                what_happened="Cannot ingest tabular data",
-                why_it_happened=f"The column_names argument must be a List (or iterable) but it was: {type(rows)}",
-                consequences="The data will not be ingested and current call will fail with an exception.",
-                countermeasures="Make sure column_names is proper List object ",
+            errors.report_and_throw(
+                UserCodeError(
+                    "Cannot ingest tabular data",
+                    f"The column_names argument must be a List (or iterable) but it was: {type(rows)}",
+                    "The data will not be ingested and current call will fail with an exception.",
+                    "Make sure column_names is proper List object ",
+                )
             )
 
         log.info(
@@ -588,17 +588,17 @@ class IngesterBase(IIngester):
                     metadata=metadata,
                 )
             except Exception as e:
-                errors.log_and_throw(
-                    to_be_fixed_by=ResolvableBy.USER_ERROR,
-                    log=log,
-                    what_happened="Failed to pre-process the data.",
-                    why_it_happened=f"User Error occurred. Exception was: {e}",
-                    consequences="Execution of the data job will fail, "
-                    "in order to prevent data corruption.",
-                    countermeasures="Check if the data sent for ingestion "
-                    "is aligned with the requirements, "
-                    "and that the pre-process plugins are "
-                    "configured correctly.",
+                errors.report_and_throw(
+                    UserCodeError(
+                        "Failed to pre-process the data.",
+                        f"User Error occurred. Exception was: {e}",
+                        "Execution of the data job will fail, "
+                        "in order to prevent data corruption.",
+                        "Check if the data sent for ingestion "
+                        "is aligned with the requirements, "
+                        "and that the pre-process plugins are "
+                        "configured correctly.",
+                    )
                 )
         return payload, metadata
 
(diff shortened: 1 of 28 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -141,34 +141,46 @@ def send_tabular_data_for_ingestion(
...
         if not isinstance(column_names, Iterable):
-            errors.log_and_throw(
-                ResolvableBy.USER_ERROR,
-                log,
-                what_happened="Cannot ingest tabular data",
-                why_it_happened=f"The column_names argument must be a List (or iterable) but it was: {type(rows)}",
-                consequences="The data will not be ingested and current call will fail with an exception.",
-                countermeasures="Make sure column_names is proper List object ",
+            errors.report_and_throw(
+                UserCodeError(
```
antoniivanov, line 174: I know it's a draft and maybe it's part of muliti part change but 

We are doing away with UserCodeError and Platform Code Error, they would become attributes. 

For example here as this is ingestion error it make sense to have "raise InvalidColumnNamesTypeException" 

and InvalidColumnNamesTypeIngestionException would inhert IngestionExpcetion which could inhert either BaseVDKEception or directly Exception

antoniivanov: Though it make sense to make the change incrementally! This is pretty big already change already. 

And I strongly discourage making refactoring and functional changes in the same commit / PR. 

I just want to make sure if we are on the same page.

DeltaMichael: I see your point. I decided to keep the old class hierarchy for now as convenience classes and added attributes to the base class. Maybe we can just keep them for now and add more specific classes down the line.

DeltaMichael: Please keep in mind that this is still WIP. I opened the PR to look at the diff more easily and to keep track of how big the change actually is.

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -358,41 +357,41 @@ def log_and_throw(
...
-        __set_error_is_logged(e)
-        raise
+
+def report_and_rethrow(error_type: ResolvableBy, exception) -> None:
+    """
+    Add exception to resolvable context and then throw it to be handled up the stack.
+    """
+    resolvable_by_actual = __error_type_to_actual_resolver(error_type)
+    setattr(exception, "to_be_fixed_by", resolvable_by_actual)
+    setattr(exception, "type", error_type)
```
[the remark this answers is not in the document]
DeltaMichael: I'll rename most of these, they have to be private anyway.

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -607,32 +363,167 @@ def __error_type_to_actual_resolver(to_be_fixed_by: ResolvableBy) -> ResolvableB
     )  # What type is the error that caused this and whom to blame, Platform or Data Jobs Developer?
```
DeltaMichael, line 363: I think we can remove this and just default to platform error. Even if we reach this case, we'll need more info on what exactly happened, so it makes sense for it to be a platform error.

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -607,32 +363,167 @@ def __error_type_to_actual_resolver(to_be_fixed_by: ResolvableBy) -> ResolvableB
...
-            "Unknown to_be_fixed_by type. "
-            "This seems like a bug. We cannot wrap exception and return original one "
+
+def log_exception(log: Logger, exception: BaseException, *lines) -> None:
+    """
+    Log message and exception
+    """
+    message = "\n".join([s for s in lines if s])
+    log.warning(message)
+    log.exception(exception)
```
antoniivanov, line 375: Why are you logging it twice? 

log.exception would log the message at ERROR level (same as log.error) and print the stack race .
(reference: https://docs.python.org/3/library/logging.html#logging.Logger.exception)


`log.warning` is very much redundant in this case

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -13,25 +13,126 @@
...
+
+    def __init__(self, *error_message_lines):
+        super().__init__(
+            ResolvableByActual.PLATFORM,
+            ResolvableBy.PLATFORM_ERROR,
+            *error_message_lines,
+        )
+
+
+class VdkConfigurationError(BaseVdkError):
```
antoniivanov, line 98: I think we should add Deprecation comment to those classes as well.

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -13,25 +13,126 @@
...
+
+    def __init__(self, *error_message_lines):
+        super().__init__(
+            CONFIGURATION_ERRORS_ARE_TO_BE_RESOLVED_BY,
+            ResolvableBy.CONFIG_ERROR,
+            *error_message_lines,
+        )
+
+
+class UserCodeError(BaseVdkError):
```
antoniivanov, line 114: I think we should add Deprecation comment to those classes as well.

projects/vdk-core/src/vdk/internal/core/errors.py
DeltaMichael: The idea is you pass a message to log at warn level, along with the exception. We use this three times in vdk-core, total, so we can re-visit it later when we audit all the log statements.

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -13,25 +13,107 @@
...
+        self._vdk_type = vdk_type
+        self._prettify_message(str(error_message))
+
+    def __str__(self):
+        return self._pretty_message
+
+    def __repr__(self):
+        return self._pretty_message
+
+    def _prettify_message(self, message):
```
duyguHsnHsn, line 66: I think this method is a little bit confusing maybe adding a docstring and adding a small example would be good here since we have a lot of magic numbers and formatting

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -13,25 +13,107 @@
...
+
+    When this exception is thrown from a data job python step, all the other steps will
+    be skipped and the data job execution will exit and be marked as success.
+    """
+
+    def __init__(self, *error_message_lines: str):
+        super().__init__(None, None, *error_message_lines)
+
+
+# RESOLVABLE CONTEXT
```
duyguHsnHsn, line 116: IMO, we should consider refactoring the  file for clearer code organisation. Currently, it contains both error definitions and the logic for error resolution and responsibility assignment.
IMO, keeping only error definitions and related utilities as 'report and rethrow' would make the module a lot easier to read and understand. We can move the error-resolving logic and classes to a new module  named 'error_resolvers.py' or something similar.
By doing this separation, we will make our codebase more intuitive, navigable (tbh reading the file was pretty hard), and maintainable, ensuring that each module has a single, clear responsibility, which will be beneficial for the development team in the long run.

PS: We might not be doing it because of compatibility issues, I am not sure but i do think maybe later on this should be changed

duyguHsnHsn (approved):
Good work! I put some comments but they are not much important

projects/vdk-core/src/vdk/internal/core/errors.py
DeltaMichael: I was thinking the same thing initially, but there's a lot of `errors.ResolvableBy` and `errors.log_and_throw` and similar statements throughout plugin code, so I decided against it. I'll add it to https://github.com/vmware/versatile-data-kit/issues/2680 as a suggestion, or open a separate ticket.

projects/vdk-core/src/vdk/internal/core/errors.py
DeltaMichael: Added docstring.

PR #2688, opened 2023-09-25
vdk-jupyter: improve init message

antoniivanov:
It seems the instructions to run the first cell are being missed often.

So I've created a markdown cell where we can put in bold the instructions so they are more noticeable.

Beyound that since we always "edit" a new notebook if hte plugin is installed. It's good idea to explain why so if a user create a notebook but doesn't know the plugin is installed , they understand why do they have all those extra things inside.

Also added an empty cell so user can get stared. Without an empty cell and , there are only 2 non-editable cells and if the user doesn't know how to create a new cell or just tries to edit the non-ediable cells, he would be very confused what to do.

Screenshot: 
image

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/initVDKConfigCell.ts
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/initVDKConfigCell.ts
@@ -33,6 +33,7 @@ export function initVDKConfigCell(notebookTracker: INotebookTracker) {
         cell: {
           cell_type: 'code',
           source: [
+            '"""\n',
             'vdk.plugin.ipython extension introduces a magic command for Jupyter.\n',
             'The command enables the user to load VDK for the current notebook.\n',
             'VDK provides the job_input API, which has methods for:\n',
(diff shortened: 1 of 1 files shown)


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/initVDKConfigCell.ts
```diff
@@ -11,34 +11,74 @@ import { INotebookTracker } from '@jupyterlab/notebook';
...
-            `VDK provides the job_input API, which has methods for:\n`,
-            `    * executing queries to an OLAP database;\n`,
-            `    * ingesting data into a database;\n`,
-            `    * processing data into a database.\n`,
-            `Type help(job_input) to see its documentation.\n`,
-            `"""\n`,
-            `%reload_ext vdk.plugin.ipython\n`,
-            `%reload_VDK\n`,
-            `job_input = VDK.get_initialized_job_input()`
+            'vdk.plugin.ipython extension introduces a magic command for Jupyter.\n',
```
duyguHsnHsn, line 36: When i try to run this cell for me it says:
vdk.plugin.ipython extension introduces a magic command for Jupyter.
                       ^
SyntaxError: invalid syntax

It is still a python cell shouldn't we have the text commented?

duyguHsnHsn (approved):

sabadzhiev (approved):

gageorgiev (approved):

PR #2699, opened 2023-09-26
vdk-jupyter: fix conversion

duyguHsnHsn:
What:
Fixed the conversion operation to add additional cell which runs the run function coming from .py step. 

Before:
notebook_steps

After: 

notebook_steps

Why:
To execute the python code in the function vdk-notebook requires it to be called (unlike how vdk generally works with .py steps).
The reason the bug was not caught is that it is only tested in the e2e tests and they are not yet added to the ci/cd.


antoniivanov:
So the e2e tests would have caught it? 

Consider still adding more smaller scope tests if it's not too hard.

gageorgiev (approved):

duyguHsnHsn:
> So the e2e tests would have caught it?
> 
> Consider still adding smaller scope tests if it's not too hard.

We have tried it but there are some problems with some jupyter classes we need to use during testing with jest. They required full mocking of the hierarchy tree so that's why we ended up having only e2e tests

PR #2706, opened 2023-09-26
vdk-dag, vdk-control-cli, airflow-provider-vdk: step using deprecated field

antoniivanov:
cache_locally was deprecated in favour of credentials_cache
    
Since the credentials_cache default is LocalFolderCredentialsCache we can leave the default 
    
This is really 3 changes in single PR - but the change would be merged in 3 separate commits for each plugin

Testing Done: existing tests passed, showing no regression


doks5 (approved):
Maybe you can add a note to the description how was this change tested?

duyguHsnHsn (approved):

PR #2709, opened 2023-09-27
vdk-notebook: set summary file path as configuration

antoniivanov:
Previously, the summary file path was hard-coded to the job directory, posing issues for certain deployments where writing to the job directory was restricted. This led to failures in some deployments.

The second problem. This design introduced a hidden and unsafe dependency between the vdk-notebook and vdk-jupyter plugins. The vdk-notebook plugin was responsible for writing the summary information, while vdk-jupyter read from the same hard-coded location, although both plugins operated independently.

This commit migrates the summary file path logic to a more appropriate location.The functionality to output summary of the job in a way that differnet consumers (plugins or tools) can read and understand is rather generic and core functionality regardless if vdk-notebook is installed or not.

After merging this would change vdk-jupyter to work with this and remove the obsolete code in vdk-noteboook


duyguHsnHsn (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/summary_output.py
```diff
@@ -0,0 +1,67 @@
...
+        config_builder.add(
+            key=JOB_RUN_SUMMARY_FILE_PATH,
+            default_value="",
+            description="The path location of the file where job run result summary is stored.",
+        )
+
+    @hookimpl(hookwrapper=True)
+    def run_job(self, context: JobContext) -> None:
+        out: HookCallResult
+        out = yield
```
DeltaMichael, line 35: Nit: This pulls the execution result out, right? Might be worth adding a comment.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
```diff
@@ -130,6 +131,7 @@ def vdk_start(plugin_registry: PluginRegistry, command_line_args: List) -> None:
     plugin_registry.add_hook_specs(JobRunHookSpecs)
     plugin_registry.load_plugin_with_hooks_impl(JobConfigIniPlugin())
     plugin_registry.load_plugin_with_hooks_impl(TerminationMessageWriterPlugin())
+    plugin_registry.load_plugin_with_hooks_impl(JobRunSummaryOutputPlugin())
```
DeltaMichael, line 134: Are we outputting the result to a file for every job? It's ok if we are, but we should probably document the behaviour.

DeltaMichael (approved):
Looks ok, please fix the Codacy warnings.

projects/vdk-core/src/vdk/internal/builtin_plugins/builtin_hook_impl.py
antoniivanov: Only if enabled (summary output location is set).

PR #2710, opened 2023-09-28
control-service: update DataJobsController with account deployment data

mrMoZ1:
what: Updated the DataJobsController and and associated service code, so that when a deployment is made or updated, the relevant data is written to the DB.

why:
Part of the Data Job configuration persistence initiative.

testing:
added test

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsController.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsController.java
@@ -18,6 +18,9 @@ import com.vmware.taurus.service.credentials.JobCredentialsService;
 import com.vmware.taurus.service.upload.JobUpload;
 import graphql.GraphQLError;
 import io.swagger.v3.oas.annotations.tags.Tag;
+import java.net.URI;
+import java.util.List;
+import java.util.function.Supplier;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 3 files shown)


mivanov1988:
I don't see the usage of the new Feature flag that you have introduced.

mivanov1988:
Does it make sense to create or update deployment when a job is created or modified?

mrMoZ1:
> Does it make sense to create or update deployment when a job is created or modified?

To me it does make sense to update the deployment when a job is modified - otherwise I think we would override it with a null value. Maybe it doesn't make sense to create an empty deployment when a job is created though.

mivanov1988:
> > Does it make sense to create or update deployment when a job is created or modified?
> 
> To me it does make sense to update the deployment when a job is modified - otherwise I think we would override it with a null value. Maybe it doesn't make sense to create an empty deployment when a job is created though.

Which fields will be updated when the job is modified?

mrMoZ1:
> > > Does it make sense to create or update deployment when a job is created or modified?
> > 
> > 
> > To me it does make sense to update the deployment when a job is modified - otherwise I think we would override it with a null value. Maybe it doesn't make sense to create an empty deployment when a job is created though.
> 
> Which fields will be updated when the job is modified?

When the update methods of the DataJobController class are called through the API there isn't any deployment information, that's why we only make sure not to override any existing deployment info.

mrMoZ1:
Closing since with recent changes DataJobs don't have an associated deployment entity anymore.

PR #2713, opened 2023-09-28
vdk-jupyter: update getting started to incldue vdksql

antoniivanov:
Update Getting started to include info on how to use `%%vdksql` cells

See the updated notebook here: https://github.com/vmware/versatile-data-kit/blob/person/aivanov/jupyter-getting-started/projects/vdk-plugins/vdk-jupyter/getting-started.ipynb


projects/vdk-plugins/vdk-jupyter/getting-started.ipynb
```diff
@@ -192,13 +192,26 @@
...
+    "\n",
+    "<h4>Python Cells</h4>\n",
+    "\n",
+    "Within the notebook, VDK provides an object - job_input, that has methods for executing queries, ingesting data, handling properties and more.<br/>\n",
+    "Type help(job_input) to learn more about it\n",
+    "\n",
+    "<h4>SQL Cells</h4> \n",
+    "\n",
+    "You can create SQL-only Cells as well by starting the cell with `%%vdksql` magic: <br/>\n",
+    "Similar to SQL steps propertie and arguments are automatically replaced (for example `{db}`)\n",
```
antoniivanov, line 207:
```suggestion
    "Similar to SQL steps properties and arguments are automatically replaced (for example `{db}`)\n",
```

duyguHsnHsn:
If we are going to encourage people to use this instead of job_input.execute_query maybe it is good to change it in the conversion, too. What do you think? Because right now the conversion Guide, Tutorial and the generated code all include ob_input.execute_query. And if we are not going to change that maybe it would be good in the getting started to show that option too

antoniivanov:
> If we are going to encourage people to use this instead of job_input.execute_query maybe it is good to change it in the conversion, too. What do you think? Because right now the conversion Guide, Tutorial and the generated code all include ob_input.execute_query. And if we are not going to change that maybe it would be good in the getting started to show that option too

Yes, we will update everywhere. This is just for getting started.

duyguHsnHsn (approved):

dakodakov (approved):

PR #2717, opened 2023-09-28
vdk-notebook: [bug fix] ignore missing id field in cell

doks5:
Currently, when the cells of an IPython notebook are loaded, it is assumed that each cell has an `id` field. However, this is true only for vdk specific notebooks, not regular ones that users may use to execute vdk commands like create, deploy, etc. In cases, where the notebook is a regular one, the job execution fails abruptly at the plugin loading stage without any meaningful message, because only json-specific errors are being handled.

Example job termination logs:

```bash
2023-09-28 19:55:51,215 [VDK] test-job [ERROR] vdk.internal.builtin_plugins.r   cli_run.py:172
create_and_run_d[id:510b0034-2249-4840-8589-1c9a3f58bc59-1695920149-24c96]- Failed executing job.
'id'
The provided Data Job will not be executed. Terminating application.
See contents of the exception and fix the problem that causes it. Most likely a prerequisite
or plugin of one of the key VDK components failed, see logs for details and ensure the
prerequisite for the failed component (details in stacktrace).

Process finished with exit code 1
```

This change makes the `id` field not required, so that regular notebooks can be ignored and not fail job executions.

Testing Done: Locally with a test data job that contained a regular IPython notebook and other regular python steps.

### DIFF
--- a/projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
+++ b/projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
@@ -17,7 +17,7 @@ class Cell:
     def __init__(self, jupyter_cell):
         self.tags = jupyter_cell["metadata"].get("tags", {})
         self.source, self.source_type = self.__extract_source_code(jupyter_cell)
-        self.id = jupyter_cell["id"]
+        self.id = jupyter_cell.get("id")
 
     @staticmethod
     def __extract_source_code(jupyter_cell):


antoniivanov (commented):
Please add a test with regular notebook and non regular

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
```diff
@@ -17,7 +17,7 @@ class Cell:
     def __init__(self, jupyter_cell):
         self.tags = jupyter_cell["metadata"].get("tags", {})
         self.source, self.source_type = self.__extract_source_code(jupyter_cell)
-        self.id = jupyter_cell["id"]
+        self.id = jupyter_cell.get("id")
```
DeltaMichael, line 20: This field isn't used for regular notebooks, right? I'm just trying to understand the implications if it's set to `None`

duyguHsnHsn:
Every cell has an id if it is a Jupyter notebook. We do not add ids, we only add vdk tag. 
I cannot understand what the difference between regular and non-regular can be

DeltaMichael:
> Every cell has an id if it is a Jupyter notebook. We do not add ids, we only add vdk tag. I cannot understand what the difference between regular and non-regular can be

@duyguHsnHsn I'm not sure how accurate this is, given that they're facing the problem in the description. I don't see a key error in the output, however. It might be a different problem altogether.

```
self.id = jupyter_cell["id"]
```

Accessing it like this will raise a `KeyError` if "id" is not present, using get will just return `None` or a specified default. We should still use the get method in any case.

duyguHsnHsn:
> > Every cell has an id if it is a Jupyter notebook. We do not add ids, we only add vdk tag. I cannot understand what the difference between regular and non-regular can be
> 
> @duyguHsnHsn I'm not sure how accurate this is, given that they're facing the problem in the description. I don't see a key error in the output, however. It might be a different problem altogether.
> 
> ```
> self.id = jupyter_cell["id"]
> ```
> 
> Accessing it like this will raise a `KeyError` if "id" is not present, using get will just return `None` or a specified default. We should still use the get method in any case.

@DeltaMichael  Yes, i agree that using the get function is the better option. But I was just curious about the use case since by default all cells are created with ids (or so says the documentation)

DeltaMichael (approved):

duyguHsnHsn (approved):

projects/vdk-plugins/vdk-notebook/src/vdk/plugin/notebook/cell.py
doks5: Back when we did the investigation, the client data job we tested with had an .ipynb file which was causing the error. We couldn't figure out why that specific notebook did not have an id.

PR #2728, opened 2023-09-30
vdk-core: pass execution result through context

DeltaMichael:
## Why?

We'd like to have just one exit point for data jobs where error logging happens and propagate all errors up to it.

## What?

Pass execution result in job context when exceptions are thrown. This way, we still get to throw the exception and have the result at a later hook, e.g. we have the execution result from run_job inside the run method, even if an exception was thrown during the run. If an exception was thrown, we can catch and log it at the exit point.

The side effect is that plugins have to check the job context for hook implementations that are wrapping if an exception was thrown.

## How was this tested?

Ran tests locally
Ran plugin tests in CI/CD

Potentially breaking change

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
@@ -54,6 +54,32 @@ class DataJobFactory:
         return DataJob(data_job_directory, core_context, name)
 
 
+def save_step_result(context: JobContext, result: StepResult):
+    step_results = context.core_context.state.get(CommonStoreKeys.STEP_RESULTS)
+    step_results.append(result)
+    context.core_context.state.set(CommonStoreKeys.STEP_RESULTS, step_results)
+    return result
+
+
+def save_execution_result(
+    context: JobContext, start_time=None, exception=None, blamee=None
+):
+    execution_status = context.core_context.state.get(CommonStoreKeys.EXECUTION_STATUS)
+    step_results = context.core_context.state.get(CommonStoreKeys.STEP_RESULTS)
+    execution_result = ExecutionResult(
+        context.name,
+        context.core_context.state.get(CommonStoreKeys.EXECUTION_ID),
+        start_time,
+        datetime.utcnow(),
+        execution_status,
+        step_results,
+        exception,
+        blamee,
+    )
+    context.core_context.state.set(CommonStoreKeys.EXECUTION_RESULT, execution_result)
+    return execution_result
+
+
 class DataJobDefaultHookImplPlugin:
     """
     Default implementation of main plugin hooks in Data Job Run Cycle.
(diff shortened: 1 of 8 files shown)


projects/vdk-core/tests/vdk/internal/builtin_plugins/run/test_data_job.py
```diff
@@ -81,22 +84,24 @@ def test_run_not_from_template(mock_store_set):
...
-        len(
-            [
-                c
-                for c in mock_store_set.call_args_list
-                if c[0][0] == StoreKey(key="vdk.template_name")
-            ]
-        )
-        == 1
-    )
+# TODO: Find a way to test this without mocking the state store
```
DeltaMichael, line 87: Not sure if we should even do this at all. This test has very little value, in my opinion.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -115,7 +116,13 @@ def run_step(context: JobContext, step: Step) -> None:
         out: HookCallResult
         out = yield
 
-        exception = out.get_result().exception
+        result: StepResult
+        if out.excinfo:
+            result = context.core_context.state.get("step_results")[-1]
```
DeltaMichael, line 121: I'm not sure if and where I should document this behaviour. It might be relevant for plugins that want to hook to `run_step` and get the step result. It's a bit convoluted.

antoniivanov (commented):
The idea looks good to me. exposing this information at the exit point and printing it there is better since we have better control over how to print it.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -54,6 +54,32 @@ def new_datajob(
         return DataJob(data_job_directory, core_context, name)
 
 
+def save_step_result(context: JobContext, result: StepResult):
+    step_results = context.core_context.state.get("step_results")
```
antoniivanov, line 58: I belive StateStore was using StoreKey as type and not string . But I assume this is for prototyping urposes.

gageorgiev:
can you explain how this is a potentially breaking change?

DeltaMichael:
> can you explain how this is a potentially breaking change?

Older versions of plugins that run against this version of core might not work correctly, because hook results are handled differently when an exception is thrown.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
DeltaMichael: Fair point, I'll add some new values to that ExecutionStoreKeys class and use them instead.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -54,6 +54,32 @@ def new_datajob(
...
+):
+    execution_status = context.core_context.state.get(CommonStoreKeys.EXECUTION_STATUS)
+    step_results = context.core_context.state.get(CommonStoreKeys.STEP_RESULTS)
+    execution_result = ExecutionResult(
+        context.name,
+        context.core_context.state.get(CommonStoreKeys.EXECUTION_ID),
+        start_time,
+        datetime.utcnow(),
+        execution_status,
+        step_results,
```
antoniivanov, line 75: We keep step_results in 2 places. 


Why not build ExecutionResult over time. 

when job starts we add it almost empty -just name, id, start time ,  
you can add the steps as they happen ?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_plugin.py
```diff
@@ -115,7 +117,13 @@ def run_step(context: JobContext, step: Step) -> None:
         out: HookCallResult
         out = yield
 
-        exception = out.get_result().exception
+        result: StepResult
+        if out.excinfo:
+            result = context.core_context.state.get(CommonStoreKeys.STEP_RESULTS)[-1]
+        else:
+            result = out.get_result()
```
antoniivanov, line 124: Replacing one line with 5 is clearly regression/worse than before. I don't think it's a good idea.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -54,6 +54,32 @@ def new_datajob(
         return DataJob(data_job_directory, core_context, name)
 
 
+def save_step_result(context: JobContext, result: StepResult):
```
antoniivanov, line 57: I think we should encapsulate things in a class . there's a lot of "knowledge" require to use this effectively

antoniivanov (changes requested):
Don't make changes in vdk-core and plugins in the same PR. 

Generally I discourage changing multiple plugins/projects in same PR unless the change is trivial (bumping version of library x in n plugins)

First is the change in core (non-breaking) then it is its adoption in separate PR in the plugins.

DeltaMichael:
> Don't make changes in vdk-core and plugins in the same PR.
> 
> Generally I discourage changing multiple plugins/projects in same PR unless the change is trivial (bumping version of library x in n plugins)
> 
> First is the change in core (non-breaking) then it is its adoption in separate PR in the plugins.

@antoniivanov I can't think of a way to make the `vdk-core` change non-breaking. It will always break backward-compatibility, because we have a whole different case where we pass execution results in the context and it needs to be handled by plugins. It makes more sense to change the plugins together with core.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
DeltaMichael: The two functions are only meant to be used inside that file. I can make them class members of DataJob, it'll be less likely someone imports them in another file.

DeltaMichael:
Closing this until I can make it backwards-compatible.

PR #2730, opened 2023-10-02
vdk-jupyter: Integration test CI step 

gageorgiev:
This PR modifies the CICD of vdk-jupyter to run the integration
tests.


projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -59,6 +59,20 @@ frontend-tests-vdk-jupyterlab-extension:
     reports:
       junit: "projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/frontend_tests_results.xml"
 
+integration-tests-vdk-jupyterlab-extension:
+  extends: .common-setup-vdk-jupyterlab-extension
+  stage: pre_release_test
```
antoniivanov, line 64: You know that you can change the image wiht `image:` tag, right?

  The default image use is `python:3.8-alpine` I can see from teh logs. 

But the warning in the logs says that it isn't supported. 
So you can change it to normal Ubuntu image or something else that is supported by playwright

gageorgiev: apparently the 'not supported' warning isn't a massive issue according to what I read

projects/vdk-plugins/vdk-jupyter/.plugin-ci.yml
```diff
@@ -59,6 +59,35 @@ frontend-tests-vdk-jupyterlab-extension:
     reports:
       junit: "projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/frontend_tests_results.xml"
 
+integration-tests-vdk-jupyterlab-extension:
+  image: "mcr.microsoft.com/playwright:v1.34.0-jammy"
+  # extends: .common-setup-vdk-jupyterlab-extension
+  stage: pre_release_test
+  script:
+    - export TZ=Europe/Minsk
+    - ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
yonitoo, line 68: Why do we need to set the timezone to ours? AFAIK, we don't do that in other CI files. I think it might potentially cause some unwanted time discrepancies in the future.

yonitoo (commented):
Left some comments, otherwise looks good.

duyguHsnHsn:
Idk, whether this will be usefull but u can check https://github.com/jupyterlab/jupyterlab-git/actions/runs/5753409851/workflow

gageorgiev:
closing in favor of #2868

PR #2747, opened 2023-10-02
vdk-impala: add Out Of Memory error handling

dakodakov:
Add Out Of Memory error handling logic in the vdk-impala error handler.


projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -441,3 +444,78 @@ def _is_pool_error(exception: Exception) -> bool:
             classname_with_package=".*OperationalError.*",
             exception_message_matcher_regex=".*admission for timeout ms in pool.*",
         )
+
+    def _handle_memory_error(self, exception, recovery_cursor):
```
antoniivanov, line 448: This is pretty long logic . We are adding more than 100 lines. I think it would be better to encapsulate the logic in separate class OutOfMemoryErrorHandler for example   . And here call it with 2 or 3 lines onlyh.

There you can also write a good nice pydoc in that new class explaining all the intricacies and the whys and the whats of your solution.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -7,6 +7,8 @@
 from vdk.internal.builtin_plugins.connection.recovery_cursor import RecoveryCursor
 from vdk.internal.core import errors
 
+MEMORY_LIMIT_PATTERN = r"Limit=(\d+\.\d+)\s*([KMGTP]B)"
```
antoniivanov, line 10: TB and PB , Considering that is memory limit per node , that would be impressive query :) 

It doesn't hurt to have. Thinking about the future

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -441,3 +444,78 @@ def _is_pool_error(exception: Exception) -> bool:
...
+            recovery_cursor.execute(f"set mem_limit={new_memory_limit};")
+        else:
+            self._log.warning(
+                f"Unable to determine current memory limit for statement."
+            )
+
+    @staticmethod
+    def _convert_to_bytes(value: str, unit: str):
+        units = {"KB": 1024, "MB": 1024**2, "GB": 1024**3, "TB": 1024**4}
+        return int(float(value) * units.get(unit, 1))
```
antoniivanov, line 508: That probably should go in some util class

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -441,3 +444,78 @@ def _is_pool_error(exception: Exception) -> bool:
...
+
+    def _get_new_memory_limit(self, exception, multiplier: float):
+        memory_limit_match = re.search(MEMORY_LIMIT_PATTERN, str(exception))
```
antoniivanov, line 511: In my experience, (far from big providedly) parsing is one of the biggest source of bugs. I always try to make sure I follow a spec or a contract or something stable. 

In this case Impala is printing this in https://github.com/apache/impala/blob/master/be/src/runtime/mem-tracker.cc#L314 

using PrettyPrinter whose implementation is [here](https://github.com/apache/impala/blob/ae14d78c8f2e8366e67aa8c39f0c02c60862905e/be/src/util/pretty-printer.h#L127) 

And from git blame I can see that code has not changed for 6-9 years. So we are probably ok. 


Also the pattern likely won't catch false positive. At least ([searching in the code](https://github.com/search?q=repo%3Aapache%2Fimpala+%22Limit%3D%22&type=code&p=5) that seems the only "Limit=" printed

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -441,3 +444,78 @@ def _is_pool_error(exception: Exception) -> bool:
...
+            )
+            recovery_cursor.execute(f"set mem_limit={new_memory_limit};")
+        else:
+            self._log.warning(
+                f"Unable to determine current memory limit for statement."
+            )
+
+    @staticmethod
+    def _convert_to_bytes(value: str, unit: str):
+        units = {"KB": 1024, "MB": 1024**2, "GB": 1024**3, "TB": 1024**4}
```
antoniivanov, line 507: nit: this could be constant.

antoniivanov (commented):
Looks good to me! My comments are mostly "clean code" type of comments.

antoniivanov (commented):
I am wondering if we want to write a functional test that simulates out of memory query. That might not be easy. 

Regardless you should test it against real system and describe how you test it in Testing Done in the PR.

mivanov1988 (approved):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -441,3 +444,76 @@ def _is_pool_error(exception: Exception) -> bool:
...
+
+            # Incrementally increase the memory limit and as a last resort try to set the memory to an extreme value
+            if recovery_cursor.get_retries() > 3:
+                #  We won't be able to handle this error by increasing limits
+                return False
+            if recovery_cursor.get_retries() > 2:
+                # An extreme case but let's try it as a last resort
+                recovery_cursor.execute("set memory_limit=512GB;")
+            elif recovery_cursor.get_retries() > 1:
+                self._update_memory_limit(exception, recovery_cursor, 2.0)
```
mivanov1988, line 480: Are the multipliers based on the current queries that we have internally?

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -107,6 +109,7 @@ def _handle_exception(self, exception, recovery_cursor) -> bool:
             or self._handle_metadata_exception_with_invalidate_and_retry(
                 exception, recovery_cursor
             )
+            or self._handle_memory_error(exception, recovery_cursor)
```
DeltaMichael, line 112: There are better patterns for this kind of error handling. You can do some kind of handler chain here. Doesn't have to be in this PR, but it wouldn't hurt to open a refactoring ticket.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
```diff
@@ -441,3 +444,76 @@ def _is_pool_error(exception: Exception) -> bool:
...
+                return False
+
+            # Incrementally increase the memory limit and as a last resort try to set the memory to an extreme value
+            if recovery_cursor.get_retries() > 3:
+                #  We won't be able to handle this error by increasing limits
+                return False
+            if recovery_cursor.get_retries() > 2:
+                # An extreme case but let's try it as a last resort
+                recovery_cursor.execute("set memory_limit=512GB;")
+            elif recovery_cursor.get_retries() > 1:
```
DeltaMichael, line 479: There should be a better way to do this. At the very minimum, you can put the threshold values in a list and do something like

```
thresholds = [1.2, 1.5, 2.0, 512]

self._update_memory_limit(exception, recovery_cursor, thresholds[recovery_cursor.get_retries()])
```

And then handle cases where we've done more than 3 retries inside `_update_memory_limit`.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
dakodakov: Done.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
dakodakov: I based the logic on several exceptions I reviewed from our internal system and this works best.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
dakodakov: Done.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
dakodakov: These multipliers were discussed during the testing with the Infra team and we agreed that they should work best. I don't think there's a "perfect" multiplier here.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
dakodakov: fixed.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
dakodakov: I would like to focus this PR on this specific issue and honestly I won't have the time to refactor this code or I don't think anyone else would "just cause"...

dakodakov:
> I am wondering if we want to write a functional test that simulates out of memory query. That might not be easy.
> 
> Regardless you should test it against real system and describe how you test it in Testing Done in the PR. (hint: you probably know but you can install it in your existing prod or test vdk env with `pip install -e <vdk-impal-dir>`)

I tested these changes on our testing impala instance together with our Infra team and the logic for this change is the result of our testing and research we did around these issues and the Impala memory configuration and settings.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_error_handler.py
antoniivanov: Why do they work best? What is the logic ?

antoniivanov (approved):
Thanks. Looks good to me.

antoniivanov:
> I tested these changes on our testing impala instance together with our Infra team and the logic for this change is the result of our testing and research we did around these issues and the Impala memory configuration and settings.

That was a bit unspecific. When something is tested manually, it's really good idea to describe it in a way that make the test reproducible by someone else. This way it can be peer-reviewed as well better.

For example. What is the configuration of the testing Impala instance used, what types of queries were executed, what specific memory and performance metrics were tracked and used for verification. In general you can summarize the research that led to the logic implemented for handling memory issues.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_memory_error_handler.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import logging
+import re
+
+from vdk.internal.core import errors
+
+MEMORY_LIMIT_PATTERN = r"Limit=(\d+\.\d+)\s*([KMGTP]B)"
+UNITS = {"KB": 1024, "MB": 1024**2, "GB": 1024**3, "TB": 1024**4}
+THRESHOLDS = [1.2, 1.5, 2.0, 512]
```
mivanov1988, line 10: Should 512 be part of the thresholds?

DeltaMichael (approved):

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_memory_error_handler.py
DeltaMichael: I proposed it as some kind of huge multiplier instead of doing

```
recovery_cursor.execute("set memory_limit=512GB;")
```

Not sure if it works the way I think it does, though, please use your best judgement. 😄

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_memory_error_handler.py
```diff
@@ -0,0 +1,105 @@
...
+                #  We won't be able to handle this error by increasing limits
+                return False
+            if recovery_cursor.get_retries() == 3:
+                # An extreme case but let's try it as a last resort
+                recovery_cursor.execute("set memory_limit=512GB;")
+            else:
+                self._update_memory_limit(
+                    exception,
+                    recovery_cursor,
+                    THRESHOLDS[recovery_cursor.get_retries()],
```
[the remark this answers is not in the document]
dakodakov: fixed.

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_memory_error_handler.py
dakodakov: This was a mistake... it's now fixed.

Issue #2682, opened 2023-09-21
[Timeboxed] Replace generic vdk errors with more domain-specific errors


DeltaMichael:
### Overview

DomainError, UserCodeError, VdkConfigurationError and PlatformServiceError should be replaced with more domain-specific exceptions 

**Note:** DomainError is used somewhere in VMWare internal vdk deployments, so we might have to leave it alone for now
Not sure if we should be replacing VdkConfigurationError either

### Acceptance criteria

1. Replace generic errors with more specific errors

PR #2752, opened 2023-10-03
vdk-core: create ingestion exceptions

antoniivanov:
Replacing generic exception names like DomainError, UserCodeError, , and PlatformServiceError with more domain-specific exceptions (in this case the domain being ingestion) can offer differnet benefits

See https://github.com/vmware/versatile-data-kit/issues/2682

VdkConfigurationError is not changed as it's pretty clear for now. I am wondering if it still might not be better ot have IngsetionConfigurationException but could not decide and omit it from this change.

TODO: I still need to test with "real" like job that fails in some of those ways and see what is the difference in logs and in how the exception can be handled

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/exception.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/exception.py
@@ -1,0 +1,181 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from typing import Any
+from typing import Optional
+
+from vdk.internal.core.errors import BaseVdkError
+from vdk.internal.core.errors import ResolvableBy
+
+
+class IngestionException(BaseVdkError):
+    """
+    Base Exception for all custom exceptions related to the ingestion process.
+    This is intended to catch general exceptions that do not fit into more specific categories.
+    """
+
+    def __init__(self, message: str, resolvable_by: Optional[ResolvableBy] = None):
+        super().__init__(None, resolvable_by, message)
+
+
+class PayloadIngestionException(IngestionException):
+    """
+    Base Exception for all payload-related issues during the ingestion process.
+    """
+
+    def __init__(
+        self,
+        payload_id: str,
+        message: str,
+        destination_table: str = "",
+        target: str = "",
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
+        """
+        :param payload_id: ID of the payload. This is a way for user to find out which payload failed to ingest.
+                           Left empty if such information is not available
+        """
+        super().__init__(message=message, resolvable_by=resolvable_by)
+        # we are purposefully are not putting payload id
+        # in the message, so it's not logged in case it's sensitive
+        self.payload_id = payload_id
+        self.destination_table = destination_table
+        self.target = target
+
+
+class EmptyPayloadIngestionException(PayloadIngestionException):
+    """
+    Raised when an empty payload is encountered during ingestion and it is not expected.
+    """
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
+        if not message:
+            message = "Payload given to ingestion method should not be empty."
+        super().__init__(payload_id="", message=message, resolvable_by=resolvable_by)
+
+
+class InvalidPayloadTypeIngestionException(PayloadIngestionException):
+    """
+    Raised when the payload provided for ingestion has an invalid type.
+    """
+
+    def __init__(
+        self,
+        payload_id: str,
+        expected_type: str,
+        actual_type: str,
+        message: Optional[str] = None,
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
+        """
+        :param expected_type: The expected type for the payload
+        :param actual_type: The actual type of the payload
+        """
+        if not message:
+            message = "Invalid ingestion payload type."
+        super().__init__(
+            message=f"{message} Expected type: {expected_type}, actual type: {actual_type}.",
+            payload_id=payload_id,
+            resolvable_by=resolvable_by,
+        )
+
+
+class JsonSerializationIngestionException(PayloadIngestionException):
+    """
+    Raised when a payload is not JSON-serializable during ingestion.
+    """
+
+    def __init__(
+        self,
+        payload_id: str,
+        original_exception: Exception,
+        message: str = "",
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
+        """
+        :param original_exception: The original exception triggering this error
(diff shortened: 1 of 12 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/exception.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    Base Exception for all custom exceptions related to the ingestion process.
+    This is intended to catch general exceptions that do not fit into more specific categories.
+    """
+
+    def __init__(self, message: str, vdk_type: ResolvableBy = None):
+        super().__init__(None, vdk_type, message)
+
+
+class IngestionConfigurationException(IngestionException):
```
antoniivanov, line 20: For now, I've kept VDKConfigurationException. But at some point I wondered if I want to have IngestionConfiguraitonException instead. What do you , reviewers think? 

I will remove this exception if we decide to stick to VDKConfigurationException

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/exception.py
```diff
@@ -0,0 +1,184 @@
...
+        message: str,
+        destination_table: str = "",
+        target: str = "",
+        vdk_type: ResolvableBy = None,
+    ):
+        """
+        :param payload_id: ID of the payload. This is a way for user to find out which payload failed to ingest.
+                           Left empty if such information is not available
+        """
+        super().__init__(message=message, vdk_type=vdk_type)
```
antoniivanov, line 46: Another thing I've been wondering. 

Do I want to pass vdk_type in each exception class ? 
Or do I just inject it dynamically in report_and_throw 

But not all exception are raised using report_and_throw perhaps.

The way ingestion works so far (and I haven't changed that), report_and_throw is used for exception at the aggregate level that are called on the main thread (e.g close()) but for code running in background threads , report_and_throw is not used (this avoids issues with concurrent access over globals in report_and_throw)

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/exception.py
```diff
@@ -0,0 +1,184 @@
...
+class IngestionConfigurationException(IngestionException):
+    """
+    Raised when there are configuration-related issues during the ingestion process.
+    """
+
+    def __init__(self, message: str, vdk_type: ResolvableBy = None):
+        super().__init__(message=message, vdk_type=vdk_type)
+
+
+class PayloadIngestionException(IngestionException):
```
DeltaMichael, line 29: I don't think these should have "Ingestion" in the name. It's enough that they extend `Ingestion Exception` or `PayloadIngestionException`.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/exception.py
DeltaMichael: We don't use `report_and_throw` for non-vdk exceptions. You probably won't find a code snippet like

```
report_and_throw(new ValueError("My message, whatever"))
```

> But not all exception are raised using report_and_throw perhaps.

This is correct.

All in all, it's not a bad idea to pass the type in both cases.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -512,12 +526,8 @@ def _payload_poster_thread(self):
...
                         exception=exception,
                     )
-            except VdkConfigurationError:
-                self._plugin_errors[VdkConfigurationError].increment()
-            except UserCodeError:
-                self._plugin_errors[UserCodeError].increment()
-            except Exception:
-                self._plugin_errors[PlatformServiceError].increment()
+            except Exception as e:
+                self._plugin_errors[type(e)].increment()
```
DeltaMichael, line 530: Not sure this will work as intended. We do this later on

```
    def __handle_results(self):
        if self._plugin_errors.get(UserCodeError, AtomicCounter(0)).value > 0:
            errors.report_and_throw(
                UserCodeError(
                    "Failed to post all data for ingestion successfully.",
                    "Some data will not be ingested.",
                    "Ensure data you are sending is aligned with the requirements",
                    "User error occurred. See warning logs for more details. ",
                )
            )
        if self._plugin_errors.get(VdkConfigurationError, AtomicCounter(0)).value > 0:
            errors.report_and_throw(
                VdkConfigurationError(
                    "Failed to post all data for ingestion successfully.",
                    "Some data will not be ingested.",
                    "Ensure job is properly configured. "
                    "For example make sure that target and method specified are correct",
                )
            )
        if (
            self._plugin_errors.get(PlatformServiceError, AtomicCounter(0)).value > 0
            or self._fail_count.value > 0
        ):
            errors.report_and_throw(
                PlatformServiceError(
                    "Failed to post all data for ingestion successfully.",
                    "Some data will not be ingested.",
                    "There has been temporary failure. ",
                    "You can retry the data job again. ",
                    "If error persist inspect logs and check ingest plugin documentation.",
                )
            )
```

It might be a better idea to map by exception type, collect the exceptions for the appropriate type and decide what to do with them.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
```diff
@@ -148,7 +148,7 @@ def send_tabular_data_for_ingestion(
             )
         else:
             errors.report_and_throw(
-                UserCodeError(
+                VdkConfigurationError(
```
DeltaMichael, line 151: What's the blamee listed as here for local runs?

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -623,63 +639,47 @@ def _execute_post_process_operations(
...
-        ):
-            errors.report_and_throw(
-                PlatformServiceError(
-                    "Failed to post all data for ingestion successfully.",
-                    "Some data will not be ingested.",
-                    "There has been temporary failure. ",
-                    "You can retry the data job again. ",
-                    "If error persist inspect logs and check ingest plugin documentation.",
-                )
+        if self._plugin_errors:
```
DeltaMichael, line 660: Do we still need `_plugin_errors` to be a map in this case?

DeltaMichael (approved):

gageorgiev (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
antoniivanov: for local run it's user error, for cloud, platform. as other vdk configuration errors

PR #2754, opened 2023-10-03
vdk-core: add error formatter configuration

DeltaMichael:
## Prerequisites

Stacked on top of https://github.com/vmware/versatile-data-kit/pull/2774

## Why?
    
Different exception message formats are required for different use cases, e.g. CLI vs. Jupyter vs. Cloud
    
## What?
    
Add a flag to vdk-core configuration that lets the user pick the formatter. Initial choices are "pretty" and "plain". The plain
format should be good enough for Jupyter and Cloud deployments. The pretty formatter is enabled by default, because CLI needs to work out of the box.
    
Formatters have been exported to fuctions in errors.py, which makes it easy to write a new formatter if necessary. Exceptions are logged only at the exit point in the `run` method, which is also where formatting takes place.
    
## How was this tested?
    
Caused some config and user errors in local runs and looked at the output
    
CI tests
    
Feature/non-breaking
    
### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
@@ -11,6 +11,7 @@ from typing import cast
 from vdk.api.job_input import IJobArguments
 from vdk.api.plugin.core_hook_spec import JobRunHookSpecs
 from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins.config.vdk_config import LOG_EXCEPTION_FORMATTER
 from vdk.internal.builtin_plugins.run.execution_results import ExecutionResult
 from vdk.internal.builtin_plugins.run.execution_results import StepResult
 from vdk.internal.builtin_plugins.run.execution_state import ExecutionStateStoreKeys
@@ -24,6 +25,7 @@ from vdk.internal.builtin_plugins.run.run_status import ExecutionStatus
 from vdk.internal.builtin_plugins.run.step import Step
 from vdk.internal.core import errors
 from vdk.internal.core.context import CoreContext
+from vdk.internal.core.errors import BaseVdkError
 from vdk.internal.core.errors import SkipRemainingStepsException
 from vdk.internal.core.statestore import CommonStoreKeys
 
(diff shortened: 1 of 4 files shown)


antoniivanov:
You can do a stacked commit.

 Change base to be the other branch . When the other branch is merged the base would automatically change to base. But this way the review shows only the person/mdilyan/error-formatter-config branch changes and not the parent

image

antoniivanov (commented):
I've been thinking I want to make the formatting pluggable. 

So that means instead of formatting in errors we will have a plugin hook to do that for us, 
 
This would enable us to build a plugin that uses some cool libraries for formatting errors like 
https://github.com/onelivesleft/PrettyErrors , https://github.com/Qix-/better-exceptions , https://github.com/cknd/stackprinter 

So I'd propose we shelve this change in favour of that.

antoniivanov:
> There is no reason we can't do this change now and open a task to make it pluggable later.

It might be redundant, on other hand might be a good alpha change to see how different formatting could work for users

antoniivanov (approved):

projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -23,6 +23,69 @@
 # ERROR TYPES
 
 
+def plain_vdk_error_formatter(ex: BaseVdkError):
+    ex._lines[0] = ex._lines[0] + "\n"
```
gageorgiev, line 27: maybe a short doc string here just for consistency's sake?

gageorgiev (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -140,11 +138,7 @@ def run_job(context: JobContext) -> ExecutionResult:
                 blamee = whom_to_blame(e, __file__, context.job_directory)
                 exception = e
                 errors.report(blamee, exception)
-                errors.log_exception(
-                    log,
-                    exception,
-                    f"Processing step {current_step.name} completed with error.",
-                )
+                log.warn(f"Processing step {current_step.name} completed with error.")
```
yonitoo, line 141: nit: isn't log.warn deprecated? Maybe switch to log.warning?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py
```diff
@@ -298,25 +292,37 @@ def run(
...
+                if config.get_value(LOG_EXCEPTION_FORMATTER) == "pretty":
+                    ex._format(errors.pretty_vdk_error_formatter)
+                else:
+                    ex._format(errors.plain_vdk_error_formatter)
             blamee = whom_to_blame(ex, __file__, job_context.job_directory)
             errors.report(blamee, ex)
-            errors.log_exception(
-                log, ex, f"Data Job {self._name} completed with error."
-            )
+            log.warn(f"Data Job {self._name} completed with error.")
```
yonitoo, line 314: nit: same deprecation concern as above.

yonitoo (approved):
LGTM. Left minor suggestions.

PR #2760, opened 2023-10-04
vdk-core: remove code duplication in ingestion router

antoniivanov:
There would some code duplication in ingester router which even cause send_object_for_ingestion and send_tabular_data to handle method different which was subtle issue.

Refactor to remove the code duplication and reduce the number of lines as a whole.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
@@ -60,38 +60,28 @@ class IngesterRouter(IIngesterRegistry, IIngester):
         target: Optional[str] = None,
         collection_id: Optional[str] = None,
     ):
-        # Use the method and target provided by customer, or load the default ones
-        # if set. `method` and `target` provided when the method is called take
-        # precedence over the ones set with environment variables.
-        method = method or self._cfg.get_value("INGEST_METHOD_DEFAULT")
-        method = method.lower() if method is not None else None
-        target = target or self._cfg.get_value("INGEST_TARGET_DEFAULT")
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
```diff
@@ -102,6 +92,7 @@ def send_object_for_ingestion(
                     f"Currently possible values are {list(self._ingester_builders.keys())}",
                 )
             )
+        return ingester
```
dakodakov, line 95: For once I agree with codacy that this is an issue.

dakodakov (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
dakodakov: Do we have tests for this class?

antoniivanov: Yes. 
 - [test_ingester_router](https://github.com/vmware/versatile-data-kit/blob/634febe37cab75ffc36cd88ef186b7c910b8e03c/projects/vdk-core/tests/vdk/internal/builtin_plugins/ingestion/test_ingester_router.py#L16)
 - [functional ingest tests](https://github.com/vmware/versatile-data-kit/tree/634febe37cab75ffc36cd88ef186b7c910b8e03c/projects/vdk-core/tests/functional/ingestion)

gageorgiev (approved):

dakodakov (approved):

PR #2769, opened 2023-10-09
vdk-core: move error classifying logic 

duyguHsnHsn:
What:
Relocated the logic for error classification from error.py to a new module, error_classifiers.py.

Why:
Over time, the error.py file has grown significantly, accumulating both deprecated and current code for error handling, reporting, and classification. This accumulation has led to a file that's lengthy and challenging to navigate, hindering both readability and maintainability.


projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -10,12 +10,7 @@
 """
 from __future__ import annotations
 
-import enum
-import logging
-import re
-from collections import defaultdict
-from enum import Enum
-from logging import Logger
+from vdk.internal.core.error_classifiers import *
```
DeltaMichael, line 13: Please make sure this isn't imported anywhere else, e.g. plugins, other core components. There are also places in the code where you might encounter something like `errors.ResolvableBy.PLATFORM_ERROR`, so please check for those as well.

I'm not sure if the change will be backwards-compatible if the stuff you moved is imported in plugins. 😕

duyguHsnHsn: As far as I know, when you move a class from one module to another but use an import statement in the original module to maintain its presence there (like using from .new_module import * in errors.py), for external code that imports package.errors.MyClass, there should be no change? I am not sure @antoniivanov what do you think?

antoniivanov: Yeah, that would work to keep backward compatible imports 

Please add a comment that indicate the purpose of this import . As we will get warnings from Codacy for unused imports so we need to know to keep it until it's needed.

antoniivanov (approved):
Looks pretty good. Thanks

projects/vdk-core/src/vdk/internal/core/errors.py
antoniivanov: Also we should plan and change those occurrences where needed.

DeltaMichael: I had no idea imports worked like that in python. Guess you learn something new every day. Thanks, guys!

DeltaMichael (approved):

projects/vdk-core/src/vdk/internal/core/error_classifiers.py
```diff
@@ -0,0 +1,313 @@
...
+ - user code-related: should be handled by the owner of the DataJob.
+"""
+from __future__ import annotations
+
+import enum
+import logging
+import re
+from collections import defaultdict
+from enum import Enum
+from logging import Logger
```
antoniivanov, line 18: nit: remove unused and unnecesary imports

antoniivanov (approved):

projects/vdk-core/src/vdk/internal/core/error_classifiers.py
duyguHsnHsn: @antoniivanov  idk why they were firstly imported like that but those imports are probably used in tests because without them the tests fail ijust wanted to put it as it was but I can move them

PR #2770, opened 2023-10-09
vdk-core: domain specific properties/secrets exceptions

antoniivanov:
The only one in properties and secrest that make sense is about the write preprocessing sequence. The other are configuraiton errors for which we (at least for now) keep using VdkConfigurationException

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/exception.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/exception.py
@@ -1,0 +1,36 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from typing import Optional
+
+from vdk.internal.core.errors import BaseVdkError
+from vdk.internal.core.errors import ResolvableBy
+
+
+class PropertiesException(BaseVdkError):
+    """
+    Base Exception for all custom exceptions related to the properties.
+    """
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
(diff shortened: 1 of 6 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/exception.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Base Exception for all custom exceptions related to the properties.
+    """
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
+        super().__init__(None, resolvable_by, message)
```
DeltaMichael, line 19: Why are we passing None for resolvable_by_actual here?

projects/vdk-core/src/vdk/internal/builtin_plugins/job_secrets/exception.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Base Exception for all custom exceptions related to the properties.
+    """
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        resolvable_by: Optional[ResolvableBy] = None,
+    ):
+        super().__init__(None, resolvable_by, message)
```
DeltaMichael, line 19: Why are we passing None for resolvable_by_actual here?

antoniivanov: resolvable_by_actual is always calculated from resolvable_by. 

I will remove it as a argument to an exception altogether, user should not really be able to set it directly.

dakodakov (approved):
LGTM

DeltaMichael (approved):

PR #2771, opened 2023-10-09
control-service: enable scheduled execution for data jobs' synchronizer

mivanov1988:
# Why
Currently, we don't have the ability to schedule the deployment synchronization process.

# What
We have scheduled the process. Only one CS instance at a time can run this process.

# Testing done
Unit tests.

### DIFF
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -1055,6 +1055,14 @@ dataJob:
         # Options are "DB" for database and "K8S" for kubernetes. The control service will read
         # properties from the specified source.
         readDataSource: "K8S"
+      synchronizationTask:
+        enabled: false
+        # The data job deployments' synchronization task interval is the time period (expressed in milliseconds)
+        # after a synchronization process has completed and before a new one is started
+        interval: 1000
+        # The data job deployments' synchronization task  initial delay is the period (expressed in milliseconds) between control service
+        # start and the first time a synchronization process is started by the control service instance
+        initialDelay: 10000
     jobImagePullPolicy: "IfNotPresent"
     initContainer:
       ## Resources set on Data Job initContainer
--- a/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
+++ b/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
@@ -1050,11 +1050,29 @@ dataJob:
       persistence:
         # Case sensitive CSV list of write to's sources. Default is "DB,K8S" The control service
         # will write data job properties to all sources present in this list.
+        # See the synchronizationTask for more information.
         writeTos: "DB,K8S"
         # Case-sensitive variable to select the truth data source for reading properties.
         # Options are "DB" for database and "K8S" for kubernetes. The control service will read
         # properties from the specified source.
         readDataSource: "K8S"
+      # Synchronization Task Behavior:
+      # Configure the behavior of synchronization task based on the following configurations:
+
+      # Case 1: If persistence.writeTo = "DB" and synchronizationTask.enabled = true:
+      # The Deployment API will write data to the database, and the synchronization task
+      # will sync the data jobs in Kubernetes.
+
+      # Case 2: If persistence.writeTo = "DB" and synchronizationTask.enabled = false:
+      # The Deployment API will update the database, but the synchronization task will be disabled,
+      # resulting in no updates in Kubernetes.
+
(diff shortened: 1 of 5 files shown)


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -1055,6 +1055,14 @@ dataJob:
         # Options are "DB" for database and "K8S" for kubernetes. The control service will read
         # properties from the specified source.
         readDataSource: "K8S"
+      synchronizationTask:
+        enabled: false
```
antoniivanov, line 1059: what happens if it's disabled  ? 

Would you be able to deploy or do anything?

mivanov1988: It depends on the writeDataSource, if it is K8S you will be able to deploy if it is DB you won't be able.

antoniivanov: So if writeDataSource is DB , synchoronizatinTask should really always be enabled. And we should really spout warning or even prevent non-working combination.

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -1055,6 +1055,14 @@ dataJob:
         # Options are "DB" for database and "K8S" for kubernetes. The control service will read
         # properties from the specified source.
         readDataSource: "K8S"
+      synchronizationTask:
```
antoniivanov, line 1058: Shouldn't deployment yaml also be edited? How are those passed to the application ?

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
antoniivanov: You can edit schema.values.json to catch that. I asked chatGPT to generate that below so I am not sure if it is correct. 

Also the syntax is terrible. But it would be nice for users to warn them from errors.

```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "http://example.com/example.json",
  "type": "object",
  "properties": {
    "synchronizationTask": {
      "type": "object",
      "properties": {
        "enabled": {
          "type": "boolean",
          "default": false
        }
      },
      "required": ["enabled"]
    },
    "deployment": {
      "type": "object",
      "properties": {
        "writeToS": {
          "type": "string",
          "enum": ["DB", "FILE"]
        }
      },
      "required": ["writeToS"]
    }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "synchronizationTask": {
            "properties": {
              "enabled": {
                "const": false
              }
            }
          },
          "deployment": {
            "properties": {
              "writeToS": {
                "const": "DB"
              }
            }
          }
        }
      },
      "then": {
        "not": {
          "properties": {
            "synchronizationTask": {
              "properties": {
                "enabled": {
                  "const": false
                }
              }
            }
          }
        },
        "description": "If synchronizationTask.enabled is false, deployment.writeToS cannot be DB"
      }
    }
  ]
}

```

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
mivanov1988: Correct, I will update it

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
mivanov1988: I am not sure about that because, in some cases (such as migration, etc.), we might want to disable only the synchronization task.

mivanov1988: Also, we plan to deprecate writeTo and readFrom properties once the Synchronization task is adopted.

antoniivanov: OK.

antoniivanov: But do add some validation that make sense.  a

mivanov1988: Added comment to the values.yaml.

antoniivanov (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
mivanov1988: I tried to incorporate the validation, but it did not work, and I don't want to spend too much time on it at this stage.

dakodakov (approved):
LGTM

PR #2774, opened 2023-10-09
vdk-core: pass exceptions from data job steps in results

DeltaMichael:
## Stack trace examples

Before - User error logged twice: https://pastebin.com/raw/7dgqRpnN
After - User error logged once: https://pastebin.com/raw/unqBCK59

## Why?

We'd like to remove repeating log statements from data jobs, e.g. the same exception getting logged in steps, step processing and job result

## What?

Pass exceptions from job steps to the execution result. Check the execution result for exceptions and raise the exception if it's not related to skipping steps.

This provides a single exit point for logging exceptions.

## How was this tested?

Tested locally by running some broken data jobs
CI

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
@@ -141,7 +141,6 @@ class CliRunImpl:
         execution_result = None
         try:
             execution_result = job.run(args)
-
             # On some platforms, if the size of a string is too large, the
             # logging module starts throwing OSError: [Errno 40] Message too long,
             # so it is safer if we split large strings into smaller chunks.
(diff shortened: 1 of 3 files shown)


duyguHsnHsn:
Did you check vdk-notebook extension as far as I remember it hooks to the run and uses the execution result, it might not be the case but please check once

antoniivanov:
can you please post a few log output from error cases that you tested with.

DeltaMichael:
> Did you check vdk-notebook extension as far as I remember it hooks to the run and uses the execution result, it might not be the case but please check once

No, but we still return the result object, so it should be fine. We can check before merging just to make sure.

DeltaMichael:
> can you please post a few log output from error cases that you tested with.

@antoniivanov Done

antoniivanov (approved):

duyguHsnHsn (approved):

PR #2781, opened 2023-10-10
control-service: asynchronous deployment deletion

mivanov1988:
Why:
As part of VEP-2272, we need to introduce a process for synchronizing data jobs from the database to Kubernetes. Currently, the process lacks the ability to delete deployments asynchronously.

What:
We have introduced an asynchronous method for deployment deletion. It is based on the logic that if the desired deployment is null and the actual deployment is not null, the process performs deployment deletion.

Testing done:
Integration tests.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
@@ -162,8 +162,7 @@ public class DataJobDeploymentCrudITV2 extends BaseIT {
 
     // Deletes deployment
     desiredJobDeploymentRepository.deleteById(testJobName);
-    dataJobsSynchronizer.synchronizeDataJob(
-            dataJob, null, actualDataJobDeployment, true);
+    dataJobsSynchronizer.synchronizeDataJob(dataJob, null, actualDataJobDeployment, true);
     Assertions.assertFalse(deploymentService.readDeployment(testJobName).isPresent());
     Assertions.assertFalse(actualJobDeploymentRepository.findById(testJobName).isPresent());
   }
--- a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
+++ b/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
@@ -159,6 +159,13 @@ public class DataJobDeploymentCrudITV2 extends BaseIT {
     Assertions.assertNotEquals(lastDeployedDateInitial, lastDeployedDateShouldBeChanged);
     Assertions.assertNotEquals(
         deploymentVersionShaShouldNotBeChanged, deploymentVersionShaShouldBeChanged);
+
+    // Deletes deployment
+    desiredJobDeploymentRepository.deleteById(testJobName);
+    dataJobsSynchronizer.synchronizeDataJob(
+            dataJob, null, actualDataJobDeployment, true);
+    Assertions.assertFalse(deploymentService.readDeployment(testJobName).isPresent());
+    Assertions.assertFalse(actualJobDeploymentRepository.findById(testJobName).isPresent());
   }
 
   @Test
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
@@ -69,7 +69,6 @@ public class DataJobsDeploymentController implements DataJobsDeploymentApi {
       if (jobName != null) {
         if (dataJobDeploymentPropertiesConfig.getWriteTos().contains(WriteTo.K8S)) {
           deploymentService.deleteDeployment(jobName);
-
         }
 
         if (dataJobDeploymentPropertiesConfig.getWriteTos().contains(WriteTo.DB)) {
@@ -96,7 +95,7 @@ public class DataJobsDeploymentController implements DataJobsDeploymentApi {
             ToModelApiConverter.toJobDeployment(teamName, jobName, dataJobDeployment);
 
         if (dataJobDeploymentPropertiesConfig.getWriteTos().contains(WriteTo.K8S)) {
-            deploymentService.patchDeployment(job.get(), jobDeployment);
+          deploymentService.patchDeployment(job.get(), jobDeployment);
         }
 
         if (dataJobDeploymentPropertiesConfig.getWriteTos().contains(WriteTo.DB)) {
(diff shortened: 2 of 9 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
```diff
@@ -51,14 +53,27 @@ public class DataJobsDeploymentController implements DataJobsDeploymentApi {
...
+
   @Override
   public ResponseEntity<Void> deploymentDelete(
       String teamName, String jobName, String deploymentId) {
     if (jobsService.jobWithTeamExists(jobName, teamName)) {
       // TODO: deploymentId not implemented
       if (jobName != null) {
-        deploymentService.deleteDeployment(jobName);
-        return ResponseEntity.accepted().build();
+        if (dataJobDeploymentPropertiesConfig
```
antoniivanov, line 66: I feel it's better this `if dataJobDeploymentPropertiesConfig.writeTo` it's in one place and not so many. 

And the appropriate DeploymentService is injected depending on dataJobDeploymentPropertiesConfig.writeTo value. 

So that may necessitate having DeploymentServiceInterface or something but that would be more maintainable, readable and easier to remove in the future if/when we decide to clean up.

mivanov1988: I am not sure how much easier it would be if we introduced an interface, the factory, and then, when we decide to clean up the code, reverted back to the same injection strategy. If you have a strong opinion I will refactor the code.

dakodakov (approved):
LGTM

antoniivanov (commented):
What happens if data job is deleted?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
antoniivanov: ok

mivanov1988:
> What happens if data job is deleted?

That is a very good point! When you delete a data job, the database will automatically remove the associated deployments from the desired and actual tables. Additionally, the old DeploymentService.deleteDeployment() will be called, and the K8s artifacts will be deleted as well. However, because the synchronization process caches the data jobs and deployments, if a user deletes a data job during synchronization, the Kubernetes artifacts will be left in the cluster, potentially leading to an inconsistent state.

In my opinion, we should remove the references from the desired and actual deployment tables to the data job table. In this scenario, we can delete the data job record from the 'data_job' table and the deployment record from the 'desired_data_job_deployment' table. Subsequently, we will have an existing deployment record in the 'actual_data_job_deployment' table, which will be deleted by the synchronization process during the next synchronization, along with all K8S-related artifacts.

@antoniivanov what do you think?

mivanov1988:
> > > What happens if data job is deleted?
> > 
> > 
> > That is a very good point! When you delete a data job, the database will automatically remove the associated deployments from the desired and actual tables. Additionally, the old DeploymentService.deleteDeployment() will be called, and the K8s artifacts will be deleted as well. However, because the synchronization process caches the data jobs and deployments, if a user deletes a data job during synchronization, the Kubernetes artifacts will be left in the cluster, potentially leading to an inconsistent state.
> > In my opinion, we should remove the references from the desired and actual deployment tables to the data job table. In this scenario, we can delete the data job record from the 'data_job' table and the deployment record from the 'desired_data_job_deployment' table. Subsequently, we will have an existing deployment record in the 'actual_data_job_deployment' table, which will be deleted by the synchronization process during the next synchronization, along with all K8S-related artifacts.
> > @antoniivanov what do you think?
> 
> Seems a good idea. It make sense that user should not be able to directly touch actual_deployment , this should be done only by the automation. Users (aka API endpoints) interact only with desired state
> 
> We need to make sure to test some corner cases:
> 
> * synchronization process is delayed or fails  (pod restarts) for some reason after the data job is deleted
It will delete the cron job on the next synchronization.
> * job is created almost immediately with the same name after starting deletion
It is okay. The cron job will be deleted, and the job will no longer have an associated one. The UI might display the existing deployment for a short period of time.
> * What would the GraphQL endpoint return (aka the UI show) Probably the job as deleted and hence nothing
Nothing happens because the job is a first-class citizen.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
```diff
@@ -159,6 +159,12 @@ public void testSynchronizeDataJob() throws Exception {
     Assertions.assertNotEquals(lastDeployedDateInitial, lastDeployedDateShouldBeChanged);
     Assertions.assertNotEquals(
         deploymentVersionShaShouldNotBeChanged, deploymentVersionShaShouldBeChanged);
+
+    // Deletes deployment
```
antoniivanov, line 163: We should really considering refactoring those long test methods into multiple methods and using @Order annotation to dictate the order . 

It's not something you need to do now, but you can add a TODO if you want for next developer to consider doing this.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobsSynchronizer.java
```diff
@@ -144,6 +144,8 @@ void synchronizeDataJob(
           desiredDataJobDeployment,
           actualDataJobDeployment,
           isDeploymentPresentInKubernetes);
+    } else if (actualDataJobDeployment != null) {
```
antoniivanov, line 147: Where are the tests that verify `synchronizeDataJobs` (all jobs) work as expected? I see tests that `synchronizeDataJob` (singular) works. But not for the full sync.

I am suspicious deleting job might not work because we are iterating using `dataJobsFromDB` but if we delete a data job then `dataJobsFromDB` would not contain our deleted job.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
mivanov1988: Yes, this is planned.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobsSynchronizer.java
mivanov1988: Correct, the synchronizeDataJobs does not cover this case.

mivanov1988: Are you ok to implement this in a separate PR?

antoniivanov: Yeah.

antoniivanov (approved):

PR #2787, opened 2023-10-11
vdk-core: simplify error message for send_**_for_ingestion

antoniivanov:
There were a couple of issues with error handling that are fixedhere

1. Inconsistent behaviour between send_xxx methods. One method was swallowing the error while the other was re-throwing it . This is fixe by both rethrowing it.

2. Remove the unecessary error message. It is redundant and doesn't provide any new information that is not availbe in the raised exception

3. When "rows" argument was not iterabot nor cursor or iterable raised custom exceptin,  it always failed with cryptic error ` AttributeError: 'BrokenIterator' object has no attribute 'fetchmany' ` This is fixed by using more appropriate if statement to cehck type of rows and raising InvalidArgumentsIngestionException if it's the wrong type.

Consolited some tests in function/ingestion module. There was no test covering the scenario of using multiple methods and it is added.

Testing Done: beyond automated test see 
 - [logs after change](https://gist.github.com/antoniivanov/a8afbc34403eee7e00cae4d1d3d00ebb) vs [logs before change](https://gist.github.com/antoniivanov/ab85ef7b5585c04fce2e76a7443b1c55) (2x shorter)


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
```diff
@@ -143,7 +143,8 @@ def __ingest_object(
                 payload, destination_table, method, target, collection_id
             )
         except Exception as e:
-            self._log.warn("Failed to send object for ingestion." f"Exception was: {e}")
+            errors.report(ResolvableBy.USER_ERROR, e)
```
DeltaMichael, line 146: Why not just call report_and_rethrow, it contains the exact same two lines.

DeltaMichael (approved):

mivanov1988 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
antoniivanov: I experimented with both and with report_and_rethrow the stack trace is more confusing. 

See the comparison https://www.diffchecker.com/U4Sgb28h/ 

The 2 newly added method calls sort of branch out . Re-raising the error keeps the stack trace more linear hence easier to understand. 

If we simply want to report the error and allow it to continue propagating as if it had not been caught at all (keeping the stack trace clean), we should use raise by itself. 

That means likely we should discourage the use of report_and_throw in favour of report() and raise

PR #2792, opened 2023-10-12
vdk-audit: Clean up some audit events

doks5:
The list of forbidden events that the vdk-audit plugin looks for, is unreasonably restricted by default. Most of the operations would not affect the security of a cluster or other client operations.

This change removes some events.

Testing Done: N/A, configuration change


mivanov1988:
I think that we should keep os.kill.

doks5:
> I think that we should keep os.kill.

Fixed, thanks

mivanov1988 (approved):

antoniivanov (approved):

PR #2794, opened 2023-10-13
control-service: add IT tests for async job deployment

mivanov1988:
I've introduced the initial version of async job deployment integration tests. This is just the starting point, and we plan to expand and refactor it further in the upcoming version based on feedback from @antoniivanov 's comment. You can find the discussion on this change in the following link: [Link to the GitHub pull request discussion](https://github.com/vmware/versatile-data-kit/pull/2781#discussion_r1358096950).


murphp15:
typo in the title, should be add in*

doks5:
> typo in the title, should be add in*

I think "it" in this case means IT (Integration Test).

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -1051,7 +1051,7 @@ dataJob:
         # Case sensitive CSV list of write to's sources. Default is "DB,K8S" The control service
         # will write data job properties to all sources present in this list.
         # See the synchronizationTask for more information.
-        writeTos: "DB,K8S"
+        writeTos: "K8S"
```
murphp15, line 1054: I'm just back from a long break off work. 

I think the the documentation here is pretty good but it would be soo much more useful if you guys said when you would you the different configurations. 

Am I correct in saying that with a fresh deployment with no exisitng jobs it would be best to use 

writeTos: "DB,K8S"
readDataSource: "DB"


If you are upgrading a control service for an existing deployment how should you set them?

murphp15 (changes requested):
I want to understand the config better before I approve

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
mivanov1988: You are correct, but the new deployment functionality is still experimental, and we don't want to break existing deployments. We will introduce a tool for populating the missing data for the existing deployments.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeploymentCrudAsyncIT.java
```diff
@@ -0,0 +1,360 @@
...
+      "datajobs.control.k8s.k8sSupportsV1CronJob=true",
+      "datajobs.deployment.configuration.persistence.writeTos=DB",
+      "datajobs.deployment.configuration.synchronization.task.enabled=true",
+      "datajobs.deployment.configuration.synchronization.task.interval.ms=1000",
+      "datajobs.deployment.configuration.synchronization.task.initial.delay.ms=0"
+    })
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+    classes = ControlplaneApplication.class)
+public class DataJobDeploymentCrudAsyncIT extends BaseIT {
```
antoniivanov, line 53: I have not looked closely. how is this different from DataJobDeploymentCrudIT ? 


Could we have one extends the other so that we do not  have so much duplication or near duplication.

mivanov1988: At the time I implemented it deletion did not exist, that's why I duplicated the test. I will update it.

mivanov1988: Refactored the code.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployerV2.java
```diff
@@ -419,7 +419,7 @@ private KubernetesService.CronJob getCronJob(
             jobDeployment.getLastDeployedBy(),
             lastDeployedDate,
             jobDeployment.getPythonVersion());
-    boolean enabled = jobDeployment.getEnabled();
+    boolean enabled = jobDeployment.getEnabled() == null || jobDeployment.getEnabled();
```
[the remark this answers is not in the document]
mivanov1988: This is the behavior of the old deployment logic. Moved the logic here - https://github.com/vmware/versatile-data-kit/pull/2794/files#diff-027ec295ea159c5a35bce00bbc225b850100d57b679affdb540aee808a7774f8R270

dakodakov (approved):
LGTM

antoniivanov (approved):

murphp15 (approved):

PR #2805, opened 2023-10-17
vdk-datasources: data sources POC

antoniivanov:
This change implements POC for Data Sources API.

It is based on some of teh requirements and reserach in https://github.com/vmware/versatile-data-kit/wiki/Ingest-Source-Research

See concepts page for explanation of data source related concepts

So what's implemented is
- Data Source APIs handling sources, streams and state
-  New Data Source is implemented by implementing IDataSource, IDataSourceConfiguraiton and IDataSourceStream
- Data Source connection management partialy
- Data Source Ingester that reads from data sources and writes to existing IIngeser
- An example data source AutoGeneratedDataSource
- An example job in the function test suite


projects/vdk-plugins/vdk-data-sources/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+### Concepts
+
+
+#### Data Source
+A Data Source is a central component responsible for establishing and managing a connection to a specific set of data. It interacts with a given configuration and maintains a stateful relationship with the data it accesses. This stateful relationship can include information such as authentication tokens, data markers, or any other form of metadata that helps manage the data connection. The Data Source exposes various data streams through which data can be read.
+
+#### Data Source Stream
+A Data Source Stream is an abstraction over a subset of data in the Data Source. It can be thought of as a channel through which data flows. Each Data Source Stream has a unique name to identify it and includes methods to read data from the stream. For example for Database based data source , each table could be a separate stream. Streams can be ingested in parallel potentially.
```
murphp15, line 21: I think a diagram here would be good

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/auto_generated.py
```diff
@@ -0,0 +1,107 @@
...
+)
+from vdk.plugin.data_sources.data_source import IDataSourceStream
+from vdk.plugin.data_sources.factory import data_source
+from vdk.plugin.data_sources.state import IDataSourceState
+
+
+@config_class(
+    name="auto_generated", description="Configuration for Auto generated data source"
+)
+class AutoGeneratedDataSourceConfiguration(IDataSourceConfiguration):
```
murphp15, line 22: Why did you not use just one class that ppl have to implement instead of 3? 
Would ppl reuse configurations with different streams of vis versa? 
Could you give an example of this?

murphp15:
I understand mostly the difference between data source and data source stream. 
However I would really like a few concrete examples. 

Can you give an example if your source was a relational database and also an example if it was blob storage?

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/data_source.py
```diff
@@ -0,0 +1,245 @@
...
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Union
+
+from vdk.plugin.data_sources.config import is_config_class
+from vdk.plugin.data_sources.state import IDataSourceState
+
+
+class IDataSourceConfiguration:
```
murphp15, line 18: what is the point in making an interface if it doesn't have any real methods in it?

murphp15: just add this check in the abstract class?

projects/vdk-plugins/vdk-data-sources/tests/functional/jobs/ingest-sources-job/step.py
```diff
@@ -0,0 +1,28 @@
...
+        "auto-generated-data"
+    )
+    auto_generated.configure(AutoGeneratedDataSourceConfiguration())
+
+    data_source_ingester = DataSourceIngester(job_input)
+
+    data_source_ingester.ingest_data_source("auto", auto_generated, method="memory")
+
+    data_source_ingester.terminate_and_wait_to_finish()
+    data_source_ingester.raise_on_error()
```
murphp15, line 28: can we not make this the default?

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/config.py
```diff
@@ -0,0 +1,164 @@
...
+            metadata_key, None
+        )
+
+    def name(self) -> str:
+        return self._field_name
+
+    def description(self) -> str:
+        return self.__get_field_metadata_value("config_description")
+
+    def is_sensitive(self) -> bool:
```
murphp15, line 108: If they tag a property as sensitive do you actually do anything with it? or is this going to be handled differntly later ?

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/data_source.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def __init__(self, data_streams_exceptions: Dict[str, Exception]):
+        super().__init__(
+            f"Data Sources failed to ingest data: {data_streams_exceptions}"
+        )
+        self.data_streams_exceptions = data_streams_exceptions
+
+
+@dataclass
+class DataSourceError:
```
murphp15, line 154: why did you not call it an exception? 
I would think that error is more serious. 

like out of memory error.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/factory.py
```diff
@@ -0,0 +1,165 @@
...
+    def inner_decorator(data_source_class: Type[IDataSource]):
+        data_source_class.__data_source_name__ = name
+        data_source_class.__data_source__config__class__ = config_class
+        return data_source_class
+
+    return inner_decorator
+
+
+@dataclass
+class DataSourceRegistryItem:
```
murphp15, line 28: Is the user who is writing an ingestion job suppose to know about this class?

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/state.py
```diff
@@ -0,0 +1,136 @@
...
+            data_source
+        ] = state
+        self._properties.set_all_properties(all_properties)
+
+
+class DataSourceStateFactory:
+    def __init__(self, storage: IDataSourceStateStorage):
+        self._storage = storage
+
+    def get_data_source_state(self, source: str) -> IDataSourceState:
```
murphp15, line 135: There really needs to be a comment here as users are expected to use this function.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/factory.py
antoniivanov: Not really

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/state.py
antoniivanov: users in the sence of other vdk developers. 

But end users are not supposed to use it directly. I will add comment though

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/config.py
antoniivanov: This is existing feature of ConfigurationBuilder. It obfuscate them in some cases when being printed.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/data_source.py
antoniivanov: What abstract class?

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/auto_generated.py
antoniivanov: Nothing stops people from doing 

```
MyDataSource(IDataSourceConfiguration, IDataSource, IDataSourceStream): 
  ....
```

They really ned to register one class in both cases. 


It seems personal preference. It's easier for me to think them when they are separate as separate steps that need to be taken. Define configuration. Split your data into streams (at least 1) , write the stream reading logic

murphp15 (approved):

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/ingester.py
```diff
@@ -0,0 +1,247 @@
...
+        return destination_table
+
+    def start_ingestion(
+        self,
+        data_source_id: str,
+        data_source: IDataSource,
+        destinations: List[IngestDestination] = None,
+        error_callback: Optional[IDataSourceErrorCallback] = None,
+    ):
+        if data_source_id not in self.__ingested_streams_set:
```
dakodakov, line 200: "ingested" streams? What if the ingestion fails? I would rename it to something something like "being_ingested" or "ingesting" streams?

projects/vdk-plugins/vdk-data-sources/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+Reading from the stream yields a sequence of Data Source Payloads
```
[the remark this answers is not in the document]
antoniivanov: VDK is not really a streaming platform. 

But what could happen depends on how the Kafka Data source is implemented 
 You can implement it where it would fetch all data until start_timestamp . In this case it should end. 
 
 But one can reuse https://pypi.org/project/pipelinewise-tap-kafka/ with vdk-singer . 

 And the way they have handled it is to use max_runtime_ms (The maximum time for the tap to collect new messages from Kafka topic) to end the ingestion batch.

projects/vdk-plugins/vdk-data-sources/README.md
```diff
@@ -0,0 +1,91 @@
...
+```python
+def run(job_input: IJobInput):
+    source = SourceDefinition(id="auto", name="auto-generated-data", config={})
+    destination = DestinationDefinition(id="auto-dest", method="memory")
+
+    with DataFlowInput(job_input) as flow_input:
+        flow_input.start(DataFlowMappingDefinition(source, destination))
+```
+
+or in config.toml file
```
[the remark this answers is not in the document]
antoniivanov: I've been considering replacing ini with TOML because 
- it supported nested structures of data 
- it supports arrays 
- it supports data types !! 
- it's somewhat similar in ini in syntax - existing ini files can be parsed by TOML parser so a change can be pretty backward compatible.

It would not have been feasible/easy to use ini format the above data flow structure. So I decided to use TOML now as an experiment to see if it's going to work for users. 

We need to move away from ini due to above reasons and we've had users who have requested to support more "modern" format. 
The other alternative is Yaml. But yaml is pretty ugly for highly nested configurations an

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/auto_generated.py
```diff
@@ -0,0 +1,107 @@
...
+        self,
+        config: AutoGeneratedDataSourceConfiguration,
+        stream_number: int,
+        start_id: int = 0,
+    ):
+        self._config = config
+        self._stream_number = stream_number
+        self._data = self._generate_test_data(start_id)
+
+    def _generate_test_data(self, start_id: int) -> List[DataSourcePayload]:
```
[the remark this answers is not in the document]
antoniivanov: Fixed. Make sense.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/auto_generated.py
```diff
@@ -0,0 +1,107 @@
...
+        generated_data = []
+        for i in range(self._config.num_records):
+            data = {
```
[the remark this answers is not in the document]
antoniivanov: Hmm, yeah I probably could have come with some more interesting example. I will leave it for now though since there are already lots of tests that expect this data. But I will change it later.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/factory.py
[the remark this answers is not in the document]
antoniivanov: It keeps the classes hat are needed to create the data source. It's not the data source. I am not sure of a better name

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/ingester.py
```diff
@@ -0,0 +1,247 @@
...
+    data_source_id: str
+    stream: IDataSourceStream
+    destinations: List[IngestDestination]
+    error_callback: Optional[IDataSourceErrorCallback] = None
+
+
+class DataSourceIngester:
+    def __init__(self, job_input: IJobInput):
+        self.__ingestion_queue = Queue()
+        self.__actual_ingester = cast(IIngester, job_input)
```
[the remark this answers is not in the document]
antoniivanov: Well, the class is called "DataSourceIngester" so the actual ingester that is going ot send the data to the target is the __actual_ingester.. 

Feel free to suggest a better name and I will rename it.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/ingester.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+            log.warning(error_message)
+            self.__stored_exceptions.put((ingest_entry.stream.name(), e))
+
+    @staticmethod
+    def _is_termination_signal(queue_item):
+        return queue_item is None
+
+    def _ingest_stream(self, ingest_entry: IngestQueueEntry):
+        for payload in ingest_entry.stream.read():
```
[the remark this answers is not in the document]
antoniivanov: When the stream decides to stop.

projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/ingester.py
antoniivanov: Sounds good

antoniivanov:
Thanks for the review @dakodakov I have  asnwered the comments above and address the changes in this PR https://github.com/vmware/versatile-data-kit/pull/2865

projects/vdk-plugins/vdk-data-sources/README.md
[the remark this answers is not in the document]
dakodakov: I generally agree with everything you said, however I would be great to have a more structured approach, e.g. use TOML everywhere, rather than use it here and use something else in a different place.

PR #2800, opened 2023-10-17
control-service: deployment controller reads from db

mrMoZ1:
what: DataJobDeploymentController now has the functionality to read from the DB.

why: Part of the persistence VEP which can be found here - https://github.com/vmware/versatile-data-kit/tree/main/specs/vep-2272-complete-data-job-configuration-persistence This PR will reduce the overhead on K8S when making API calls to the control-service.

testing: added unit tests.

PR #2815, opened 2023-10-19
control-service: fix post deployment test

mrMoZ1:
what: When calling the deploymentList method of the DataJobsDeploymentController class and a job without deployments exists, the API will now return an empty list instead of 404.

why: Recent changes in https://github.com/vmware/versatile-data-kit/pull/2800 refactored the method to return 404 instead of empty list in this situation, however this breaks the post deployment test.

testing: added unit test.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
@@ -20,7 +20,6 @@ import com.vmware.taurus.service.deploy.DeploymentServiceV2;
 import com.vmware.taurus.service.diag.OperationContext;
 import com.vmware.taurus.service.model.JobDeploymentStatus;
 import io.swagger.v3.oas.annotations.tags.Tag;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dakodakov (approved):

mivanov1988:
I don't think this is a proper solution to the regression. You should rework the code https://github.com/vmware/versatile-data-kit/pull/2800.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java
```diff
@@ -111,7 +112,12 @@ public ResponseEntity<List<DataJobDeploymentStatus>> deploymentList(
       String teamName, String jobName, String deploymentId, DataJobMode dataJobMode) {
     // TODO: deploymentId and mode not implemented
     if (jobsService.jobWithTeamExists(jobName, teamName)) {
-      return deploymentAsList(jobName.toLowerCase());
+      var response = deploymentAsList(jobName.toLowerCase());
+      if (response.getStatusCodeValue() == 404) {
```
mivanov1988, line 116: I don't think this is a proper solution to the regression. You should rework the code https://github.com/vmware/versatile-data-kit/pull/2800.

mivanov1988 (approved):

dakodakov (approved):

PR #2816, opened 2023-10-19
Person/miroslavi/improve async deployment deletion

mivanov1988:
# Why
When the data job is deleted, both desired deployment records are deleted, but the actual deployment records and Kubernetes cron jobs are still present.

# What
Extended the deployment synchronization to cover this scenario.

# Testing done
Integration tests.


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobsSynchronizer.java
```diff
@@ -68,35 +67,25 @@ public class DataJobsSynchronizer {
           "${datajobs.deployment.configuration.synchronization.task.initial.delay.ms:10000}")
   @SchedulerLock(name = "synchronizeDataJobsTask")
   public void synchronizeDataJobs() {
```
antoniivanov, line 69: synchronizeDataJobs is doing multiple things and getting pretty long. Consider breaking it into smaller methods that handle separate concerns separately.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
```diff
@@ -200,6 +200,19 @@ public void testSynchronizeDataJobs() throws Exception {
     OffsetDateTime lastDeployedDateInitial = actualDataJobDeployment.getLastDeployedDate();
     Assertions.assertNotNull(deploymentVersionShaInitial);
     Assertions.assertNotNull(lastDeployedDateInitial);
+
+    jobsRepository.deleteById(testJobName);
```
antoniivanov, line 204: Nit: why not use JobService.delete() ? Or even the controller ? Isn' that more appropriate and realistic . 


Also we can test jobService.Delete ; jobService.create. (same-job-name) same-job-name should have zero deployments (eventually) . The test should pass because JobService.delete also deletes the desired deployment. But better tested for regressions than not.

antoniivanov (approved):
Looks good to me. Thanks for addressing those changes. 

Please make sure vdk-heartbeat passes reliably (3 times in sequence for example). 

I am concerned that changing from syncrhouos to more async/eventual consistence model might introduce some intermittent issues so do not rely on passing immediately and monitor the control service pipeline a few days after the merge.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
mivanov1988: We are going to remove this test and use the controller as part of the DataJobDeploymentCrudAsyncIT.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobsSynchronizer.java
mivanov1988: Agree, I tried to reduce it a bit in this PR.

PR #2820, opened 2023-10-20
vdk-core: test exception propagation to user code

DeltaMichael:
## Why

As part of our run logs initiative, we stopped wrapping non-vdk errors in vdk errors. This should result in errors coming from libraries, templates, etc. being propagated to user code. Users should then be able to handle those errors. This behaviour has not been tested, however.

## What

Add a functional test that creates a template and runs it. The templates tries to read an empty csv using pandas, which results in a pandas-specific exception. The exception is then handled in a data job that executes the template.

## How was this tested

Ran the test locally

Feature/non-breaking


murphp15 (approved):
Looks good. very simple

antoniivanov (approved):

Issue #2802, opened 2023-10-17
Create a bound logger in vdk-core


DeltaMichael:
## Overview

Vdk users should have the option of binding additional context to their log statements. This should be provided out of the box by vdk-core.

Example:

```python
import logging

class BoundLogger(logging.LoggerAdapter):
    def process(self, msg, kwargs):
        # merge bound extra dict with existing exra dict if any
        if "extra" in kwargs:
            kwargs["extra"] = {**self.extra, **kwargs["extra"]}
        else:
            kwargs["extra"] = self.extra
        return msg, kwargs

def bind_logger(log, context):
     return BoundLogger(log, context)
```

Somewhere else in core, plugin or data job code

```python
from vdk.internal.core.logging import bind_logger
import logging

log = logging.getLogger(__name__)

bound = bind_logger(log, {"bound_key": "bound_value"})
bound.warning("From vdk_initialize", extra={"first_key": "first_value", "second_key": "second_value"})
bound.warning("Something else from vdk_initialize")
```

Then, with vdk-structlog installed and a metadata config like

```sh
logging_metadata=timestamp,level,file_name,line_number,vdk_job_name,bound_key,second_key
```

We should see the bound_key and second_key values in the logs. bound_key should be present for both statements. second_key should be present for just the first statement.

## Acceptance criteria

1. Above scenario is implemented and works with vdk-structlog

PR #2823, opened 2023-10-23
vdk-core: VdkBoundLogger

gageorgiev:
This change introduces the VdkBoundLogger and addresses #2802.

Testing done: locally tested, ran a job with the following job code:
```py
def run(job_input):
    import logging
    from vdk.internal.builtin_plugins.config.log_config import bind_logger

    bound = bind_logger(logging.getLogger(__name__), {"bound_key": "bound_value"})
    bound.warning("From vdk_initialize", extra={"first_key": "first_value", "second_key": "second_value"})
    bound.warning("Something else from vdk_initialize")
```

And the following config:
```
logging_metadata=timestamp,level,file_name,line_number,vdk_job_name,bound_key,second_key
```
 And observed the following logs:
```
2023-10-23 15:17:42,547 [VDK] [WARNI]                10.py :5    gg-job4 bound_value second_value - From vdk_initialize
2023-10-23 15:17:42,548 [VDK] [WARNI]                10.py :6    gg-job4 bound_value  - Something else from vdk_initialize
```

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
@@ -204,53 +204,6 @@ def ensure_logging_for_current_test() -> None:
         configure_loggers(fname, str(time()))
 
 
-class VdkBoundLogger(logging.LoggerAdapter):
-    """
-    This class wraps around a Logger to append additional context to log records.
-    Logger-wide context can be included by providing it in the `context` parameter of the
-    `bind_logger` method.
-    Log statement-specific context can be included by including an `extra` kwarg to the log call and
-    providing a dictionary object (or any object which implements __getitem__ and __iter__) with the
-    required additional log fields.
-    It is intended to be instantiated using the `bind_logger` function and providing it with the current
-    log object and the additional context which will be included in dictionary form.
-
-    Example use:
-
-    import logging
-    from vdk.internal.builtin_plugins.config.log_config import bind_logger
-
-    bound_logger = bind_logger(logging.getLogger(__name__), {"logger_wide_bound_key": "logger_wide_bound_value"})
-    bound.warning("From vdk_initialize", extra={"first_key": "first_value"})
-    bound.warning("Something else from vdk_initialize")
-
-    Now, the LogRecord produced by the first log statement will have both
-    "logger_wide_bound_key": "logger_wide_bound_value" and "first_key": "first_value" included in its __dict__,
-    while the second one will only have "first_key": "first_value" in it.
-
-    Note that these will not be automatically included in the output log and must be processed downstream by
-    something like vdk-structlog.
-    """
-    def process(self, msg, kwargs):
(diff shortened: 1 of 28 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
```diff
@@ -204,6 +204,54 @@ def ensure_logging_for_current_test() -> None:
         configure_loggers(fname, str(time()))
 
 
+class VdkBoundLogger(logging.LoggerAdapter):
```
antoniivanov, line 207: Move this to a separate file/module. 

log_config module is already doing too many things. It should be refactored to remove things and not add new things. 

But also VdkBoundLogger is a public (plugin-facing) interface so it's going ot be reused a lot of times. You want it in a file with only its own imports otherwise users (modules) importing VDKBoundLogger ot use it will import bunch of unnecessary imports for them


Also the "config" builtin_plugin doesn't seem the best place. 

Overall we have a definition for https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#public-python-interfaces here. 

Last, plugins have been really allowed to use anything under `vdk.internal.core` or `vdk.interna.plugin` but generally discouraged to use things under vdk.internal.builtin_plugins namespaces

So best place would be in vdk.internal.core.logging.py module

gageorgiev: gotcha

projects/vdk-core/src/vdk/internal/core/logging.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Now, the LogRecord produced by the first log statement will have both
+    "logger_wide_bound_key": "logger_wide_bound_value" and "first_key": "first_value" included in its __dict__,
+    while the second one will only have "first_key": "first_value" in it.
+
+    Note that these will not be automatically included in the output log and must be processed downstream by
+    something like vdk-structlog.
+    """
+
+    def process(self, msg, kwargs):
```
DeltaMichael, line 35: Can you please add type hints everywhere?

DeltaMichael (approved):

projects/vdk-core/src/vdk/internal/core/logging.py
```diff
@@ -0,0 +1,52 @@
...
+    `bind_logger` method.
+    Log statement-specific context can be included by including an `extra` kwarg to the log call and
+    providing a dictionary object (or any object which implements __getitem__ and __iter__) with the
+    required additional log fields.
+    It is intended to be instantiated using the `bind_logger` function and providing it with the current
+    log object and the additional context which will be included in dictionary form.
+
+    Example use:
+
+    import logging
```
antoniivanov, line 20: We use reStructuredText in python docs, so make sure to format it as code block so it shows up well in Help - https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

antoniivanov (approved):

PR #2824, opened 2023-10-23
vdk-heartbeat: Introduce additional sleep when checking deployments

doks5:
With the introduction of the deployment synchronizer, the behaviour of the Control Service, when a data job is deployed, has changed slightly. Now, there is a slight delay between a job status changing from Enabled to Disabled, and the status being reflected in the output of the `vdk deploy --show` command. This causes the vdk heartbeat to fail, as it expects the changes to be applied immediately.

This change adds a new check when the `check_deployments()` method is called, whose purpose is to verify that the job status has changed from enabled to disabled, and sleep for a bit if this is not the case.

Testing Done: Executed the heartbeat locally with the command
```bash
vdk-heartbeat -f test_basic_config.ini
```

### DIFF
--- a/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
+++ b/projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
@@ -206,6 +206,12 @@ class JobController:
             if not deployments:
                 log.info("No deployments so far. Will wait 30 seconds and try again.")
                 time.sleep(30)
+            elif deployments[0]["enabled"] != enabled:
+                log.info(
+                    "Job deployment should be disabled, but is still enabled. Will wait for 30 seconds and try again."
+                )
+                time.sleep(30)
+                deployments = None
         assert deployments, f"Job {self.config.job_name} deployment is missing"
         assert (
             deployments[0]["enabled"] == enabled


mivanov1988 (approved):

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
```diff
@@ -206,6 +206,12 @@ def check_deployments(self, enabled=True, timeout_seconds=600):
             if not deployments:
                 log.info("No deployments so far. Will wait 30 seconds and try again.")
                 time.sleep(30)
+            elif deployments[0]["enabled"] != enabled:
```
antoniivanov, line 209: it's better try eveyr few seconds and timeout on 30 seconds instead of just sleeping

antoniivanov (approved):

projects/vdk-heartbeat/src/vdk/internal/heartbeat/job_controller.py
doks5: 30 seconds may not be enough, so setting a timeout at it is something I'd rather avoid.
Thank you for the review

antoniivanov: Well, the timeout must be configurable of course. 

The point is that having a fixed delay is not a good pattern because hard-coding a fixed timeout in tests can lead to inefficiencies and false negatives

 Even if the deployment happens in 1 second you still wait 30 seconds. 
If the deployment happens 35 seconds you are not flexible enough to cover it and the test would fail unnecessarily

PR #2825, opened 2023-10-23
vdk-core: fix postgres and greenplum tests

yonitoo:
What: Due to the introduction of some vdk-core log changes, similar unit test for vdk-postgres and vdk-greenplum started failing and are breaking the CI/CD.

Why: To fix the CI/CD.

Testing Done: locally.


murphp15 (approved):

murphp15:
Nice that it is a better message too.

gageorgiev (approved):

PR #2829, opened 2023-10-24
control-service: integration test for async job deploy

mrMoZ1:
what: Add integration test that checks the status of the database when using the new data jobs deployment synchronizer.

why: part of the https://github.com/vmware/versatile-data-kit/tree/main/specs/vep-2272-complete-data-job-configuration-persistence VEP

testing: n/a


mivanov1988:
As we discussed the test had to be reworked as per @antoniivanov 's comment - https://github.com/vmware/versatile-data-kit/pull/2781#discussion_r1358096950, but I don't think this is the case. Also, you should verify all deployment fields not just 3-4 after each operation through the APIs.

mrMoZ1:
> As we discussed the test had to be reworked as per @antoniivanov 's comment - [#2781 (comment)](https://github.com/vmware/versatile-data-kit/pull/2781#discussion_r1358096950), but I don't think this is the case. Also, you should verify all deployment fields not just 3-4 after each operation through the APIs.

The test couldn't be refactored into multiple smaller test cases, since there is apparently some intrinsic state that is lost when executing different test cases. I refactored the larger test cases to call different methods (which were supposed to be ordered tests) and it works that way. When I try to make them into different test cases they fail. 

I am also checking all fields except for resources - since they aren't supported yet and if I set them via the POST request the test fails :
https://github.com/vmware/versatile-data-kit/blob/ff1156a389df735e923d2874bc818636a5425167/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/datajobs/DataJobsDeploymentController.java#L203-L217

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
```diff
@@ -5,6 +5,14 @@
 
 package com.vmware.taurus.service.deploy;
 
+import static com.vmware.taurus.datajobs.it.common.WebHookServerMockExtension.TEST_TEAM_NAME;
```
mivanov1988, line 8: DataJobDeploymentCrudITV2.java will be decommissioned in favor of DataJobDeploymentCrudAsyncIT.java, you should refactor the BaseJobDeploymentCrudITV2.java instead.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeploymentCrudAsyncIT.java
```diff
@@ -24,6 +29,61 @@
...
-  protected void beforeDeploymentDeletion() {}
+  protected void afterDeploymentDeletion() {
+    var desiredDataJobDeployment = desiredJobDeploymentRepository.findById(testJobName);
+    var actualDataJobDeployment = actualJobDeploymentRepository.findById(testJobName);
+
+    checkDesiredDeploymentDeleted(desiredDataJobDeployment);
+    checkActualDeploymentDeleted(actualDataJobDeployment);
+  }
+
+  private void checkDesiredDeployment(Optional<DesiredDataJobDeployment> desiredDataJobDeployment) {
```
mivanov1988, line 54: If it is possible the test shouldn't use repositories and services but only the controllers/APIs.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java
```diff
@@ -109,7 +111,7 @@ public void testDataJobDeploymentCrud() throws Exception {
     Assertions.assertFalse(StringUtils.isBlank(testJobVersionSha));
 
     // Setup
-    String dataJobDeploymentRequestBody = getDataJobDeploymentRequestBody(testJobVersionSha);
+    String dataJobDeploymentRequestBody = getDataJobDeploymentRequestBody(testJobVersionSha, "3.9");
```
mivanov1988, line 114: Make sure that the test verifies all fields through the controller/API after each deployment change.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeploymentCrudAsyncIT.java
```diff
@@ -24,6 +29,61 @@
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
     classes = ControlplaneApplication.class)
 public class DataJobDeploymentCrudAsyncIT extends BaseDataJobDeploymentCrudIT {
+
+  @Autowired private DesiredJobDeploymentRepository desiredJobDeploymentRepository;
+  @Autowired private ActualJobDeploymentRepository actualJobDeploymentRepository;
+
+  @Override
+  protected void beforeDeploymentDeletion() {
+    var desiredDataJobDeployment = desiredJobDeploymentRepository.findById(testJobName);
```
murphp15, line 38: why don't you use existsById here and then you can do the assertions here, removing line 41 and 42.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/service/deploy/DataJobDeploymentCrudITV2.java
```diff
@@ -74,6 +73,14 @@ public class DataJobDeploymentCrudITV2 extends BaseIT {
 
   @Autowired private DeploymentService deploymentService;
 
+  private boolean jobEnabled;
```
murphp15, line 76: I think I would prefer it if you weren't creating fields here and instead just using local variables.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/DataJobDeploymentCrudIT.java
```diff
@@ -77,4 +77,9 @@ protected void beforeDeploymentDeletion() throws Exception {
         .andExpect(status().isOk())
         .andExpect(jsonPath("$.vdk_version", is("release")));
   }
+
+  @Override
+  protected void afterDeploymentDeletion() {
```
murphp15, line 82: Presuably the rows wouldn't be in the database here either? 
Maybe we could just have the implementation in the base class.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java
mivanov1988: I still don't see this verification. For example, the test deploys a job with specific resources, and then doesn't even check them.

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java
```diff
@@ -65,30 +74,131 @@ public void setup() throws Exception {
   }
 
   @Test
-  public void testDataJobDeploymentCrud() throws Exception {
+  @Order(1)
```
murphp15, line 77: why do they need to be in order ?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java
```diff
@@ -65,30 +74,131 @@ public void setup() throws Exception {
...
+    deleteDeploymentNoUser();
+
+    deleteDeploymentWrongTeam();
+
+    executeDeleteDeployment(jobDeploymentName);
+  }
+
+  @Test
+  @Order(4)
+  public void testDataJobDeleteSource() throws Exception {
```
murphp15, line 145: I don't see you deleting anything in this test?

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java
```diff
@@ -65,30 +74,131 @@ public void setup() throws Exception {
...
+  @Test
+  @Order(3)
+  public void testDataJobDeploymentCrudOperations() throws Exception {
+
+    var testDataJobVersion = uploadJobWithProperUser();
+    var testJobVersionSha = getTestJobVersionSha(testDataJobVersion);
+    var dataJobDeploymentRequestBody = getDataJobDeploymentRequestBody(testJobVersionSha, "3.9");
+    String jobDeploymentName = JobImageDeployer.getCronJobName(testJobName);
+
+    buildAndDeployJobWithNoUser(dataJobDeploymentRequestBody);
```
murphp15, line 110: Why do you need to test this again if you just tested it separately?

murphp15 (approved):

projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java
mrMoZ1: I didn't modify/refactor this test it's present in the original test file - https://github.com/vmware/versatile-data-kit/blob/df1ea41a89f493b12e6be56d73b64261aa214a2a/projects/control-service/projects/pipelines_control_service/src/integration-test/java/com/vmware/taurus/datajobs/it/BaseDataJobDeploymentCrudIT.java#L337-L350 
I'm not quite sure what it's supposed to do.

mrMoZ1: deleted this test since it was redundant.

mivanov1988 (approved):

PR #2831, opened 2023-10-24
vdk-core: implement config option for logging execution result

DeltaMichael:
## Why?

User reasearch indicates that the execution result is too verbose for local runs. Users don't expect a lot of output for successful jobs and expect error output for failing jobs.

## What?

Introduce the LOG_EXECUTION_RESULT config option that enables/disables displaying the end result.

### Config Error Output

https://pastebin.com/raw/MvJX0u9m

### User Error Output

https://pastebin.com/raw/0MxZp6Zg

### Success Output

https://pastebin.com/raw/GTNBiEFG

## How was this tested?

Ran locally with successful and failing jobs

Feature/non-breaking


gageorgiev (approved):

antoniivanov:
Please add tests? 

I think the easiest way to test them would be with functional test: 

```
    runner = CliEntryBasedTestRunner()

    result: Result = runner.invoke(["run", util.job_path("fail-job-indirect-library")])
```

the result object contain stderr and std out which you can inspect for verification.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -141,24 +161,8 @@ def create_and_run_data_job(
...
-                    f"Data Job execution summary: {json.dumps(temp_exec_result, indent=2)}"
-                )
-
-                chunks = math.ceil(len(string_exec_result) / 5000)
-                for i in self.__split_into_chunks(exec_steps=steps, chunks=chunks):
-                    log.info(f"Execution Steps: {json.dumps(i, indent=2)}")
-
-            else:
-                log.info(f"Data Job execution summary: {execution_result}")
+            if context.configuration.get_value("LOG_EXECUTION_RESULT"):
```
antoniivanov, line 164: Where is the option "LOG_EXECUTION_RESULT" defined ?

There should be in vdk_configure hook definition : 
```
def vdk_configure(self, config_builder: ConfigurationBuilder)
    config_builder.add(....
```

DeltaMichael: We split the PR in two, as initially suggested by @dakodakov

https://github.com/vmware/versatile-data-kit/pull/2850/files

murphp15: so how do you know if a job completed successfully if this is disabled by default?

murphp15: A problem we have is that config is not super easy to discover. so ppl might not know about this.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -120,6 +120,26 @@ def __warn_on_python_version_disparity(
                         """
                     )
 
+    def __log_exec_result(self, execution_result):
```
antoniivanov, line 123: Please add type hints

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -120,6 +120,26 @@ def __warn_on_python_version_disparity(
...
+        # On some platforms, if the size of a string is too large, the
+        # logging module starts throwing OSError: [Errno 40] Message too long,
+        # so it is safer if we split large strings into smaller chunks.
+        string_exec_result = str(execution_result)
```
antoniivanov, line 127: By contract str(x) is not guaranteed to return parseable object so later json.loads may fail.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -120,6 +120,26 @@ def __warn_on_python_version_disparity(
...
+        if len(string_exec_result) > 5000:
+            temp_exec_result = json.loads(string_exec_result)
```
antoniivanov, line 129: what if string_exec_result is not proper json ?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -120,6 +120,26 @@ def __warn_on_python_version_disparity(
...
+            steps = temp_exec_result.pop("steps_list")
```
antoniivanov, line 130: Why pop? and not get?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
antoniivanov, line 123: Those methods seem more suitable for some "utils" module.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
DeltaMichael: Refactoring is out of scope for this PR. I'm just copying existing code to a new function. We can open a ticket to take a look at this at a later time.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
DeltaMichael: Refactoring is out of scope for this PR. I'm just copying existing code to a new function. We can open a ticket to take a look at this at a later time.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
DeltaMichael: Refactoring is out of scope for this PR. I'm just copying existing code to a new function. We can open a ticket to take a look at this at a later time.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
DeltaMichael: Refactoring is out of scope for this PR. I'm just copying existing code to a new function. We can open a ticket to take a look at this at a later time.

DeltaMichael: Added

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
antoniivanov: This is a new code as far as I can see. Moved from there? 
The code may fail. What are the guarantees that `json.loads(string_exec_result)` would not raise an exception . Logging  should not cause job to fail. 


Also we generally do not do refactoring dedicated stories unless they are big enough. Refactoring should be considered as a task as part of the implementation

DeltaMichael: Please look at the diff again

Screenshot 2023-11-14 at 16 36 43

DeltaMichael: To address your other points:

> Also we generally do not do refactoring dedicated stories unless they are big enough. Refactoring should be considered as a task as part of the implementation

The task is to put logging of the execution result behind a config option. I'm not about to introduce potentially breaking changes as part of this task. We can do it as part of another story and it's completely fine.

> The code may fail. What are the guarantees that json.loads(string_exec_result) would not raise an exception . Logging should not cause job to fail.

`str()` will try to call the object's `__str__()` method. If it's not implemented, it will fall back to `repr()` which calls the object's `__repr__()` method. In our case, we're calling `str()` on an `ExecutionResult` object.

Digging into the code a little, we can see it has a `__repr__` method. https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py#L127

And the `__repr__` method handles the exact case you're referring to and falls back to the python pretty printer. https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/execution_results.py#L164

It seems to me it will produce some kind of json if possible. https://docs.python.org/3/library/pprint.html#prettyprinter-objects

We're guaranteed to get an ExecutionResult object every time, because exceptions are handled in the `run` function. https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py#L327

antoniivanov: > We're guaranteed to get an ExecutionResult object every time, because exceptions are handled in the `run` function. https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/run/data_job.py#L327

For now. But by contract str and repr are not guaranteed to be json parseable. So if someone changes their implementation , that would break. Though one can hope and expect tests would catch this. 

But you are correct. As this existing code and not newly introduced one is subject to separate refactoring (if we decide to do it).

PR #2832, opened 2023-10-24
control-service: fix data job image building

mivanov1988:
Why
Changing Python versions without updating gitCommitSha doesn't trigger the build, causing data job issues due to different VDK images.

What
We improved the logic to check whether the Python version has changed, and if it has, the image building is triggered.

Testing done:
Unit tests.


dakodakov (approved):
LGTM

antoniivanov (approved):

PR #2833, opened 2023-10-24
vdk-events: Add ingest and anonymize workshop

antoniivanov:
We are planning a workshop that would guide people through the process of ingesting data and creating a configurable plugin to anonymize data before ingesting it into a database.

This is its home page and materials


antoniivanov:
I will merge it, This was reviewed live.

Issue #2803, opened 2023-10-17
vdk-structlog: create e2e tests


DeltaMichael:
## Overview

Further changes to vdk-structlog require tests that will ensure the current behaviour is upheld. We should have e2e tests that do the following

1. Run a simple data job with log statements in it
2. Parse stderr and ensure
3. Ensure configured metadata is present
4. Ensure configured logging format is upheld

These branch into several cases, for example

**Case 1**
Logs outside data job code an hooks should not have the vdk_data_job_name key, even if it's configured.
Logs inside data job code and hooks should have it

**Case 2**
Logs from bound loggers should display the bound context metadata wherever they're used and if the bound keys are added to the metadata configuration.
Logs outside of bound loggers should not display the bound context metadata.

**Case 3**
Extra params with should be displayed if the extra param keys are added to the metadata configuration.
Logs that do not pass extra params should not display the extra keys or values, even if they're added to the metadata configuration.

**Case 4**
Bound loggers should work with extra params. If extra params are passed to a bound logger, then they should be displayed together with the bound params, if both are part of the metadata config.

**Case 5**
Logs should respond to logging metadata configuration, e.g. if the timestamp is added to the config, it should be displayed. If it's not added, it should be omitted. This goes for both default values that are part of the log record, e.g. line number, file name, etc. and custom values passed through bound loggers and extra_params

## Acceptance criteria
1. Implement e2e testing for vdk-structlog plugin
2. Add the above cases as a minimum

Issue #2810, opened 2023-10-19
vdk-structlog: cannot remove timestamp when using json format


DeltaMichael:
## Steps to reproduce

1. Install vdk-structlog and configure as follows

```
logging_metadata=level,file_name,line_number,vdk_job_name
logging_format=json
```

2. Run an example data job

**Expected**

Timestamp not visible in log messages

**Actual**

Timestamp still visible in log messages

## Acceptance criteria
1. Fix the bug described above
2. Check this for other metadata fields
3. Create e2e tests that check for it if there aren't any

gageorgiev:
this issue appears for a number of other fields, maybe we should expand the scope of it

gageorgiev:
this issue appears for a number of other fields, maybe we should expand the scope of it

PR #2838, opened 2023-10-26
vdk-structlog: Tests

gageorgiev:
This change includes tests for the vdk-structlog plugin and addresses #2803.


projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
```diff
@@ -1,33 +1,81 @@
...
-from vdk.plugin.test_utils.util_funcs import cli_assert_equal
 from vdk.plugin.test_utils.util_funcs import CliEntryBasedTestRunner
 from vdk.plugin.test_utils.util_funcs import jobs_path_from_caller_directory
 
-"""
-This is a sample test file showing a recommended way to test new plugins.
-A good way to test a new plugin is how it would be used in the command that it extends.
-"""
 
+def test_structlog():
```
DeltaMichael, line 12: Can you also add a similar test using the JsonFormatter? Filtering is a little bit different there, so it should also be tested end-to-end.

projects/vdk-plugins/vdk-structlog/tests/jobs/job-with-bound-logger/10_dummy.py
```diff
@@ -0,0 +1,25 @@
...
+            "bound_test_key": "bound_test_value",
+            "excluded_bound_test_key": "excluded_value",
+        },
+    )
+
+    log.info("Log statement with no bound context")
+    bound_logger.info("Log statement with bound context")
+    bound_logger.info(
+        "Log statement with bound context and extra context",
+        extra={"extra_test_key": "extra_test_value"},
```
duyguHsnHsn, line 24: and here

duyguHsnHsn (approved):
I left minor comments, it loos good imo

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
```diff
@@ -1,33 +1,81 @@
...
-        cli_assert_equal(0, result)
+
+        test_log = [
+            x
+            for x in result.output.split("\n")
+            if "Log statement with bound context and extra context" in x
+        ][0]
+
+        # check that log level and job name do not appear if we have not included them in logging_metadata
+        assert "[INFO ]" not in test_log
```
DeltaMichael, line 80: Can you also test this for each canonical metadata field, e.g. timestamp, file_name, line_number, etc.? These fields are basically hiding %-format specifiers, so it would be nice to see that each of them filter correctly.

You can maybe do this using a loop and re-setting the environment for each execution, then checking that the metadata field is missing.

DeltaMichael (changes requested):
Please address my comments, I don't think the PR covers everything in the current state.

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
gageorgiev: did this for all except timestamps because I need to work around with regex and it overcomplicates the test

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
```diff
@@ -1,33 +1,120 @@
...
+                )
+
+                test_log = [
+                    x
+                    for x in result.output.split("\n")
+                    if "Log statement with bound context and extra context" in x
+                ][0]
+
+                # check that the removed_field in not shown in the log
+                assert STOCK_FIELD_REPRESENTATIONS[removed_field] not in test_log
```
DeltaMichael, line 86: I think you should also assert that the shown fields are indeed shown. Otherwise, it can get messy. For example, I know that the line number is not going to show up in the json logs like this `:23`. So just asserting that it's not there doesn't add anything to the test. We have to assert that the fields that weren't removed are present as well.

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
```diff
@@ -1,33 +1,120 @@
...
-        cli_assert_equal(0, result)
+JOB_NAME = "job-with-bound-logger"
+BOUND_TEST_KEY = "bound_test_key"
+BOUND_TEST_VALUE = "bound_test_value"
+EXTRA_TEST_KEY = "extra_test_key"
+EXTRA_TEST_VALUE = "extra_test_value"
+EXCLUDED_VALUE = "excluded_value"
+
+
+STOCK_FIELDS = ["level", "file_name", "line_number", "vdk_job_name"]
```
DeltaMichael, line 19: This is missing the timestamp, but I know that filtering it doesn't work for json logs yet. Just adding it here as a note, there's a separate ticket for it. https://github.com/vmware/versatile-data-kit/issues/2810

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
gageorgiev: included two sets of of field representations for console and json each

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
```diff
@@ -1,33 +1,157 @@
...
+                )
+
+                test_log = [
+                    x
+                    for x in result.output.split("\n")
+                    if "Log statement with bound context and extra context" in x
+                ][0]
+
+                # check that the removed_field in not shown in the log
+                assert stock_field_reps[removed_field] not in test_log
```
DeltaMichael, line 116: Can you also please assert that the shown fields are in the log?

```python
for field in shown_fields:
    assert stock_field_reps[field] in test_log
```

DeltaMichael (approved):

projects/vdk-plugins/vdk-structlog/tests/test_plugin.py
```diff
@@ -1,33 +1,157 @@
...
+        "level": '"level": "INFO"',
+        "file_name": '"filename:":"10_dummy.py"',
+        "line_number": '"lineno": 23',
+        "vdk_job_name": f'"vdk_job_name": "{JOB_NAME}"',
+    },
+}
+
+
+def test_structlog_console():
+    for log_format in [
```
antoniivanov, line 42: Please split the test into arrange, test, verity and move most of the arrange in helper method. (https://java-design-patterns.com/patterns/arrange-act-assert/)

antoniivanov: Also if you want to run same test with multiple parameters, the way to do is it to use `@pytest.mark.parametrize` see https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

antoniivanov (commented):
You seem to be testing too many things in a single test method. Split it around. 

Rename the test_plugin to something more telling. test_structlog_plugin for example.

PR #2840, opened 2023-10-26
vdk-core: fix bug in error classification

DeltaMichael:
## Why

Errors coming from libraries imported in user code are getting classified as platform errors in the execution result This is caused by the job_input_error_classifier checking for an incorrect attribute of the error object.

## What

Export the _vdk_resolvable_by and _vdk_resolvable_actual attributes to constants and replace wherever they're used in code

## How has this been tested

Functional test

Bugfix

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
@@ -31,9 +31,7 @@ def whom_to_blame(
     :return: ResolvableBy.PLATFORM_ERROR if exception was recognized as Platform Team responsibility.
              errors.ResolvableBy.USER_ERROR if exception was recognized as User Error.
     """
-    if isinstance(exception, errors.BaseVdkError) or hasattr(
-        exception, "to_be_fixed_by"
-    ):
+    if is_classified(exception):
         return errors.find_whom_to_blame_from_exception(exception)
     if is_user_error(exception, data_job_path):
         return errors.ResolvableBy.USER_ERROR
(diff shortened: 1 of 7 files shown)


gageorgiev (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/run/job_input_error_classifier.py
```diff
@@ -32,7 +32,7 @@ def whom_to_blame(
              errors.ResolvableBy.USER_ERROR if exception was recognized as User Error.
     """
     if isinstance(exception, errors.BaseVdkError) or hasattr(
-        exception, "to_be_fixed_by"
+        exception, errors.ATTR_VDK_RESOLVABLE_BY
```
antoniivanov, line 35: It's better to use error_classifiers.get_exception_resolvable_by(exception) which basically does hasattr(xx) but it's encapsulated 

```
if isinstance(exception, errors.BaseVdkError) or error_classifiers.get_exception_resolvable_by(exception): 
```

projects/vdk-core/src/vdk/internal/core/error_classifiers.py
```diff
@@ -310,3 +310,7 @@ def MSG_WHY_FROM_EXCEPTION(exception: Exception) -> str:
         return f"An exception occurred, exception message was: {message}"
 
     return _get_exception_message(exception)
+
+
+ATTR_VDK_RESOLVABLE_BY = "_vdk_resolvable_by"
```
doks5, line 315: Why setting these here, and not at the top?

doks5 (approved):

PR #2841, opened 2023-10-26
vdk-core: remove redundant logs

DeltaMichael:
## Why

As part of the run logs initiative, we should audit all log statements and error messages in vdk-core and make sure there aren't any redundancies, e.g. overly verbose messages or messages that don't add value.

## What

Remove usage of the below constants wherever possible

```
MSG_CONSEQUENCE_DELEGATING_TO_CALLER__LIKELY_EXECUTION_FAILURE
MSG_COUNTERMEASURE_FIX_PARENT_EXCEPTION 
MSG_CONSEQUENCE_TERMINATING_APP
```

Re-word log statements and exception messages that are too verbose

## How was this tested?

CI

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/src/vdk/internal/core/errors.py
+++ b/projects/vdk-core/src/vdk/internal/core/errors.py
@@ -10,6 +10,8 @@ It defines classes and methods for handling exceptions, and ensuring that there 
 """
 from __future__ import annotations
 
+from logging import Logger
+
 from vdk.internal.core.error_classifiers import *
 
 # Due to the expansion of this file's responsibilities over time,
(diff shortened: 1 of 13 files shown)


projects/vdk-core/src/vdk/internal/core/errors.py
```diff
@@ -10,6 +10,8 @@
 """
 from __future__ import annotations
 
+from logging import Logger
+
```
gageorgiev, line 14: why this?

gageorgiev (approved):

dakodakov (commented):
LGTM

projects/vdk-core/src/vdk/internal/core/errors.py
DeltaMichael: left over, will remove it

DeltaMichael: Actually, nvm, it's for the type hints. I'm leaving it as is.

dakodakov (approved):

PR #2843, opened 2023-10-27
vdk-duckdb: fix ingestion

antoniivanov:
Ingestion was not working properly due to multiple issues so it's being fixed.

- The plugin now seamlessly integrates with the VDK connection interface, making it easier to manage DuckDB connections instead of relying on custom DuckDB Connection 
- The ingestion insertion mechanism fixes:
     - The ingestion method was not registered (so job_input.send(method='duckdb') was not working)
     - Change the insertion to be batch and not one by one (which is very slow)
     - added ingestion tests to cover regressions and verify it's working
 - Renamed duckdb_file to database (which is what duckdb name for that parameter is)


mrMoZ1:
Could you update the PR description. Right now it's a bit vague.

mrMoZ1 (approved):
LGTM

PR #2846, opened 2023-10-27
vdk-core: ensure early logs are available

antoniivanov:
When vdk command is started logs do not really show up until click command start getting executed. All logs before that (from plugin hooks) seemed ignored.

I couldn't figure out for sure while click_log does that.

But this changes it so that early logs can be seen on console . The log level of the early logs is controlled by LOG_LEVEL_VDK option (which control the log level of VDK logs , since early logs are really vdk logs it make sense)

By default LOG_LEVEL_VDK is WARNING, so if more verbose logs are set it should be changed to INFO.

Testing Done:
```
export LOG_LEVEL_VDK=INFO
vdk run example
```
logs look like
https://gist.github.com/antoniivanov/bdd935927c1149b74944c58e96b50ee3

```
unset LOG_LEVEL_VDK
vdk run example
```
logs look like
https://gist.github.com/antoniivanov/afce585f8cd03ac4e09e9010dcdd2b9a

Tested also with vdk -v DEBUG ... to see debug logs show up correctly (with bot set and unset LOG_LEVEL_VDK)

Tested other commands vdk deploy, login, info that logs look like before (aka no unnecessary logs)

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
@@ -83,8 +83,8 @@ def configure_initial_logging_before_anything():
         is run which adds more context to the logs and initializes syslog handler (if configured to do so)
     """
     log_level = "WARNING"
-    if os.environ.get("LOG_LEVEL_VDK", None):
-        log_level = os.environ.get("LOG_LEVEL_VDK")
+    if os.environ.get(LOG_LEVEL_VDK, None):
+        log_level = os.environ.get(LOG_LEVEL_VDK)
     elif os.environ.get("VDK_LOG_LEVEL_VDK", None):
         log_level = os.environ.get("VDK_LOG_LEVEL_VDK")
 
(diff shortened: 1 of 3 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
```diff
@@ -65,6 +67,30 @@ def _parse_log_level_module(log_level_module):
...
+
+    Logging lifecycle:
+    1. This function is executed at the entry point of the app and configures default logging format and level
+    2. After vdk_initialize hook is executed and CLI command starts then `vdk --verbosity` option is taken (if set)
+        and click logging configuration takes over
+    3. When initialize_job hook (applicable for vdk run only) is executed then below configure_loggers function
+        is run which adds more context to the logs and initializes syslog handler (if configured to do so)
+    """
+    log_level = "WARNING"
+    if os.environ.get("LOG_LEVEL_VDK", None):
```
doks5, line 86: Why not using the imported `LOG_LEVEL_VDK` constant, imported at the top of the file?

doks5 (approved):

mivanov1988 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
antoniivanov: No reason, not to use it.

PR #2853, opened 2023-10-31
control-service: update ingress

mivanov1988:
Why
Currently, there is no way to exclude the user interface (UI) from the Ingress configuration.

What
To address this issue, we introduced a new check within the Ingress configuration. This check allows for determining whether the UI is enabled or not

Testing done:
helm template


dakodakov (approved):
ship it

PR #2859, opened 2023-11-01
vdk-data-source-git: data source for git POC

antoniivanov:
Extracts content from Git repositories along with associated file metadata. See README for more details


projects/vdk-plugins/vdk-data-source-git/src/vdk/plugin/data_source_git/git_source.py
```diff
@@ -0,0 +1,88 @@
...
+        description="SSH key to use when cloning the repo."
+        "Leave empty if no authentication is needed",
+        default="",
+    )
+
+
+@data_source(name="git", config_class=GitDataSourceConfiguration)
+class GitDataSource(IDataSource):
+    """
+    Data source who is only generating some dummy data for testing purposes.
```
murphp15, line 41: Remove this comment.

projects/vdk-plugins/vdk-data-source-git/src/vdk/plugin/data_source_git/git_source.py
```diff
@@ -0,0 +1,88 @@
...
+    git_url: str = config_field(description="Git URL that would be cloned. ")
+    git_ssh_key: Optional[str] = config_field(
+        description="SSH key to use when cloning the repo."
+        "Leave empty if no authentication is needed",
+        default="",
+    )
+
+
+@data_source(name="git", config_class=GitDataSourceConfiguration)
+class GitDataSource(IDataSource):
```
murphp15, line 39: This datasource will always only have a single stream?

murphp15 (approved):

projects/vdk-plugins/vdk-data-source-git/src/vdk/plugin/data_source_git/git_source.py
antoniivanov: At this iteration of the data source implementation. All singer data sources are really relational data sources (json / dictionary) I needed a data source that is blobs of data so that's why I developed that. To find out the limitations of our ingestion . Eventually it might make sense to allow users to configure streams - maybe branches, or directories or something else. But not at this first iteration. 

It's development status is pre-alpha currently or maybe alphase as per https://martin-thoma.com/software-development-stages/


PS : The main limitation I found is that we want the payloads to be json serializable so we don't really accept "bytes" in the payload.

projects/vdk-plugins/vdk-data-source-git/src/vdk/plugin/data_source_git/git_source.py
antoniivanov, line 41:
```suggestion
    Data source implementation is fetching data from git repository and associated metadata.
    For now it provides only 1 stream (for the whole repo). 
    Eventually, it might make sense to allow users to configure streams - maybe branches, or directories or something else. 
```

PR #2863, opened 2023-11-02
vdk-jupyterlab-extensions: update dependencies

murphp15:
# What

Upgrade dependencies so a build will will always create a runnnable jupyter lab. 

# How was this tested?
Tested the build locally and everything works good.


antoniivanov (approved):
What was wrong with the prevoius versions?

duyguHsnHsn (approved):

PR #2868, opened 2023-11-02
vdk-jupyter: run tests in CI 

murphp15:
# Why? 
We need to run jupyter tests in CICD to make sure there are no failures/regressions. 

Please note i have commented out failing tests: https://github.com/vmware/versatile-data-kit/issues/2870

# How has this been tested? 

Passing now. 
In this PR I tested failures are reported as expected https://github.com/vmware/versatile-data-kit/pull/2872


yonitoo (approved):
LGTM! What turned out to be the problem after all? I see we use a different image now. Was that it?

murphp15:
@yonitoo 
I'm not exactly sure what was causing the issue. I just went through every step and removed anything that wasn't totally necessary and then it started working.

duyguHsnHsn:
I see that 3 jobs are skipped do we know the reason why? 
Screenshot 2023-11-03 at 12 57 40

murphp15:
@duyguHsnHsn yeah I will fix that in a separate pr. I have the fix locally

duyguHsnHsn:
@paul-dimitrov how do we get alerted for those?

murphp15:
@duyguHsnHsn the gitlab job will be marked as a fail and it will be handle as normal?

duyguHsnHsn (approved):

Issue #2807, opened 2023-10-18
vdk-jupyter: adapt the job polling mechanism in the Jupyter UI


yonitoo:
**What is the feature request? What problem does it solve?**
The VDK JupyterLab extension encounters the following problem: it receives a 504 Gateway Time-out error while waiting for a long process to conclude. In order to address this, we introduce a polling mechanism that would allow us to execute jobs in the background and provide a way to track their status. 

**Suggested solution**
The idea is to adapt the TaskRunner interface to the UI. We want to use it upon sending a API request to the backend in order to start the job and then track its status until completion/error and receive an informative output in the end.

**Additional context**
It is a follow-up of the solution provided in #2789.

Issue #2806, opened 2023-10-18
vdk-jupyter: introduce a polling mechanism in the backend that would execute tasks in the background and track their status


yonitoo:
**What is the feature request? What problem does it solve?**
The VDK JupyterLab extension encounters the following problem: it receives a 504 Gateway Time-out error while waiting for a long process to conclude. This error likely comes from an intermediate proxy or the server configuration due to prolonged inactivity.

**Suggested solution**
The idea is to introduce a new class TaskRunner that would provide a polling mechanism. It would have an interface similar to this:

- start_task(id, task_handler) -> T/F (or something else); task_handler is a lambda function that would execute what the post methods of the current derived APIHandler classes do
- get_status(id) -> complex output dict/structure/class that would contain at least the following keys: "status", "message" and "error". It would be used to check the status of the currently running task and if the task has finished - proper message and error.

We would check the status of the ongoing task every x seconds or following some other pattern, having some threshold number of status checks. After the last one, if the task is still ongoing, we will return some kind of error to the user.

**Additional context**
It is a follow-up of the solution provided in #2789.

Issue #2789, opened 2023-10-11
Gateway Timeout in VDK UI for Long-Running Jobs Freezes Jupyter Lab Session[Solution]


duyguHsnHsn:
Solves: https://github.com/vmware/versatile-data-kit/issues/2741

#### Problem Narrative:
The issue at hand involves the VDK JupyterLab extension that initiates a server-side process expected to run for an extended period, sometimes over an hour (VDK operations like Run). However, the extension encounters a problem: it receives a 504 Gateway Time-out error while waiting for the process to conclude.This error isn't originating from the extension itself but likely from an intermediate proxy or the server configuration, indicative of a severed connection due to prolonged inactivity based on timeout settings.

#### Proposed solution:
_Note:_ Here when we talk about server, we do not mean Control Service but the Python component of the extension.
Asynchronous Requests with Status Checks (Polling)
Given these constraints, the recommended strategy pivots to an asynchronous model, characterised by the following steps:

1. Asynchronous Requests with Status Checks: The process begins with the client (React part of JupyterLab extension) sending a request to initiate the long-running task. Instead of keeping the request open(Python part of the extension), the server responds immediately with an acknowledgment that includes a unique task identifier for future status update.
2. Initiate the Process: This approach necessitates server-side adjustments(in the extension, not in Control Service) to ensure it doesn't wait for task completion before responding. Instead, it must return a task ID or something similar, enabling the client to track the task's progress independently of the original request.
3. Client-Side Polling: Armed with the task ID or something similar, the client implements a polling mechanism. Periodically, it dispatches a lightweight status inquiry to the server, using the task ID as a reference. This routine continues at regular intervals, ensuring the client remains updated regarding the task's status while avoiding connection inactivity.
4. Handle Task Completion: Upon receiving server confirmation that the task has concluded, the client issues a final request to fetch the resultant data or undertake subsequent steps as necessary.

 By decoupling the task's duration from the request-response cycle, this solution mitigates the risk of premature connection termination due to timeouts, ensuring that users receive the outcomes of their long-running tasks without disruption.

PR #2869, opened 2023-11-03
vdk-jupyter: introduce Task Runner, a polling mechanism that runs tasks in the background and tracks their status

yonitoo:
What: Introduce Task Runner, a polling mechanism that runs tasks in the background and tracks their status.
Implements [#2806](https://github.com/vmware/versatile-data-kit/issues/2806) and [#2807](https://github.com/vmware/versatile-data-kit/issues/2807).

Why: Address the problem described in [#2789](https://github.com/vmware/versatile-data-kit/issues/2789).

Testing Done: CI/CD is passing. Introduced relevant tests


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
```diff
@@ -100,18 +99,6 @@ def __prepare_vdk_run_command(arguments, path):
...
-        """
-        Execute `delete job`.
-        :param name: the name of the data job that will be deleted
-        :param team: the team of the data job that will be deleted
-        :return: message that the job is deleted
-        """
-        cmd = JobDelete(RestApiUrlConfiguration.get_rest_api_url())
-        cmd.delete_job(name, team)
-        return f"Deleted the job with name {name} from {team} team. "
-
```
gageorgiev, line 114: why are we deleting this?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/task_runner.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2021-2023 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+import threading
+
+
+class TaskRunner:
+    def __init__(self):
```
gageorgiev, line 7: A doc string maybe?

murphp15:
I would much prefer if you pushed the formatting in a separate PR.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -55,11 +131,22 @@ type jobRequestResult = {
...
+        {
+          body: JSON.stringify(getJobDataJsonObject()),
+          method: 'POST'
+        }
+      );
+      if (initialResponse.error) {
+        showError(initialResponse.error);
+        return { message: initialResponse.message, isSuccessful: false };
+      }
+      const finalResult = await pollForTaskCompletion('RUN');
```
murphp15, line 145: What about if two ppl run tasks in a notebook at the same time? 
does the task id not need to be unique to that run? 
the same would be true for the convert notebook. 
what happens if two ppl are trying to convert a notebook at the same time?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -47,6 +63,66 @@ type jobRequestResult = {
...
+  message: string | IJobConvertToNotebookMessage;
+  /**
+   * Error message
+   */
+  error: string;
+};
+
+const pollForTaskCompletion = async (
+  taskId: string,
+  maxAttempts = 43200,
```
murphp15, line 87: These don't seem like good defaults

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -20,6 +20,22 @@ const showError = async (error: any) => {
   );
 };
 
+interface IJobConvertToNotebookMessage {
+  code_structure: string[];
+  removed_files: string[];
+}
+
+type jobConvertToNotebookRequestResult = {
```
murphp15, line 28: This is probably a little subjective so you don't need to make this change. 

But I would probably make two types. 

jobConvertToNotebookSuccess {
message: IJobConvertToNotebookMessage
}
jobConvertToNotebookFailure{
message: string
}

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/vdk_ui.py
duyguHsnHsn: @gageorgiev we actually removed the delete option from the UI, and I forgot to remove it from here 
The reason we decided to remove delete from Jupyter UI is that it is kind of "dangerous" operation and it is better for people to remov jobs from the Operations UI or using the cli

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/task_runner.py
duyguHsnHsn: Here we definitely need a good docstring

duyguHsnHsn: As far as I see this class is responsible only for running a single task right? 
Do we want it to handle multiple tasks? Maybe we need to talk it over again

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -47,6 +63,66 @@ type jobRequestResult = {
...
+  error: string;
+};
+
+const pollForTaskCompletion = async (
+  taskId: string,
+  maxAttempts = 43200,
+  interval = 10000,
+  errorThreshold = 5
+): Promise<TaskStatusResult> => {
+  let errorCounter = 0;
```
duyguHsnHsn, line 91: Why do we count errors here? 
What type of errors are expected?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -47,6 +63,66 @@ type jobRequestResult = {
...
+        { method: 'GET' }
+      );
+      if (
+        result.task_type === taskId &&
+        (result.status === 'completed' || result.error)
+      ) {
+        return result;
+      }
+    } catch (error) {
+      showError(error);
```
duyguHsnHsn, line 105: Here we show the error but we keep trying to run the operation wouldn't it somehow confuse the user? 
Does the error say the VDK operation will continue running if not maybe it should be added as some message

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -47,6 +63,66 @@ type jobRequestResult = {
...
+        };
+      }
+    }
+    await new Promise(resolve => setTimeout(resolve, interval));
+  }
+  return {
+    task_type: taskId,
+    status: 'error',
+    message: '',
+    error: 'Task timed out'
```
duyguHsnHsn, line 122: Here we return that the task has timed out but what happens in the server, will the TaskRunner still be running the job itself or it will be terimated too?

yonitoo:
> I would much prefer if you pushed the formatting in a separate PR.

I didn't realise it became such big part of the change. I might move it to a separate PR, yes.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: My suggestion here was based on the idea that we got a limit of 12 hours for a job to run. That's why I decided it would make sense to initially try to poll every 10 seconds for these 12 hours which would result in 4320 polls. (added one more 0 here it seems.)

This is initial idea but maybe exponential backoff IMO would fit good here.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: I can introduce the 2 types you proposed, yeah, sounds good.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: My intuition here is that we do polling to get the status of the job and if there is some temporary (connectivity for example) issue while a task is in progress, then we would give it a retry for some time until we eventually stop trying (5 attempts). That's why I introduced this error threshold for such cases.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: We don't support running multiple tasks simultaneously at this point. When we decide to, we can modify the task runner appropriately and introduce UID or some unique prefix/suffix to the task id.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: The argument is the same as above. If we don't find it reasonable enough to handle such cases using the error threshold, we can directly return the error on each occasion and stop the operation on the first error.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
murphp15: but this will run on the jupyter server. so does one user block it for all other users? 
That seems like awful functionality.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: The timeout case is designed to happen after 12 hours pass without any success. This is also our limit for the VDK operations, right? If I am mistaken, correct me and I will add some server-side timeout mechanism as well.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
murphp15: But different users will have the same task id on the server ?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
murphp15: Ok, that reasoning is probably ok for a first iteration. 
But you need to add a comment here explaining that.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: @yonitoo I am not sure about the limit of the VDK operations if it is the same, I suppose it is not a problem but i do think 12 hours is too much for this maybe @antoniivanov can share his opinion too? 12 h for UI imo too much maybe 5-6 h is better?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
```diff
@@ -81,13 +81,14 @@ export async function showConvertJobToNotebookDialog(
...
-      let { message, status } = await jobConvertToNotebookRequest();
-      if (status) {
-        const transformjobResult = JSON.parse(message);
-        notebookContent = initializeNotebookContent(
-          transformjobResult['code_structure'],
-          transformjobResult['removed_files']
-        );
+      let { message, isSuccessful } = await jobConvertToNotebookRequest();
+      if (isSuccessful) {
+        if (typeof message !== 'string') {
```
DeltaMichael, line 86: Nit: It might be worth merging this condition with the above one, e.g.

```
if (isSuccessful && typeof message !== 'string') {
// do stuff
}
```

DeltaMichael: Also, it's not really clear why we check the message type.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/task_runner.py
DeltaMichael: Is there a use case for running multiple tasks from a single notebook? If yes, then this should definitely be updated to poll multiple tasks.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
yonitoo: I think when we introduce the 2 types Paul proposed, it will be more clear.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: Actually, then we would have to go and refactor all the empty string messages because it would not recognise "string" as their type when replaced by the 2 new types. I will not add them for now.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/components/ConvertJobToNotebook.tsx
yonitoo: Actually, I will add a comment to explain why we do this.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/task_runner.py
yonitoo: I don't think I got this out of the discussion prior to the implementation. I thought we want to support single operation at a time. I guess it's a matter of when do we want to start handling multiple tasks but I think for now handling one should work just well and then update it when we want to handle multiple tasks.
@antoniivanov what is your view on this?

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
duyguHsnHsn: @yonitoo I understand, but i think we somehow should manage those errors, what if it is some other essential error in this way we might lose track of it?

yonitoo: Well, then we can initially just make it fail instantly and later change it.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/task_runner.py
yonitoo: I'd suggest we go with a first iteration of supporting 1 task at a time and improve it as a next step.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: Made it this way.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: Done. Made use of UUID to do so.

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/vdk_jupyterlab_extension/handlers.py
```diff
@@ -103,15 +124,26 @@ class DownloadJobHandler(APIHandler):
...
+        if task_id:
+            self.finish(json.dumps({"message": f"Task {task_id} started", "error": ""}))
+        else:
+            self.finish(
+                json.dumps(
+                    {
+                        "message": "Task DOWNLOAD failed to start",
+                        "error": "Another task is already running",
+                    }
+                )
```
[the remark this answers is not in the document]
yonitoo: Thanks, done.

murphp15 (approved):

PR #2874, opened 2023-11-03
vdk-structlog: fix filtering of metadata fields for json

DeltaMichael:
## Why?

Filtering by timestamp, level and vdk_job_name does not work for json formatting

## What?

- Introduce logic that inserts the correct fields in the json formatter log record
- Move the metadata filters after the filters that add the job name
- Fix bug in ConsoleMetadataFilters
- Add timestamp to tests
- Enable json tests

## How was this tested?

Ran functional tests locally

Bugfix

## Follow-up

https://github.com/vmware/versatile-data-kit/issues/2875

### DIFF
--- a/projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
+++ b/projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
@@ -75,6 +75,7 @@ class StructlogPlugin:
             description=(
                 f"Possible values: {STRUCTLOG_LOGGING_METADATA_ALL_KEYS}"
                 "User-defined key-value pairs added to the logger's context will be displayed after the metadata, but before the message"
+                "Keys for user-defined key-value pairs have to be added in this config option for the values to be displayed in the metadata"
             ),
         )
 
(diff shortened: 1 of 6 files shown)


projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -121,8 +122,8 @@ def initialize_job(self, context: JobContext) -> None:
 
         handler = logging.StreamHandler(sys.stderr)
         handler.setFormatter(formatter)
-        handler.addFilter(metadata_filter)
```
murphp15, line 124: why is the order changed?

murphp15 (approved):

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: Filters execute in sequence. If you add the job name after you filtered the metadata, filtering the job name in the metadata will not work. You'll be stuck with it til the end of the execution.

DeltaMichael: This is currently a problem with `vdk_step_name` and `vdk_step_type`. I opened a separate ticket for it https://github.com/vmware/versatile-data-kit/issues/2875

duyguHsnHsn (approved):

Issue #2568, opened 2023-08-17
Audit existing log statements in vdk-plugins


DeltaMichael:
## Overview

Based on user feedback, our info logging is too verbose. This means there are probably logs at the info level and below that should be moved to debug and above. We should search through our code base and move such logs to the appropriate level.

In addition, we have error messages that are too verbose and don't provide much value. These messages should be shortened.

## Acceptance criteria
1. Search for verbose logs in vdk-plugins
2. Move logs to higher levels
3. Remove unnecessary text from error messages

PR #2878, opened 2023-11-03
vdk-plugins: Audit log statements

gageorgiev:
This PR goes over every log statement in each VDK plugin and shortens ones which need to be shorter, deletes some which are deemed unnecessary, and lowers the level of other ones with the aim to reduce log clutter and improve readability.
Addresses #2568

Testing done: CI


yonitoo (approved):
lgtm

sabadzhiev (approved):

PR #2887, opened 2023-11-09
vdk-structlog: LTSV formatting

gageorgiev:
This PR incorporates the LTSV formatting which was present in the vdk-logging-format plugin to vdk-structlog, rendering vdk-logging- format unneccessary.

Testing done: unit tests cover LTSV formatting

### DIFF
--- a/projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/formatters.py
+++ b/projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/formatters.py
@@ -42,6 +42,22 @@ class JsonFormatter(jsonlogger.JsonFormatter):
             log_record["level"] = record.levelname
 
 
+class LtsvFormatter(Formatter):
+    """
+    Formats logs in the LTSV format.
+    
+    """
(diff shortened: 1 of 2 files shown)


projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/formatters.py
```diff
@@ -42,6 +42,23 @@ def add_fields(
             log_record["level"] = record.levelname
 
 
+class LtsvFormatter(Formatter):
+    """
+    Formats logs in the LTSV format.
```
murphp15, line 47: I would like if the not acronym value was used in the comment

murphp15 (approved):
Looks good

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/formatters.py
duyguHsnHsn: I agree here

duyguHsnHsn (approved):

PR #2897, opened 2023-11-14
control-service: write resources to db

mrMoZ1:
what: If no resources specified when deploying data jobs with the DataJobsSynchronizer we will now write the default resources values to the DB.

why: Previously these values would only be present in K8S.

testing: added unit test, modified integration test.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobDefaultConfigurations.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobDefaultConfigurations.java
@@ -6,6 +6,8 @@
 package com.vmware.taurus.service.deploy;
 
 import com.vmware.taurus.service.KubernetesService;
+import java.text.NumberFormat;
+import java.text.ParseException;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
@@ -30,5 +32,33 @@ public class DataJobDefaultConfigurations {
 
   public KubernetesService.Resources dataJobLimits() {
     return new KubernetesService.Resources(cpuLimits, memoryLimits);
+  }
+
+  public float getCpuRequests() throws ParseException {
+    return NumberFormat.getInstance().parse(cpuRequests).floatValue();
+  }
+
+  public float getCpuLimits() throws ParseException {
+    return NumberFormat.getInstance().parse(cpuLimits).floatValue();
+  }
+
+  public int getMemoryRequests() throws ParseException {
+    var memoryRequest = NumberFormat.getInstance().parse(memoryRequests).intValue();
+    return getMemoryInMi(memoryRequests, memoryRequest);
+  }
+
+  public int getMemoryLimits() throws ParseException {
+    var memoryLimit = NumberFormat.getInstance().parse(memoryLimits).intValue();
+    return getMemoryInMi(memoryLimits, memoryLimit);
+  }
+
+  private int getMemoryInMi(String memory, int amount) {
+    if (memory.endsWith("Gi")) {
+      return amount * 1024;
+    }
+    if (memory.endsWith("G")) {
+      return amount * 1000;
+    }
+    return amount;
   }
 }
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobDefaultConfigurations.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobDefaultConfigurations.java
@@ -43,22 +43,12 @@ public class DataJobDefaultConfigurations {
   }
 
   public int getMemoryRequests() throws ParseException {
-    var memoryRequest = NumberFormat.getInstance().parse(memoryRequests).intValue();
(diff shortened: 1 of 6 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DataJobDefaultConfigurations.java
```diff
@@ -31,4 +33,32 @@ public KubernetesService.Resources dataJobRequests() {
...
+
+  private int getMemoryInMi(String memory, int amount) {
+    if (memory.endsWith("Gi")) {
+      return amount * 1024;
+    }
+    if (memory.endsWith("G")) {
+      return amount * 1000;
+    }
+    return amount;
+  }
```
antoniivanov, line 63: Ideally, our code should account for a wide range of expected scenarios. In this case this includes accommodating different memory units such as MB, KB, and even TB, in addition to GB. Users might need these variations, and our code should be flexible enough to handle them.

Also It seems like this type of function is a prime candidate for a Utility class, given its very reusable nature.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/K8SMemoryConversionUtils.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2021-2023 VMware, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.vmware.taurus.service.deploy;
+
+public class K8SMemoryConversionUtils {
+
+  public static int getMemoryInMi(String memory, int amount) {
```
antoniivanov, line 10: It's minor thing but it bugs me. I would expect the value to be first and then the unit (1 KB not KB 1) 

ALso memory is misleading a bit variable name. Better if it is "memoryUnit"

```suggestion
  public static int getMemoryInMi(int amount, String memoryUnit) {
```

antoniivanov (approved):
Looks good to me.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployerV2.java
```diff
@@ -229,10 +229,76 @@ private ActualDataJobDeployment updateCronJob(
           DeploymentModelConverter.toActualJobDeployment(
               desiredDataJobDeployment, desiredDeploymentVersionSha, lastDeployedDate);
     }
-
+    checkDefaultDeploymentResources(actualJobDeployment);
```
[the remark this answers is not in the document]
mivanov1988: I think we should incorporate this logic into the `DeploymentModelConverter.toActualJobDeployment()` and get the resources from `KubernetesService.CronJob desiredCronJob` since they are already set.

mivanov1988:
I think we can verify this change through the integration test.

mrMoZ1:
> I think we can verify this change through the integration test.

We are, the integration test specifically checks the resources' amounts.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployerV2.java
```diff
@@ -346,34 +345,37 @@ private KubernetesService.CronJob getCronJob(
...
+            .map(cpuRequestM -> cpuRequestM + "m")
             .orElse(defaultConfigurations.dataJobRequests().getCpu());
-
+    checkDefaultCpuRequest(jobDeployment.getResources(), cpuRequest);
     String cpuLimit =
         Optional.ofNullable(jobDeployment.getResources())
             .map(DataJobDeploymentResources::getCpuLimitCores)
             .filter(cpuLimitCores -> cpuLimitCores > 0)
-            .map(String::valueOf)
+            .map(cpuLimitM -> String.valueOf(cpuLimitM.intValue()))
```
mivanov1988, line 361: Why intValue?

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageDeployerV2.java
```diff
@@ -346,34 +345,37 @@ private KubernetesService.CronJob getCronJob(
...
     String cpuRequest =
         Optional.ofNullable(jobDeployment.getResources())
             .map(DataJobDeploymentResources::getCpuRequestCores)
             .filter(cpuRequestCores -> cpuRequestCores > 0)
-            .map(String::valueOf)
+            .map(cpuRequestM -> String.valueOf(cpuRequestM.intValue()))
+            .map(cpuRequestM -> cpuRequestM + "m")
             .orElse(defaultConfigurations.dataJobRequests().getCpu());
-
+    checkDefaultCpuRequest(jobDeployment.getResources(), cpuRequest);
```
mivanov1988, line 356: In my opinion, a better place for these assignments is in the

`DeploymentModelConverter.toActualJobDeployment(desiredDataJobDeployment, desiredDeploymentVersionSha, lastDeployedDate);` 

Since the purpose of this method is just to construct the job.

mivanov1988 (approved):

PR #2916, opened 2023-11-17
vdk-jupyter: fix bug for failed requests and improve error handling

yonitoo:
What: Bug fix. Since the Task Runner change got in, if you tried to deploy a job without keytab, you would not have been informed about the error in any way and the Task Runner state wouldn't have been cleaned, therefore, not being able to handle any other request. Only a restart of the session would have solved this which is not good. 
To address this I:
* Modified the Task Runner start_task() method to proceed after 'failed' last request.
* Improved the error handling - show the error of failing request which was not shown and made the backend handlers able to detect any errors.

Why: To make it work in a better way and not to have to restart the session on each bad request.

Testing Done: Introduced relevant tests. The rest of the tests are passing.

### DIFF
--- a/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
+++ b/projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
@@ -104,11 +104,13 @@ const pollForTaskCompletion = async (
         `taskStatus?taskId=${taskId}`,
         { method: 'GET' }
       );
-      if (
-        result.task_id === taskId &&
-        (result.status === 'completed' || result.error)
-      ) {
-        return result;
+      if (result.task_id === taskId) {
+        if (result.status !== 'running') {
+          if (result.status === 'failed') {
+            showError(result.error);
+          }
+          return result;
+        }
       }
     } catch (error) {
       showError(error);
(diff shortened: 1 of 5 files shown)


projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
```diff
@@ -104,11 +104,13 @@ const pollForTaskCompletion = async (
         `taskStatus?taskId=${taskId}`,
         { method: 'GET' }
       );
-      if (
-        result.task_id === taskId &&
-        (result.status === 'completed' || result.error)
-      ) {
-        return result;
+      if (result.task_id === taskId) {
+        if (result.status !== 'running') {
```
DeltaMichael, line 108: Why check this if you're going to check `result.status === 'failed'` after?

DeltaMichael (approved):

projects/vdk-plugins/vdk-jupyter/vdk-jupyterlab-extension/src/serverRequests.ts
yonitoo: 3 scenarios for status:
running - continue;
completed - return result;
failed - show error and return result.

antoniivanov (commented):
I don't really see tests covering the scenario

gageorgiev (approved):
looks good, add a test or two and go ahead

PR #2922, opened 2023-11-20
vdk-structlog: Test for compatibility for log_level_module propagation 

gageorgiev:
This change ensures that vdk-structlog tests covers the log_level_module
functionality compabitibility with vdk-structlog log configuration.

Testing done: unit tests


projects/vdk-plugins/vdk-structlog/tests/jobs/job-with-bound-logger/config.ini
```diff
@@ -0,0 +1,2 @@
+[vdk]
+LOG_LEVEL_MODULE=vdk.internal.builtin_plugins.run.file_based_step=WARNING
```
antoniivanov, line 2: When you're writing tests, especially unit tests, the goal is to make them as deterministic and isolated as possible. This means that a test should not depend on external factors or states that are outside of the test's control. Unless the test wants to verify that file_based_step is never logged at warning :) 

Now imagine if someone changes the log message or the log level. No reason this test should fail if they do so.

So it would be better if you add a log statement yourself (e.g in the data job) and verify it's not logged. You can control the module name with logging.getLogger(module-name) so  you can set it to anything.

gageorgiev: that was my initial approach but I kept having issues with the consistency between running the job and running the test (if I run the job through the CLI, the log wouldn't appear, but it would when running the test), I'll try your suggested approach since I haven't tried changing the module name and only used __name__

antoniivanov: Cool. 

A quick advice from me: 

Please try to get into a habit of writing comments about Why something is done (not what or how). 
It's not just my advice [here](https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/) and [here](https://www.jackfranklin.co.uk/blog/code-comments-why-not-how/)

DeltaMichael: @gageorgiev If you still have problems with the above suggestion, you can inline your own plugin that just outputs a log statement and plug it for that one test. I think either is fine.

DeltaMichael (approved):

yonitoo (approved):

Issue #2903, opened 2023-11-15
vdk-structlog: support the default console logger and logging configuration


DeltaMichael:
## Overview

We should support the default console logger and logging format configuration. A user should be able to select the logging format as `console` and pass a format string like `(asctime)s %(name)-12s %(levelname)-8s %(message)s`

## Acceptance Criteria
1. vdk-structlog supports default console logger and logging configuration

PR #2936, opened 2023-11-23
vdk-structlog: default formatter

duyguHsnHsn:
Solves: [vdk-structlog: support the default console logger and logging configuration#2903](https://github.com/vmware/versatile-data-kit/issues/2903)


projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/constants.py
```diff
@@ -4,6 +4,7 @@
 
 STRUCTLOG_LOGGING_METADATA_KEY = "logging_metadata"
 STRUCTLOG_LOGGING_FORMAT_KEY = "logging_format"
+STRUCTLOG_LOGGING_CUSTOM_FORMAT_KEY = "logging_custom_format"
```
antoniivanov, line 7: This is applicable only for Console format. 

So really it should be STRUCTLOG_CONSOLE_LOG_PATTERN 

The name is inspired by the similar variable in spring logging - https://docs.spring.io/spring-boot/docs/2.1.13.RELEASE/reference/html/boot-features-logging.html

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
```diff
@@ -113,6 +113,34 @@ def test_stock_fields_removal(log_format):
                 assert re.search(stock_field_reps[shown_field], test_log) is not None
 
 
+@pytest.mark.parametrize("log_format", ["console"])
```
antoniivanov, line 116: It's better to have separate test file for each module. 

e.g test_formatter.py test_structlog_plugin.py ... 
 

- This way the test module are more maintainable (no one long file) 
- It's easier to find relevant tests (module X tests are in test_x)

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -134,14 +151,19 @@ def initialize_job(self, context: JobContext) -> None:
 
     @hookimpl(hookwrapper=True)
     def run_job(self, context: JobContext) -> Optional[ExecutionResult]:
```
antoniivanov, line 153: This method is starting to doo too many things (too many responsibilities) 

Pllease refactor it. There are tips in https://refactoring.guru/smells/long-method

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
duyguHsnHsn: I will do this in a separate PR along with the refactoring of the existing code

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/formatters.py
```diff
@@ -153,8 +155,12 @@ def create_formatter(logging_format: str, metadata_keys: str) -> [Formatter, Fil
...
-        formatter = ConsoleFormatter(
-            fmt=StructlogMetadataBuilder(metadata_keys).build_console_format()
-        )
+        if custom_console_format:
+            format_string = custom_console_format
+        else:
+            format_string = StructlogMetadataBuilder(
+                metadata_keys
+            ).build_console_format()
+        formatter = ConsoleFormatter(fmt=format_string)
```
gageorgiev, line 164: We should probably do some validation on the format string, and disregard it if it doesn't pass. Otherwise a misconfigured format string will cause difficult to debug problems with logs down the line.

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
```diff
@@ -113,6 +113,34 @@ def test_stock_fields_removal(log_format):
...
+def test_custom_console_format(log_format):
+    custom_format_string = "%(asctime)s %(name)-12s %(levelname)-8s %(message)s"
+
+    with mock.patch.dict(
+        os.environ,
+        {
+            "VDK_LOGGING_METADATA": "timestamp,level,file_name,line_number,vdk_job_name",
+            "VDK_LOGGING_FORMAT": log_format,
+            "STRUCTLOG_CONSOLE_LOG_PATTERN": custom_format_string,
```
DeltaMichael, line 125: I'm not sure this is the correct env variable. Shouldn't it be `VDK_LOGGING_CUSTOM_FORMAT`?

DeltaMichael: Also, can you please add a description of the behavior to README.md. It should say that VDK_LOGGING_CUSTOM_FORMAT overrides VDK_LOGGING_METADATA and VDK_LOGGING_FORMAT.

DeltaMichael: I don't like the way this test is set up. It can pass because of the additional configuration even if something is not right. It's better to have two tests: one that tests if the custom format works and another that tests if it overrides the other settings correctly.

DeltaMichael (changes requested):
Please address my comments and add a description to the PR.

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
duyguHsnHsn: @DeltaMichael about the naming, please check the previous comments of Antoni I got it from there.
I will change the README but firstly will go over the tests

DeltaMichael: @duyguHsnHsn My issue is not with the naming, you're setting a variable that doesn't exist. Please check again.

This

```python
"STRUCTLOG_CONSOLE_LOG_PATTERN": custom_format_string,
```

should be this

```python
"VDK_LOGGING_CUSTOM_FORMAT": custom_format_string,
```

because of this

```
STRUCTLOG_CONSOLE_LOG_PATTERN = "logging_custom_format"
```

`STRUCTLOG_CONSOLE_LOG_PATTERN` is just a const name in this case. You should be passing the value behind it, prefixed with `VDK`.

duyguHsnHsn: Oh, you are right, sorry for the confusion

DeltaMichael (approved):

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/formatters.py
duyguHsnHsn: We talked it over with the team and decided it would be an overkill. And also there is not a common way to do it, i will need to generate a regex which is not really safe since it might not work for every possible option.

PR #2939, opened 2023-11-24
vdk-structlog: incorporate vdk-core built-in logging plugin into vdk-structlog

yonitoo:
What:
Incorporate vdk-core built-in logging plugin into vdk-structlog by moving the logging configuration and relevant tests, adding deprecation warning and keeping the vdk-core functionality behind a flag which is activated upon vdk-structlog installation.

Why: To avoid reconfiguring the logging config, thus losing information, by vdk-structlog but instead just configure it once.

Testing Done: Existing tests pass and the vdk-core logging tests are added to vdk-structlog and are passing (except test_log_config()).

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
@@ -5,6 +5,7 @@ import os
 import re
 import socket
 import types
+import warnings
 from sys import modules
 from typing import cast
 
@@ -81,13 +82,19 @@ def configure_initial_logging_before_anything():
     3. When initialize_job hook (applicable for vdk run only) is executed then below configure_loggers function
         is run which adds more context to the logs and initializes syslog handler (if configured to do so)
     """
-    log_level = "WARNING"
-    if os.environ.get(LOG_LEVEL_VDK, None):
-        log_level = os.environ.get(LOG_LEVEL_VDK)
-    elif os.environ.get("VDK_LOG_LEVEL_VDK", None):
-        log_level = os.environ.get("VDK_LOG_LEVEL_VDK")
+    warnings.warn(
+        "The vdk-core logging configuration is deprecated and will be removed in a future version. "
+        "Please use vdk-structlog for logging configuration.",
+        DeprecationWarning,
+    )
+    if not os.environ.get("VDK_USE_STRUCTLOG"):
+        log_level = "WARNING"
+        if os.environ.get(LOG_LEVEL_VDK, None):
+            log_level = os.environ.get(LOG_LEVEL_VDK)
+        elif os.environ.get("VDK_LOG_LEVEL_VDK", None):
+            log_level = os.environ.get("VDK_LOG_LEVEL_VDK")
 
-    logging.basicConfig(format="%(message)s", level=logging.getLevelName(log_level))
+        logging.basicConfig(format="%(message)s", level=logging.getLevelName(log_level))
 
 
 def configure_loggers(
--- a/projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
+++ b/projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
@@ -1,28 +1,155 @@
 # Copyright 2021-2023 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 import logging
+import os
 import sys
 from typing import List
 from typing import Optional
 
-from pythonjsonlogger import jsonlogger
 from vdk.api.plugin.hook_markers import hookimpl
 from vdk.api.plugin.plugin_registry import HookCallResult
 from vdk.api.plugin.plugin_registry import IPluginRegistry
+from vdk.internal.builtin_plugins.config import vdk_config
 from vdk.internal.builtin_plugins.run.execution_results import ExecutionResult
 from vdk.internal.builtin_plugins.run.execution_results import StepResult
 from vdk.internal.builtin_plugins.run.job_context import JobContext
 from vdk.internal.builtin_plugins.run.step import Step
+from vdk.internal.core import errors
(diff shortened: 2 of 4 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
```diff
@@ -81,13 +82,19 @@ def configure_initial_logging_before_anything():
...
         is run which adds more context to the logs and initializes syslog handler (if configured to do so)
     """
-    log_level = "WARNING"
-    if os.environ.get(LOG_LEVEL_VDK, None):
-        log_level = os.environ.get(LOG_LEVEL_VDK)
-    elif os.environ.get("VDK_LOG_LEVEL_VDK", None):
-        log_level = os.environ.get("VDK_LOG_LEVEL_VDK")
-
-    logging.basicConfig(format="%(message)s", level=logging.getLevelName(log_level))
+    warnings.warn(
```
DeltaMichael, line 85: Please don't make changes to core and plugins in the same PR. Plugin tests don't run against vdk-core that was built from source and this often leads to confusion. This particular change won't cause a problem, because there are no tests related to it.

Please move the core change to a separate PR. Maybe we don't even need to merge it right now, because structlog is not fully ready yet.

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
```diff
@@ -56,6 +63,63 @@
...
+                attempt_id="attempt-id",
+                log_level_module="a.b.c=INFO;foo.bar=ERROR",
+            )
+            configured_loggers = mock_dict_config.call_args[0][0]["loggers"]
+            assert configured_loggers["a.b.c"]["level"] == "INFO"
+            assert configured_loggers["foo.bar"]["level"] == "ERROR"
+
+
+def test_log_plugin_exception():
+    print("This")
```
DeltaMichael, line 98: Nit: remove this line before merging.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -87,8 +216,33 @@ def vdk_configure(self, config_builder: ConfigurationBuilder):
...
+            default_value="UDP",
+            description=f"Socket type for SysLog log forwarding connection. Currently possible values are "
+            f"{list(SYSLOG_SOCK_TYPE_VALUES_DICT.keys())}",
+        )
+
     @hookimpl
     def vdk_initialize(self, context: CoreContext):
+        # Use this plugin for logging config and warn the user about the vdk-core logging deprecation
+        os.environ["VDK_USE_STRUCTLOG"] = "1"
+        configure_initial_logging()
```
DeltaMichael, line 245: Can you please post log outputs of jobs using all formats, e.g. one for json, one fore console, one for ltsv? 

I have a hunch that this might mess up parts of the initial log. But I also see we still eject the formatter later on, so not sure. It would be nice to see the full output.

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
```diff
@@ -56,6 +63,63 @@
 }
 
 
+def test_parse_log_level_module():
```
DeltaMichael, line 66: It would be nice to have some tests for syslog output. If you think that's too involved for this PR, it can become a separate ticket.

yonitoo: Yeah, I will include such tests as part of another PR.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
```diff
@@ -81,13 +82,19 @@ def configure_initial_logging_before_anything():
...
+        log_level = "WARNING"
+        if os.environ.get(LOG_LEVEL_VDK, None):
+            log_level = os.environ.get(LOG_LEVEL_VDK)
+        elif os.environ.get("VDK_LOG_LEVEL_VDK", None):
+            log_level = os.environ.get("VDK_LOG_LEVEL_VDK")
+
+        logging.basicConfig(format="%(message)s", level=logging.getLevelName(log_level))
 
 
 def configure_loggers(
```
antoniivanov, line 100: You should add the check in log_config.py where configure_loggers is called as well no?

projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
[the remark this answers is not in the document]
yonitoo: Alright, I see, I will reword this warning.

yonitoo: ok, moving it to another PR.

projects/vdk-plugins/vdk-structlog/tests/test_structlog.py
yonitoo: Done.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
yonitoo: Added the check as part of the PR #2944.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
yonitoo: Added 1 for JSON and 1 for console to the PR description.

yonitoo: Here it is: https://pastebin.com/8NnfF0En

DeltaMichael (changes requested):
We should have further discussion around this. I really messed up the task definition. 😕

DeltaMichael:
Issue was resolved here.

PR #2941, opened 2023-11-24
control-service: deployment cannot be suspended

mivanov1988:
Why:
The deployment cannot be suspended in the case of a previously failed one.

What:
Added additional logic to the DeploymentServiceV2 to handle this particular case.

Testing done:
IT and unit tests.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentServiceV2.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentServiceV2.java
@@ -160,8 +160,9 @@ public class DeploymentServiceV2 {
           dockerRegistryService.dataJobImage(
               desiredJobDeployment.getDataJobName(), desiredJobDeployment.getGitCommitSha());
 
-      if (jobImageBuilder.buildImage(
-          imageName, dataJob, desiredJobDeployment, actualJobDeployment, sendNotification)) {
+      // If the job suspension operation is performed, we don't want to build the image.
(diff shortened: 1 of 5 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentServiceV2.java
```diff
@@ -188,6 +176,53 @@ public void updateDeployment(
...
+
+    if (Boolean.FALSE.equals(desiredJobDeployment.getEnabled())) {
+      suspendDeployment(actualJobDeployment, isJobDeploymentPresentInKubernetes);
+      log.trace(
+          "Skipping the data job [job_name={}] deployment due to the suspended deployment",
+          dataJob.getName());
+      return false;
+    }
+
+    if (DeploymentStatus.USER_ERROR.equals(desiredJobDeployment.getStatus())
```
doks5, line 200: Out of curiosity, if a deployment enters an error state, how can it get out of this state, given that it looks like it will always be skipped?

doks5 (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/DeploymentServiceV2.java
mivanov1988: If you redeploy the data job through the API, the status will be reset.

doks5: Got it, thanks

dakodakov (approved):
LGTM

Issue #2813, opened 2023-10-19
vdk-structlog: incorporate vdk-core built-in logging plugin into vdk-structlog


DeltaMichael:
## Overview

vdk-core has a built-in plugin for logging configuration. This is not ideal, because vdk-structlog has to eject the previous configuration and re-configure. This loses parts of the configuration that are present in the core plugin but not in structlog. The config done in core should be moved to vdk-structlog and core plugin should be deprecated.

## Acceptance Criteria
1. Configuration from core plugin is moved to vdk-structlog
2. Core plugin is deprecated (deprecation should follow https://github.com/vmware/versatile-data-kit/wiki/Coding-Standard#deprecations)
3. Core plugin functionality put behind feature flag, e.g. installing  vdk-structlog sets the core flag and disables the functionality

PR #2944, opened 2023-11-27
vdk-core: add logging plugin warning and check if the vdk-structlog plugin is used

yonitoo:
What: This is part of #2813 which is addressed by #2939. Add logging plugin warning and check if the vdk-structlog plugin is used by checking the dedicated env var VDK_USE_STRUCTLOG.

Why: To notify the user about the logging configuration changes and to guide them to use vdk-structlog instead.


gageorgiev (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
```diff
@@ -81,13 +82,17 @@ def configure_initial_logging_before_anything():
     3. When initialize_job hook (applicable for vdk run only) is executed then below configure_loggers function
         is run which adds more context to the logs and initializes syslog handler (if configured to do so)
     """
-    log_level = "WARNING"
-    if os.environ.get(LOG_LEVEL_VDK, None):
-        log_level = os.environ.get(LOG_LEVEL_VDK)
-    elif os.environ.get("VDK_LOG_LEVEL_VDK", None):
-        log_level = os.environ.get("VDK_LOG_LEVEL_VDK")
+    warnings.warn(
```
DeltaMichael, line 85: I don't think we need this until we've released the plugin.

DeltaMichael (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/config/log_config.py
yonitoo: ok, removed it.

PR #2948, opened 2023-11-27
vdk-oracle: support type inference

DeltaMichael:
## Why?

Inferring the type in case a string is passed in the payload instead of the corresponding data type improves the user experience. Type conversion can be relegated to vdk instead of doing pre-processing of large payloads inside data jobs

## What?

Add column data types to column cache
In case a string is passed, infer the python data type based on the Oracle column data type.
In case math.nan is passed, write NULL to the Oracle table

## How was this tested?

Ran functional tests locally
CI/CD

Feature/non-breaking

### DIFF
--- a/projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
+++ b/projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import datetime
 import logging
+import math
 from decimal import Decimal
 from typing import Any
 from typing import Dict
@@ -22,7 +23,7 @@ class IngestToOracle(IIngesterPlugin):
         self.conn: PEP249Connection = connections.open_connection("ORACLE").connect()
         self.cursor: ManagedCursor = self.conn.cursor()
         self.table_cache: Set[str] = set()  # Cache to store existing tables
-        self.column_cache: Dict[str, str] = {}  # New cache for columns
+        self.column_cache: Dict[str, Dict[str, str]] = {}  # New cache for columns
 
     @staticmethod
     def _get_oracle_type(value: Any) -> str:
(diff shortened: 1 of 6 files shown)


projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -94,19 +97,37 @@ def _add_columns(self, table_name: str, payload: List[Dict[str, Any]]) -> None:
...
-        if type(value) is Decimal:
+    def _cast_to_correct_type(self, table: str, column: str, value: Any) -> Any:
+        def cast_string_to_type(t: str, v: str) -> Any:
+            if t == "FLOAT" or t == "DECIMAL":
+                return float(v)
+            if t == "NUMBER":
+                v = v.capitalize()
+                return bool(v) if v in ["True", "False"] else int(v)
+            if "TIMESTAMP" in t:
+                return datetime.datetime.strptime(v, "%Y-%m-%dT%H:%M:%S")
```
DeltaMichael, line 119: Not sure if this should be configurable somehow.

gageorgiev (approved):
looks good, left some comments, make sure to address the codacy warnings, especially the isinstance one

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -94,19 +97,37 @@ def _add_columns(self, table_name: str, payload: List[Dict[str, Any]]) -> None:
...
     # TODO: https://github.com/vmware/versatile-data-kit/issues/2930
-    def _cast_to_correct_type(self, value: Any) -> Any:
-        if type(value) is Decimal:
+    def _cast_to_correct_type(self, table: str, column: str, value: Any) -> Any:
+        def cast_string_to_type(t: str, v: str) -> Any:
+            if t == "FLOAT" or t == "DECIMAL":
+                return float(v)
+            if t == "NUMBER":
+                v = v.capitalize()
+                return bool(v) if v in ["True", "False"] else int(v)
```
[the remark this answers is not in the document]
gageorgiev: gotcha

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
duyguHsnHsn: I was thinking the same but I am not sure whether this would really matter for the users, if not i think it is better to stay like this bc changing it would complicate the code and the usage

DeltaMichael: Maybe we can leave it like this for now. This is the default oracle date time format. I'm more concerned if someone wants to pass in just time or just a date. Oracle supports this, but it gets complicated on our end in this case.

PR #2970, opened 2023-12-12
vdk-oracle: support thick mode by default

DeltaMichael:
## Why?

Internal deployments use thick mode and have the oracle driver bundled in their docker image.

## What?

Add a config option for thick mode and set the default to True

## How was this tested?

Ran code connection against a version 10c database, which requires thick mode
CI

Feature/non-breaking


mivanov1988 (approved):

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_configuration.py
```diff
@@ -74,3 +78,8 @@ def add_definitions(config_builder: ConfigurationBuilder):
             default_value=None,
             description="The password secret key if using secrets to connect to Oracle",
         )
+        config_builder.add(
+            key=ORACLE_THICK_MODE,
+            default_value=True,
+            description="Use oracle thick mode, default is True. Set to False to disable oracle thick mode",
```
gageorgiev, line 84: I'd love a short summary of what thick mode is since it seems to be something specific to oracledb and I had to google it.

gageorgiev (approved):

PR #2978, opened 2023-12-18
control-service: getting all jobs should now be consistent

mrMoZ1:
what: The control-service now uses LinkedHashMap when populating the graphQL filter and search criteria strategies.
Which ensures consistent filtering and sorting.

why: The previous usage of a HashMap would not guarantee any ordering of the key set which would lead to user issues when querying data from the API such as differently sorted lists of data jobs for the same query. This would cause more problems when we include pagination on a larger query - such as vdk list -a - users would see duplicated data, or missing data. Example of the same query with different ordering results can be seen below:

image

This was run on a local instance of the control-service with dummy data.

testing: Added unit test.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/GraphQLDataFetchers.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/GraphQLDataFetchers.java
@@ -182,11 +182,13 @@ public class GraphQLDataFetchers {
                     f -> f,
                     filter ->
                         strategyFactory.findStrategy(
-                            JobFieldStrategyBy.field(filter.getProperty())), (v1, v2) -> {
+                            JobFieldStrategyBy.field(filter.getProperty())),
+                    (v1, v2) -> {
                       throw new IllegalArgumentException(
                           "No duplicate keys allowed in Filter Strategy Map. Duplicate was: "
                               + v1.getStrategyName());
-                    }, LinkedHashMap::new));
+                    },
+                    LinkedHashMap::new));
 
     // compute criteria
     filterStrategyMap.forEach(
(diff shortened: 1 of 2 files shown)


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/GraphQLDataFetchers.java
```diff
@@ -176,7 +182,13 @@ private Criteria<V2DataJob> populateCriteria(List<Filter> filterList) {
                     f -> f,
                     filter ->
                         strategyFactory.findStrategy(
-                            JobFieldStrategyBy.field(filter.getProperty()))));
+                            JobFieldStrategyBy.field(filter.getProperty())),
+                    (v1, v2) -> {
+                      throw new IllegalArgumentException(
```
mivanov1988, line 187: What's the point of this exception? Won't it be thrown if there are duplicates?

mrMoZ1: This is the current default behavior of the toMap function we use. However when supplying a map provider we need to also provide a duplicate handling function:
https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-java.util.function.BinaryOperator-java.util.function.Supplier-

Current behavior:
https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-

Which states in the docs that it throws IllegalStateException on duplicates.
So to summarize the only thing that we are changing is the map type provider to LinkedHashMap provider.

mivanov1988 (approved):

dakodakov (approved):

PR #2986, opened 2024-01-02
vdk-coverity-scan: double the limit for resources

duyguHsnHsn:
What: Double the limit fr resources for coverity scan
Why: Due to recent problems with the cicd https://gitlab.com/vmware-analytics/versatile-data-kit/-/jobs/5841662488


murphp15:
I don't think this is working

antoniivanov (approved):

sabadzhiev:
Not a problem anymore.

PR #3019, opened 2024-01-10
vdk-oracle: separate connection string into host, port, sid

DeltaMichael:
## Why?

Passing the whole connection string is sometimes not convenient

## What?

Add separate config options for host, port and SID Keep the connection string option and make it override the new options

## How was this tested?

CI

Feature/non-breaking


projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_configuration.py
```diff
@@ -57,6 +69,24 @@ def add_definitions(config_builder: ConfigurationBuilder):
...
+        )
+        config_builder.add(
+            key=ORACLE_PORT,
+            default_value="1521",
+            is_sensitive=True,
+            description="The port of the Oracle database. Note: This option is overridden by ORACLE_CONNECTION_STRING",
+        )
+        config_builder.add(
+            key=ORACLE_SID,
+            default_value=None,
```
antoniivanov, line 86: Is there reasonable default value?

PR #3032, opened 2024-01-16
vdk-oracle: add new config options (host,port,sid,thick_mode_lib_dir)

antoniivanov:
We are adding 3 options host,port,sid that were being originally added as part of PR https://github.com/vmware/versatile-data-kit/pull/3019

The issue in the PR is that SID was incorrecctly passed as "service_name" to the underlying oracle db connect method. But it is actuall "SID" (system identifier) .

Setting it correctly there works fine.

I also had to add thick_mode_lib_dir since it was impossible to get thick mode working in Mac without passing it on.

Testing Done:
	- added   test_oracle_connect_execute_without_connection_string
	- tested in production environment against real (non local) with
thick mode database with a data job which executed query and send data
for ingestion successfully.

Currently the automated tests do not cover the thick mode scenario due to extra setup required and this will be added later.


projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_connection.py
```diff
@@ -14,24 +16,52 @@
...
         self._oracle_password = password
+        self._host = host
+        self._port = port
+        self._sid = sid
         self._oracle_connection_string = connection_string
         self._thick_mode = thick_mode
+        self._thick_mode_lib_dir = thick_mode_lib_dir
 
-    def _connect(self) -> PEP249Connection:
+    def _connect(self) -> Connection:
```
DeltaMichael, line 39: The abstract method signature is

```python
    @abstractmethod
    def _connect(self) -> PEP249Connection:
        """
        this method MUST be implemented by inheritors and should return PEP249 Connection object (unmanaged)
        """
        raise NotImplementedError
```

I'm aware of the discrepancy, I'm just curious if we should change the abstract method type hint as well.

DeltaMichael (approved):

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_configuration.py
```diff
@@ -64,22 +72,43 @@ def add_definitions(config_builder: ConfigurationBuilder):
...
+            description="Use oracle thick mode. Set to False to disable oracle thick mode."
+            "If set to true you may need to manually install client library. "
+            "Refer to "
+            "https://python-oracledb.readthedocs.io/en/latest/user_guide/initialization.html#enablingthick "
+            "More info: https://python-oracledb.readthedocs.io/en/latest/user_guide/appendix_b.html",
+        )
+        config_builder.add(
+            key=ORACLE_THICK_MODE_LIB_DIR,
+            default_value=None,
+            description="This is the location of the Oracle Client library used when tick mode is enabled. "
```
yonitoo, line 110: typo: tick -> thick

yonitoo (approved):
LGTM!

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_connection.py
antoniivanov: No, because each database provides its own class without a common abstract class actually though all follow a common specification(protocol) 

Our PEP249Connection is actually meant ot be a protocol as defined in https://peps.python.org/pep-0544/ which his used in cases of duck typing like those.

Changing the subclass is merely for convenience and clarity.

PR #3038, opened 2024-01-17
vdk-core: Short job summary

gageorgiev:
This PR includes a short job summary printed after the main one, which only lists the job status and the status of each individual steps.

Testing done: manual

Example:
```
2024-01-17 19:38:27,382 [VDK] [INFO ]           cli_run.py :152  gg-job4   - Job execution result: SUCCESS
Steps list:
10.py - SUCCESS
20.py - SUCCESS
30.py - SUCCESS
```

Job with error:
```
2024-01-18 11:17:46,090 [VDK] [INFO ]           cli_run.py :152  gg-job4   - Job execution result: ERROR
Steps list:
10.py - SUCCESS
20.py - ERROR
```


antoniivanov:
Can you also print exampe of logs with short summary in case of failure in some steps?

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -141,6 +141,24 @@ def __log_exec_result(self, execution_result: ExecutionResult) -> None:
         else:
             log.info(f"Data Job execution summary: {execution_result}")
 
+        self.__log_short_job_summary(execution_result)
```
DeltaMichael, line 144: This code will log two job summaries if we set LOG_EXECUTION_RESULT=True https://github.com/vmware/versatile-data-kit/blob/dd916d25adce6ce0596016f2d1cb34023936746f/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py#L183C1-L183C1

The call to `__log_short_job_summary` should be in this else clause instead.
https://github.com/vmware/versatile-data-kit/blob/dd916d25adce6ce0596016f2d1cb34023936746f/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py#L185

gageorgiev: The new summary will only be shown when the old one is shown, do we want it to appear instead of the old one?
The else clause here is irrelevant to this, the if-else here is a bugfix for jobs with massive step lists having their summary not print on some OS-es.

gageorgiev: addressed comment

DeltaMichael (approved):

antoniivanov (approved):

yonitoo (approved):

PR #3039, opened 2024-01-19
vdk-core: remove verbose logs

antoniivanov:
Those looks can be extremely spammy if you try to ingest many records (millions) , as it's called for every record.
We can add some kind of stats logging which is done every minute or so to track progress.

This can cause significant slow down of the ingestion and the data job. 

Looking at the logs send by 500 jobs in one deployment of VDK. Logs from `ingester_router` represent more than 80% of hte logs sent.


duyguHsnHsn (approved):

dakodakov:
Hello,

please don't merge this as our service operators as well as our users use these logs as means to debug ingestion data jobs. Offline we discussed potentially adding a "IngestionStatsLogger" which logs ingestion statistics periodically, but before such a functionality is introduced, let's not merge this change.

PR #3045, opened 2024-01-22
vdk-oracle: escape special chars in column names

DeltaMichael:
## Why?

Special characters in column names currently cause errors. They should be supported in case the plugin is used for tables with unorthodox column names

## What?

Support special characters in oracle column names
Refactor column caching logic

Special chars are escaped as advised here
https://docs.oracle.com/en/error-help/db/ora-00904/

## How was this tested?

Ran functional tests locally
CI/CD

Feature/non-breaking

Export identifier normalization to separate function


projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -18,12 +20,90 @@
 log = logging.getLogger(__name__)
 
 
+# Functions for escaping special characters
+def _is_plain_identifier(identifier: str) -> bool:
+    class InvalidOracleIdentifierError(BaseException):
```
antoniivanov, line 25: The custom InvalidOracleIdentifierError is great. Please move it outside the _is_plain_identifier function. It’s more conventional to define custom exceptions at the module level rather than inside a function otherwise importing and catching is more complex.

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -18,12 +20,90 @@
...
+        raise InvalidOracleIdentifierError(
+            f"{identifier} is not a valid Oracle identifier. "
+            f"https://docs.oracle.com/en/error-help/db/ora-00904/"
+        )
+
+    # https://docs.oracle.com/en/error-help/db/ora-00904/
+    # Alphanumeric that doesn't start with a number
+    # Can contain and start with $, # and _
+    regex = "^[A-Za-z\\$#_][0-9A-Za-z\\$#_]*$"
+    return True if re.fullmatch(regex, identifier) else False
```
antoniivanov, line 39: nit: this could be simplified to 

```suggestion
    return bool(re.fullmatch(regex, identifier))
```

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -18,12 +20,90 @@
...
+            self._tables[table.upper()] = {
+                col: ("DECIMAL" if data_type == "NUMBER" and data_scale else data_type)
+                for (col, data_type, data_scale) in result
+            }
+        except Exception as e:
+            # TODO: https://github.com/vmware/versatile-data-kit/issues/2932
+            log.error(
+                "An exception occurred while trying to cache columns. Ignoring for now."
+            )
+            log.exception(e)
```
antoniivanov, line 73: It's redudant to use 2 logs. `log.exception("An exception occurred while trying to cache columns. Ignoring for now.", e)`

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -18,12 +20,90 @@
 log = logging.getLogger(__name__)
 
 
+# Functions for escaping special characters
+def _is_plain_identifier(identifier: str) -> bool:
+    class InvalidOracleIdentifierError(BaseException):
+        pass
+
+    # Raise error identifier it starts with a number
```
antoniivanov, line 28: Can't it start with a number of it's in double quotes??

antoniivanov (approved):
Looks good to me. None major comments. Mostly minor clean code suggestions.

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
DeltaMichael: Theoretically it can't start with a number as stated here https://docs.oracle.com/en/error-help/db/ora-00904/

It worked when it started with a special character, though, so who knows. I'll test quickly and see what happens.

PR #3055, opened 2024-01-26
vdk-structlog: add default logging format values

DeltaMichael:
## Why?

For users who do not wish to configure each structlog option separately, introduce presets for different cases.

## What?

Introduce CLOUD and LOCAL presets.

The CLOUD preset hardcodes the following options

```sh
export VDK_STRUCTLOG_FORMAT=console
export VDK_STRUCTLOG_CONSOLE_LOG_PATTERN="%(asctime)s [VDK] %(vdk_job_name)s [%(levelname)-5.5s] %(name)-30.30s %(filename)20.20s:%(lineno)-4.4s %(funcName)-16.16s[id:%(attempt_id)s]- %(message)s"
```

LOCAL just uses the defaults set in the configuration code

Refactor structlog plugin to eliminate repeating code
Remove syslog config tests
Fix bug with vdk fields crashing logging for custom format
Add tests for vdk fields with custom format

## How was this tested?

Functional tests
CI

Feature/non-breaking


projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -77,6 +77,66 @@ def process(self, msg, kwargs):
...
+    formatter, metadata_filter = create_formatter(
+        logging_formatter, metadata_keys, custom_format_string
+    )
+    return formatter, metadata_filter
+
+
+def _configure_non_root_log_levels(context: CoreContext):
+    if logging.getLogger().getEffectiveLevel() == logging.NOTSET:
+        log_level = os.environ.get(
+            "LOG_LEVEL_VDK", os.environ.get("VDK_LOG_LEVEL_VDK", "INFO")
```
antoniivanov, line 105: Why are we using environment varaible and not `context.configuration`?

projects/control-service/projects/job-base-image/Dockerfile-data-job-base
```diff
@@ -27,6 +27,11 @@ RUN set -ex \
       && ldconfig \
       && apt-get purge -y --auto-remove curl unzip
 
+# Setup default logging with vdk-structlog
```
antoniivanov, line 30: I am uncertain this is the correct place. I think you want to ensure that all jobs deployed by the control service have "Cloud" settings but  this is just one base image of many. So it won't work really. You have to set it in every base image which is not feasible. 
 

I was thinking LOG_CONFIG_TYPE which can act as "aggregator" of logging configuration values 

LOG_CONFIG_TYPE="CLOUD" would set format and all config options that are suitable for cloud in a single place in vdk-struct-log

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: Not sure, I copied this over. I'll change it and test before we merge.

projects/control-service/projects/job-base-image/Dockerfile-data-job-base
DeltaMichael: So I should have something like below?

```python
    @hookimpl
    def vdk_initialize(self, context: CoreContext):
        config_type = context.configuration.get_value("log_config_type")
        if config_type == "CLOUD":
            # hardcode format, etc.
```

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: I think it's because we're overriding a core option. Not really sure. @yonitoo can you please help?

projects/control-service/projects/job-base-image/Dockerfile-data-job-base
antoniivanov: Yeah , something like that.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
yonitoo: While I was transferring the functionality from vdk-core back then, I think there was a case where the context config LOG_LEVEL_VDK value was None (not set) => vdk_log_level = None => the non-root loggers depending on this value were None. Therefore, I added the same vdk-core setting of initial log level by the env vars and everything worked out well. However, if it would work without this, it would be better, I agree.

yonitoo (approved):
Overall, LGTM. Just the bit with the log level env vars has to be checked if it could be improved. Also, at first I had a little concern about passing the whole CoreContext to some of the functions but it makes sense.

antoniivanov:
I just wanted to clarify the intent behind this PR. I don't think the description makes it clear (the why section ). So I will ask to update it, please. 



The way I see it the primary goal of this PR is to offer a mechanism that allows users to easily configure their logging setup, tailored specifically for cloud-based or local runs. So we introduce opinionated configuration options that we believe represent the best practices for cloud and local scenarios. So instead of users figuring how to configure all 5-10 options . They just set config_type="CLOUD" or LOCAL.

Of course these configurations are not rigid. Users retain the flexibility to modify any aspect of the logging configuration, such as the `log_format`, `logging_metadata,` to better suit their specific needs. This approach ensures a balance between providing a solid starting point and maintaining user flexibility in customization

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: This came from core and we don't need it here. It has to do with logging before vdk is initialized and all the plugins loaded. I removed it and move the calls to this method after the root logger is configured. Should be fine.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -77,7 +81,71 @@ def process(self, msg, kwargs):
 """
 
 
+class StructlogConfig:
+    def __init__(self, configuration: Configuration):
+        self._configuration = configuration
+        self._is_cloud = configuration.get_value(STRUCTLOG_CONFIG_PRESET) == "CLOUD"
```
antoniivanov, line 87: preset configuration should have the least priority. 

Users should be able to override it either in config.ini or environment variables. 

Currently, it's the opposite if I understand correctly.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -77,7 +81,71 @@ def process(self, msg, kwargs):
...
+
+    def get_value(self, key):
+        if self._is_cloud:
+            if key == STRUCTLOG_LOGGING_FORMAT_KEY:
+                return "console"
+            if key == STRUCTLOG_CONSOLE_LOG_PATTERN:
+                return DETAILED_LOGGING_FORMAT
```
antoniivanov, line 94: Wouldn't be more convenient to have to have structure like this 

```
presets ={
  "CLOUD": { "logging_format:" : "console", "log_pattern" : DETAILED_LOGGING_FORMAT } , 
  "LOCAL": { "...
}

This is more maintainable , extensible way to define them.

File: projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py [hunk 5]
```diff
@@ -77,7 +81,71 @@ def process(self, msg, kwargs):
 """
 
 
+class StructlogConfig:
```
antoniivanov, line 84: There is a common pattenr for configuration classes in vdk 

See 
- https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/dag_plugin_configuration.py. 
- https://github.com/vmware/versatile-data-kit/blob/eae895ac2b41c735cfefe301b4e2cfeba55932c0/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration.py#L20
- https://github.com/vmware/versatile-data-kit/blob/eae895ac2b41c735cfefe301b4e2cfeba55932c0/projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/trino_config.py#L21
- https://github.com/vmware/versatile-data-kit/blob/eae895ac2b41c735cfefe301b4e2cfeba55932c0/projects/vdk-core/src/vdk/internal/builtin_plugins/notification/notification_configuration.py#L43

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -77,14 +81,152 @@ def process(self, msg, kwargs):
...
+                vdk_config.LOG_LEVEL_MODULE.lower(): configuration.get_value(
+                    vdk_config.LOG_LEVEL_MODULE.lower()
+                ),
+            },
+        }
+
+        self._config = presets[configuration.get_value(STRUCTLOG_CONFIG_PRESET)]
+
+        for key in configuration.list_config_keys():
+            value = configuration.get_value(key)
```
antoniivanov, line 136: Are you sure the capital case would match? You may consider user key.lower if unsure

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -96,246 +238,177 @@ def vdk_configure(self, config_builder: ConfigurationBuilder):
...
+            default_value=None,
             description=(
                 f"Controls the logging output format. Possible values: {STRUCTLOG_LOGGING_FORMAT_POSSIBLE_VALUES}"
             ),
         )
 
         config_builder.add(
             key=SYSLOG_HOST_KEY,
-            default_value=DEFAULT_SYSLOG_HOST,
+            default_value=None,
```
antoniivanov, line 259: It's better to leave the default values here because 

A) They are used to infer the type and convert to correct type (though default type is str) 
B) They show up in `vdk config-help` and in other help and configuration documentation we will generate 

It seems get using Configuration.get_value(key) ([link to code](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/core/config.py#L73)) there's no way to know if value is set or it's the default passed here. 

But we need to solve that without breaking current good practices above. 

I don't have many good ideas. 
```
    def get_value(self, key: ConfigKey, fallback_value: ConfigValue) -> ConfigValue:
                   key = _normalize_config_key(key)
               
        default_value = fallback_value if fallback_value else self.__config_key_to_default_value.get(key)
        value = self.__config_key_to_value.get(key, default_value)
        return value
```
 
 This would allow you do `configuration.get_value(key, presets.getdefault(preset_key, {}).get(key, None))` 
 
 I am not sure if this is the best design. It may work, but it can be a bit complex and may not be the most straightforward approach. 
 
 Alterantively we can introduce is_value_set(key) function? 
 
 (If you do any changes in vdk-core, it must be in separate PR and merged before this one is merged)
 
  What do you think? 
  
 There's need of more thinking here. This is not that straightforward task.

antoniivanov (changes requested):

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: This is ok, because all keys in the configuration get lower-cased for some reason.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: I think this should be a separate ticket.

antoniivanov: If we merge this PR as it is , I think it will leave technical debt as it would introduce some inconsistencies in how configurations are handled (above points A) And B) ) . 

And consistency is pretty crucial for maintainability, usability, and avoiding potential bugs.
First we should agree on that. 

We could merge it if we have a clear plan for revisiting it (aka story/issue) within the near future ( next sprint).

DeltaMichael: Ok, let's do this first instead https://github.com/vmware/versatile-data-kit/pull/3076

I have some capacity left over for this sprint, so it should be fine.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
```diff
@@ -77,7 +81,146 @@ def process(self, msg, kwargs):
 """
 
 
+class StructlogConfig:
```
antoniivanov, line 84: I realy think the structlog_plugin module is very long. 

It is the plugin's entry point (basically similar to "main" method for an application), so it should only bootstrap things and delegate most of the logic to other modules. 

So please let's please StructlogConfig in separate module (aka file). 

I'd refactor StructlogPlugin to be shorter as well. But that's out of scope here.

antoniivanov (approved):
It looks good to me. 

The only thing I'd simplify the structlog_plugin module as I suggested below. 
But I don't feel that strongly to block merging the change over that.

projects/vdk-plugins/vdk-structlog/src/vdk/plugin/structlog/structlog_plugin.py
DeltaMichael: I can do it in a separate PR if that's ok.

PR #3062, opened 2024-01-29
vdk-core: relevant info in step result

DeltaMichael:
## Why?

Sometimes the exact errors in user code are hard to find in the stack trace

## What?

Print relevant information from stack trace in step results

Example:

On error in step 20_python_step.py, the short result summary looks like this

```
2024-01-29 15:14:14,736 [VDK] hello-world [INFO ] vdk.internal.builtin_plugins.r           cli_run.py:166  __log_short_exec[id:3a22ec2e-d6e6-4f94-96c0-122955930cbb-1706534052-abffd]- Job execution result: ERROR
Step results:
10_sql_step.sql - SUCCESS
20_python_step.py - ERROR
  File "/Users/mdilyan/Projects/vdk-playground/hello-world/20_python_step.py", line 27, in run
    raise ValueError("No value in this")
ValueError: No value in this
```

## How was this tested?

Ran locally
Functional test

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import re
 import sys
+import traceback
 from typing import cast
 from typing import Dict
 from typing import List
@@ -18,6 +19,7 @@ from vdk.internal.builtin_plugins.config.job_config import JobConfig
 from vdk.internal.builtin_plugins.run import job_input_error_classifier
 from vdk.internal.builtin_plugins.run.data_job import DataJobFactory
 from vdk.internal.builtin_plugins.run.execution_results import ExecutionResult
+from vdk.internal.builtin_plugins.run.execution_results import StepResult
 from vdk.internal.builtin_plugins.run.execution_tracking import (
     ExecutionTrackingPlugin,
 )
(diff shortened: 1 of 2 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -143,14 +144,41 @@ def __log_exec_result(self, execution_result: ExecutionResult) -> None:
 
     @staticmethod
     def __log_short_exec_result(execution_result):
+        def extract_relevant_lines(step):
```
antoniivanov, line 147: Add type hints. What is the type of step

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -143,14 +144,41 @@ def __log_exec_result(self, execution_result: ExecutionResult) -> None:
 
     @staticmethod
     def __log_short_exec_result(execution_result):
```
antoniivanov, line 146: Add type hints please if you can even though it's not introduced now

 I must have missed to comment on that in previous review.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -143,14 +144,41 @@ def __log_exec_result(self, execution_result: ExecutionResult) -> None:
...
+                for i, line in enumerate(call_list):
+                    # Check if the step name is in the line
+                    if step.name in line:
+                        out.append(line)
+                        j = i + 1
+                        # Pull in subsequent relevant lines
+                        # that do not come from another file
+                        while j < len(call_list) and not re.match(
+                            r'^\s*File "', call_list[j]
+                        ):
```
antoniivanov, line 161: How robust is this logic? Is there exception format specification we can trust and refer to. 

I really don't want to put anything in core that may break in different version of python or in some corner cases stack traces.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -143,14 +144,41 @@ def __log_exec_result(self, execution_result: ExecutionResult) -> None:
...
+        def extract_relevant_lines(step):
+            out = []
+            if step.exception:
+                call_list = traceback.format_tb(step.exception.__traceback__)
+
+                for i, line in enumerate(call_list):
+                    # Check if the step name is in the line
+                    if step.name in line:
+                        out.append(line)
+                        j = i + 1
```
antoniivanov, line 156: Please don't use j and i. 

Use sensible names that say what is j and i. Even Index or line_index, call_index or stack_depth are better.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
```diff
@@ -143,14 +144,41 @@ def __log_exec_result(self, execution_result: ExecutionResult) -> None:
 
     @staticmethod
     def __log_short_exec_result(execution_result):
+        def extract_relevant_lines(step):
+            out = []
+            if step.exception:
+                call_list = traceback.format_tb(step.exception.__traceback__)
```
antoniivanov, line 150: Is it possibel something to fail? is `__traceback__` always available. 

I am asking because If something fails, it's better to still log something (with less detail) than to crash teh program

DeltaMichael: Calling `traceback.format_tb(None)` just returns an empty list. `__traceback__` should always be available, but I think we're safe if it's not.

projects/vdk-core/src/vdk/internal/builtin_plugins/run/cli_run.py
DeltaMichael: I couldn't find any specific docs that describe the traceback format. There are PEPs related to enhancing the traceback, but that's pretty much it.

We could refer to the module code itself. https://github.com/python/cpython/blob/main/Lib/traceback.py#L52
Or to the interpreter code https://github.com/python/cpython/blob/main/Python/traceback.c#L711

They don't seem to change very often, so we should be ok.

duyguHsnHsn (approved):
It looks good to me

antoniivanov (approved):

PR #3063, opened 2024-01-29
vdk-core: Update IngesterRouter to ignore pre-processors

doks5:
Currently, the IntesterRouter applies all available pre-processors, irrespective if the ingested payload actually needs to go through them.

This change introduces a new configuration variable, `INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD`, that allows a user to specify for which ingestion methods should the pre-processors be ignored. The new configuration accepts a comma-separated string of the form "<ingest-method1>, <ingest-method2>". This way, a data job can have the following logic:
```python
job_input.send_object_for_ingestion(payload, destination_table="test_table", method="file")

job_input.send_object_for_ingestion(payload2, destination_table="test_table", method="http")
```
and based on the value in "INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD", the available pre- processors could be applied, for example, to only the payload whose `method` is **http**.

Testing Done: Added test

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
@@ -60,6 +60,18 @@ class IngesterConfigurationPlugin:
             """,
         )
         config_builder.add(
+            key="INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD",
+            default_value=None,
+            description="""A string of coma-separated ingestion methods, for
+            which the payload preprocess sequence is not to be applied.
+            Example:
+                   INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD="file, http"
+            This configuration can be used in cases, where a data job ingests
+            payloads through multiple methods and the preprocessors need to be
+            used only for some payloads.
+            """,
+        )
+        config_builder.add(
             key="INGEST_PAYLOAD_POSTPROCESS_SEQUENCE",
             default_value=None,
             description="""A string of coma-separated ingestion post-
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_router.py
@@ -189,6 +189,26 @@ class IngesterRouter(IIngesterRegistry, IIngester):
                 "INGEST_PAYLOAD_POSTPROCESS_SEQUENCE"
             )
 
+            if self._cfg.get_value(
+                "INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD"
+            ):
+                ignore_seq = [
+                    i.strip()
+                    for i in self._cfg.get_value(
+                        "INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD"
+                    ).split(",")
+                ]
+                if method in ignore_seq:
+                    self._cached_ingesters[method] = IngesterBase(
+                        self._state.get(ExecutionStateStoreKeys.JOB_NAME),
+                        self._state.get(CommonStoreKeys.OP_ID),
+                        ingester_plugin,
+                        IngesterConfiguration(config=self._cfg),
+                        pre_processors=[],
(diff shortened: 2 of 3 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -59,6 +59,18 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
             INGEST_PAYLOAD_PREPROCESS_SEQUENCE and INGEST_METHOD_DEFAULT.
             """,
         )
+        config_builder.add(
+            key="INGEST_IGNORE_PAYLOAD_PREPROCESS_SEQUENCE_FOR_METHOD",
```
antoniivanov, line 63: I am not sure about this?  It seems complicated from user point of view and not sure it is needed. It seems a bit limited also. 

For example ignoring completely preprocesing sequence is incorrect as it would break plugin like [vdk-lineage](https://github.com/vmware/versatile-data-kit/tree/main/projects/vdk-plugins/vdk-lineage#vdk-lineage) which wants to be run for every method 

Let's think through the problem you are trying to solve by explaining it more clearly.
What other options would you consider?

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
```diff
@@ -59,6 +59,18 @@ def vdk_configure(self, config_builder: ConfigurationBuilder) -> None:
...
+            default_value=None,
+            description="""A string of coma-separated ingestion methods, for
```
antoniivanov, line 65: Make sure to update first IngesterConfigurationPlugin doc
https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py#L9

antoniivanov (changes requested):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_configuration_plugin.py
DeltaMichael: I think the problem comes when we have pre-processors that are related to databases. Like preprocessors that change queries and payload entires or pre-processors that call the generic cursor object somehow. I think those preprocessors should be filtered based on the ingestion method. For example, if the ingestion method is set to "oracle", but we have an impala pre-processor, that pre-processor should not run. A pre-processor that just does a get request and logs the result should be completely fine to run every time.

doks5: Pre-processors are meant to execute operations that modify the payload that is being ingested. If there are any operations like lineage or telemetry data collection, these are supposed to be implemented as part of post-processors, not pre-processors.
This change only affects pre-processors, not post-processors.

doks5: @DeltaMichael I agree that the optimal solution would be to filter the pre-processors that are specific only to DB-related ingestion methods. However, I couldn't figure out how could that happen without over-complicating the configuration, like for example:
```
INGEST_IGNORE_PAYLOAD_PREPROCESSORS_FOR_METHOD="oracle--http,impala-preprocess; impala--oracle-preprocess"
```
Also, this would cause conflicts between user-defined configurations in data job configs with configurations at the vdk distribution-level set by system administrators.

antoniivanov: Let me share a few ideas. 

1. Tagging Pre-Processors with ingestion method compatibility 

1.1  Do it Inside the preprocessing sequence you just add an if 
```
def pre_process..(method, ...)
    if method != 'impala':
         break;
```

1.2 Do it more configuration driven 
```
preprocessing_hooks_compat: { "hook1: [ "method1"] , "hoo2: all } 
```

2. Change the configuration structure
```
preprocess_hooks = [all: "hook0", {oracle : hook1, hook2} , {impala : hook3}, { all: hook5}
```

1.1 doesn't require any changes. Add the "if" in the preprocessing hooks. 

Longer term something like 2 may provide the most flexibility.

PR #3065, opened 2024-01-30
examples: add Fetch And Embed Data Job Example

### LINKED ISSUE #2992: Decide on which API should be used for embedding chunks of text
To build a RAG pipeline we need to be able to create embeddings from text chunks. 
The idea of building these pipelines is that they can be used as part of a private AI solution, because of this we can't just use huggingface (public) or openai's apis to generate the embeddings. We need to use a privately hosted model in VMWare to generate these embeddings.  
Even though we are targetting a private AI solution embedding apis should be reasonably interchangeable.

It is important not to reinvent the wheel here and look at what langchain are doing.

**Definition of done**
There should be an example data job that runs and embeds chunks of text

yonitoo:
What:
Add Data Job that fetches public Confluence data and then creates embeddings out of it.

Why:
To have a job that creates embeddings out of Confluence documents.

Closes #2992


examples/fetch-embed-job-example/10_fetch_confluence_space.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def run(job_input: IJobInput):
+    log.info(f"Starting job step {__name__}")
+
+    confluence_url = "https://yoansalambashev.atlassian.net/"
+    # since the Confluence space is public, no need to generate API token
+    token = ""
+    space_key = "RESEARCH"
+    output_csv = "documents.csv"
```
antoniivanov, line 40: I know that it's prototype/poc but some simple good coding practices apply and should be followed in prototypes (building good habits) 

It's better to create config.py or util.py and put common variables and configs there - e.g documnents_csv_file_location. which is reused in later step.

examples/fetch-embed-job-example/20_clean_and_embed_data.py
```diff
@@ -0,0 +1,84 @@
...
+from vdk.api.job_input import IJobInput
+
+nltk.download("stopwords")
+nltk.download("wordnet")
+nltk.download("punkt")
+
+log = logging.getLogger(__name__)
+
+
+def clean_text(text):
```
antoniivanov, line 20: Write a function description mainly why this is needed. 
Not so much "what" (which is clear).

antoniivanov (approved):
looks good.

examples/fetch-embed-job-example/10_fetch_confluence_space.py
```diff
@@ -0,0 +1,49 @@
...
+    with open(filename, mode="w", newline="", encoding="utf-8") as file:
+        writer = csv.writer(file)
+        for doc in documents:
+            writer.writerow([doc.page_content])
+
+
+def run(job_input: IJobInput):
+    log.info(f"Starting job step {__name__}")
+
+    confluence_url = "https://yoansalambashev.atlassian.net/"
```
antoniivanov, line 36: nit: Let's put configuraitons in properties ?

```suggestion
    confluence_url = job_input.get_property("confluence_url", "https://yoansalambashev.atlassian.net/")
```

examples/fetch-embed-job-example/10_fetch_confluence_space.py
yonitoo: addressed

examples/fetch-embed-job-example/20_clean_and_embed_data.py
yonitoo: addressed

examples/fetch-embed-job-example/20_clean_and_embed_data.py
```diff
@@ -0,0 +1,84 @@
...
+
+import nltk
+from nltk.corpus import stopwords
+from nltk.stem import WordNetLemmatizer
+from sentence_transformers import SentenceTransformer
+from vdk.api.job_input import IJobInput
+
+nltk.download("stopwords")
+nltk.download("wordnet")
+nltk.download("punkt")
```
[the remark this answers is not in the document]
yonitoo: addressed

examples/fetch-embed-job-example/10_fetch_confluence_space.py
yonitoo: done

examples/fetch-embed-job-example/requirements.txt
```diff
@@ -0,0 +1,5 @@
+atlassian-python-api
+langchain_community
+lxml
+nltk
+sentence-transformers
```
duyguHsnHsn, line 5: i think you should fix here

examples/fetch-embed-job-example/20_clean_and_embed_data.py
```diff
@@ -0,0 +1,102 @@
...
+def run(job_input: IJobInput):
+    log.info(f"Starting job step {__name__}")
+
+    input_csv = DOCUMENTS_CSV_FILE_LOCATION
+    # output_cleaned_csv = 'documents_cleaned.csv'
+    data_job_dir = pathlib.Path(job_input.get_job_directory())
+    output_embeddings = data_job_dir / EMBEDDINGS_PKL_FILE_LOCATION
+
+    temp_dir = job_input.get_temporary_write_directory()
+    nltk_data_path = temp_dir / "nltk_data"
```
duyguHsnHsn, line 83: maybe here add a comment what those lines do

yonitoo: added a comment

examples/fetch-embed-job-example/requirements.txt
yonitoo: I think the '-' or '_' cases for sentence transformers package are interchangeable.

duyguHsnHsn (approved):

PR #3070, opened 2024-01-31
vdk-examples: example job with confluence reader

duyguHsnHsn:


antoniivanov (approved):

yonitoo (approved):

antoniivanov (approved):

examples/confluence-reader/README.md
```diff
@@ -0,0 +1,52 @@
...
+        "deleted": false
+    },
+    {
+        "metadata": {
+            "title": "Another Page Title",
+            "id": "Another Page ID",
+            "source": "Another Source URL"
+        },
+        "page_content": "Another Page Content Text",
+        "deleted": false
```
antoniivanov, line 40: It's not super critical at this point . But it's better if "deleted" is part of metadata. 

And page_content can be named just "data" 

This is so that this can map seamlessly to [DataSourcePayload](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-data-sources/src/vdk/plugin/data_sources/data_source.py#L47) object later when creating data source plugin.

duyguHsnHsn: @antoniivanov i will sync with yoan for this since afaik he worked with this scheme

PR #3072, opened 2024-02-01
vdk-core: Add 'method' to pre_ingest_process API

doks5:
This change adds a `method` argument to the pre_ingest_process() and post_ingest_process()
APIs, in order to allow plugin developers to add logic specific only to certain ingestion
methods.

Testing Done: CI/CD

### DIFF
--- a/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
+++ b/projects/vdk-core/src/vdk/api/plugin/plugin_input.py
@@ -387,6 +387,7 @@ class IIngesterPlugin:
         target: Optional[str] = None,
         collection_id: Optional[str] = None,
         metadata: Optional[IngestionMetadata] = None,
+        method: Optional[str] = None,
     ) -> Tuple[List[Dict], Optional[IngestionMetadata]]:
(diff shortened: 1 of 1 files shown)


projects/vdk-core/src/vdk/api/plugin/plugin_input.py
```diff
@@ -387,6 +387,7 @@ def pre_ingest_process(
         target: Optional[str] = None,
         collection_id: Optional[str] = None,
         metadata: Optional[IngestionMetadata] = None,
+        method: Optional[str] = None,
```
antoniivanov, line 390: Please add this to the post_ingest_process. 

They should be consistent.

antoniivanov (approved):
Looks good. thanks.

projects/vdk-core/src/vdk/api/plugin/plugin_input.py
doks5: Done. Thanks

dakodakov (approved):
LGTM

PR #3076, opened 2024-02-05
vdk-core: add is_default() function to config

DeltaMichael:
## Why?

There are cases where we want to know we're using the default value for a config option. One such case is configuration presets

https://github.com/vmware/versatile-data-kit/pull/3055

## What?

Add is_default() to the config object.

## How was this tested?

Unit tests

Feature/non-breaking

### DIFF
--- a/examples/confluence-reader/confluence_document.py
+++ b/examples/confluence-reader/confluence_document.py
@@ -1,6 +1,7 @@
 # Copyright 2021-2024 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+
 class ConfluenceDocument:
     def __init__(self, metadata, data, deleted=False):
         """
--- a/projects/vdk-core/src/vdk/internal/core/config.py
+++ b/projects/vdk-core/src/vdk/internal/core/config.py
@@ -124,6 +124,17 @@ class Configuration:
         """
         return self.__config_key_to_sensitive.get(key)
 
+    def is_default(self, key: ConfigKey) -> bool:
+        """
+        Return True if the configuration value for a given key uses the default value.
+        If set_value is called for a specific key, this will return False, even if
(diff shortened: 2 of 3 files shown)


gageorgiev (approved):

examples/confluence-reader/confluence_document.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2021-2024 VMware, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+
```
antoniivanov, line 4: nit: please revert :) I tihnk it's by mistake

projects/vdk-core/src/vdk/internal/core/config.py
```diff
@@ -124,6 +124,19 @@ def is_sensitive(self, key: ConfigKey) -> bool | None:
...
+        Return True if the configuration value associated with a given key
+        is the same as the default value
+
+        :param key: the configuration key (e.g db_host, service_uri, etc.)
+        :return: the value corresponding to the configuration key
+        """
+        key = _normalize_config_key(key)
+        default_value = self.__config_key_to_default_value.get(key)
+        value = self.__config_key_to_value.get(key, default_value)
+        return value == default_value
```
antoniivanov, line 138: That would not work . If the default for some option is is "XXX" and the user also sets it to "XXX". 
That would erroneously return that the we are using the default . And it should not.

antoniivanov (changes requested):

examples/confluence-reader/confluence_document.py
DeltaMichael: Pre-commit added this line for some reason. Maybe it's because I called it from the root of the project dir.

antoniivanov (approved):

Issue #3089, opened 2024-02-07
Refine the text cleaning before embedding the documents in the RAG pipeline


yonitoo:
## Overview
Our [current text cleaning method ](https://github.com/vmware/versatile-data-kit/blob/main/examples/embed-ingest-job-example/20_clean_and_embed_json_data.py#L19) converts the text to lower case, removes punctuation, lemmatizes and removes the stop words from the text. As discussed [HERE](https://github.com/vmware/versatile-data-kit/pull/3085#issuecomment-1929483614), the transformer models (in our case SentenceTransformer) doesn't require such extensive preprocessing, it's even suggested to not do it as this way some context might be lost.

**Suggested solution**
Drop the lemmatization and stop words removal from the cleaning.
Double-check if the lower case conversion isn't done by default by [the transformer model](https://huggingface.co/sentence-transformers/all-mpnet-base-v2) we are using.
The cleaning step is something you would expect to have in a pipeline, so we need to figure out how to handle it properly.
Decide on what text cleaning logic might be relevant and add it.

## Acceptance criteria
Remove the extensive NLP preprocessing (lemmatization and stop words removal).
Add relevant text cleaning logic.

PR #3093, opened 2024-02-07
examples: add chunker job to support configurable chunking

### LINKED ISSUE #3084: Introduce document chunking mechanism prior to embedding within the RAG pipeline
This feature request aims to address the challenge of processing large Confluence documents within the RAG pipeline. Currently, the whole documents are embedded and ingested into the database but their size sometimes exceed the token limit (4096 tokens) imposed by the LLM used for inference. This story aims to introduce a proper document chunking mechanism to overcome this problem.

**Suggested solution**
Develop a chunking logic that can split large Confluence documents into smaller, manageable chunks. Each chunk should be of reasonable token size to comply with the LLM token limit. In order to keep the context, chunk overlap between consecutive chunks should be added. 
The chunk size and the overlap size should be configurable.

## Acceptance criteria
Have a configurable chunking mechanism that splits the texts prior to embedding.
Modify the schema and ingest the chunks properly in the vector db.
Extend [the existing data job](https://github.com/vmware/versatile-data-kit/tree/main/examples/embed-ingest-job-example).

yonitoo:
What: Extend the pgvector-embedder by adding configurable chunking mechanism.

Why: Until now, the whole documents were embedded and ingested into the database but their size sometimes exceed the token limit imposed by the LLM used for inference. This change introduces a configurable document chunking mechanism to overcome this problem.

Testing Done: Ran the pipeline jobs locally.

Closes #3084 

Signed-off by: Yoan Salambashev <yoan.salambashev@broadcom.com>


examples/pgvector-chunk-embedder/20_clean_and_embed_json_data.py
```diff
@@ -0,0 +1,114 @@
...
+    for doc in documents:
+        if "data" in doc:
+            cleaned_text = clean_text(doc["data"])
+            cleaned_documents.append([cleaned_text])
+
+    print(len(cleaned_documents))
+    return cleaned_documents
+
+
+def split_documents_into_chunks(
```
antoniivanov, line 53: I think it would be better to make it a separate job (or step at least). 
You can generate new data file  with data chunked as output. 
This file can be picked off pgvector-embedder

so the pipeline can be 
[confluence-reader] -> [chunker] -> [pgvector-embedder] 

My current understanding is that there is a lot of customization in how data is chunked.
[This article](https://medium.com/@anuragmishra_27746/five-levels-of-chunking-strategies-in-rag-notes-from-gregs-video-7b735895694d) is some outline of some strategies 
And it is more often than not somewhat coupled with the source. For example confluence type of chunking might be utiliza strategy of chunking based on headers (sections) in the page. 

we can have some default chunking strategy that is source agnostic - which is what you are building here as it make sense to have one to get started.

yonitoo: I've changed the structure of the job by extracting the cleaning, chunking and embedding as separate job steps. 
Added a property accounting for the chunking strategy that is easily extensible - by default it is the current fixed size chunking.

examples/pgvector-chunk-embedder/README.md
```diff
@@ -0,0 +1,48 @@
...
+            "deleted": <is the content being deleted in the source>
+        },
+        "data": "Content Text"
+    },
+]
+```
+
+# Chunking the data prior to embedding
+
+The example supports configurable chunking - the CHUNK_SIZE and CHUNK_OVERLAP are configured in config.py.
```
antoniivanov, line 24: Can the same chunking logic be applied if we have another vector store (e.g pinecone, FAISS) . In other words would it work?

If it works, then how can be decoupled?

antoniivanov: How can we vary the chunking? 

Should we consider make it possible to have more structured chunking? For example in HTML maybe you chunk by headers (h1, h2). In Wiki by sections . In Git by functions/ classes .

duyguHsnHsn (approved):
Looks amazing! Good work

examples/chunker/10_clean_data.py
```diff
@@ -0,0 +1,65 @@
...
+import re
+
+from config import CLEANED_DOCUMENTS_JSON_FILE_LOCATION
+from config import DOCUMENTS_JSON_FILE_LOCATION
+from vdk.api.job_input import IJobInput
+
+log = logging.getLogger(__name__)
+
+
+def clean_text(text):
```
murphp15, line 15: why do you think this is needed? I really think this is a bad idea. I left a similar comment on a differnt PR https://github.com/vmware/versatile-data-kit/pull/3117#discussion_r1491023553. 
Please take a look.

yonitoo: Well, as I also came to a similar conclusion last week, we created #3089 to come up with a better approach as the current one is not great - I would say 100% lemmatization and stop-words removal should be removed. About the punctuation and the lower case I'm not sure. I don't think it would necessarily damage the model but we might get rid of them as well.

yonitoo: Done - removed the cleaning part.

examples/chunker/config.ini
```diff
@@ -0,0 +1,12 @@
...
+
+[vdk]
+db_default_type=POSTGRES
+ingest_method_default=POSTGRES
+postgres_dbname=
+postgres_dsn=
+postgres_host=
+postgres_password=
+postgres_user=
+ingester_wait_to_finish_after_every_send=True
```
antoniivanov, line 12: is that needed now?

projects/control-service/projects/versions-of-external-dependencies.gradle
```diff
@@ -31,7 +31,7 @@ project.ext {
             'com.cronutils:cron-utils'                                           : 'com.cronutils:cron-utils:9.2.1',
             'net.lingala.zip4j:zip4j'                                            : 'net.lingala.zip4j:zip4j:2.11.5',
             'net.javacrumbs.shedlock:shedlock-provider-jdbc-template'            : 'net.javacrumbs.shedlock:shedlock-provider-jdbc-template:5.11.0',
-            'net.javacrumbs.shedlock:shedlock-spring'                            : 'net.javacrumbs.shedlock:shedlock-spring:5.11.0',
+            'net.javacrumbs.shedlock:shedlock-spring'                            : 'net.javacrumbs.shedlock:shedlock-spring:5.10.0',
```
antoniivanov, line 34: Do you needd to rebase the branch? How did this show up. 

When updating the branch please use the rebase option (not the merge - in the Github UI the merge is the default one you need to click to select rebase)

antoniivanov (commented):
How did you test it?  

Looks ok. I just want to make sure it's tested properly before approving. All my other comments are minor things about documentation or adding comments.

antoniivanov (approved):
Let's merge it in whatever stable state it could be. 

But a few things to note: 

The current fixed splitting is splitting the text in the middle of a word/sentence which I am not sure how good it would affect the RAG app. We should test end to end. 

It would be really good idea to add the metadata to each chunk (when creating embeddings ). 

We should see to change the ID - as you can have multiple chunks per document so the ID should be composite 

The default CHUNK_SIZE should be probably like 2048 and default overlap 64. As most LLMs have token limit like 4000.

examples/chunker/config.ini
yonitoo: dropped them

examples/chunker/README.md
```diff
@@ -0,0 +1,34 @@
+# Embed And Ingest Data Job Example
+
+The following Versatile Data Kit example allows you to chunk document data (in certain format).
+
+# Expected input format
```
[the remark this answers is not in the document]
yonitoo: done

projects/control-service/projects/versions-of-external-dependencies.gradle
yonitoo: I might have added it mistakenly, fixed now

PR #3098, opened 2024-02-09
examples: RAG Question-answering Web Service

gageorgiev:
This PR introduces a script which  allows for the quick deployment of a local service which can answer questions based on the context provided from a configured vector database.

I've left the table names hardcoded, looking for suggestions as to what the best approach for making them configurable is - putting them in the config.ini file did not seem appropriate.

Testing done: local


support/rag/qa-webservice/private_ai_api.py
```diff
@@ -0,0 +1,109 @@
...
+from nltk.stem import WordNetLemmatizer
+from sentence_transformers import SentenceTransformer
+import configparser
+from pgvector.psycopg2 import register_vector
+import re
+import requests
+import json
+from openai import OpenAI
+
+# TODO: figure out how to make the parts configurable, i.e. embedding model could be configured here
```
antoniivanov, line 20: It's a chat_api or qa_api? Is that correct ? 

So private_ai_api is not a precise name within the context of the PR.

support/rag/qa-webservice/private_ai_api.py
```diff
@@ -0,0 +1,109 @@
...
+# but it would also need to be the same at the document ingestion step so that the similarity search
+# can work
+
+
+def clean_text(text):
```
antoniivanov, line 25: Please, move all the reused logic about chunking and embedding in separate file. So it's clearly delianated. 

This would make it easier to reason about it and think how to re-use it properly (without copy pasting)

support/rag/qa-webservice/private_ai_api.py
```diff
@@ -0,0 +1,109 @@
...
+    setup_nltk(".")
+
+    config = configparser.ConfigParser()
+    config.read("api_config.ini")
+
+    # Embed the question 
+    model = SentenceTransformer("all-mpnet-base-v2")
+    embedding = model.encode(clean_text(question.question), show_progress_bar=True)
+
+    # DB connection
```
antoniivanov, line 77: I once got an advice. If ineed to write a comment to clarify what a section is about . It's. better to encapsulate this in a separate function with proper name and write the comment (if needed) as a function doc.

antoniivanov (approved):
Looks good to me. None of the comments are major ones.

examples/rag-chat-api/reqs.txt
```diff
@@ -0,0 +1,7 @@
+pgvector
```
antoniivanov, line 1: the standard name in python is requirements.txt isn't it? 


Let's use it.

PR #3101, opened 2024-02-11
vdk-server: fix ingress settings

antoniivanov:
In 5369fc623 ingress configuraiton was made more configurable but the defaults were removed and that broke vdk-server

Now installation fails with error like

-Stderr output: Error: INSTALLATION FAILED: release vdk failed, and has been uninstalled due to atomic being set: Ingress.extensions "vdk-pipelines-control-service" is invalid: spec: Invalid value: []networking.IngressRule(nil): either `defaultBackend` or `rules` must be specified

This change also
 - Addressed a TODO about command description 
 - Added missing dependency.

Testing Done: vdk server --install succeeded locally. Separate [PR](https://github.com/vmware/versatile-data-kit/pull/3103) is
working on adding automated tests.


projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
```diff
@@ -51,8 +51,18 @@ class Installer:
     git_server_admin_email = "vdkuser@vmware.com"
     git_server_repository_name = "vdk-git-repo"
 
-    def __init__(self):
+    def __init__(self, server_id: str = ""):
+        self.git_server_image = "gogs/gogs:0.12"
```
DeltaMichael, line 55: Is it a good idea to hardcode this?

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/server_plugin.py
```diff
@@ -37,7 +38,13 @@
     is_flag=True,
     help="Returns whether a local Control Service is currently installed.",
 )
-def server(install, uninstall, status):
+@click.option(
+    "-si",
+    "--server-id",
```
DeltaMichael, line 43: Can you please update the description of the PR to mention what exactly the change is? I think it adds a bunch of config options and adds this flag, but it's not super obvious.

projects/vdk-plugins/vdk-server/setup.py
```diff
@@ -14,7 +14,13 @@
     description="Versatile Data Kit SDK plugin that facilitates the installation of a local Control Service.",
     long_description=pathlib.Path("README.md").read_text(),
     long_description_content_type="text/markdown",
-    install_requires=["vdk-core", "click-spinner", "docker", "kubernetes"],
+    install_requires=[
+        "vdk-core",
+        "click-spinner",
+        "docker",
+        "kubernetes",
+        "vdk-plugin-control-cli",
```
yonitoo, line 22: Why are we adding the vdk-plugin-control-cli now?

yonitoo (approved):
Overall LGTM!

projects/vdk-plugins/vdk-server/setup.py
antoniivanov: Because it was a missed dependency. vdk-server depends on vdk-control-cli for setting default control API URL. 
But the dependency was not added.

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/server_plugin.py
antoniivanov: Removed that. Thanks for pointing it out. It was not supposed to be in this change. It's part of my experiments for getting tests to work

projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
antoniivanov: It was hard-coded on line 247. I simply moved it here. My thought was to make it configurable. But that would be in separate change.

PR #3109, opened 2024-02-14
control-service: fix data job deployment cpu conversion

mivanov1988:
Why
Currently, we can configure data job CPU only in milicores.

What
Added an ability to provide cpu resources in either milicores or cores. Also, switched the database CPU unit to cores as it used to be.

Testing done:
Unit tests


dakodakov (approved):
LGTM

gageorgiev (approved):

projects/control-service/projects/pipelines_control_service/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
antoniivanov, line 1: Why are the gradle jar and properties in source controL? I think those are dynamically downloaded on demand when needed.

PR #3114, opened 2024-02-15
SC-3163 : Added a popup code to remove a sensitive key from the prope…

vivekdf:
Made changes to remove sensitive information getting appended to the log file, during properties write operation.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
@@ -78,8 +78,8 @@ class DataJobsServiceProperties(IProperties):
             )  # keeps the outer scope originally-referenced dictionary intact
             for client in self._write_preprocessors:
                 try:
-                    if (properties.get('cipher_key')) is not None:
-                        properties.pop('cipher_key')
+                    if (properties.get("cipher_key")) is not None:
+                        properties.pop("cipher_key")
                     properties = client.write_properties(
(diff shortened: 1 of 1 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
```diff
@@ -78,6 +78,8 @@ def set_all_properties(self, properties: Dict[str, PropertyValue]) -> None:
             )  # keeps the outer scope originally-referenced dictionary intact
             for client in self._write_preprocessors:
                 try:
+                    if (properties.get("cipher_key")) is not None:
```
DeltaMichael, line 81: This prevents the key from getting passed to the backend service. It won't get persisted at all. I don't think this is what we want here.

DeltaMichael (changes requested):
Please provide more information about the exact problem you're trying to solve. Your commit message and PR description should follow this template. https://github.com/vmware/versatile-data-kit/blob/main/support/git-commit-template.txt

Please check our contributing guide for more information. https://github.com/vmware/versatile-data-kit/blob/main/CONTRIBUTING.md#submit-a-pull-request

projects/vdk-core/src/vdk/internal/builtin_plugins/job_properties/datajobs_service_properties.py
dakodakov: Agreed.

DeltaMichael:
Closing due to inactivity.

PR #3117, opened 2024-02-15
examples: adopt RAG examples for remote execution

antoniivanov:
To be able to run the RAG DAG in a deployment we need non-local file local storage. 
The POC was build to pass data between jobs using a local file. Since I want to deploy the jobs I need a way to pass data between them without that since they do not share a file system. Postgres based storage was created for that. 

So moved created one and adopt it. 
It's currently copied in both jobs.  I will refactor it away after this PR. 

I also ended up removing NLKT from everywhere. And also few doc fixes


murphp15:
can you elaborate on what non-local file local storage means ?

examples/confluence-reader/common/database_storage.py
```diff
@@ -0,0 +1,90 @@
...
+            return result.rowcount > 0
+
+    @staticmethod
+    def _serialize_content(content: Union[str, bytes, Any]) -> tuple[bytes, str]:
+        if isinstance(content, bytes):
+            return content, "bytes"
+        elif isinstance(content, str):
+            return content.encode(), "string"
+        else:
+            # Fallback to pickle for other types
```
murphp15, line 81: I'd probably prefer json instead of pickle. but just personal opinion.

examples/confluence-reader/fetch_confluence_space.py
```diff
@@ -189,3 +192,6 @@ def run(job_input: IJobInput):
         data_file,
         confluence_reader.fetch_all_pages_in_confluence_space(parent_page_id),
     )
+
+    storage = DatabaseStorage(get_value(job_input, "storage_connection_string"))
```
murphp15, line 196: you are saving all information to one row in the database?

examples/confluence-reader/fetch_confluence_space.py
murphp15, line 1: I am confused by this class.
you are saving to a file first and then you are writing it to the database? 

Why not write directly to the database?

murphp15:
can you remove the duplicated classes? 
I think it is fine to say that a folder needs to be put on the python path for it to work.

examples/confluence-reader/common/database_storage.py
```diff
@@ -0,0 +1,90 @@
...
+from sqlalchemy import String
+from sqlalchemy import Table
+from sqlalchemy.exc import IntegrityError
+
+
+class DatabaseStorage(IStorage):
+    def __init__(self, connection_string: str):
+        self.engine = create_engine(connection_string)
+        self.metadata = MetaData()
+        self.table = Table(
```
murphp15, line 24: Would it be a bit cleaner if you created a class which represents content and used the orm stuff more directly.

for example on line 35 you would create instances of content instead of using the insert.

examples/confluence-reader/fetch_confluence_space.py
antoniivanov: Yeah. Because before it was saved to a file and I just replace the line to save to a file , to save it in DB. 
 
I supposed I was too lazy to write a for loop

examples/confluence-reader/fetch_confluence_space.py
antoniivanov: That's becasue I reused the previous implementation by just adding the lines about saving to db. 
and making no other changes. I just want it to get it to run in cloud.

murphp15: maybe just add a TODO comment explaining this.

examples/confluence-reader/common/database_storage.py
antoniivanov: Mee too btw. I don't remember now why I needed pickle. 

I will fix a lot of things around later. I just want to make sure I have a version that works end to end well.

examples/pgvector-embedder/20_clean_and_embed_json_data.py
```diff
@@ -4,10 +4,8 @@
 import logging
 import re
 
-import nltk
+from common.database_storage import DatabaseStorage
```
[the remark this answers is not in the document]
yonitoo: I answered in [this comment](https://github.com/vmware/versatile-data-kit/pull/3093#discussion_r1491141057)

antoniivanov:
> can you elaborate on what non-local file local storage means ?

Updated description of PR

murphp15 (approved):

examples/pgvector-embedder/20_clean_and_embed_json_data.py
```diff
@@ -18,29 +16,20 @@ def clean_text(text):
...
-    (focusing on relevant words), and lemmatization (reducing words to their base form).
-    Such preprocessing is crucial for effective NLP analysis.
+    converting to lowercase (uniformity),
 
     :param text: A string containing the text to be processed.
     :return: The processed text as a string.
     """
     text = text.lower()
     # remove punctuation and special characters
     text = re.sub(r"[^\w\s]", "", text)
```
duyguHsnHsn, line 26: Weren't we going to remove this part as well?

examples/rag-chat-api/clean_embed_text.py
```diff
@@ -19,32 +19,9 @@ def clean_text(text):
     text = text.lower()
     # remove punctuation and special characters
```
duyguHsnHsn, line 20: same as above question

duyguHsnHsn (approved):

PR #3118, opened 2024-02-15
vdk-examples: llama data job

duyguHsnHsn:
A job that uses LLamaIndex to fetch data from confluence and to create embeddings to local postgres database with pgvector

Further reading: https://github.com/vmware/versatile-data-kit/wiki/VDK-and-LLamaIndex


murphp15:
I am confused here. I see another example using the langchain confluence reader and also we have a confluence reader plugin. 

why are we reading from confluence in 3 different ways?

PR #3121, opened 2024-02-15
vdk-postgres: batch inserts during ingestion

antoniivanov:
Insert data is painfully slow even small amounts making even testing and pocs hard. So this is a quick and slight optimization Before insert 300 rows took 15 minutes. Now it's about 1 minute

Testing Done: made sure existing test pass. Tested with a job sending 300 rows in 2 tables.

### DIFF
--- a/projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
+++ b/projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
@@ -61,7 +61,7 @@ class IngestToPostgres(IIngesterPlugin):
     ) -> (str, list):
         """
         Prepare the SQL query and parameters for bulk insertion.
-        
+
         Returns insert into destination table tuple of query and parameters;
         E.g. for a table dest_table with columns val1, val2 and payload size 2, this method will return:
         'INSERT INTO dest_table (val1, val2) VALUES (%s, %s)',
--- a/projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
+++ b/projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
@@ -61,6 +61,11 @@ class IngestToPostgres(IIngesterPlugin):
     ) -> (str, list):
         """
         Prepare the SQL query and parameters for bulk insertion.
+        
+        Returns insert into destination table tuple of query and parameters;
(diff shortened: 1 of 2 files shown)


projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
```diff
@@ -57,28 +58,19 @@ def ingest_payload(
...
-        'INSERT INTO dest_table (val1, val2) VALUES (%s, %s), (%s, %s), (%s, %s)', ['val1', 'val2']
-
-        :param destination_table: str
-            the name of the destination table
-        :param cursor: PEP249Cursor
-            the database cursor
-        :param payload: List[dict]
-            the payloads to be ingested
-        :return: Tuple[str, Tuple[str]]
-            tuple containing the query and parameters
```
gageorgiev, line 73: why delete the docstring?

gageorgiev (approved):
would love to see a before/after on ingesting the same confluence pages we discussed, so I have an idea as to how much of an improvement this is with respect to our POC

murphp15 (approved):

projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
```diff
@@ -57,28 +58,19 @@ def ingest_payload(
...
+        Prepare the SQL query and parameters for bulk insertion.
```
antoniivanov, line 63:
```suggestion
        Prepare the SQL query and parameters for bulk insertion.
        
        Returns insert into destination table tuple of query and parameters;
        E.g. for a table dest_table with columns val1, val2 and payload size 2, this method will return:
        'INSERT INTO dest_table (val1, val2) VALUES (%s, %s)',
        [('val1', 'val2'), ('val1', 'val2')]
```

projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py
antoniivanov: The param comments are redundant.

 But the description was good. I returned it (after adjusting it to the new interface).

antoniivanov:
> would love to see a before/after on ingesting the same confluence pages we discussed, so I have an idea as to how much of an improvement this is with respect to our POC


@gageorgiev . It's in the description :
> Before insert 300 rows took 15 minutes. Now it's about 1 minute

PR #3125, opened 2024-02-16
vdk-core: support overriding configs with secrets

DeltaMichael:
## Why?

VDK doesn't provide a way to set sensitive configuration like passwords, such as trino_password. The only way to currently do this is by adding config keys and fetching the values from secrets.

## What?

Add a plugin that reconfigures the Configuration object in CoreContext based on secrets. Do this in the initialize_job hook. In this setup, secrets override options set by regular configs. For example if you set trino_password to "password" in config.ini, but also have a secret called trino_passowrd="another password", the value of trino_password will be "another_passowrd".

**Note:** The Configuration class and the CoreContext class are annotated with `@dataclass(frozen=True)`. This enforces encapsulation, so in order to mutate the Configuration object after it's created, we have to add more public methods to the Configuration class.

https://docs.python.org/3/library/dataclasses.html#frozen-instances

## How was this tested?

Functional test
CI/CD

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/config/secrets_config.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/config/secrets_config.py
@@ -1,0 +1,14 @@
+# Copyright 2023-2024 Broadcom
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2021-2024 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins.run.job_context import JobContext
+
+
+class SecretsConfigPlugin:
+    @hookimpl(trylast=True)
+    def initialize_job(self, context: JobContext):
+        secrets = context.job_input.get_all_secrets()
+        for key, value in secrets.items():
+            context.core_context.configuration.override_value(key, value)
--- a/projects/vdk-core/src/vdk/internal/core/config.py
+++ b/projects/vdk-core/src/vdk/internal/core/config.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import logging
-from collections import OrderedDict
 from dataclasses import dataclass
 from dataclasses import field
 from typing import Any
@@ -69,6 +68,11 @@ class Configuration:
     def __getitem__(self, key: ConfigKey):
         key = _normalize_config_key(key)
         return self.get_value(key)
+
+    def override_value(self, key: ConfigKey, value: ConfigValue):
+        key = _normalize_config_key(key)
+        if key in self.__config_key_to_default_value:
+            self.__config_key_to_value[key] = value
 
     def get_value(self, key: ConfigKey) -> ConfigValue:
         """
--- a/projects/vdk-plugins/vdk-plugin-control-cli/tests/test_secrets_plugin.py
+++ b/projects/vdk-plugins/vdk-plugin-control-cli/tests/test_secrets_plugin.py
@@ -61,14 +61,15 @@ def test_secrets_plugin_no_url_configured():
     ):
         runner = CliEntryBasedTestRunner(vdk_plugin_control_cli, secrets_plugin)
 
-        result: Result = runner.invoke(
-            ["run", jobs_path_from_caller_directory("simple-job")]
-        )
-        # job that do not use secrets should succeed
-        cli_assert_equal(0, result)
+        # TODO: Uncomment this after vdk-core release
+        # result: Result = runner.invoke(
+        #     ["run", jobs_path_from_caller_directory("simple-job")]
+        # )
+        # # job that doesn't explicitly use secrets should fail due to secrets overrides in core
+        # cli_assert_equal(1, result)
 
         result: Result = runner.invoke(
             ["run", jobs_path_from_caller_directory("secrets-job")]
         )
-        # but jobs that do use secrets should fail.
+        # job that explicitly uses secrets should also fail
         cli_assert_equal(1, result)
--- a/projects/vdk-core/README.md
+++ b/projects/vdk-core/README.md
@@ -35,6 +35,46 @@ vdk --help
 Plugins are a powerful way of extending or adapting Versatile Data Kit for all kinds of use-cases.
 For more information see [Plugins doc](../vdk-plugins/README.md).
 
+### A note on configs and secrets
+
+[Secrets also double as config keys.](https://github.com/vmware/versatile-data-kit/pull/3125). At the end of the
+`initialize_job` hook, the secrets provider is called and all secrets keys are fetched. Any configs that match a
+secrets key have their values overridden. However, if a plugin caches some config values in the `initialize_job`
+hook and re-uses them later, those values will not be overridden by secrets.
+
+**Example**
+
+This will not work
+
+```shell
+vdk secrets --set-prompt mytoken
+```
+
+```python
+    @hookimpl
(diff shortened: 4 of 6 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/config/secrets_config.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021-2024 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins.run.job_context import JobContext
+
+
+class SecretsConfigPlugin:
+    @hookimpl(trylast=True)
+    def initialize_job(self, context: JobContext):
```
DeltaMichael, line 9: This is a bit of a hack, not sure if it's a good idea. There was no way to mutate the configuration in an earlier hook, because it gets frozen once it gets built and there is no access to secrets, because the haven't been initialized yet. Also, importing JobContext in `vdk_config.py` causes a circular.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/secrets_config.py
```diff
@@ -0,0 +1,13 @@
...
+        secrets = context.job_input.get_all_secrets()
+        for key, value in secrets.items():
+            context.core_context.configuration.override_with_secret(key, value)
+        context.core_context.configuration.lock_overrides()
```
DeltaMichael, line 13: Also not sure about this, there's potential for abuse.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/secrets_config.py
```diff
@@ -0,0 +1,13 @@
...
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins.run.job_context import JobContext
+
+
+class SecretsConfigPlugin:
+    @hookimpl(trylast=True)
+    def initialize_job(self, context: JobContext):
+        secrets = context.job_input.get_all_secrets()
+        for key, value in secrets.items():
+            context.core_context.configuration.override_with_secret(key, value)
```
DeltaMichael, line 12: Not sure about this either, there's potential for abuse.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/secrets_config.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021-2024 VMware, Inc.
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.plugin.hook_markers import hookimpl
+from vdk.internal.builtin_plugins.run.job_context import JobContext
+
+
+class SecretsConfigPlugin:
+    @hookimpl(hookwrapper=True)
+    def initialize_job(self, context: JobContext):
+        yield  # make sure secrets factory methods are added
```
murphp15, line 10: how does this work?

projects/vdk-core/src/vdk/internal/core/config.py
```diff
@@ -188,6 +203,7 @@ def add(
         """
         key = _normalize_config_key(key)
         self.__config_key_to_default_value[key] = default_value
+        self.__config_key_to_can_override[key] = True
```
murphp15, line 206: what is the rule? new config properties can be overridden? 
Can you give examples of properties that can be overridden and ones that can't

DeltaMichael: You can override any config key-value pair with a secret, as long as you have a secret with the same key.

murphp15: ok if you can override anyone then why not remove this? and we can just override everyone?

antoniivanov:
I was wondering if we need to finalize the configuration at vdk_configure.
Now below proposal is bigger and it may require a bit broader discussion (and is more costly so might not be worth at this time). 
So I am ok if we ship the current implementation as it is.


But this is not the first time we've had to create workarounds due to the limitation that configuration options can only be set during that phase. In the [JobConfigIniPlugin](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/internal/builtin_plugins/config/vdk_config.py#L217), we check if the command is run and search for the config.ini file.

Perhaps we could allow for more dynamic configuration?

As it is currently, vdk_configure is used to define the available configuration options, but the actual values can be provided or overridden later by dynamically added providers.

```
@hookimpl
def vdk_configure(context): 
    context.add("team", description, default, ...)   
```

Configuration providers can be dynamically added at any point in the application's lifecycle.

```
@hookimpl
def run_job(context): 
    context.configuration.add_configuration_provider(
        name="secrets", 
        get_function=lambda context, key: context.secrets.get_secret(key), 
        priority_after="config_ini"  # The default priority is LIFO.
    )
```

projects/vdk-core/src/vdk/internal/core/config.py
```diff
@@ -65,11 +64,25 @@ class Configuration:
...
+    __config_key_to_can_override: dict[ConfigKey, bool] = field(default_factory=dict)
 
     def __getitem__(self, key: ConfigKey):
         key = _normalize_config_key(key)
         return self.get_value(key)
 
+    def __can_override(self, key: ConfigKey):
+        return self.__config_key_to_can_override.get(key, False)
+
+    def override_with_secret(self, key: ConfigKey, value: ConfigValue):
```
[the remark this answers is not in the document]
DeltaMichael:
```python
newbuilder = ConfigurationBuilder(current_configuration)
newbuilder.set_value ...
context.config = newbuilder.get()   
```

I tried this and it doesn't work, because the context class is also a frozen dataclass. I went with your suggestion to rename the method. Also removed the locking. Seems a bit pointless.

murphp15 (approved):

DeltaMichael:
Follow-up https://github.com/vmware/versatile-data-kit/issues/3156

projects/vdk-plugins/vdk-plugin-control-cli/tests/test_secrets_plugin.py
```diff
@@ -61,14 +61,15 @@ def test_secrets_plugin_no_url_configured():
...
-        result: Result = runner.invoke(
-            ["run", jobs_path_from_caller_directory("simple-job")]
-        )
-        # job that do not use secrets should succeed
-        cli_assert_equal(0, result)
+        # TODO: Uncomment this after vdk-core release
+        # result: Result = runner.invoke(
+        #     ["run", jobs_path_from_caller_directory("simple-job")]
+        # )
+        # # job that doesn't explicitly use secrets should fail due to secrets overrides in core
```
antoniivanov, line 68: I think job that do not use secrets should succeed still. 

Simply it won't fetch any configuration from secrets.

DeltaMichael: The test checks what happens when you set `VDK_SECRETS_DEFAULT_TYPE = control-service` but do not provide a url. If you don't set the secrets default type, the job will succeed. The reason it fails is that we no call get_all_secrets() on each job run, so if secrets are misconfigured somehow, it will fail.

antoniivanov (approved):

DeltaMichael:
Follow-ups

https://github.com/vmware/versatile-data-kit/issues/3156
https://github.com/vmware/versatile-data-kit/issues/3210

PR #3143, opened 2024-02-21
versatile-data-kit: Support for Py3.12

gageorgiev:
This PR adds a CICD job checking that components and plugins can be built and tested in a Py3.12 environment.

Testing done: cicd

### DIFF
--- a/projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
+++ b/projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
@@ -35,6 +35,10 @@ build-py311-quickstart-vdk:
   extends: .build-quickstart-vdk
   image: "python:3.11"
 
+build-py312-quickstart-vdk:
+  extends: .build-quickstart-vdk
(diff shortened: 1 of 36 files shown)


projects/vdk-plugins/quickstart-vdk/.plugin-ci.yml
```diff
@@ -35,6 +35,10 @@ build-py311-quickstart-vdk:
   extends: .build-quickstart-vdk
   image: "python:3.11"
```
antoniivanov, line 36: I think we should remove 3.11 tests? We want to run tests with only the first and latest support python version

antoniivanov (approved):
Looks good to me. For those plugins that have 3.12 tests we do not need 3.11 tests. 

I think only vdk-core is tested with every version (as it is the core component :) )

yonitoo (approved):
LGTM! Only the removal of 3.11 tests remains, maybe it could be done as part of another PR as well.

Issue #3142, opened 2024-02-21
Build a data job with larger python dependencies like sentence-transformers


murphp15:
**What is the feature request? What problem does it solve?**
Build a data job with a large python dependency. Make sure you document all the config changes needed for it to be done. So for the demo we can build the job with the required dependencies. 

Here is changes I previously made on a branch to allow data jobs with big dependencies to be built: https://github.com/vmware/versatile-data-kit/compare/person/murphp15/build_jobs_with_big_dependencies?expand=1

Any changes required here should be merged to master for this ticket to be considered closed.

antoniivanov:
Btw if this  https://github.com/vmware/versatile-data-kit/pull/3151 doesn't really work well. 

You can cheat and install the sentence-transformers as a step in the job . 
I tried it out and it worked. 


```
import logging
import subprocess
import sys

from vdk.api.job_input import IJobInput

log = logging.getLogger(__name__)


def run(job_input: IJobInput):
    install_using_subprocess(job_input)
    pass


def install_using_subprocess(job_input):
    try:
        target_directory = job_input.get_temporary_write_directory() / "lib"
        subprocess.check_call(
            [sys.executable, '-m', 'pip', 'install', "--no-cache-dir", "sentence-transformers", '--target',
             str(target_directory)])

        if target_directory not in sys.path:
            sys.path.append(str(target_directory))
        log.info("Installed sentence-transformers")
    except Exception as e:
        log.exception("Failed to install library", e)
```

PR #3151, opened 2024-02-21
control-service: optimize job builder pip install

antoniivanov:
Pip caching is never useful for Docker builds, and it makes the image bigger and cause the process to consume more memory which can lead to OOM errors.


murphp15:
Brilliant! 
Does that make this ticket (https://github.com/vmware/versatile-data-kit/issues/3142) redundant ?

murphp15 (approved):

mivanov1988 (approved):

antoniivanov:
> Brilliant! Does that make this ticket (#3142) redundant ?

I haven't verified it is enough so no for now. But it should help

PR #3152, opened 2024-02-21
vdk-dag: improve error handling and error messages

antoniivanov:
To make it easier to debug VDK DAG jobs some improvements in error handling are done.
- Stop using deprecated UserCodeError in favour of domain oriented exceptoin class
- Improve the formatting of the error message in case of failure.
- Renamed DAG_LOCAL_RUN_JOB_PATH to DAGS_LOCAL_RUN_JOB_PATH to bring it inline with other configuration options [like here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/dag_plugin_configuration.py#L6). Eventually we should expose it the same way but not in this PR.

Testing Done: existing tests for regression.
Here is a comparison before and now in the formatting: https://gist.githubusercontent.com/antoniivanov/8f87cdef1c610fcd24361197bcba45dc/raw/c31b990de2c0b1f49ac5989419cfc6ab12aac934/gistfile1.txt

### DIFF
--- a/projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/local_executor.py
+++ b/projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/local_executor.py
@@ -135,7 +135,8 @@ class LocalDataJob:
             return JobStatus.SUCCEEDED.value
 
     def _update_message_with_summary(self, content: str):
-        self._message = {"summary": json.loads(content), "logs": self._log_file}
+        self._message = json.loads(content)
+        self._message["logs"] = self._log_file
 
 
 class LocalDataJobRunException(Exception):
@@ -158,9 +159,10 @@ class LocalDataJobExecutor(IDataJobExecutor):
         candidates = [
             os.getcwd(),
         ]
+        # TODO: expose this using the plugin configuration mechanisms (which infers also from env. vars among others)
         candidates += [
             part
-            for part in os.environ.get("DAG_LOCAL_RUN_JOB_PATH", "").split(":")
+            for part in os.environ.get("DAGS_LOCAL_RUN_JOB_PATH", "").split(":")
             if part
         ]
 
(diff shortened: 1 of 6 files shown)


projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/local_executor.py
```diff
@@ -158,9 +159,10 @@ def __find_job_path(job_name: str):
         candidates = [
             os.getcwd(),
         ]
+        # TODO: expose this using the plugin configuration mechanisms (which infers also from env. vars among others)
         candidates += [
             part
-            for part in os.environ.get("DAG_LOCAL_RUN_JOB_PATH", "").split(":")
+            for part in os.environ.get("DAGS_LOCAL_RUN_JOB_PATH", "").split(":")
```
gageorgiev, line 165: was this a bug previously or was the var name changed here accidentally; I don't see any other corrections to the var name elsewhere

gageorgiev (approved):

projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/local_executor.py
antoniivanov: Renamed DAG_LOCAL_RUN_JOB_PATH to DAGS_LOCAL_RUN_JOB_PATH to bring it inline with other configuration options [like here](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-dag/src/vdk/plugin/dag/dag_plugin_configuration.py#L6). Eventually we should expose it the same way but not in this PR.

duyguHsnHsn (approved):
The code looks much cleaner now! Nice

Issue #3030, opened 2024-01-16
vdk secrets --set silently fails to set secret


antoniivanov:
**Describe the bug**
When attempting to set a secret using the `vdk secrets --set-prompt ` command, the operation incorrectly reports success even when it fails due to authentication issues. This behavior is misleading and can lead to confusion about the actual state of the secret.

**Steps To Reproduce**
1. Execute the command vdk -v debug secrets --set-prompt "xxx_password" -n aivanov-xxx -t team.
2. The command does not update the secret due to authentication issues (token expired and cannot be refreshed) but still succeeds.

**Expected behavior**
The command should fail and return an error message indicating the failure due to authentication issues.

**Log Snippet**
Command: `vdk -v debug secrets --set-prompt "xxx_password"  -n aivanov-xxx -t supercollider`
> ...
Passed parameters to function <function secrets_command at 0x123b80f40>: (), {'set_prompt': ('xxx_password',), 'name': 'aivanov-xxx', 'team': 'supercollider', 'set': (), 'delete': (), 'delete_all_job_secrets': False, 'overwrite_all_job_secrets': None, 'get': None, 'list': False, 'rest_api_url': 'https://dp.supercollider.vmware.com', 'output': 'TEXT'}
secrets passed options: name: aivanov-xxx, team: supercollider, set: (), set_prompt: ('xxx_password',), get: None, list: False, delete: () rest_api_url: https://dp.supercollider.vmware.com, output: TEXT
xxx_password: 
Rest API URL: https://dp.supercollider.vmware.com
Reading vdk credential file: /Users/aivanov/.sc.vdk/.vdk.internal/vdk-cred.json ...
Acquire access token (it's either expired or missing) ...
Using auth type credentials to acquire access token
Save vdk credential file: /Users/aivanov/.sc.vdk/.vdk.internal/vdk-cred.json 
Save vdk credential file: /Users/aivanov/.sc.vdk/.vdk.internal/vdk-cred.json 
Refresh API token against https://console.cloud.vmware.com/csp/gateway/am/api/auth/api-tokens/authorize 
Adding auto refresh key word arguments {}.
Prepared refresh token request body grant_type=refresh_token&refresh_token=xxx-xxx
Requesting url https://console.cloud.vmware.com/csp/gateway/am/api/auth/api-tokens/authorize using method POST.
[...]

PR #3154, opened 2024-02-22
vdk-control-cli: Added a generic exception to address token expiry and re-login

vivekdf:
Fix for: https://github.com/vmware/versatile-data-kit/issues/3030

Added check for handing exception and logging in information, in scenarios where token is expired or corrupted.


projects/vdk-control-cli/src/vdk/internal/control/rest_lib/rest_client_errors.py
```diff
@@ -87,5 +87,18 @@ def decorated(*args, **kwargs):
                     "and points to the correct host.",
                 )
                 raise vdk_ex from ex
+            except Exception as ex:
+                log.debug(
+                    f"An gemeral Exception occurred in {fn.__module__}.{fn.__name__}",
```
DeltaMichael, line 92: Nit: Grammar

```python
f"A general Exception occurred in {fn.__module__}.{fn.__name__}"```

File: projects/vdk-control-cli/src/vdk/internal/control/rest_lib/rest_client_errors.py [hunk 2]
```diff
@@ -87,5 +87,18 @@ def decorated(*args, **kwargs):
...
+                    f"The Exception class was: {ex}",
+                )
+
+                vdk_ex = VDKException(
+                    what=self.__what,
+                    why=str(ex),
+                    consequence=self.__consequences,
+                    countermeasure="Authentication Error and needs to be logged in again",
```
DeltaMichael, line 100: This message is not very clear, please reword it.

projects/vdk-control-cli/src/vdk/internal/control/rest_lib/rest_client_errors.py
```diff
@@ -87,5 +87,18 @@ def decorated(*args, **kwargs):
                     "and points to the correct host.",
                 )
                 raise vdk_ex from ex
+            except Exception as ex:
+                log.debug(
+                    f"An gemeral Exception occurred in {fn.__module__}.{fn.__name__}",
+                    f"The Exception class was: {ex}",
+                )
+
+                vdk_ex = VDKException(
```
antoniivanov, line 96: We are only wrapping the original exception. Is there any benefit to do that? 

We want to avoid wrapping exceptions unless needed because that leads to a bigger logs (since now we have the original and he previous expcetion), user cannot handle it as easily (catch it). 

Also assuming all exceptions are authentication-related is certainly not correct. There might be other exceptions - bugs or the command may fail for other reason. Also they may already be hanled and formatted appropriately 

This decorator was created to catch rest API-specific exceptions and make the errors more user friendly to the user.

antoniivanov: I think this might work though. 

The reason no error is printed in those cases is indeed because the exception passes through  `except Exception as ex:` 

And click (the library we use) would print only errors from exceptions that inherit click.UsageError
And the authentication  exception is [VDKAuthException](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_exception.py#L5)
which does not inherit click.UsageError so click prints nothing. 

Since you are wrapping it in VDKException, this would work since VDKException inherits click.UsageError 

But if the original exception  [VDKAuthException](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-control-api-auth/src/vdk/plugin/control_api_auth/auth_exception.py#L5) already contains what, why, consequences, countermeasures, and putting them again here would be ugly . 

It would be better to pass the original exception message to VDKException (we may need a new constructor  directly) without changing it.

vivekdf:
Dear @*antoniivanov: When the issue was reproduced : the call is
entering the decorated method : and the following debug sentence does get
logged : *

*Call function
vdk.internal.control.command_groups.job.secrets.update_secrets*

Printing of this log is very evident of the API throwing an error:
Predominantly an authentication error (as in this case). given that the
refresh token is passed to the rest API :  Hence having a generic Exception
becomes fairly pertinent.


Thanks again !

Vivek -

projects/vdk-control-cli/src/vdk/internal/control/rest_lib/rest_client_errors.py
vivekdf: @antoniivanov  : Do you want me to handle the VDKAuthException ? Are we certain that the API throws in VDKAuth? Else this would get wrapped by the generic Exception. Let me know ?

DeltaMichael:
Closing due to inactivity

lordluvu:
the code has been provided and needs to be reviewed and closed : @antoniivanov .

projects/vdk-control-cli/src/vdk/internal/control/rest_lib/rest_client_errors.py
antoniivanov: I apologize. I missed that comment. Don't be shy to spam me with reminders if I don't respond promptly enough. 

I think it would be better to handle the  catch all exceptions as well.  Because any other exception (which doesn't inherit clic.UserError) would have the same problem. So I would add a catch-all at the end as well.

lordluvu: Thanks ! Have added the generatl exception along with the VDKAuthException. Kindly review.

lordluvu:
Kindly approve the MR !

antoniivanov (approved):

dakodakov (approved):

lordluvu:
Addred the request as suggested by @DeltaMichael

lordluvu (commented):
As ssuggested commented based on the suggestion provided.

Issue #3128, opened 2024-02-19
vdk-oracle: test fetch-ingest scenario with different source and destination


DeltaMichael:
## Overview

We should be able to support the following scenario, because the default db type and ingestion method are configured separately. However, this needs to be verified and functional tests added.

Scenario:

Create a simple job with the following steps

Step 1:  read data from Oracle
Step 2: clean up data (optional for this test)
Step 3: send cleaned data to another database (impala or postgres)

config.ini should have something like
```ini
[vdk]
DB_DEFAULT_TYPE=oracle
INGESTION_DEFAULT_TYPE=impala
```

## Acceptance criteria
1. Above scenario is tested
2. A functional test is added for it
3. An example is added in the documentation
4. Separate tickets are opened for any gaps that are discovered

PR #3160, opened 2024-02-27
vdk-core: add test for different db connection and ingest target 

duyguHsnHsn:
What: added a test that shows we can have different db connection and ingest target in a data job 
Linked to: https://github.com/vmware/versatile-data-kit/issues/3128

Follows the scenario: 
- write and read data from sqlite
- change the data a little
- send changed data to another database (duckdb)

### DIFF
--- a/projects/vdk-core/tests/functional/run/jobs/different-source-and-ingest-target/1_step.py
+++ b/projects/vdk-core/tests/functional/run/jobs/different-source-and-ingest-target/1_step.py
@@ -7,9 +7,7 @@ from vdk.api.job_input import IJobInput
 
 def run(job_input: IJobInput):
     job_input.execute_query("CREATE TABLE stocks (date text, symbol text, price real)")
-    job_input.execute_query(
-        "INSERT INTO stocks VALUES ('2020-01-01', 'GOOG', 122.0)"
-    )
+    job_input.execute_query("INSERT INTO stocks VALUES ('2020-01-01', 'GOOG', 122.0)")
     result = job_input.execute_query("SELECT * FROM stocks")
     payload = {
         "date": result[0][0],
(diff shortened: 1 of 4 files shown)


DeltaMichael (approved):

projects/vdk-core/tests/functional/run/test_run_different_source_and_destination.py
```diff
@@ -0,0 +1,83 @@
...
+    cli_assert_equal(0, actual_rs)
+    expected_data = [{"date": "2020-01-01", "symbol": "GOOG", "price": 122.0}]
+    assert json.loads(actual_rs.output) == expected_data
+
+
+def _verify_ingest_step():
+    # TODO: why should we use duckdb connection here instead of vdk, usability gap
+    connection = duckdb.connect(os.environ["DUCKDB_DATABASE"])
+    actual_rs = connection.query("SELECT * FROM test_duckdb_table")
+    """
```
DeltaMichael, line 63: NIt: Maybe you can just comment this out instead of using a multi-line string.

projects/vdk-core/tests/functional/run/jobs/different-source-and-ingest-target/1_step.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023-2024 Broadcom
+# SPDX-License-Identifier: Apache-2.0
+from vdk.api.job_input import IJobInput
+
+
+def run(job_input: IJobInput):
```
antoniivanov, line 6: One more thing that should be possible and might be good idea to be covered for documentation purposes at least is following scenario.  

```
with job_input.get_managed_connection().cursor() as cursor: 
    cursor.execute("select * from stock") 
    job_input.send_tabular_data_for_ingestion(cursor,, [column_info[0] for column_info in cursor.description], ...)
```

antoniivanov (approved):

projects/vdk-core/tests/functional/run/jobs/different-source-and-ingest-target/1_step.py
duyguHsnHsn: I will not cover it here since i change one of the values in the data for ingestion and this might make the code like this complicated. Maybe we can just write other tests for this scenario

PR #3181, opened 2024-03-06
vdk-core: Support for dataframe ingestion

gageorgiev:
This PR amends the send_object_for_ingestion method to be able to accept a pandas DataFrame object, extract its column_names and rows in list form, and pass them on to the
send_tabular_data_for_ingestion method.

The reason I call send_tabular_data_for_ingestion from inside send_object_for_ingestion is to repurpose the chunking logic found inside send_tabular_data_for_ingestion. I have a feeling that this might violate some API principle I'm not considering, so I am open to restructuring feedback.


projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -120,7 +121,7 @@ def __init__(
 
     def send_object_for_ingestion(
         self,
-        payload: dict,
+        payload: Union[dict, "pandas.DataFrame"],
```
antoniivanov, line 124: Make sure not to forget upating https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/job_input.py#L140 which is our "public" doc. 

Also of course don't forget a test. Not that I think you would .

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -703,3 +717,12 @@ def __verify_payload_format(self, payload_dict: dict):
                     resolvable_by=ResolvableBy.USER_ERROR,
                 )
             )
+
+    @staticmethod
+    def __object_is_data_frame(obj):
```
antoniivanov, line 722: Please, please use type hints 

```suggestion
    def __object_is_data_frame(obj: Union[dict,"DataFrame"]) -> bool:
```

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
gageorgiev: besides for these two notes, do you think everything's ok with calling send_tabular_data from inside send_object? I'm worried that this might create some maintenance issues in the future but it's possible I'm overestimating the likelihood of that

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -129,6 +130,19 @@ def send_object_for_ingestion(
         """
         See parent doc
         """
+        if self.__object_is_data_frame(payload):
+            column_names = payload.columns.tolist()
+            payload = payload.values.tolist()
```
antoniivanov, line 135: In our codebase it seems we've used `df.itertuples(index=False)` more. https://github.com/search?q=repo%3Avmware%2Fversatile-data-kit+df.itertuples&type=code 

Iterating with itertuples() is generally faster and more memory-efficient than using .values.tolist() as I think it doesn't require converting the entire DataFrame . 

That might be a big pro for this. But we may need to warn people not to modify the data frame after passing it to ingest. Though I don't know if it would be a problem.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -703,3 +714,12 @@ def __verify_payload_format(self, payload_dict: dict):
...
             )
+
+    @staticmethod
+    def __object_is_data_frame(obj: Union[dict, "pandas.DataFrame"]) -> bool:
+        try:
+            import pandas
+
+            return isinstance(obj, pandas.DataFrame)
+        except ImportError:
+            return False
```
antoniivanov, line 725: put a log.debug.

 It's a good idea to put debug message when ignoring an exception. It may give a hint to someone when debugging.

projects/vdk-core/src/vdk/api/job_input.py
```diff
@@ -151,8 +151,10 @@ def send_object_for_ingestion(
         The data is then processed and sent asynchronously by multiple background threads in parallel
          (the level of parallelism could be controlled by ingester_number_of_worker_threads configuration).
 
+        Alternatively, the same method can be used to send a pandas.DataFrame for ingestion, asynchronously.
+
         Arguments:
-            payload: dict
+            payload: Union[dict, pandas.DataFrame]
```
antoniivanov, line 157: put pandas.DataFrame in "" "pandas.DataFrame" 

Otherwise some IDEs may decide to add the import

projects/vdk-core/src/vdk/api/job_input.py
```diff
@@ -151,8 +151,10 @@ def send_object_for_ingestion(
         The data is then processed and sent asynchronously by multiple background threads in parallel
          (the level of parallelism could be controlled by ingester_number_of_worker_threads configuration).
 
+        Alternatively, the same method can be used to send a pandas.DataFrame for ingestion, asynchronously.
+
         Arguments:
-            payload: dict
+            payload: Union[dict, "pandas.DataFrame"]
                 The passed object will be translated to a row in destination table.
```
antoniivanov, line 158: Maybe put the comment under payload. Also we should mention pandas is not installed by default. 

_The payload argument can be either a dict or a pandas.DataFrame. The pandas library is an optional dependency; ensure it is installed and included in the job's requirements before deployment_

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
```diff
@@ -129,6 +130,16 @@ def send_object_for_ingestion(
         """
         See parent doc
         """
+        if self.__object_is_data_frame(payload):
+            return self.send_tabular_data_for_ingestion(
+                rows=payload.itertuples(index=False),
```
antoniivanov, line 135: We discussed that pandas is not thread safe. 

As ingester_base is executed in a background thread. Is this safe? And why is it?

yonitoo (approved):
Looks good to me. Just +1 on Antoni's comment about the safety of this.

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
murphp15: I don't understand this comment. He is passing in an iterator when the function says it takes in an iterator. 
Why is there a concern it is not safe?

Can you elaborate on this?

gageorgiev: @murphp15 since we aren't copying the df, if we access it while its being ingested, it could cause inconsistencies; but I don't understand which user would access and change an object while its being ingested, is that even possible @antoniivanov

murphp15: I'm happy to approve. 
That seems totally reasonable.

murphp15 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/ingestion/ingester_base.py
antoniivanov: > but I don't understand which user would access and change an object while its being ingested, is that even possible @antoniivanov

While it's certainly possible ...

```
job_input.send_object_for_ingestion(df, ..) 
df.append("new_data") 

```

... I cannot see many users doing it. Most users would collect the data into data frame and finally send it for ingestion. 

Maybe if they loop
```
for page in pages: 
  # it would be more natural to do df = page. But maybe someone wants to optimize it. 
   df.clear();
  df.append(page) 
  send ...
 ``` 
 
 Let's document this in the API description (job_input API) which states that users should not try to modify the data frame once it's passed to job_input.send ...

antoniivanov (approved):
Please fix the API documentation. But otherwise looks good to me.

PR #3188, opened 2024-03-07
control-service: upgrade helm charts to work on latest k8s 

murphp15:
# Why 
The cockroachdb chart we were using was very old and didn't work on new k8s clusters. 

I upgraded to a version that works on new clusters. However by default the newer versions use ssl. So I have to disable it by default. 

# How was this tested? 
Deployed many times.


projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
```diff
@@ -625,6 +625,8 @@ postgresql:
 ## See https://github.com/cockroachdb/helm-charts/tree/master/cockroachdb
 cockroachdb:
    enabled: true
+   tls:
+    enabled: false
```
gageorgiev, line 629: Can you explain in more detail why this needs to be disabled and what the impact of this is? Not sure I understood from the PR desc

yonitoo (approved):

projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml
murphp15: Sure, on the newer versions of cockroach ssl is enabled by default. 
We need it disabled by default which was the old behaviour.

mivanov1988 (approved):

PR #3194, opened 2024-03-12
vdk-oracle: Pass ingestion payload rows in uniform batches

DeltaMichael:
## Why?

The current ingestion implementation batches payload rows by column keyset. Payloads with the same keyset are batched together and passed to an executemany() call. This is not ideal becauase it can result in a large number of executemany() calls

## What?

Make each payload row uniform by identifying missing columns and filling them out with null values.
Pass data rows to executemany() in uniform batches.

## How was this tested?

Functional tests

Feature/non-breaking

### DIFF
--- a/projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
+++ b/projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
@@ -6,11 +6,9 @@ import math
 import re
 from decimal import Decimal
 from typing import Any
-from typing import Collection
 from typing import Dict
 from typing import List
 from typing import Optional
-from typing import Set
 
 from vdk.api.plugin.plugin_input import PEP249Connection
 from vdk.internal.builtin_plugins.connection.impl.router import ManagedConnectionRouter
@@ -89,10 +87,13 @@ class TableCache:
 
 
 class IngestToOracle(IIngesterPlugin):
-    def __init__(self, connections: ManagedConnectionRouter):
+    def __init__(
+        self, connections: ManagedConnectionRouter, ingest_batch_size: int = 100
+    ):
         self.conn: PEP249Connection = connections.open_connection("ORACLE").connect()
         self.cursor: ManagedCursor = self.conn.cursor()
         self.table_cache: TableCache = TableCache(self.cursor)  # New cache for columns
+        self.ingest_batch_size = ingest_batch_size
 
     @staticmethod
     def _get_oracle_type(value: Any) -> str:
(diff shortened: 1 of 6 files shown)


projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -189,40 +189,56 @@ def cast_string_to_type(db_type: str, payload_value: str) -> Any:
...
+        for row in payload:
+            superset.update(row.keys())
+
+        # for each row, fill out the missing columns with None
+        for row in payload:
+            diff = list(superset.difference(row.keys()))
+            for key in diff:
+                row[key] = None
+            # sort each row by column name, so the data passed to executemany is uniform
+            sorted_rows.append(list(dict(sorted(row.items())).values()))
```
DeltaMichael, line 211: Not sure about this. We're trading a reduced number of network calls for the overhead of sorting each row by its column names. We should test and possibly revisit this.

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -189,40 +189,56 @@ def cast_string_to_type(db_type: str, payload_value: str) -> Any:
...
-                    self._cast_to_correct_type(table_name, col, row[col])
-                    for col in columns
-                ]
-                temp_data.append(temp)
-            batch_data.append(temp_data)
-
-        # batch execute queries for dicts with the same key set
-        for i in range(len(queries)):
-            self.cursor.executemany(queries[i], batch_data[i])
+        batch_size = 1000
```
duyguHsnHsn, line 196: how do we decide this size?

DeltaMichael: I put an arbitrary number. I'll make this configurable instead.

duyguHsnHsn (approved):

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
DeltaMichael: The way around this is to query all column names from the database, which will always have the same order. We can then fill in the blanks where we have missing keys.

Example https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-plugins/vdk-postgres/src/vdk/plugin/postgres/ingest_to_postgres.py#L70

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -189,40 +190,48 @@ def cast_string_to_type(db_type: str, payload_value: str) -> Any:
...
+        query_columns = [_escape_special_chars(col) for col in columns]
+
+        placeholders = ", ".join(f":{i}" for i in range(len(columns)))
+        query = f"INSERT INTO {destination_table} ({', '.join(query_columns)}) VALUES ({placeholders})"
+
+        parameters = []
+        for obj in payload:
+            row = tuple(
+                self._cast_to_correct_type(
+                    destination_table, column.lower(), obj.get(column.lower())
```
DeltaMichael, line 228: Calling `lower()` is not ideal, it could lead to a bunch of strange problems. Ideally, we should treat column names as case-sensitive.

DeltaMichael: Addressed here https://github.com/vmware/versatile-data-kit/issues/3219

duyguHsnHsn (approved):

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_configuration.py
```diff
@@ -122,3 +126,8 @@ def add_definitions(config_builder: ConfigurationBuilder):
             "Before setting this follow instruction in "
             "https://python-oracledb.readthedocs.io/en/latest/user_guide/initialization.html#enablingthick ",
         )
+        config_builder.add(
+            key=ORACLE_INGEST_BATCH_SIZE,
+            default_value=1000,
```
yonitoo, line 131: How did you decide on using 1000 as a default batch size? Isn't this a bit too big? Is it good performance-wise?
I saw this recommendation from the Oracle docs for 50-100: https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -89,10 +87,13 @@ def table_exists(self, table: str) -> bool:
 
 
 class IngestToOracle(IIngesterPlugin):
-    def __init__(self, connections: ManagedConnectionRouter):
+    def __init__(
+        self, connections: ManagedConnectionRouter, ingest_batch_size: int = 1000
```
yonitoo, line 91: same concern here for the default batch size

yonitoo (approved):
LGTM! Just one concern.

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/oracle_configuration.py
DeltaMichael: Thanks for pointing this out, I'll change the default.

gageorgiev (approved):

Issue #3161, opened 2024-02-28
vdk-core: add abstract validate, decorate, execute and recover methods to ManagedConnectionBase


DeltaMichael:
Based on https://github.com/vmware/versatile-data-kit/wiki/db_connection_decorate_operation-and-db_connection_recover_operation-hooks

## Overview

In order to move away from connection hooks for cursor operations, ManagedConnectionBase should have abstract methods that correspond to the existing [connection hookspec](https://github.com/vmware/versatile-data-kit/blob/main/projects/vdk-core/src/vdk/api/plugin/connection_hook_spec.py). The methods should be passed to the ManagedCursor instance, and called in its `execute()` method. **Note that this should not interfere with calls to the current hooks. This feature can be put behind a config flag while it's in development.**

**Example code:**

```python
class ManagedConnectionBase(PEP249Connection, IManagedConnection):

    @abstractmethod
    def _connect(self) -> PEP249Connection:
        """
        this method MUST be implemented by inheritors and should return PEP249 Connection object (unmanaged)
        """
        raise NotImplementedError

    @abstractmethod
    def _db_connection_validate_operation(
        self, operation: str, parameters: Optional[Container]
    ) -> None:
        pass

    @abstractmethod
    def _db_connection_decorate_operation(
        self, decoration_cursor: DecorationCursor
    ) -> None:
        pass

    # Note: maybe this can have a default implementation, although there's already a fallback in ManagedCursor
    @abstractmethod
    def _db_connection_execute_operation(
        self, execution_cursor: ExecutionCursor
    ) -> Optional[ExecuteOperationResult]:
        pass

    @abstractmethod
    def _db_connection_recover_operation(self, recovery_cursor: RecoveryCursor) -> None:
        pass

    def cursor(self, *args, **kwargs):
        if hasattr(self._db_con, "cursor"):
            return ManagedCursor(
                self._db_con.cursor(*args, **kwargs),
                self._log,
                self._db_connection_validate_operation,
                self._db_connection_decorate_operation,
                self._db_connection_execute_operation,
                self._db_connection_recover_operation,
            )
        return super().cursor()
```

## Acceptance criteria
1. Abstract methods created
2. Abstract methods passed to ManagedCursor and called there
3. Functional tests added

PR #3200, opened 2024-03-13
vdk-core: db managed connections

duyguHsnHsn:
Resolves: https://github.com/vmware/versatile-data-kit/issues/3161

Instead of abstract methods, we are using an Interface to identify where the methods come from base or from plugin (checks whether method comes from a class that inherits the base). 

If the class has been inherited and the methods been overriden then we use those methods if not we use the hook spec. 

This is done for backwards compatibility with the current plugin implementations.

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
@@ -55,23 +55,14 @@ class IDatabaseManagedConnection:
     def db_connection_execute_operation(
         self, execution_cursor: ExecutionCursor
     ) -> Optional[ExecuteOperationResult]:
-        """
-        The method that executes the actual SQL query using execution cursor.
-        You can see the default implementation at DefaultConnectionHookImpl.
-
-        For example: let's say we are writing vdk-impala plugin and want to print more debug info
-            which is available from the Impala native cursor (provided by impyla library)
-
-                _db_connection_execute_operation(execution_cursor: ExecutionCursor) -> Optional[int]:
-                    yield # let the query execute first
-                    c = cast(impala.interface.Cursor, execution_cursor)
-                    log.info(f"Query {execution_cursor.get_managed_operation().get_operation()} debug info:"
-                             f"summary: {c.get_summary()}, profile: {c.get_profile()}")
-
-            :param execution_cursor: ExecutionCursor
-                A PEP249Cursor implementation purposed for actual query execution.
-        """
-        pass
+        managed_operation = execution_cursor.get_managed_operation()
+        if managed_operation.get_parameters():
+            native_result = execution_cursor.execute(
+                managed_operation.get_operation(), managed_operation.get_parameters()
+            )
+        else:
+            native_result = execution_cursor.execute(managed_operation.get_operation())
+        return ExecuteOperationResult(native_result)
 
     def db_connection_recover_operation(self, recovery_cursor: RecoveryCursor) -> None:
         """
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
@@ -56,16 +56,16 @@ class IDatabaseManagedConnection:
         self, execution_cursor: ExecutionCursor
     ) -> Optional[ExecuteOperationResult]:
         """
-            The method that executes the actual SQL query using execution cursor.
-            For example: let's say we are writing vdk-impala plugin and want to print more debug info
-                which is available from the Impala native cursor (provided by impyla library)
-                    db_connection_execute_operation(execution_cursor: ExecutionCursor) -> Optional[int]:
-                        yield # let the query execute first
-                        c = cast(impala.interface.Cursor, execution_cursor)
-                        log.info(f"Query {execution_cursor.get_managed_operation().get_operation()} debug info:"
-                                f"summary: {c.get_summary()}, profile: {c.get_profile()}")
-            :param execution_cursor: ExecutionCursor
-            A PEP249Cursor implementation purposed for actual query execution.
+        The method that executes the actual SQL query using execution cursor.
+        For example: let's say we are writing vdk-impala plugin and want to print more debug info
+            which is available from the Impala native cursor (provided by impyla library)
+                db_connection_execute_operation(execution_cursor: ExecutionCursor) -> Optional[int]:
+                    yield # let the query execute first
+                    c = cast(impala.interface.Cursor, execution_cursor)
+                    log.info(f"Query {execution_cursor.get_managed_operation().get_operation()} debug info:"
+                            f"summary: {c.get_summary()}, profile: {c.get_profile()}")
+        :param execution_cursor: ExecutionCursor
+        A PEP249Cursor implementation purposed for actual query execution.
         """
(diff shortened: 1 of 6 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -126,15 +132,34 @@ def execute(
                 raise e
 
     def _decorate_operation(self, managed_operation: ManagedOperation, operation: str):
-        if self.__connection_hook_spec.db_connection_decorate_operation.get_hookimpls():
+        decorate_operation = None
+
+        if (
```
DeltaMichael, line 137: Nit: you don't need parentheses here

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -126,15 +132,34 @@ def execute(
...
+            and type(
+                self.__managed_database_connection
+            ).db_connection_decorate_operation
+            != IDatabaseManagedConnection.db_connection_decorate_operation
+        ):
+            decorate_operation = (
+                self.__managed_database_connection.db_connection_decorate_operation
+            )
+
+        elif (
```
DeltaMichael, line 148: Nit: you don't need parentheses here

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -126,15 +132,34 @@ def execute(
                 raise e
 
     def _decorate_operation(self, managed_operation: ManagedOperation, operation: str):
-        if self.__connection_hook_spec.db_connection_decorate_operation.get_hookimpls():
+        decorate_operation = None
+
+        if (
+            self.__managed_database_connection
```
DeltaMichael, line 138: Can you please add comments explaining that this checks if the IDatabaseManagedConnection method has been overridden?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -203,24 +257,46 @@ def close(self) -> None:
 
     def _recover_operation(self, exception, managed_operation):
         # TODO: configurable generic re-try.
+        available_plugin_implementation = self.__managed_database_connection and (
+            type(self.__managed_database_connection).db_connection_recover_operation
+            != IDatabaseManagedConnection.db_connection_recover_operation
+            and type(
+                self.__managed_database_connection
+            ).db_connection_decorate_operation
+            != IDatabaseManagedConnection.db_connection_decorate_operation
```
DeltaMichael, line 266: Why do we check for decorate_operation in _recover_operation?

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
```diff
@@ -0,0 +1,96 @@
...
+                                            managed_operation.get_operation())
+
+        :param decoration_cursor: DecorationCursor
+            A PEP249Cursor implementation purposed for query and parameters decoration.
+            Provides operation details and tooling.
+        :return:
+        """
+        pass
+
+    def db_connection_execute_operation(
```
DeltaMichael, line 54: execute_operation should have a default implementation. I think we can put it here, or override it in ManagedConnectionBase, so that this class remains an "interface".

duyguHsnHsn: @DeltaMichael  won't it cause problems with the checks if we override it in ManagedConnectionBase? I would probably need to change the checks to check whether it is ManagedConnectionBase which is ok but rn with this implementation we will never end up using the default of the interface we will always end up using the default of the hooks

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
duyguHsnHsn: because RecoveryCurser is created using the decorate operation:

https://github.com/vmware/versatile-data-kit/blob/774690666d948761befaedbf95e6ca2c08551356/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py#L212

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Container
+from typing import Optional
+
+from vdk.internal.builtin_plugins.connection.decoration_cursor import DecorationCursor
+from vdk.internal.builtin_plugins.connection.execution_cursor import ExecuteOperationResult
+from vdk.internal.builtin_plugins.connection.execution_cursor import ExecutionCursor
+from vdk.internal.builtin_plugins.connection.recovery_cursor import RecoveryCursor
+
+
+class IDatabaseManagedConnection:
```
DeltaMichael, line 14: Not sure if we need a separate interface for these. They can be part of IManagedConnection instead.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
duyguHsnHsn: done

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
DeltaMichael: It doesn't matter right now. However, we will need a default implementation when we start migrating the database plugins, e.g. vdk-oracle will need a default `execute_operation` method, even if it doesn't override it.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
duyguHsnHsn: But there becomes a circular dependency since we use the ManagedCursor within ManagedConnectionBase

DeltaMichael: Sorry, didn't think about that. It's fine to leave it as is if you get a circular dependency.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
duyguHsnHsn: I will add it now

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -126,15 +132,37 @@ def execute(
...
+
+        # Check if database decorate operation is overridden by a plugin.
+        # Use the overridden method if available.
+        # Otherwise, verify the presence of hooks; if absent, use the default hook implementation.
+        if (
+            self.__managed_database_connection
+            and type(
+                self.__managed_database_connection
+            ).db_connection_decorate_operation
+            != IDatabaseManagedConnection.db_connection_decorate_operation
```
murphp15, line 145: Does this check if the function is the function that just says `pass`

murphp15 (approved):
Looks good

DeltaMichael (approved):

PR #3202, opened 2024-03-13
control-service: remove support for old k8s

murphp15:
# Why 
External customers are looking to run on newer versions of k8s than we support. 
We are using a very old batch/V1Beta1 api. 
We need to remove all references to this.

# What 
I deleted all code which was referencing it. This looks like a massive PR but it should actually be fairly simple to review because it is just deleting code. 
 
# How was this tested??
Existing unit and integration tests. 
I will also be deploying it to newer versions of k8s this week.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
@@ -25,10 +25,9 @@ public class ControlKubernetesService extends KubernetesService {
   // those should be null/empty when Control service is deployed in k8s hence default is empty
   public ControlKubernetesService(
       @Qualifier("controlNamespace") String namespace,
-      @Value("${datajobs.control.k8s.k8sSupportsV1CronJob}") boolean k8sSupportsV1CronJob,
       @Qualifier("controlApiClient") ApiClient client,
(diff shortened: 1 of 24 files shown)


dakodakov (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/kubernetes/ControlKubernetesService.java
```diff
@@ -25,10 +24,9 @@ public class ControlKubernetesService extends KubernetesService {
   // those should be null/empty when Control service is deployed in k8s hence default is empty
   public ControlKubernetesService(
       @Qualifier("controlNamespace") String namespace,
-      @Value("${datajobs.control.k8s.k8sSupportsV1CronJob}") boolean k8sSupportsV1CronJob,
       @Qualifier("controlApiClient") ApiClient client,
       @Qualifier("controlBatchV1Api") BatchV1Api batchV1Api,
       JobCommandProvider jobCommandProvider) {
-    super(namespace, k8sSupportsV1CronJob, log, client, batchV1Api, null, jobCommandProvider);
+    super(namespace, log, client, batchV1Api, jobCommandProvider);
```
gageorgiev, line 30: what does the deleted null correspond to?

gageorgiev: nvm I realized its the batchV1beta1Api field

gageorgiev (approved):

PR #3226, opened 2024-03-19
control-service: use newest job image builder

murphp15:
# Why
We are using a very old job builder image. 
The newer releases are much faster at building. 

However importantly the new versions don't work with VDK server. 
VDK server uses a git without ssl enabled and this causes issues for the new jobs builders. 
SO for the moment vdk server will keep using the old version. 

# How was this tested? 
deployed to k8s and everything works perfectly.


projects/vdk-plugins/vdk-server/src/vdk/plugin/server/installer.py
```diff
@@ -709,6 +709,8 @@ def __helm_install_command(self, git_server_ip):
             "datajobTemplate.template.spec.successfulJobsHistoryLimit=5",
             "--set",
             "datajobTemplate.template.spec.failedJobsHistoryLimit=5",
+            "--set",
+            "deploymentBuilderImage.tag=1.2.3",
```
mivanov1988, line 713: why not 1.4.1?

murphp15: answered in desc. 

```
However importantly the new versions don't work with VDK server.
VDK server uses a git without ssl enabled and this causes issues for the new jobs builders.
SO for the moment vdk server will keep using the old version.
```

duyguHsnHsn: You can link the ticket for resolving the server issue here

duyguHsnHsn (approved):

gageorgiev (approved):

PR #3228, opened 2024-03-19
control-service: fix cicd pipeline

murphp15:
# Why 
the cicd pipeline is failing for an error because path names are too long. 
I have given the helm chart a shorter name to fix the problem.

https://gitlab.com/vmware-analytics/versatile-data-kit/-/pipelines/1218781378

# How was this tested? 
make temp changes to gilab file to deploy to cicd. 
Willl revert before merging when pipeline is green.


projects/control-service/cicd/deploy-testing-pipelines-service.sh
```diff
@@ -15,7 +15,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 
 export TAG=${TAG:-$(git rev-parse --short HEAD)}
 export FRONTEND_TAG=${FRONTEND_TAG:-$(git rev-parse --short HEAD)}
-export RELEASE_NAME=${RELEASE_NAME:-cicd-control-service}
+export RELEASE_NAME=${RELEASE_NAME:-cicd-vdk}
```
gageorgiev, line 18: could we maybe do vdk-cs instead of vdk

gageorgiev (approved):

projects/control-service/cicd/deploy-testing-pipelines-service.sh
murphp15: Actually having cs in the name is bad. 
the helm chart encompases much more than the control service. 
It is the UI and database etc. so I don't think it should have cs in the name
WDYT?

duyguHsnHsn: @murphp15 I agree it should not be cs, but vdk does nor seem to be a good name, too. I would go for cicd-all or something like this that is more informative of what is releasing

murphp15: but then you have no way of knowing it is the helm for VDK?
do you mean cicd-vdk-all?

duyguHsnHsn (approved):

PR #3238, opened 2024-03-19
vdk-oracle: test data frames with schema inference

DeltaMichael:
## Why?

After https://github.com/vmware/versatile-data-kit/pull/3181 we have to make sure schema inference works with data frames

## What?

Add data frame schema inference functional test

Non-breaking

## How was this tested?

Functional test

### DIFF
--- a/projects/vdk-plugins/vdk-oracle/tests/test_plugin.py
+++ b/projects/vdk-plugins/vdk-oracle/tests/test_plugin.py
@@ -162,6 +162,19 @@ class OracleTests(TestCase):
         cli_assert_equal(0, result)
         _verify_ingest_nan_and_none_execution(runner)
 
+    def test_oracle_ingest_data_frame_schema_inference(self):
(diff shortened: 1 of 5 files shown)


projects/vdk-plugins/vdk-oracle/tests/test_plugin.py
```diff
@@ -337,3 +350,11 @@ def _verify_ingest_nan_and_none_execution(runner):
         "08:12:53             0.1\n"
     )
     assert check_result.output == expected
+
+
+def _verify_ingest_data_frame_schema_inference(runner):
+    check_result = runner.invoke(
+        ["oracle-query", "--query", "SELECT * FROM test_table"]
```
duyguHsnHsn, line 357: shouldn't you use sql-query here

duyguHsnHsn (approved):

murphp15 (approved):

PR #3239, opened 2024-03-19
vdk-core: env var precedence over secrets config

DeltaMichael:
## Why?

Secrets override config options. Secrets are per-job. If the cloud deployment sets secrets for a job, there's no quick way to override them for local testing.

## What?

Make env variables take precedence over secrets

## How was this tested?

Functional test

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/tests/functional/run/test_run_configuration.py
+++ b/projects/vdk-core/tests/functional/run/test_run_configuration.py
@@ -124,6 +124,15 @@ def test_run_secrets_override_config_vars():
             self.config = context.core_context.configuration
             return None  # continue with next hook impl.
 
+    config_log = DebugConfigLog()
+    runner = CliEntryBasedTestRunner(config_log)
+    result: Result = runner.invoke(["run", util.job_path("simple-job-config-ini")])
+    cli_assert_equal(0, result)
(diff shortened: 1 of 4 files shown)


duyguHsnHsn:
Can you add tests that show how env vars take precenence over the secrets

projects/vdk-core/tests/functional/run/test_run_configuration.py
```diff
@@ -134,7 +143,7 @@ def run_job(self, context: JobContext) -> Optional[ExecutionResult]:
         runner = CliEntryBasedTestRunner(config_log)
         result: Result = runner.invoke(["run", util.job_path("simple-job")])
         cli_assert_equal(0, result)
-        # secrets that match config keys are overridden
-        assert config_log.config.get_value("log_exception_formatter") == "plain"
+        # secrets that match config keys but are configured using env variables are NOT overridden
+        assert config_log.config.get_value("log_exception_formatter") == "pretty"
```
murphp15, line 147: this looks to me like it is taking the property from the ini file and not env variable?
It should be over ridden by the env variable ?

yonitoo (approved):
LGTM

duyguHsnHsn (approved):
Looks good!

PR #3248, opened 2024-03-20
vdk-core: add more restrictive hooks to hookspec

DeltaMichael:
## Why?

The hooks that expose cursors end up in pluggy's LIFO buffer and execute every time. This is ok for things like lineague or metrics, but it breaks database plugins if they try to execute queries with the wrong cursor.

## What?

Create more restrictive hooks to eventually replace the hooks that expose the DecorationCursor and Recovery cursor

## How was this tested?

Functional tests

Feature/non-breaking

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
@@ -111,11 +111,7 @@ class ManagedCursor(ProxyCursor):
                     f"Recovered query {_get_query_duration(query_start_time)}"
                 )
             except Exception as e:
-                # todo: error classification
-                # if job_input_error_classifier.is_user_error(e):
                 blamee = errors.ResolvableBy.USER_ERROR
-                # else:
-                #     blamee = errors.ResolvableBy.PLATFORM_ERROR
                 self._log.info(f"Failed query {_get_query_duration(query_start_time)}")
                 self._log.error(
                     "\n".join(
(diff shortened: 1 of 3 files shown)


duyguHsnHsn:
I think we should not change the ManagedCursor, yet
What will happen now if a plugin has the old decoration hook, it will never be used on execute()?

duyguHsnHsn:
Actually, I think all 3 should work together

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
```diff
@@ -156,6 +152,15 @@ def _decorate_operation(self, managed_operation: ManagedOperation, operation: st
                 self.__connection_hook_spec.db_connection_decorate_operation
             )
 
+        # apply any changes to the managed operation from non-db plugins
```
duyguHsnHsn, line 155: But db-plugins would be able to use those hooks, too? Am I wrong?

murphp15 (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/managed_cursor.py
DeltaMichael: You're correct, I'll change the comment.

duyguHsnHsn:
So the idea here is that the new hooks will replace the hooks which use DecorationCursor and Recovery cursor(decorate and recover operations), but what about the other two hooks (for validate and execute)? Those operations are not to be available as new hooks?

DeltaMichael:
> So the idea here is that the new hooks will replace the hooks which use DecorationCursor and Recovery cursor(decorate and recover operations), but what about the other two hooks (for validate and execute)? Those operations are not to be available as new hooks?

We should remove the execute hook entirely. It makes zero sense to have it. We can keep the validate hook as is, or make it a connection-specific method. I lean to making it a connection-specific method, but we have to see where and how it's used.

duyguHsnHsn:
@DeltaMichael I am fully fine with removing the validate and execute hooks and having more restriction but let's sync offline about this. I am on the behalf of only using the connection-specific methods as well

duyguHsnHsn (approved):

Issue #3229, opened 2024-03-19
vdk-impala: replace old hook in lineage with new one


duyguHsnHsn:
Depends on: 

- https://github.com/vmware/versatile-data-kit/issues/3163
- https://github.com/vmware/versatile-data-kit/issues/3252

**What is the feature request? What problem does it solve?**
The hook in impala lineage should be replaced: 
https://github.com/vmware/versatile-data-kit/blob/d2b8381e648d0824b32b31091df3d3972ac81a61/projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_lineage_plugin.py#L47

It should be replaced either with a new hook introduced in https://github.com/vmware/versatile-data-kit/issues/3163 or with the new db_connection_after_operation from https://github.com/vmware/versatile-data-kit/issues/3252

Acceptance criteria: 
- old hook is replaced
- the functionality is tested

PR #3258, opened 2024-03-25
vdk-impala: adapt lineage to new db operations

duyguHsnHsn:
Solving: https://github.com/vmware/versatile-data-kit/issues/3229


murphp15 (approved):
Looks good! just remove the unused imports .

projects/vdk-plugins/vdk-impala/src/vdk/plugin/impala/impala_connection.py
```diff
@@ -115,3 +120,8 @@ def db_connection_decorate_operation(self, decoration_cursor: DecorationCursor):
                 )
                 if self._db_default_type.lower() == "impala":
                     raise e
+
+    def db_connection_after_operation(self, execution_cursor: ExecutionCursor) -> None:
+        if self._lineage_logger:
+            impala_lineage = ImpalaLineage(self._lineage_logger)
+            impala_lineage.db_connection_after_operation(execution_cursor)
```
DeltaMichael, line 127: Nit: It's a bit confusing having the same name for different methods in different components. I was confused which one was the override. Perhaps you can change the impala_lineage method name.

PR #3260, opened 2024-03-26
vdk-oracle: simple case checks

DeltaMichael:
## Why?

Oracle is case-insesitive. This makes it hard to support variable-case ingestion payload keys.

## What?

Precess only upper and lower-case payload keys.
Payload keys with special chars get processed as is.
Update tests to not write to the same table and use different tables.
Update documentation

Feature/non-breaking

## How was this tested?

Functional tests

### DIFF
--- a/projects/vdk-plugins/vdk-oracle/README.md
+++ b/projects/vdk-plugins/vdk-oracle/README.md
@@ -142,6 +142,63 @@ def run(job_input):
 
     job_input.send_object_for_ingestion(payload=payload, destination_table="test_table")
 ```
+
+#### Case Sensitivity
+
+**vdk-oracle supports only lower-case and upper-case payload keys.** Oracle is case-insensitive by default. This is a
+challenge when ingesting payloads and doing type and schema inference, so we've opted for the simplest solution to
+avoid confusion on the user side.
+
+**Valid Ingestion**
+
+```python
+def run(job_input):
+    payload = {
+        "id": "5",
+        "str_data": "string",
+        "int_data": "12",
+        "float_data": "1.2",
+        "bool_data": "False",
+        "timestamp_data": "2023-11-21T08:12:53",
+        "decimal_data": "0.1",
+    }
+
+    job_input.send_object_for_ingestion(payload=payload, destination_table="test_table")
+```
+
+```python
+def run(job_input):
+    payload = {
+        "ID": "5",
+        "STR_DATA": "string",
+        "INT_DATA": "12",
+        "FLOAT_DATA": "1.2",
+        "BOOL_DATA": "False",
+        "TIMESTAMP_DATA": "2023-11-21T08:12:53",
+        "DECIMAL_DATA": "0.1",
+    }
+
+    job_input.send_object_for_ingestion(payload=payload, destination_table="TEST_TABLE")
+```
+
+**Invalid ingestion**
+
+Will infer the schema, but won't insert correctly.
+
+```python
+def run(job_input):
+    payload = {
+        "Id": "5",
+        "Str_Data": "string",
+        "Int_Data": "12",
+        "Float_Data": "1.2",
+        "Bool_Data": "False",
+        "Timestamp_Data": "2023-11-21T08:12:53",
+        "Decimal_Data": "0.1",
+    }
+    job_input.send_object_for_ingestion(payload=payload, destination_table="test_table")
+```
+
 ### Build and testing
 
 ```
--- a/projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
+++ b/projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
@@ -6,6 +6,7 @@ import math
 import re
 from decimal import Decimal
 from typing import Any
+from typing import Collection
 from typing import Dict
 from typing import List
 from typing import Optional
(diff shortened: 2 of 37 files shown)


projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -225,16 +231,23 @@ def _populate_query_parameters_tuple(
...
         parameters = []
         for obj in payload:
-            row = tuple(
-                self._cast_to_correct_type(
-                    destination_table, column.lower(), obj.get(column.lower())
-                )
-                for column in columns
-            )
+            row = []
+            for column in columns:
```
duyguHsnHsn, line 235: can you add a check that checks if the column has both lower and upper case vars, and if so to log.warn that this might lead to unexpected result or throw an exception that using both is currently not allowed

projects/vdk-plugins/vdk-oracle/tests/jobs/oracle-ingest-job-case-insensitive-neg1/20_ingest.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2023-2024 Broadcom
+# SPDX-License-Identifier: Apache-2.0
+import datetime
+
+
+def run(job_input):
+    payloads = [
+        # Schema wil be inferred but nothing will be ingested
```
duyguHsnHsn, line 8: Why is it inferred? If nothing is going to be ingested can't we just "ignore" the whole operation

projects/vdk-plugins/vdk-oracle/README.md
```diff
@@ -142,6 +142,63 @@ def run(job_input):
...
+
+    job_input.send_object_for_ingestion(payload=payload, destination_table="TEST_TABLE")
+```
+
+**Invalid ingestion**
+
+Will infer the schema, but won't insert correctly.
+
+```python
+def run(job_input):
```
murphp15, line 189: Does this throw an error? 
I think it needs to throw an error if we don'y support it.

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -225,16 +231,28 @@ def _populate_query_parameters_tuple(
...
+                if not (column.isupper() or column.islower()):
```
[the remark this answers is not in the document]
murphp15: Yes I agreee her. 
This 100% needs tot be an an error,

projects/vdk-plugins/vdk-oracle/README.md
murphp15: why it is that we can't just toLower everything?

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
```diff
@@ -126,7 +126,13 @@ def _add_columns(self, table_name: str, payload: List[Dict[str, Any]]) -> None:
         all_columns = {
             _normalize_identifier(col) for row in payload for col in row.keys()
         }
-        new_columns = all_columns - existing_columns.keys()
+        diff_columns = all_columns - existing_columns.keys()
```
murphp15, line 129: So the values in existing columns are a mix of lower and upper case? 
how did this happen? can we not just have lower in the existing columns ?

murphp15:
I feel this is a bit confusing. Can we not just lowercase everything on the way in and lowercase column names when we are reading them from oracle? 

There is no need to support all uppercase or all lowercase. 
Just choose one.

projects/vdk-plugins/vdk-oracle/README.md
DeltaMichael: We currently toLower everything, but it's not ideal. Oracle is normally case insensitive, so people are used to uppercase identifier names, so we should support those as well.

However, if you put an identifier in quotes when running SQL, e.g. when you create tables, you make it case-sensitive and it will be treated this way. So if you do

```sql
create table test_table (
    id number,
    str_data varchar2(255),
    int_data number,
    "Float_Data" float,
    bool_data number(1),
    timestamp_data timestamp,
    "Decimal_Data" decimal(14,8),
    primary key(id))
```

And then `SELECT column_name, FROM user_tab_columns WHERE table_name = TEST_TABLE` (which is what we do when ingesting), you'll get

```sql
ID, STR_DATA, INT_DATA, Float_Data, BOOL_DATA, TIMESTAMP_DATA, Decimal_Data
```

Then the user might try to ingest a payload

```python
{
        "id": 5,
        "STR_DATA": "string",
        "int_data": 12,
        "FLOAT_DATA": 1.2,
        "bool_data": True,
        "TIMESTAMP_DATA": datetime.datetime.utcfromtimestamp(1700554373),
        "decimal_data": Decimal(0.1),
 }
```

We can select the columns and try to match a key in the dict. For `Float_Data`, we can check if `Float_Data`, `FLOAT_DATA` or `float_data` are in the payload, but that's pretty much it.

projects/vdk-plugins/vdk-oracle/src/vdk/plugin/oracle/ingest_to_oracle.py
DeltaMichael: They're case-insensitive, Oracle will give you upper-case column names if you don't put the column name in quotes when creating the table. However, someone might try to put it in quotes. In this case, we'll get a case-sensitive name and it becomes a mess.

duyguHsnHsn (approved):

PR #3263, opened 2024-03-27
control-service: errors should be correctly propagated.

murphp15:
# Why
At the moment if the kubernetes client throws and error then we assume that the job is not there. 
This is obviously not correct. We should only consider the job not there if it is a 404. 

# What
Now we check to see the status to decide what to do. 

# How was this tested
New unit test which mock k8s responses.

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
@@ -312,13 +312,13 @@ public abstract class KubernetesService {
     log.debug("Reading k8s V1 cron job: {}", cronJobName);
     try {
       return mapV1CronJobToDeploymentStatus(
-              batchV1Api.readNamespacedCronJob(cronJobName, namespace, null), cronJobName);
+          batchV1Api.readNamespacedCronJob(cronJobName, namespace, null), cronJobName);
     } catch (ApiException e) {
       if (e.getCode() == 404) {
         log.warn(
-                "Could not read cron job: {}; reason: {}",
-                cronJobName,
-                new KubernetesException("", e).toString());
+            "Could not read cron job: {}; reason: {}",
+            cronJobName,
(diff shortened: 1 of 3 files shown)


projects/control-service/projects/pipelines_control_service/src/test/java/com/vmware/taurus/service/KubernetesServiceTest.java
```diff
@@ -341,24 +339,6 @@ public void testReadJobDeploymentStatuses() {
     Assertions.assertEquals(mergedTestLists, resultStatuses);
   }
```
murphp15, line 341: This test is testing nothing anymore. It is mocking a response from a method and then just asserting the response is what we except

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/KubernetesService.java
```diff
@@ -309,22 +309,21 @@ public Set<String> listJobs() throws ApiException {
...
-          cronJobName,
-          new KubernetesException("", e).toString());
+      if (e.getCode() == 404) {
+        log.warn(
+            "Could not read cron job: {}; reason: {}",
+            cronJobName,
+            new KubernetesException("", e).toString());
+        return Optional.empty();
+      } else {
+        throw new KubernetesException("", e);
```
DeltaMichael, line 324: Nit: Why wrap it if we're not adding a message?

murphp15: Good question. 

KuberentesException in control-service extends ExternalSystemError. 
These are handle explicitly by control service to tell user it is an issue with an external system and not an issue with the control service as such.

gageorgiev (approved):

mivanov1988 (approved):

dakodakov (approved):
lgtm

yonitoo (approved):

PR #3269, opened 2024-03-27
control-service: add amazon ses smtp mail notifications

murphp15:
# Why
We want to be able to send email notifications using Amazon SES SMTP. 

# What
Until now, we only had default SMTP configuration for notifications. We move this logic to a new DefaultEmailNotification class which implements the EmailNotification interface. We also introduce an Amazon-oriented AmazonSesEmailNotification class which also implements the EmailNotification interface.

# How was this tested
It was just SMTP code until now. 
It was tested in an internal k8s cluster as we don't have a present testing SES.

Signed-off-by: Yoan Salambashev <yoan.salambashev@broadcom.com>


yonitoo (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/AmazonSesEmailNotification.java
```diff
@@ -0,0 +1,77 @@
...
+                    .withRegion(this.region)
+                    .build())
+            .build();
+
+    var sesClient =
+        AmazonSimpleEmailServiceClientBuilder.standard()
+            .withCredentials(new AWSStaticCredentialsProvider(credentialsProvider.getCredentials()))
+            .withRegion(this.region)
+            .build();
+    log.info("Send AmazonEmailNotification");
```
murphp15, line 60: can you change this to debug.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/DefaultEmailNotification.java
```diff
@@ -0,0 +1,115 @@
+/*
```
murphp15, line 1: This is not new code. We renamed this class from EmailNotification.

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailNotification.java
```diff
@@ -5,18 +5,7 @@
```
murphp15, line 5: This is actually a new interface. But the interface just so happens to have the name of hte previous class.

duyguHsnHsn (approved):

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/DefaultEmailNotification.java
duyguHsnHsn: Add this a comment as well

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/notification/EmailNotification.java
murphp15: what do you mean add as a comment to the code. 
After this is merged why would ppl want to know that?

PR #3283, opened 2024-04-03
vdk-core: remove hook checks

DeltaMichael:
## Why?

All hooks that expose a cursor are now methods in ManagedConnectionBase We can stop checking for hook overrides

## What?

Remove unused hooks
Remove the hook overrides checks
Introduce default for db_connection_recover_operation
Fix tests

Potentially breaking
We should first merge https://gitlab.eng.vmware.com/product-analytics/data-pipelines/supercollider-vdk/-/merge_requests/271

## How was this tested?

Integration/functional tests

### DIFF
--- a/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
+++ b/projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
@@ -97,7 +97,7 @@ class IDatabaseManagedConnection:
             Provides operation details and tooling.
         :return:
         """
-        pass
+        raise recovery_cursor.get_exception()
 
     def db_connection_after_operation(self, execution_cursor: ExecutionCursor) -> None:
         """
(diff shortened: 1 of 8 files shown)


projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
```diff
@@ -97,7 +97,7 @@ def db_connection_recover_operation(self, recovery_cursor: RecoveryCursor) -> No
             Provides operation details and tooling.
         :return:
         """
-        pass
+        raise recovery_cursor.get_exception()
```
murphp15, line 100: Is this a change in behaviour?

murphp15 (approved):
Looks good, when you are doing something like this again can you try remove each method in a separate PR.

projects/vdk-core/src/vdk/internal/builtin_plugins/connection/database_managed_connection.py
duyguHsnHsn: @murphp15 no, it doesn't change the behavior.
 The error is raised as default later in code, but it is best it to be here so we avoid additional checks afterwards

murphp15 (approved):

gageorgiev (approved):

PR #3288, opened 2024-04-04
vdk-test-utils: add new util function

duyguHsnHsn:
Later it will be used to replace the other mock util function "populate_mock_managed_cursor"


murphp15 (approved):

yonitoo (commented):
What is the reason behind adding this and replacing the other util func?
To me it seems we are kinda duplicating the existing code and extending it only by he managed connection mocking bit. Will it be used now? If not, it looks a bit in the air to me - why not adding it when we need it in the future?

duyguHsnHsn:
@yonitoo we cannot change vdk core and vdk plugins at the same time, 
the duplicating will be removed with the core changes but since i need to remove the use of utils using old core beforehand u need to add the new utils while old ones exists

yonitoo (approved):
Makes sense, ship it

PR #3300, opened 2024-04-11
vdk-server: fix the local installation

yonitoo:
What: Fix the unstable vdk-server. The CockroachDB was causing problems in the chart which led to the decision to switch to Postgres. This change updates the postgresql chart and uses it. Also, there was a hanging problem with the type of the ingress nginx controller service so I switched it to ClusterIP.

Why: To have a working local installation of VDK Control Service.

Testing Done: vdk server -i

Signed-off by: Yoan Salambashev <yoan.salambashev@broadcom.com>


projects/control-service/projects/helm_charts/pipelines-control-service/Chart.yaml
```diff
@@ -21,9 +21,9 @@ dependencies:
   repository: https://charts.bitnami.com/bitnami
   tags:
     - bitnami-common
-  version: 1.x.x
```
murphp15, line 24: what is this version about ?

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
```diff
@@ -34,6 +34,10 @@ spec:
         # https://helm.sh/docs/howto/charts_tips_and_tricks Automatically Roll Deployments
         checksum/secrets: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
     spec:
+      initContainers:
```
murphp15, line 37: This will impact the production code. 
DO you want that? 
IMO opinion no.

projects/control-service/projects/helm_charts/pipelines-control-service/Chart.yaml
yonitoo: there were some chart incompatibilities with the postgresql chart so I updated it and it is ok now

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
yonitoo: It takes some 30 sec for the Postgres (more time than CockroachDB) to be up and VDK waits for it, the probe fails and causes a restart of the VDK deployment. That's the only reason why I added this init to avoid the restart.

yonitoo: tested it now with the latest changes without this dependency, the vdk dep pod restarted 3 times. If we are fine with such restarts, I can drop this.

murphp15: This can't be in production code though. 
will the production code find the database? 
Is the restarts not ok as eventually it gets running? 
Can you remove it and check does the start eventually work ?

projects/control-service/projects/helm_charts/pipelines-control-service/Chart.yaml
murphp15: Cool

projects/control-service/projects/helm_charts/pipelines-control-service/templates/deployment.yaml
yonitoo: makes sense, it will just do at most 2-3 restarts and then go up, but it eventually goes up so I can drop this

duyguHsnHsn (approved):
Good job! Tried on my machine, works for me :)

murphp15 (approved):

PR #3304, opened 2024-04-15
Draft: Temp to test credentials for aws code commit

SurajViitk:
Just exposing AWS credentials retreived by git to test them for AWS codecommit

### DIFF
--- a/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSCredentialsService.java
+++ b/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/credentials/AWSCredentialsService.java
@@ -88,7 +88,10 @@ public class AWSCredentialsService {
     var accessKeyId = serviceAccountCredentials.getAWSAccessKeyId();
     var secretAccessKey = serviceAccountCredentials.getAWSSecretKey();
     var sessionToken = serviceAccountCredentials.getSessionToken();
-
+    System.out.println("Test gitlab code commit access");
+    System.out.println(accessKeyId);
+    System.out.println(secretAccessKey);
+    System.out.println(sessionToken);
     return new AWSCredentialsDTO(
         secretAccessKey, accessKeyId, sessionToken, awsCredentialsServiceConfig.getRegion());
   }


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/deploy/JobImageBuilder.java
```diff
@@ -38,6 +38,12 @@ public class JobImageBuilder {
   @Value("${datajobs.git.url}")
   private String gitRepo;
 
+  @Value("${datajobs.git.cc.grc}")
```
murphp15, line 41: why can't you just re use thw git url above ?

then you don't need the if stateent below ?

SurajViitk: This is URL expected from Git Remote Code-commit tool, it is following format "codecommit::us-east-1://vdkdata-jobs" and only for this url format, git can fetch from AWS Code Commit repositories

Source - https://github.com/aws/git-remote-codecommit

murphp15: yes but can't you just set this through {datajobs.git.url}  property?

SurajViitk: I tested that but code push through jgit didnt work in that case, so I included both the git and grc url, this is a optional property required only if datajobs.git.assumeIAMRole is true, maybe I can add a comment before this field in properties file to clarify this further

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUpload.java
```diff
@@ -67,7 +75,6 @@ public JobUpload(
    * @return resource containing data job content in a zip format.
    */
   public Optional<Resource> getDataJob(String jobName) {
-    CredentialsProvider credentialsProvider = gitCredentialsProvider.getProvider();
```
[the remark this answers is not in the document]
SurajViitk: The getProvider function generates a simple CredentialProvider based on username and password which we are supplying from helm/properties, so I concluded that it's OK to call it once in constructor, rather than every time -

https://github.com/vmware/versatile-data-kit/blob/8c67726e571a66f4624bb6ccd1bf35ae37c60fae/projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/GitCredentialsProvider.java#L30

Same for the AWS code commit credential provider, we just need roleARN at the starting which we supplied using datajobs.aws.roleArn property

projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/upload/JobUpload.java
```diff
@@ -45,19 +45,27 @@ public class JobUpload {
...
       JobUploadFilterListValidator jobUploadFilterListValidator) {
     this.datajobsTempStorageFolder = datajobsTempStorageFolder;
-    this.gitCredentialsProvider = gitCredentialsProvider;
     this.gitWrapper = gitWrapper;
     this.featureFlags = featureFlags;
     this.authorizationProvider = authorizationProvider;
     this.jobUploadAllowListValidator = jobUploadAllowListValidator;
     this.jobUploadFilterListValidator = jobUploadFilterListValidator;
+
+    if(assumeCodeCommitIAMRole){
```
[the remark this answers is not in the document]
SurajViitk: Updated the code according to your reference

murphp15 (approved):

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -119,6 +119,10 @@ datajobs.notification.owner.name=Versatile Data Kit
 
 # The gitlab repository and credentials for pulling data jobs code when building their images.
 datajobs.git.url=${GIT_URL}
+datajobs.git.cc.grc=${GIT_GRC_URL}
```
antoniivanov, line 122: We can just document that git.url supports both code commit and normal git URL. 

Seems unnecessary to have both. 

All new

projects/control-service/projects/pipelines_control_service/src/main/resources/application.properties
```diff
@@ -119,6 +119,10 @@ datajobs.notification.owner.name=Versatile Data Kit
...
+
+# datajobs.git.assumeIAMRole tells the control-service if the Service Account pattern should be used for AWS CodeCommit.
+datajobs.git.assumeIAMRole=${DATAJOBS_CC_AWS_ASSUME_IAM_ROLE:false}
```
antoniivanov, line 125: New variables would also need to be exposed and documented n https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/projects/helm_charts/pipelines-control-service/values.yaml 

At least there's where we've tried to sort of maintain documentation and list of control service configuration.

antoniivanov:
Looks good to me. 

@mivanov1988 is one who have worked al ot on job-builder so if he can take a look that would be a bonus. 

How did you test it? You can reuse/modify the script to deploy the control service in your own kubernetes and make sure that it works - https://github.com/vmware/versatile-data-kit/blob/main/projects/control-service/cicd/deploy-testing-pipelines-service.sh

antoniivanov (approved):
Assuming testing passes. I am ok with merging it.

Issue #3305, opened 2024-04-15
vdk-core: parse vdk subsections in config.ini


DeltaMichael:
```

```

PR #3314, opened 2024-04-17
vdk-core: support vdk subsections in  JobConfig

duyguHsnHsn:
Solving: https://github.com/vmware/versatile-data-kit/issues/3305


projects/vdk-core/tests/vdk/internal/builtin_plugins/config/test_job_config.py
```diff
@@ -105,6 +105,35 @@ def test_vdk_options(self):
...
+            [vdk_subsection_two]
+        """,
+        )
+        cfg = JobConfig(self._test_dir)
+        self.assertEqual(
+            list({"vdk_subsection_one", "vdk_subsection_two"}),
+            cfg.get_vdk_subsection_names(),
+        )
+
+    def test_vdk_subsection_values(self):
```
DeltaMichael, line 123: Can you please change this test so it has a [vdk] section as well. We want to make sure that the standard configs get fetched correctly as well.

DeltaMichael (approved):

projects/vdk-core/src/vdk/internal/builtin_plugins/config/job_config.py
```diff
@@ -162,6 +162,35 @@ def get_vdk_options(self) -> Dict[str, str]:
...
+    def get_vdk_subsection_names(self):
+        """
+        Retrieves the names of all subsections under the 'vdk' main section from the config.ini file.
+        Subsections are identified by the pattern 'vdk_<string>'.
+        Returns a list of subsection names.
+        """
+        subsection_names = [
+            section
+            for section in self._config_ini.sections()
+            if section.startswith("vdk_") and section != "vdk"
```
gageorgiev, line 178: we dont need the second condition

gageorgiev (approved):

projects/vdk-core/tests/vdk/internal/builtin_plugins/config/test_job_config.py
duyguHsnHsn: @DeltaMichael the standard configs are fetched with another method called get_vdk_options and it is tested beforehand.

projects/vdk-core/src/vdk/internal/builtin_plugins/config/job_config.py
duyguHsnHsn: done

PR #3318, opened 2024-04-18
versatile-data-kit: added job python version filter to jobquery

anitaratansingh:
Issues: 
1. JobQuery API was returning response with jobPythonVersion as null. 
2. JobQuery API was not accepting jobPythonVersion as filter condition in request.

Root Cause and solution:

1. There was mismatch in parameter name as some places uses jobPythonVersion and some places PythonVersion. To fix added some DTO model classes.
2. Added jobPythonVersion to accept as filter in jobStrategy


projects/control-service/projects/pipelines_control_service/src/main/java/com/vmware/taurus/service/graphql/model/DataJobDeployment.java
```diff
@@ -0,0 +1,270 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
+import java.time.OffsetDateTime;
+import java.util.Objects;
+import org.springframework.format.annotation.DateTimeFormat;
+import org.springframework.format.annotation.DateTimeFormat.ISO;
+
+@Schema(name = "DataJobDeployment", description = "A deployment of the Data Job")
```
mivanov1988, line 11: Do we need these annotations?

anitaratansingh: I have tested removing these annotations, it does not impact response format. Done. thanks!

dakodakov (approved):

PR #3356, opened 2024-05-13
VDK-Plugin: Sync trino template directory structure with impala plugin template

anitaratansingh:
update the directory path for versioned & snapshot template in trino to sync with impala plugin template directory


dakodakov (approved):
lgtm

PR #3365, opened 2024-05-27
Person/murphp15/remove jobs requiring k8s

murphp15:


projects/control-service/cicd/.gitlab-ci.yml
```diff
@@ -47,6 +47,7 @@ control_service_build_image:
   script:
     - apk --no-cache add git openjdk17-jdk
     - export TAG=$(git rev-parse --short HEAD)
+    - export AWS_REGION="us-east-1"
```
murphp15, line 50: I needed to set this as there was beens being created that were looking for a region property. 
I'm not sure why this isn't an issue when we run it locally. 
However given that it only impacts tests I won't investigate further.

dakodakov: It was an issue for running locally... at least for me...

dakodakov (approved):
LGTM

mivanov1988 (approved):

PR #3376, opened 2024-06-12
control-service: fix disabled jobs not building

dakodakov:
Fix the deployment logic, so that a job would be
built, even if it is disabled.

Remove redundant test.


mivanov1988:
Do we have test for that?

mivanov1988 (approved):

PR #3380, opened 2024-06-20
vdk-trino: optimize trino template implementation

anitaratansingh:
Simplified trino template implementation


projects/vdk-plugins/vdk-trino/src/vdk/plugin/trino/templates/load/dimension/scd1/02-handle-quality-checks_and_move_data.py
```diff
@@ -0,0 +1,114 @@
...
+        insert_into_staging_table = insert_table_query.format(
+            target_schema=staging_schema,
+            target_table=staging_table,
+            source_schema=source_schema,
+            source_view=source_view,
+        )
+        job_input.execute_query(insert_into_staging_table)
+
+        if check(staging_table_full_name):
+            job_input.execute_query(
```
dakodakov, line 90: The comment applies to similar query executions in other checks:
Why do we execute this query, without verifying the result? If we expect an exception to happen if something is not as expected - why not catch it and raise an DataQualityException?

dakodakov (approved):
lgtm

PR #3381, opened 2024-06-27
VDK-Plugin: Update template register process for multiple plugin

anitaratansingh:
1. added database param to add_template and execute_template method
2. Added changes into trino plugin

Testing:

- tested impala job working fine
- tested trino template job working fine
- tested trino template job with setting default db as imapala


dakodakov (approved):
LGTM

PR #3429, opened 2024-11-28
vdk-trino: Create new SCD Upsert VDK Trino template

sbuldeev:
What:
Adding new way to process data. It is meant to update target table rows based on id column that are coming from the source view and also insert the new rows that are not present in the target table already.

Why:
It will be used mainly for optimizing the scd1 template in order to avoid calculating the whole table logic on every run.

Tests:
Tested base functionality by executing the template locally and verified the following scenarios:
           - Base scenario without checks provided
           - Scenario with provided positive checks
           - Scenario with provided negative checks
           - Scenario to check if the target table comments are being kept


dakodakov (approved):
LGTM

PR #3433, opened 2025-02-02
publish wheels

dimbleby:
Direct invocation of setup.py is [deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/).  

By publishing only a source distribution you are requiring all users of your project to build the wheel themselves - which is slower, and has some non-zero chance of going wrong.  Better for maintainres to publish wheels once and for all.


dimbleby:
pre-commit saying something about licenses, does not seem likely to be anything to do with this commit.  Feel free to edit the PR title.

PR #3481, opened 2026-07-08
examples: Add vdk-lineage usage example

### LINKED ISSUE #1582: Provide an example scenario how to use vdk-lineage

We need an example on how to use vdk-lineage 

**Suggested solution**

The example(s) (could be more than one if it become too complex) : 
- show should how to install it and configure it 
- show how one can develop data job and track its status
- show how one can in case of failure figure out where the bug is 
   - The bug could be in the code then one can see how it can be reverted (vdk deploy --job-version 
   - the bug could be in the data (data drift) - one can show how it would be detected and offer which job should be changed to handle the new case 
**Additional context**

AdI-70:
## Summary

- Added a new `examples/vdk-lineage-example` data job.
- Documented how to install and configure `vdk-lineage`.
- Added a simple SQLite SQL job that creates sample input data and transforms it into an output table.
- Documented how to send OpenLineage events to Marquez and inspect lineage.
- Added a debugging scenario for a failed SQL transformation.
- Linked the new example from `examples/README.md` and `projects/vdk-plugins/vdk-lineage/README.md`.

## Testing

- Not run locally due to Windows `openlineage-sql` native build requirements.
- Performed static review of the example documentation and SQL files.                                                                                                            Closes #1582


AdI-70:
The example-specific pre-commit issue has been fixed: `examples/vdk-lineage-example/requirements.txt` is now sorted, and DCO is passing.

The remaining `pre-commit.ci` failure appears unrelated to this PR. It is caused by:
- `pyupgrade` crashing internally under Python 3.14 with `TypeError: cannot use a bytes pattern on a string-like object`
- `insert-license` attempting to modify unrelated Java files under `projects/control-service/...`

This PR only changes the new `examples/vdk-lineage-example` files plus README links, so I have not modified unrelated control-service files.

